PR #55, opened 2018-01-22
fix guide examples code, rust not support '#' comment

ava57r:


fafhrd91:
I don't think we should remove "# ", those are for reducing noise from examples.
all example get compiled in travis, ignore'ed examples won't compile.

fafhrd91:
cargo doc and mdbook, both support "# "

ava57r:
Did sample a few files.
Did it to check. In the examples looks unusual.

fafhrd91:
"#" is usual for doc strings. as I said, "#" just hides noise from documentation

xilec:
@fafhrd91, just compare rendered view [old](https://github.com/actix/actix-web/blob/master/guide/src/qs_2.md) and [new](https://github.com/andreevlex/actix-web/blob/9b79083b2647baa96831fd9bdcaec699b5c84c2f/guide/src/qs_2.md) variants.

fafhrd91:
you should render guide with mdbook

https://actix.github.io/actix-web/guide/qs_2.html

fafhrd91:
everything that is commented with "# ", is not relevant to documentation

for example:

```rust
# extern crate actix_web;
# use actix_web::*;
# fn index(req: HttpRequest) -> &'static str {
#    "Hello world!"
# }
# fn main() {
   Application::new()
       .resource("/", |r| r.f(index));
# }
```

in this example those lines exists only to satisfy compiler, and not relevant to documentation itself.
I like this comments because travis build compiles all examples and makes sure documentation is relevant.

ava57r:
I realized a great thing, thanks for Your detailed and clear answer. I haven't written enough code infrastructure rust. Now it will use in its work.

fafhrd91:
"cargo test --doc" supports "# "
to test docs or readme you can use https://crates.io/crates/skeptic

fafhrd91:
actix-web build script with skeptic config https://github.com/actix/actix-web/blob/master/build.rs

PR #65, opened 2018-02-05
Allow returning failure::Error from handlers

radix:
This was a little bit roundabout, because I couldn't directly implement `ResponseError for failure::Error` -- because `failure::Error` does not implement `failure::Fail` (or `std::Error`, for that matter). Here's the explanation: https://github.com/withoutboats/failure/issues/151

So instead, this implements `From<failure::Error>` for `Error` and `ResponseError` for `failure::Compat<T>`. The `From` implementation calls `error.compat()` to get the `failure::Compat` value which *does* implement `Fail`.

My main question is: is the `from` implementation ideal? I'm not sure if I should have done something with `backtrace`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -10,6 +10,7 @@ use std::error::Error as StdError;
 use cookie;
 use httparse;
 use futures::Canceled;
+use failure;
 use failure::{Fail, Backtrace};
 use http2::Error as Http2Error;
 use http::{header, StatusCode, Error as HttpError};
@@ -92,6 +93,16 @@ impl<T: ResponseError> From<T> for Error {
             None
         };
         Error { cause: Box::new(err), backtrace: backtrace }
+    }
+}
+
+impl<T> ResponseError for failure::Compat<T>
+    where T: fmt::Display + fmt::Debug + Sync + Send + 'static
+{ }
+
+impl From<failure::Error> for Error {
+    fn from(err: failure::Error) -> Error {
+        Error { cause: Box::new(err.compat()), backtrace: None }
     }
 }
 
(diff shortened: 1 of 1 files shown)


fafhrd91:
I think it worth to implement custom wrapper type for `failure::Error` specifically for backtrace. but we can do this later.

what do you think?

radix:
I don't really know. I guess I should write some tests to see what the behavior of the backtrace is.

fafhrd91:
backtrace is useful if you start your process with RUST_BACKTRACE=1, then you can see source of the error

radix:
yeah, I understand backtraces in general... but I realize now that my PR has a bug, since the current Debug implementation for `actix_web::error::Error` looks like this:

```rust
        if let Some(bt) = self.cause.backtrace() {
            write!(f, "{:?}\n\n{:?}", &self.cause, bt)
        } else {
            write!(f, "{:?}\n\n{:?}", &self.cause, self.backtrace.as_ref().unwrap())
        }
```

this means when printing out Error objects created from `failure::Compat` it's blowing up on the `unwrap()` in this code, because I'm not setting a `backtrace` to `Some` value. So I guess this assumes you're either getting a `Backtrace` from the `cause` or it MUST be set otherwise.

Now that I look at the other `From<T: ResponseError> for Error`, I think I realize what I need to do.

fafhrd91:
`failure::Error` guarantees that it has backtrace, so we should use it.
I designed `actix_web::Error` similar to `failure::Error`

radix:
Ok, I pushed a fix. The `From<failure::Error>` is now simpler and doesn't duplicate with the `From<T>`, which means it has correct backtrace handling now (I think). Added a very simple test that was failing until I fixed it.

fafhrd91 (approved):
lgtm

fafhrd91:
you should have write permission, merge whenever travis is ready

radix:
ok, will do, but I'm going to bed so it will be in the morning. Thanks very much for helping me with this PR!

fafhrd91:
thanks for PR!

PR #67, opened 2018-02-07
Avoid using `Path` to calculate URIs, because it doesn't do the right thing on Windows

radix:
I'm not *totally* sure that this is 100% correct, but it makes the tests pass for me on Windows.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -283,8 +283,14 @@ impl<S> Handler<S> for StaticFiles {
 
             if path.is_dir() {
                 if let Some(ref redir_index) = self.index {
-                    let new_path = format!("{}{}/{}",
-                        req.path(), relpath.to_string_lossy(), redir_index);
+                    // TODO: Don't redirect, just return the index content.
+                    // TODO: It'd be nice if there were a good usable URL manipulation library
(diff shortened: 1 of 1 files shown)


fafhrd91:
I don't think this is complete fix. Could you add test case with request to some nested files.
something similar to test_redirect_to_index, but for existing nested file (for example examples/basic/Cargo.toml)

fafhrd91:
You can use path iterative https://doc.rust-lang.org/std/path/struct.Path.html#method.iter
And then join it with ‘/‘

radix:
Thanks for the advice @fafhrd91 . I have updated the PR with your suggestion, and added a new test.

I know we discussed improving the index handling support on gitter but this is at least a quicker improvement.

fafhrd91:
looks good.

lets do index changes as separate PR

fafhrd91:
wait, I think there is bug

fafhrd91 (approved):

fafhrd91:
I think `Directory` implementation has similar bug

fafhrd91:
merge whenever you are ready

radix:
Okay. I was waiting for appveyor but it seems to be stuck. I'll merge and then try to look in to the other errors on windows.

PR #71, opened 2018-02-12
Wait for spawned thread

rbtcollins:
A spawned thread doesn't block the main thread exiting unless explicitly joined.
The demo as written in the guide simply exits immediately at the moment.

### DIFF
--- a/guide/src/qs_2.md
+++ b/guide/src/qs_2.md
@@ -79,13 +79,14 @@ fn index(req: HttpRequest) -> &'static str {
 }
 
 fn main() {
-# thread::spawn(|| {
+# let child = thread::spawn(|| {
     HttpServer::new(
         || Application::new()
             .resource("/", |r| r.f(index)))
         .bind("127.0.0.1:8088").expect("Can not bind to 127.0.0.1:8088")
         .run();
-# });
+  });
+# child.join().expect("failed to join server thread");
 }
 ```
 
--- a/guide/src/qs_2.md
+++ b/guide/src/qs_2.md
@@ -71,7 +71,7 @@ Here is full source of main.rs file:
 
 ```rust
 # use std::thread;
-# extern crate actix_web;
+extern crate actix_web;
 use actix_web::*;
 
 fn index(req: HttpRequest) -> &'static str {
@@ -79,14 +79,16 @@ fn index(req: HttpRequest) -> &'static str {
 }
 
 fn main() {
-# let child = thread::spawn(|| {
+#  // In the doctest suite we can't run blocking code - deliberately leak a thread
+#  // If copying this example in show-all mode make sure you skip the thread spawn
+#  // call.
+#  thread::spawn(|| {
     HttpServer::new(
         || Application::new()
             .resource("/", |r| r.f(index)))
         .bind("127.0.0.1:8088").expect("Can not bind to 127.0.0.1:8088")
         .run();
-  });
-# child.join().expect("failed to join server thread");
+#  });
 }
 ```
 


fafhrd91:
Code in guide intentionally written this way. Mdbook hides  “ #” lines, but they are runnable, I added this line to make code snippet compileable, so I can test them in ci. Your change will block tests execution.

Here is link to rendered guide https://actix.github.io/actix-web/guide/qs_2.html

fafhrd91:
this is not a bug

rbtcollins:
As written it does not fit the prose that instructs you to run it and
browse to 127.0.0.1:8088 because it exits immediately and thus you will
receive a connection error.

Either the prose is wrong, or the example is wrong, or the expectation that
the prose world not mislead one is wrong.

radix:
@rbtcollins hey, fancy meeting you here!

The example *as rendered* does actually work -- notice that the `#` at the beginning of the threading lines prevents those lines from being rendered. If you copy/paste the code from the rendered example, no threading is involved, it's only creating an HttpServer directly and calling `run()` in the main thread.

rbtcollins:
@radix its a small small world. So, If on https://actix.github.io/actix-web/guide/qs_2.html, you click on the -><- symbol, and then copy the shown text to main.rs, and run it, it will do what I said above.

If you don't click on -><- and try to build and run it, it will fail too like so:
```
   Compiling actix-web-test v0.1.0 (file:///Z:/src/experiments/actix-web)
error[E0432]: unresolved import `actix_web`
 --> src\main.rs:1:5
  |
1 | use actix_web::*;
  |     ^^^^^^^^^ Maybe a missing `extern crate actix_web;`?

error[E0433]: failed to resolve. Use of undeclared type or module `HttpServer`
 --> src\main.rs:8:5
  |
8 |     HttpServer::new(
  |     ^^^^^^^^^^ Use of undeclared type or module `HttpServer`

error[E0433]: failed to resolve. Use of undeclared type or module `Application`
 --> src\main.rs:9:12
  |
9 |         || Application::new()
  |            ^^^^^^^^^^^ Use of undeclared type or module `Application`

error[E0412]: cannot find type `HttpRequest` in this scope
 --> src\main.rs:3:15
  |
3 | fn index(req: HttpRequest) -> &'static str {
  |               ^^^^^^^^^^^ not found in this scope

warning: unused import: `actix_web::*;`
 --> src\main.rs:1:5
  |
1 | use actix_web::*;
  |     ^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

error: aborting due to 4 previous errors

error: Could not compile `actix-web-test`.

To learn more, run the command again with --verbose.
```

Should I have clicked on -><- ? Perhaps not. But there you go. I did that because the earlier examples didn't have the needed imports for me to bring them across to my scratchpad and build them, so I started poking around to find something useful and the -><- widget gave me what I needed.

So I repeat: *as rendered* in *either form*, the example as given does not deliver its purpose.

I don't know why the thread.spawn() call is in there if its not needed: I'm going to guess that its a doctest hack: you're spawning off a deliberately leaked thread during the example validation in order to not have to send a shutdown message to the HttpServer actor.

fafhrd91:
your PR wont pass ci tests because it blocks test runner.

is there any other options?

radix:
I didn't know about the "show hidden lines" button!

maybe just add comments that say "don't include these lines when copying into your own code" or something?

rbtcollins:
Ok, so I now understand the problem: the test runner is being blocked but we want a blocked environment outside the test runner. If we had an oracle that could detect that we could put it in the code, and it would then not block the runner.

Alternatively and more simply we could add a comment explaining whats going on, as well as fixing the trivial rendered example so it runs. I'll update the branch to do that (if you could unclose it please :)).

fafhrd91:
Thanks!

ubnt-intrepid:
In doctest, there is an option to prevent execution of compiled test by specifying `no_run` as a prefix of code block:

```rust
/// ```no_run
/// println!("Here is an example code");
/// ```
pub fn foo() {}
```

If there is no other reason, I think that it is more reasonable to use `no_run` instead of use `thread::spawn` in this case.

fafhrd91:
that is good to know! but I'd like to run examples

ubnt-intrepid:
Ok, I'd understood the situation.

PR #72, opened 2018-02-13
Use AtomicUsize properly

rbtcollins:
doing a read+write on an atomic int will lose updates from other threads.


fafhrd91:
thanks!

PR #80, opened 2018-02-23
Basic Ranges header support

ghost:
As mentioned earlier, I reopen this PR, with modifications you asked, also fixed some other things, and formatted a bit.

The only thing I'm not 100% sure is the following:

```if let Ok(ranges) = HttpRange::parse(rangeheader.to_str().unwrap_or(""), file_metadata.len())```

In theory if the header is not convertable to str(not possible in theory, but I'm taking your side: better safe than sorry)  somehow, it will give back a empty string on error which cant be parsed therefore the parse function is failing and finally generating a 416 Error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ script:
        USE_SKEPTIC=1 cargo test --features=alpn
     else
        cargo clean
-       cargo test
+       cargo test -- --nocapture
        # --features=alpn
     fi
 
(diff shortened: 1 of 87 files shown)


fafhrd91:
could you rebase this PR.

regarding `HttpRange::parse()` if it can not parse header we should return 400 response

fafhrd91:
sorry, i did some work on static files, it provides more functionality

ghost:
No problem, I'll refactor the code as well.

fafhrd91:
Any updates?

fafhrd91:
No activity

PR #87, opened 2018-02-28
fix session mut borrow lifetime

adwhit:
This fixes a lifetime issue where it is impossible to call `session.set(...)` more than once. e.g. calling
```rust
session.set("user_id", 1)?;
session.set("token", "1234")?;
```
throws an error 
```
error[E0499]: cannot borrow `session` as mutable more than once at a time
```
 I believe the problem is that the extra `'a` in the signature ties the lifetime of the borrow to the lifetime of the `Session` object itself.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## 0.4.1 (2018-03-xx)
+
+* Fix Session mutable borrow lifetime
+
+
 ## 0.4.0 (2018-02-28)
 
 * Actix 0.5 compatibility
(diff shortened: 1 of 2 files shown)


fafhrd91:
oh, thanks!

fafhrd91 (approved):

fafhrd91:
could you also add entry to CHANGES.md

adwhit:
Done. BTW thanks for your hard work, this is a great project

fafhrd91:
thanks for fix! the project getting bigger, I need all help I can get :)

PR #89, opened 2018-03-02
add csrf filter middleware

niklasf:
Hi. Many applications need a middleware that filters cross-site requests (to prevent CSRF). Here's an attempt to implement that.

### DIFF
--- a/src/middleware/csrf.rs
+++ b/src/middleware/csrf.rs
@@ -1,0 +1,266 @@
+//! A filter for cross-site request forgery (CSRF).
+//!
+//! This middleware is stateless and [based on request
+//! headers](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Verifying_Same_Origin_with_Standard_Headers).
+//!
+//! By default requests are allowed only if one of these is true:
+//!
+//! * The request method is safe (`GET`, `HEAD`, `OPTIONS`). It is the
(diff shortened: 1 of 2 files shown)


fafhrd91:
looks pretty good! 

I haven't done any browser related services for long time. do people use csrf tokens nowadays?

niklasf:
Yes, that's definitely an option (and OWASP still [recommends](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#CSRF_Specific_Defense) using both headers and some kind of tokens). Headers alone are sufficient, though, and supported in browsers [for a long time](https://datatracker.ietf.org/doc/rfc6454/?include_text=1).

fafhrd91:
I think csrf middleware should support tokens otherwise it's functionality is similar to cors middleware.

niklasf:
Should I perhaps try to merge this into the CORS middleware (and leave the token approach for future work)?

That would probably involve something like adding `CorsBuilder::same_origin()` which would be similar to `CorsBuilder::allowed_origin()` (but not announced in the `Acess-Control-Allow-Origin` header).

fafhrd91:
hmm, I see. let's just merge it.

fafhrd91 (approved):

fafhrd91:
thanks!

niklasf:
Cool, thanks. I guess it's different enough from the CORS middleware to warrant seperation.

fafhrd91:
yes, especially if we add tokens later

Issue #92, opened 2018-03-04
Unix Domain Sockets support for server


messense:
Should be doable with [tokio-uds](https://docs.rs/tokio-uds/0.1.7/tokio_uds/)

fafhrd91:
you can use uds already. I didn't test though

https://actix.github.io/actix-web/actix_web/struct.HttpServer.html#method.start_incoming

messense:
Would be nice to be able to do

```rust
use actix_web::*;

fn main() {
    HttpServer::new(
        || Application::new()
             .resource("/", |r| r.h(httpcodes::HttpOk)))
        .bind("/tmp/myserver.socket").expect("Can not bind")
        .run();
}
```

or something like `.bind_uds(path: AsRef<Path>)`

fafhrd91:
sounds reasonable. would you provide PR?

messense:
Sure, I'll give it a try.

messense:
Turns out it's much more involved than I thought. To add unix domain socket support, I have to modify the `HttpServer` struct

https://github.com/actix/actix-web/blob/1ab676d7eb379cae62d9926808c82af2fe34a929/src/server/srv.rs#L31-L47

I think we have two options, first we can add a new field `unix_sockets: HashMap<net::SocketAddr, std::os::unix::net::UnixListener>`, second we can modify the existing `sockets` field to make the HashMap value an `enum`, for example:

```rust
enum SocketListener {
   Tcp(net::TcpListener),
   Unix(std::os::unix::net::UnixListener),
}
```

For the first option we may need to copy the `start_accept_thread` code to something like `start_accept_thread_unix_socket` to add unix socket specific code and push to `self.accept` in `HttpServer::start` method and etc.

For the second option, we can modify the `start_accept_thread` signature to

```rust
fn start_accept_thread(
    sock: SocketListener, addr: net::SocketAddr, backlog: i32,
    srv: mpsc::UnboundedSender<ServerCommand>, info: Info,
    mut workers: Vec<(usize, mpsc::UnboundedSender<Conn<net::TcpStream>>)>)
    -> (mio::SetReadiness, sync_mpsc::Sender<Command>)
```

Turns out this will get much more complicated since it needs to be able to register/deregister `mio` `TcpListener` and `mio_uds` `UnixListener` to `Poll` and support the `accept_std` API.

fafhrd91:
I think `enum SocketListener` is better approach, it is simpler to extend in the future.

[...]

PR #95, opened 2018-03-05
Add unix domain socket example

messense:
Ref #92 

```bash
$ curl --unix-socket /tmp/actix-uds.socket http://localhost/
Hello world!
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ script:
       cd examples/tls && cargo check && cd ../..
       cd examples/websocket-chat && cargo check && cd ../..
       cd examples/websocket && cargo check && cd ../..
+      cd examples/unix-socket && cargo check && cd ../..
     fi
   - |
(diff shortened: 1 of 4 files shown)


messense:
Although this will only one thread for handling incoming connections according to the [documentation](https://actix.github.io/actix-web/actix_web/struct.HttpServer.html#method.start_incoming).

And it does not delete the socket file (`/tmp/actix-uds.socket`) when stopping the server so it will fail to start next time you run it unless you delete the socket file manually.

fafhrd91 (approved):

fafhrd91:
I thought you want to add something like `.bind_uds` or something like that

messense:
Yes, I still want that. But getting a working example of current code will help me iterate the API design and will be much clear when reviewing that PR someday.

fafhrd91:
Sounds good

PR #98, opened 2018-03-06
Update qs_14.md

flip111:
fix missing semicolon

```
error: expected one of `!`, `+`, `::`, or `;`, found `fn`
  --> src/main.rs:29:5
   |
27 |     type Result = Result<User, Error>
   |                                      - expected one of `!`, `+`, `::`, or `;` here
28 | 
29 |     fn handle(&mut self, msg: CreateUser, _: &mut Self::Context) -> Self::Result
   |     ^^ unexpected token
```


fafhrd91:
thanks!

fafhrd91:
one of the reason why `ignore` is just bad :)

PR #102, opened 2018-03-06
add tests with large random bodies for gzip

mockersf:
failing tests around gzip bodies in response and request


fafhrd91:
Thanks! That is helpful

PR #106, opened 2018-03-06
give a url in the log when starting

niklasf:
This (or something similar) would be nice because the URLs are then clickable in many terminal emulators.


fafhrd91:
thanks

PR #109, opened 2018-03-07
make session an optional feature

kingoflolz:
enabled by default so should not be a breaking change.

(would be great if you can push out a new release)


messense (approved):
LGTM, most stateless REST API servers may not need session.

fafhrd91:
Thanks!

I will push new release today or tomorrow

fafhrd91 (approved):

PR #112, opened 2018-03-07
Bind server to RawFd

### LINKED ISSUE #105: Start server from RawFd
Cargo watch does not work because the sometimes the socket that's running will not be closed in time. That's why catflap exists. The problem is that the actix-web server does not allow for servers to be created from a raw file descriptor.

https://github.com/passcod/catflap  

There are integrations with hyper and iron but not with actix-web.

ghost:
Attempt to resolve #105

### DIFF
--- a/src/server/srv.rs
+++ b/src/server/srv.rs
@@ -19,6 +19,9 @@ use native_tls::TlsAcceptor;
 #[cfg(feature="alpn")]
 use openssl::ssl::{AlpnError, SslAcceptorBuilder};
 
+#[cfg(target_family="unix")]
+use std::os::unix::io::{FromRawFd, RawFd};
+
 use helpers;
 use super::{IntoHttpHandler, IoStream};
 use super::{PauseServer, ResumeServer, StopServer};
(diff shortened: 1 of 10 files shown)


src/server/srv.rs
```diff
@@ -173,9 +176,22 @@ impl<H> HttpServer<H> where H: IntoHttpHandler + 'static
         self.sockets.keys().cloned().collect()
     }
 
+    #[cfg(target_family="unix")]
+    /// The socket to bind to
+    ///
+    /// To bind multiple sockets this method can be called multiple times.
+    pub fn bind_socket(mut self, sock: RawFd) -> io::Result<Self> {
```
fafhrd91, line 183: Let’s rename it to a bind_raw_fd. I am not sure about safety though

fafhrd91:
I don't think we need bind_raw_fd at all.
could you rewrite your example using HttpServer::listen() method

fafhrd91:
Ping

fafhrd91:
reopen if you would like to complete this pr

ffimnsr:
this didn't get to release that's why I was searching on how to bind fd

fafhrd91:
@ffimnsr Rawfd is not supported atm. Should be easy to support, but it should be added to actix-server first

PR #113, opened 2018-03-07
Let CSRF filter catch cross-site upgrades

niklasf:
To prevent CSWSH. Applications that want to allow cross-site websockets can opt-in.


fafhrd91:
Looks good! But could you add simple test for upgrade

niklasf:
Of course, done.

fafhrd91:
Thanks!

PR #115, opened 2018-03-09
Add protobuf support

kingxsp:
add google protobuf support.


fafhrd91:
that is pretty cool!

fafhrd91 (approved):

fafhrd91:
oh, I merged it too fast :)

could you add some tests

messense:
I think this should be feature gated and off by default to avoid pulling in too many dependencies?

fafhrd91:
i agree, i will add feature later

kingxsp:
@fafhrd91  ok! I will write test in protobuf.rs.  protobuf-example run is ok！

fafhrd91:
It is more about make sure everything will work in future

fafhrd91:
after careful consideration of all implications of protobuf code, i decided to move it to example.
i am definitely interested in proper protobuf integration, but this needs more thinking. also i think protobuf code belongs to separate library, something like actix-protobuf

kingxsp:
I waiting for actix-protobuf. I hope to come soon.

fafhrd91:
@kingxsp I don't have much time at the moment but I can create repo and give you write access. what do you think?

kingxsp:
@fafhrd91 ok! let me try. Before this, do you have any ideas or suggestions? I think the current example of protobuf.rs is good, move this file to the new repo and add the test and example is done.

fafhrd91:
@kingxsp I added repo for you, you need to accept invitation

PR #118, opened 2018-03-11
Use Vec instead of HashMap to store sockets in HttpServer

messense:
It seems that there's no need to use a `HashMap`, it also relaxes the `Hash` trait bound on `SocketAddr`. Unix socket [`SocketAddr`](https://doc.rust-lang.org/std/os/unix/net/struct.SocketAddr.html) does not impl `Hash`.


kingxsp:
Use HashMap is correct.

fafhrd91 (approved):
Hash concern makes sense

PR #119, opened 2018-03-11
Add default resource for StaticFiles

### LINKED ISSUE #57: StaticFiles handler will not invoke default_resource on 404
When using `fs::StaticFiles::new(".", false)` default resource will not be called upon hitting 404.

Presumably because we only in handler for non-default resource

DoumanAsh:
I'd like to hear opinion on my approach to default resource.

Currently tests are not working since it expects HttpResponse (and which also provides extra information about result)

Maybe you could advise test strategy for just `Reply`?

Closes #57

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -21,11 +21,11 @@ use mime_guess::get_mime_type;
 use header;
 use error::Error;
 use param::FromParam;
-use handler::{Handler, Responder};
+use handler::{Handler, RouteHandler, WrapHandler, Responder, Reply};
 use httpmessage::HttpMessage;
 use httprequest::HttpRequest;
 use httpresponse::HttpResponse;
-use httpcodes::{HttpOk, HttpFound, HttpMethodNotAllowed};
+use httpcodes::{HttpOk, HttpFound, HttpNotFound, HttpMethodNotAllowed};
 
 /// A file with an associated name; responds with the Content-Type based on the
 /// file extension.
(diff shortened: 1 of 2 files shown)


fafhrd91:
I think approach is fine. 

I am not sure about `Resource`. I think it is too flexible.

DoumanAsh:
Is there a proper way to test resulting `Reply` though?

For `Resource` though I only aimed for consistency with `Application::default_resource`

fafhrd91:
1. for `Reply` you can convert it to a response and then inspect it or you can use TestServer.

2. `Resource` can lead to confusion, you can register route with wrong predicates. It is better to use Handler

DoumanAsh:
Rebase, feel free to take a look

fafhrd91:
I think it looks good. One question, should we make StaticFiles generic over application state?

DoumanAsh:
I was thinking to avoid complicating default handler.
Do you think we need?
I can add if you'd like

fafhrd91:
Let’s add it, just for sake of symmetry

fafhrd91 (approved):

fafhrd91:
Could you also add entry to changes

DoumanAsh:
As per your request: updated changes and added State to Static Files

src/fs.rs
```diff
@@ -447,28 +428,32 @@ impl StaticFiles {
...
     /// showing files listing.
-    pub fn index_file<T: Into<String>>(mut self, index: T) -> StaticFiles {
+    pub fn index_file<T: Into<String>>(mut self, index: T) -> StaticFiles<S> {
         self.index = Some(index.into());
         self
     }
+
+    /// Sets default resource which is used when no matched file could be found.
+    ///
+    /// Contrary to Application, it cannot have state.
```
fafhrd91: remove this

fafhrd91:
lgtm, just fix comment for default_handler

you should have write permission

PR #121, opened 2018-03-13
Send Query Parameters in client requests

glademiller:


src/client/writer.rs
```diff
@@ -112,7 +112,9 @@ impl HttpClientWriter {
 
             // status line
             let _ = write!(buffer, "{} {} {:?}\r\n",
-                           msg.method(), msg.uri().path(), msg.version());
+                           msg.method(), 
+                           msg.uri().path_and_query().map(|u| u.as_str()).unwrap_or(""),
```
fafhrd91: should we use .unwrap_or("/") instead of .unwrap_or("")  ?

glademiller: I was trying to match the functionality here https://docs.rs/http/0.1.5/src/http/uri/mod.rs.html#339 IE before we called path and it returned "" if there was no path.

fafhrd91: i think empty string is a bug

glademiller: I could see how it could be a problem. I'll return "/"

fafhrd91 (approved):

fafhrd91:
thanks!

PR #154, opened 2018-04-02
Use https (not http) url for meritbadge

dholbert:
Right now this readme file uses an HTTP url to reference a meritbadge image, which ends up producing "broken https" UI on the crates.io page https://crates.io/crates/actix-web. This patch just upgrades this to an HTTPS url (which still works), to avoid that problem. (Literally a 1-character change, changing "http" to "https" in "http://meritbadge.herokuapp.com/actix-web")


dholbert:
Here's a screenshot of roughly what the degraded HTTPS UI looks like here, in Firefox (screenshot taken from another project that has the same problem, but I get the same results at https://crates.io/crates/actix-web ):
screenshot of degraded HTTPS UI

fafhrd91:
thanks!

PR #156, opened 2018-04-03
Added support for configuring a named resource

ghost:
On latest master you can access a resource's name, but I couldn't see an easy way to create a named resource for use in middleware, etc.

This change just adds a named_resource function for creating a resource with the specified name.


fafhrd91:
You can use [name()](https://actix.github.io/actix-web/actix_web/dev/struct.ResourceHandler.html#method.name)

Here is [example](https://actix.github.io/actix-web/actix_web/struct.HttpRequest.html#method.url_for)

fafhrd91:
could you add doc string and test

fafhrd91 (changes requested):
please add doc string and test

fafhrd91:
No activity, closing

PR #170, opened 2018-04-09
Public, signed and private cookies

### LINKED ISSUE #159: Add private cookie sessions
At the moment, cookie-based sessions are secure but not private. This means that the client is able to view the session data, but not alter it.

While this is usually good enough, it would be nice to enable private cookies (a feature which already exists in [cookie-rs](https://github.com/alexcrichton/cookie-rs)) so that the user can neither read nor alter session cookies.

adwhit:
This is one way to allow the use of public, signed and private cookies with `CookieSessionBackend`.

closes #159

### DIFF
--- a/src/middleware/session.rs
+++ b/src/middleware/session.rs
@@ -381,15 +381,6 @@ pub struct CookieSessionBackend(Rc<CookieSessionInner>);
 
 impl CookieSessionBackend {
 
-    /// Construct new signed `CookieSessionBackend` instance.
-    ///
-    /// Panics if key length is less than 32 bytes.
-    #[deprecated(since="0.5.0", note="use `signed` or `private` instead")]
-    pub fn new(key: &[u8]) -> CookieSessionBackend {
-        CookieSessionBackend(
-            Rc::new(CookieSessionInner::new(key, CookieSecurity::Signed)))
-    }
-
     /// Construct new *signed* `CookieSessionBackend` instance.
     ///
     /// Panics if key length is less than 32 bytes.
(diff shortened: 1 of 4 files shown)


src/middleware/session.rs
```diff
@@ -316,7 +328,21 @@ impl CookieSessionInner {
                 if cookie.name() == self.name {
                     let mut jar = CookieJar::new();
                     jar.add_original(cookie.clone());
-                    if let Some(cookie) = jar.signed(&self.key).get(&self.name) {
+
+                    let cookie_opt = match self.security {
+                        CookieSecurity::Public => {
+                            // annoyingly this must be handled separately since jar.get
```
fafhrd91: you should open issue in upstream crate got this

fafhrd91:
what is use case for public session?

adwhit:
I thought about not having it. But sometime it can be useful. If you just use your cookie to carry a session token which is validated through e.g. redis on the server side, then it can be more readable, and just as secure, to have a public cookie which is just `{ token: ABCDE12345 }`

fafhrd91:
but that case client can do brute force attack on tokens

adwhit:
Well, usually it wouldn't be possible to brute-force an API token, they're often 256+ bits. (And most servers would have rate-limiting). But I am happy to remove if you prefer?

fafhrd91:
lets remove it.

adwhit:
Rebased

fafhrd91:
great! lets wait for travis and then I'll merge it! 

thanks!

adwhit:
Quick thought, since `CookieSessionBackend::new` was a new (breaking) API change for 0.5, it doesn't really make sense to deprecate it. Should it just be removed instead?

fafhrd91:
Sounds good. Let’s remove it

fafhrd91:
thanks again!

PR #173, opened 2018-04-12
fix end-of-stream handling in parse_payload

jannic:
parse_payload can be called with a pre-filled buf.

In this case, it's totaly fine for read_from_io to return
sync::Ready(0) while buf is not empty. This is not an
PayloadError::Incomplete.

So, move the check for PayloadError::Incomplete down to the
decoding code: If the decoder is not ready, but the input stream
is finished, PayloadError::Incomplete will be returned.

### DIFF
--- a/src/client/parser.rs
+++ b/src/client/parser.rs
@@ -81,16 +81,11 @@ impl HttpResponseParser {
         if self.decoder.is_some() {
             loop {
                 // read payload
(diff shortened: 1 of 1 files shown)


fafhrd91 (approved):

fafhrd91:
very good, thanks!

PR #184, opened 2018-04-17
Fix route in App::resource example

ivanovaleksey:
Hello, I guess the route should be updated.
Also, it is said [here](https://docs.rs/actix-web/0.5.2/actix_web/struct.HttpRequest.html#method.match_info) `:param` syntax should be used.
Is `HttpRequest::match_info` documentation outdated or I am missing on that?


fafhrd91:
thanks! I'll update doc string

PR #189, opened 2018-04-20
Update README links to use new guide

drklee3:
Seems like there were a few links that weren't updated when the guide was moved to a separate repo. This updates those links, hopefully to the correct pages.


fafhrd91 (approved):

fafhrd91:
Thanks!

Issue #191, opened 2018-04-21
panic when Websocket is closed with no error code


fuchsnj:
I setup a very basic "echo" websocket server. When I connect and then close the connection, I get a panic.

Websocket client (javascript console in Chrome/Firefox)
```javascript
let ws = new Websocket("ws://127.0.0.1/ws");
ws.close(); //if I call ws.close(1000) instead, everything works fine
```
The code that tries to read an error code and causes a panic is here: https://github.com/actix/actix-web/blob/v0.5.4/src/ws/mod.rs#L313

I see the exact same behavior in Chrome and Firefox. If I provide a status code to the `close()` method such as 1000, everything seems to work fine.

I am running actix-web `v0.5.4`

Relevant RFC notes (https://tools.ietf.org/html/rfc6455#page-41 section 7.1.5)
```
If this Close control frame contains no status code, _The WebSocket
   Connection Close Code_ is considered to be 1005.
```

Stacktrace
```
thread 'arbiter:"b333e870-835c-416a-8d36-1cd189cb731b":"actor"' panicked at 'assertion failed: 1 <= nbytes && nbytes <= 8 && nbytes <= buf.len()', /home/nathan/.cargo/registry/src/github.com-1ecc6299db9ec823/byteorder-1.2.2/src/lib.rs:1858:9
stack backtrace:
   0:     0x5562ec72d18f - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::h2fb330a06456ac7e
                               at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1:     0x5562ec7328d0 - std::sys_common::backtrace::print::hf6ccec7cfd12ed67
                               at libstd/sys_common/backtrace.rs:71
                               at libstd/sys_common/backtrace.rs:59
   2:     0x5562ec71f663 - std::panicking::default_hook::{{closure}}::h464de9af91e4ca8e
                               at libstd/panicking.rs:205
   3:     0x5562ec71f3ed - std::panicking::default_hook::hcde1a6167c711c5f
                               at libstd/panicking.rs:221
   4:     0x5562ec71fbea - std::panicking::rust_panic_with_hook::h8dcdd9a7e80a2917
                               at libstd/panicking.rs:457
   5:     0x5562ec6e04c7 - std::panicking::begin_panic::h657bd638087f2a51
                               at /checkout/src/libstd/panicking.rs:391
[...]

PR #196, opened 2018-04-21
Websocket close reason

### LINKED ISSUE #193: Websocket CloseCode Empty/Status is ambiguous
The `CloseCode::Empty` and `CloseCode::Status` are basically the same code.

According to the RFC, if a client sends no close code (`CloseCode::Empty`) this is equivalent to 1005 (`CloseCode::Status`). This can be confusing since it might not be obvious which one of these values you should check against.

I suggest removing `CloseCode::Empty` and using an `Option::None` instead.
I would also recommend changing `CloseCode::Status` to `CloseCode::NoStatus` to reflect what it actually means.

These are both breaking changes, but if you like them I can make a PR for this.

fuchsnj:
Closes #193 

Fixes:
- Fix a panic in the websocket client if it receives a close frame without a close code (similar to #191)

Breaking Changes:
- Remove `CloseCode::Empty` and `CloseCode::Status`, which are replaced with `Option::None`
- Add `CloseReason` which includes both the `CloseCode` and description. Since the description is not sent unless a `CloseCode` is given, this allows the description to not be specified when no `CloseCode` is given
- When you receive a Close frame, the description is now parsed

I wanted to make the interface for `close()` methods look like
```rust
pub fn close<T: Into<CloseReason>>(&mut self, reason: Option<T>)
```
then you could use it like this...
```rust
ws.close(None); //close with no status
ws.close(Some(CloseCode::Normal)); //close with only status code
ws.close(Some((CloseCode::Normal, "some description"))); //close with both
```

but I couldn't find a nice way to allow `None` without declaring `T`. If you know of a good alternative, I can modify this PR.

### DIFF
--- a/src/ws/frame.rs
+++ b/src/ws/frame.rs
@@ -11,7 +11,7 @@ use payload::PayloadHelper;
 
 use ws::ProtocolError;
 use ws::mask::apply_mask;
-use ws::proto::{CloseCode, OpCode};
+use ws::proto::{CloseCode, CloseReason, OpCode};
 
 /// A struct representing a `WebSocket` frame.
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


src/ws/frame.rs
```diff
@@ -29,24 +28,22 @@ impl Frame {
...
-                raw[..]
-                    .iter()
-                    .chain(reason.as_bytes().iter())
-                    .cloned(),
-            )
-        };
-
-        Frame::message(payload, OpCode::Close, true, genmask)
+    pub fn close(reason: Option<CloseReason>, genmask: bool) -> Binary {
+	    let payload = match reason {
```
messense: Some indentations seem odd, mixing space and tab I guess? It would be better to use the same style as the rest of the codebase.

fafhrd91: `cargo fmt` should help

fuchsnj: `cargo fmt` tried to change a lot of code, including stuff I didn't touch, so I just manually fixed this.

fafhrd91: Do you use rustfmt-preview?

fuchsnj: I tried rustfmt-preview, and also the rustfmt built into the IntelliJ Rust plugin (Not sure how it's different) but they both tried to change a lot of code I didn't touch.

fafhrd91: ok, np. I will run fmt.

fafhrd91 (approved):
lgtm

but lets wait with merging for couple days. i'd like to release 0.5.5 in couple days, and then switch master to 0.6

fafhrd91:
@fuchsnj you should have permission to merge

fafhrd91 (approved):

PR #204, opened 2018-04-27
Add Content-Disposition to NamedFile (fixes #172)

### LINKED ISSUE #172: Add Contet-Disposition header support to NamedFile
At the moment responder implementation does not add content-disposition header to a NamedFile response

svenstaro:
I used the raw strings for the headers instead of stuff in `content_disposition.rs` because that is commented out and I don't know why that is.


DoumanAsh:
Hej @svenstaro in past we had some talk about this header with @fafhrd91 
Sadly I was a bit busy to work on it myself, but we had idea about simple heuristic to deterine whether file should be inline or attachment.
Things like images should more like to be as `inline` rather than attachment.

And more importantly it would be nice to have a way for user to configure it in place of automatic settings my actix-web itself.

svenstaro:
Alright fair enough but for now this would be an improvement over the current state for sure.

fafhrd91:
@svenstaro could you use inline for images and text

you can check mimetype by `get_mime_type(&ext.to_string_lossy(())`

fafhrd91 (changes requested):
could you use inline for images and text. you can check mimetype by get_mime_type(&ext.to_string_lossy(())

svenstaro:
@fafhrd91 fixed with tests.

fafhrd91:
thanks

PR #208, opened 2018-05-02
Add WsWriter trait

therustmonk:
`WsWriter` trait is a common interface for writing to a websocket and
it's implemented for both: `WebScoketContext` and `ClientWriter`.

`WsWriter` useful for creating generic implementations of `StreamHandler` which could handle incoming (server) and outgoing (client) sides. If you know how to create universal `Actor` for websockets compatible with two directions in another way then tell me, please.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,6 +175,7 @@ pub use httprequest::HttpRequest;
 pub use httpresponse::HttpResponse;
 pub use json::Json;
 pub use scope::Scope;
+pub use ws::WsWriter;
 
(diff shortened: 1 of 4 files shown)


fafhrd91 (approved):

fafhrd91:
Could you explain reason why you want to unify client and server ws impl?

I think change is good though!

therustmonk:
I want to unify client and server to create p2p network over WebSocket (similar like `libp2p` has), but want to use actors instead.

Client's handler equals a handler of server in this case.

PR #219, opened 2018-05-10
correct order of format arguments in readme example 

benjamingroeber:
The example currently prints
Hello {id}! id:{name}
whereas it should be
Hello {name}! id:{id}


fafhrd91:
thanks! :)

PR #221, opened 2018-05-12
Include mention of http client in README.md

skorgu:
The http client was mentioned in one of the release notes but is not discoverable from the readme. For now, just link to the mod.rs which contains a brief usage example in comments.


fafhrd91:
thanks!

Issue #60, opened 2018-01-27
Ranges header support


ghost:
This week I took a deep dive in the RFC and thought about the possibilites. I have a solution now, but I don't feel it's complete, so I need your advice about the implementation.

RFC defines two types of response when the ranges header is present: a simple one, and when you request multiple ranges which must give you back a "Multipart" response.

Implementing the latter one has the advantage of fully conforming the RFC, but has security implications(overlapping ranges, etc). The former one is much easier if you modify the Responder. In most cases it's enough. You can watch&seek a movie in your browser, also you can get a working ipxe server(this was my goal with this). Implementing this for the NamedFIle is trivial, but what if we need the ranges support for files when they come from another location.(database like MongoDB or CouchDB, S3, another webserver etc) In this case it's maybe easier to add the support in a main component.

What do you think about this and which is the preferred way for this project?

fafhrd91:
first of all let's support only NamedFile, supporting arbitrary sources is too complex, and way out of scope of this ticket. 

lets add simple ranges for now. I am planing to add http client with multipart support after that adding multipart ranges would be much easier

fafhrd91:
i think #227 is enough for now

PR #227, opened 2018-05-16
NamedFile: added basic ranges header support, added content-length support

ghost:
I added support for the content-length for the NamedFile, since this can be important with big files, since it helps the client to decide to get a whole file or specific chunks from it.

I also added some ranges support as mentioned in #60 . I'm not expecting this getting merged as-is, since I had to modify a helper-type, it works however but not something I'm happy with right now. I shared it with you with the hopes of you'll have a better idea how to do implement this.

I also modified the disposition of the Video types, since it's making sense to watch a video in your browser. To me at least.

There's also some bugginess with the "accept-ranges: bytes" header, since it causing some error messages, and I didn't found the root cause of these, even with a packetcapturing. I'll work on this some more for sure.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,13 +31,13 @@ environment:
       CHANNEL: beta
     # Nightly channel
     - TARGET: i686-pc-windows-gnu
-      CHANNEL: nightly-2017-12-21
+      CHANNEL: nightly
     - TARGET: i686-pc-windows-msvc
-      CHANNEL: nightly-2017-12-21
+      CHANNEL: nightly
     - TARGET: x86_64-pc-windows-gnu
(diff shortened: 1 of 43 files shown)


fafhrd91:
looks good. but could you add test for range response

ghost:
*sigh* oh dear github...

Added test for status code, content-length header, and  range response header.

fafhrd91:
something is wrong with PR, rebase or squash maybe?

fafhrd91:
ping

ghost:
I'm working on it. Somehow a pull messed up my git tree.

ghost:
Should be fine by now.

fafhrd91:
thanks!

PR #231, opened 2018-05-19
NamedFile: range upgrade

ghost:
Range header got fixed, from now on it should work as intended. Requested range is now inclusive. 

Range header in a response changed to Content-Range: Range is request only, Contene-Range is response only. [I should read more](https://tools.ietf.org/html/rfc7233)

Enabled Accept-Ranges since tests showed me the error messages didn't had any connection with this header. I found some other interesting stuff here, I'll open issues if I can't find the culprit.

Also added tests for content-range(new), content-length(placed in a different fn, added new tests), and the test for HTTP Status 416 was missing.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -289,30 +289,28 @@ impl Responder for NamedFile {
                 resp.set(header::ETag(etag));
             });
 
-        // TODO: Debug, enabling "accept-ranges: bytes" causes problems with
-        // certain clients when not using the ranges header.
-        //resp.header(header::ACCEPT_RANGES, format!("bytes"));
+        resp.header(header::ACCEPT_RANGES, "bytes");
 
         let mut length = self.md.len();
         let mut offset = 0;
 
-        // check for ranges header
+        // check for range header
         if let Some(ranges) = req.headers().get(header::RANGE) {
             if let Ok(rangesheader) = ranges.to_str() {
                 if let Ok(rangesvec) = HttpRange::parse(rangesheader, length) {
-                    length = rangesvec[0].length - 1;
+                    length = rangesvec[0].length;
                     offset = rangesvec[0].start;
                     resp.header(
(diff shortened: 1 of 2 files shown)


src/fs.rs
```diff
@@ -300,26 +298,21 @@ impl Responder for NamedFile {
...
-                    resp.header(header::RANGE, format!("*/{}", length));
+                    resp.header(header::CONTENT_RANGE, format!("bytes */{}", length));
                     return Ok(resp.status(StatusCode::RANGE_NOT_SATISFIABLE).finish());
                 };
             } else {
                 return Ok(resp.status(StatusCode::BAD_REQUEST).finish());
             };
         };
 
+        resp.header(
```
fafhrd91: do we always need `content-range` header?

ghost:
You're absolutely right. Modified.

src/fs.rs
```diff
@@ -289,37 +289,35 @@ impl Responder for NamedFile {
                 resp.set(header::ETag(etag));
             });
 
-        // TODO: Debug, enabling "accept-ranges: bytes" causes problems with
-        // certain clients when not using the ranges header.
-        //resp.header(header::ACCEPT_RANGES, format!("bytes"));
+        resp.header(header::ACCEPT_RANGES, format!("bytes"));
```
fafhrd91: remove `format!`, `.header()` works with static str

fafhrd91:
did you test ranges in browser?

ghost:
Yes. Works as intended. Depends on the filetype too.

fafhrd91 (approved):

ghost:
Thanks!

Fun fact: Does not work with all the browsers apparently: https://bugzilla.mozilla.org/show_bug.cgi?id=1418430

fafhrd91:
thanks

PR #239, opened 2018-05-23
Add ability to set encoding for exact NamedFile.

max-frai:
Simple api for passing content encoding to httpresponse builder.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -38,6 +38,7 @@ pub struct NamedFile {
     md: Metadata,
     modified: Option<SystemTime>,
     cpu_pool: Option<CpuPool>,
+    encoding: Option<ContentEncoding>,
     only_get: bool,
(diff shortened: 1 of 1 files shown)


src/fs.rs
```diff
@@ -114,6 +117,19 @@ impl NamedFile {
...
+    // Set content encoding for serving this file
+    #[inline]
+    pub fn set_content_encoding(mut self, enc: ContentEncoding) -> Self {
+        self.encoding = Some(enc);
+        self
+    }
+
+    // Get content encoding used for serving this file
+    #[inline]
+    pub fn content_encoding(&self) -> Option<ContentEncoding> {
```
fafhrd91, line 129: I don't think we need to export this function

fafhrd91:

fafhrd91:
thanks

PR #264, opened 2018-05-31
Re-export actix dependency. Closes #260

### LINKED ISSUE #260: Re-export actix

DoumanAsh:
I'll use CI to verify my build as locally i have some troubles to build ring

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -194,7 +194,9 @@ pub use httprequest::HttpRequest;
 pub use httpresponse::HttpResponse;
 pub use json::Json;
 pub use scope::Scope;
-pub use actix;
+pub mod actix {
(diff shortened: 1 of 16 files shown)


src/lib.rs
```diff
@@ -194,6 +194,7 @@ pub use httprequest::HttpRequest;
 pub use httpresponse::HttpResponse;
 pub use json::Json;
 pub use scope::Scope;
+pub use actix;
```
fafhrd91: I think, this should be 

```rust
pub mod actix {
   pub use actix::prelude::*;
}
```

fafhrd91 (changes requested):
lets re-export `actix::prelude::*`

fafhrd91 (approved):

PR #266, opened 2018-06-01
Fix some ResourceHandler docs

killercup:
Re-enables code blocks as doc tests to prevent them failing in the future, and also adds some new code examples with the corresponding methods being used (in addition to the "this is a shortcut to" snippets).


fafhrd91:
thanks!

Issue #263, opened 2018-05-30
Use `&mut self` instead of `&self` for Middleware trait


fafhrd91:
i think we should switch to `&mut self` for Middleware trait, it is safe

joshleeb:
I'd like to give this one a shot 😄

I started playing around with changing `&self` to `&mut self` for the `start` function of the `Middleware` trait. 

It looks the `mws: Rc<Vec<Box<Middleware<S>>>>` value in `PipelineInfo`, `ComposeInfo`, and similar structs will need to be changed to `Rc<RefCell<Vec<Box<Middleware<S>>>>>` or `Rc<UnsafeCell<Vec<Box<Middleware<S>>>>>`. I mention `UnsafeCell` because it's used with `req: UnsafeCell<HttpRequest<S>>`.

I don't have a heap of experience working with `RefCell` (and even less with `UnsafeCell`) so I'm not sure if this is the best way to implement this change. 

@fafhrd91 WDYT? Or is there a better way to do this?

fafhrd91:
let's use `RefCell`. don't forget to update change log.

thanks!

fafhrd91:
fixed by #267

fafhrd91:
@joshleeb what to work on something else?

joshleeb:
Yea! What have you got in mind?

fafhrd91:
1. there is bug in middleware handling #255  2 and 3 points
2. #254  - this one is more involved

joshleeb:
Great 😄 thanks @fafhrd91

PR #267, opened 2018-06-02
Update Middleware Trait to Use `&mut self`

joshleeb:
Ref. #263 - Use `&mut self` instead of `&self` for Middleware trait.

This PR updates the function signatures for the `Middleware` trait, changing `&self` to `&mut self`.


fafhrd91 (approved):
lgtm

fafhrd91:
thanks!

PR #268, opened 2018-06-02
No more missing docs

killercup:
Sets `#![warn(missing_docs)]` and adds missing API docs. These were written without much knowledge of the actix-web internals! Please review carefully!

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -102,9 +102,12 @@ where
     A: Actor<Context = Self>,
 {
     #[inline]
+    /// Create a new HTTP Context from a request and an actor
     pub fn new(req: HttpRequest<S>, actor: A) -> HttpContext<A, S> {
         HttpContext::from_request(req).actor(actor)
     }
+
+    /// Create a new HTTP Context from a request
     pub fn from_request(req: HttpRequest<S>) -> HttpContext<A, S> {
         HttpContext {
             inner: ContextImpl::new(None),
@@ -113,7 +116,9 @@ where
             disconnected: false,
         }
     }
+
     #[inline]
+    /// Set the actor of this context
     pub fn actor(mut self, actor: A) -> HttpContext<A, S> {
         self.inner.set_actor(actor);
         self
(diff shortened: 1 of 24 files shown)


src/context.rs
```diff
@@ -239,12 +244,14 @@ where
     }
 }
 
+/// Consume a future
 pub struct Drain<A> {
```
killercup, line 248: I have no idea what this actually does

src/error.rs
```diff
@@ -521,12 +521,16 @@ impl ResponseError for UriSegmentError {
 /// Errors which can occur when attempting to generate resource uri.
 #[derive(Fail, Debug, PartialEq)]
 pub enum UrlGenerationError {
+    /// Resource not found
     #[fail(display = "Resource not found")]
```
killercup, line 525: Memo to self: Can't failure infer the display text from a doc comment?

fafhrd91:
Oh! Thanks!

src/fs.rs
```diff
@@ -422,15 +422,19 @@ type DirectoryRenderer<S> =
 /// A directory; responds with the generated directory listing.
 #[derive(Debug)]
 pub struct Directory {
+    /// Base directory
     pub base: PathBuf,
+    /// Path of subdirectory to generate listing for
```
killercup, line 427: Not a very good explanation

src/httpcodes.rs
```diff
@@ -5,7 +5,7 @@ use httpresponse::{HttpResponse, HttpResponseBuilder};
 
 macro_rules! STATIC_RESP {
     ($name:ident, $status:expr) => {
-        #[allow(non_snake_case)]
+        #[allow(non_snake_case, missing_docs)]
```
killercup, line 8: macros are evil but you could play with transforming the `$name:ident` to a `$name:expr` (via `$name:tt`) and putting it in a `#[doc = $name]` attr

fafhrd91 (approved):

killercup (commented):
Most of these comments just give a little bit of context, but in the future could really come to shine if they also included examples and links to relevant guides or other APIs

fafhrd91:
i think this is good

PR #270, opened 2018-06-02
Specialize ResponseError for PayloadError

### LINKED ISSUE #257: Shouldn't FormConfig::limit returns 413 instead of 500?
Right now exceeding the limit set by [FormConfig::limit](https://docs.rs/actix-web/0.6.10/actix_web/dev/struct.FormConfig.html#method.limit) makes the request return [500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500). Shouldn't it be better to use [413 Payload Too Large](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) or maybe [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)?

DoumanAsh:
Closes #257

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -294,7 +294,14 @@ impl From<IoError> for PayloadError {
 }
 
 /// `InternalServerError` for `PayloadError`
-impl ResponseError for PayloadError {}
+impl ResponseError for PayloadError {
+    fn error_response(&self) -> HttpResponse {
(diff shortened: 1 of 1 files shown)


fafhrd91 (approved):

fafhrd91:
lgmt

PR #274, opened 2018-06-03
add default value for header User-Agent in requests

mockersf:
According to https://tools.ietf.org/html/rfc7231#section-5.5.3:
> A user agent SHOULD send a User-Agent field in each request unless specifically configured not to do so.

This adds a default `User-Agent` header to requests, with value `Actix-Web`. It can be overridden by the user, and won't be added if the `ClientRequestBuilder` `default_headers` is false

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 
 * Use `&mut self` instead of `&self` for Middleware trait
 
+* Added header `User-Agent: Actix-web/<current_version>` to default headers when building a request
 
 ### Removed
 
--- a/src/client/request.rs
+++ b/src/client/request.rs
@@ -615,7 +615,10 @@ impl ClientRequestBuilder {
                 true
             };
             if !contains {
-                self.header(header::USER_AGENT, "Actix-web");
+                self.header(
+                    header::USER_AGENT,
+                    concat!("Actix-web/", env!("CARGO_PKG_VERSION")),
+                );
             }
         }
 
--- a/src/client/request.rs
+++ b/src/client/request.rs
@@ -506,7 +506,7 @@ impl ClientRequestBuilder {
     }
 
     /// Do not add default request headers.
-    /// By default `Accept-Encoding` header is set.
+    /// By default `Accept-Encoding` and `User-Agent` headers are set.
     pub fn no_default_headers(&mut self) -> &mut Self {
         self.default_headers = false;
         self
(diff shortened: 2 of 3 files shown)


DoumanAsh:
Just to clarify, `SHOULD` doesn't mean it is mandatory, only that it is recommended.
But I would suggest against this change as actix is library first of all, we should not default user's identity as user may wish not to set any

P.s. the most common use case for setting `User-Agent` is to pretend to be a browser.
Other than browsers I doubt that anyone would care about other clients (aside from may be special case of search engine crawlers)

mockersf:
Yes, it's only `SHOULD`, so not mandatory.

It is also often used for statistics and monitoring.

Python's `requests` set it to `python-requests/2.18.1`, `curl` is `curl/7.54.0`.

As you prefer, I can continue to add it manually if you don't want it in the library 😄

DoumanAsh:
it is up to @fafhrd91 i guess

fafhrd91:
I think we should add it. 

@mockersf should we use version of rust and actix?

DoumanAsh:
If we're going to include it by default, should we allow user to disable it?

In some cases client might wish to not be identifiable by library it uses(for example if there are known vulnerabilities)

mockersf:
Version of actix can be added manually in the string but it would increase the work for each release, or retrieved with `env!("CARGO_PKG_VERSION");` but this would mean the `&str` won't be valid for the `'static` lifetime and setting it will require a few allocations, which we can probably live with

Getting the rust version is more complicated, it requires a build script or a dependency on [`rustc_version_runtime`](https://crates.io/crates/rustc_version_runtime) (which itself uses a build script)

@DoumanAsh it can be removed either with [`builder.no_default_headers()`](https://docs.rs/actix-web/0.6.10/actix_web/client/struct.ClientRequestBuilder.html#method.no_default_headers) or by setting another value for this header

DoumanAsh:
Ok that's good.

As for version I think we should follow suit of python's requests and just use actix's version?

fafhrd91:
i think we can read version of the package in build.rs

fafhrd91:
https://doc.rust-lang.org/cargo/reference/environment-variables.html

fafhrd91:
here is how to use it https://doc.rust-lang.org/std/macro.env.html

mockersf:
just found the [`concat!`](https://doc.rust-lang.org/std/macro.concat.html) macro to concat literal strings into another `&'static str`

fafhrd91 (approved):

fafhrd91:
@mockersf could you add entry to changes.md

mockersf:
done!

fafhrd91:
thanks!

PR #288, opened 2018-06-08
Update TechEmpower benchmarks to round 16

memoryruins:
Well done :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ You may consider checking out
 
 ## Benchmarks
 
-* [TechEmpower Framework Benchmark](https://www.techempower.com/benchmarks/#section=data-r15&hw=ph&test=plaintext)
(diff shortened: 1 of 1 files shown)


fafhrd91:
thanks!

pickfire:
Now there is techempower benchmark round 17 but the performance drop slightly, should we update it?

But somehow actix-raw in data-updates section are way faster than the others.

fafhrd91:
Yes, we can update. Db benchmarks are better because now they use async pg driver

pickfire:
@fafhrd91 Nice, now it seems that db are way faster than the others. Do you know what is the reason on the minor performance drop on plain text benchmark?

fafhrd91:
who knows, maybe power fluctuation or solar flare, or may someone stepped on network cord

PR #291, opened 2018-06-09
Add synchronous adaptors for client side

### LINKED ISSUE #269: Sync interface for actix_web::client
- Add `actix_web::client::sync` module
- Should be able to be used within sync code

DoumanAsh:
Closes #269 

Current design requires user to keep `Client` struct alive to get body.
We can change it by moving these methods to access body into `Client` though I don't like such API.

On other hand we could just return error `ClientResponse` when worker thread is down.

I'd prefer to avoid implicit thread

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ time = "0.1"
 encoding = "0.2"
 language-tags = "0.2"
 lazy_static = "1.0"
+crossbeam-channel = "0.2"
 url = { version="1.7", features=["query_encoding"] }
 cookie = { version="0.10", features=["percent-encode"] }
 brotli2 = { version="^0.3.2", optional = true }
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -31,6 +31,7 @@ mod pipeline;
 mod request;
 mod response;
 mod writer;
+pub mod sync;
 
 pub use self::body::{ClientBody, ClientBodyStream};
 pub use self::connector::{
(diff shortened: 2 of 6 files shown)


fafhrd91:
i think we need to start thread with async client

fafhrd91:
it this point i dont want to provide sync api.

jgillich:
Hey so is there no easy way to use the client in a synchronous context (Juniper executor)? I'm using reqwest currently but it comes with lots of dependencies, I'd rather use the built-in actix client.

DoumanAsh:
@jgillich You'd need to elaborate a bit, if you want to use actix-web client, as long as you can spawn its future, you should be fine.
For example if you're on the same thread as your system or some custom arbiter you can spawn future using https://docs.rs/actix/0.7.10/actix/fn.spawn.html

jgillich:
Yes, sorry I'm a bit clueless about futures. What I have is pretty close to the [Juniper example](https://github.com/actix/examples/tree/master/juniper), I use a SyncArbiter to spawn Juniper executor threads. Juniper itself is fully synchronous, and I'd like to use the actix-web client inside a [resolver](https://github.com/actix/examples/blob/master/juniper/src/schema.rs#L32).

DoumanAsh:
@jgillich since it is sync arbiter then you cannot use async... I guess your option would to use `Future::wait` as each handler would be on separate thread, but to be honest the best would be to have async actor, not sure if it is possible with Juniper

ivnsch:
I'm new to actix-web (and still to Rust). Not sure I understand. I'm looking for an equivalent to `reqwest::blocking::Client` (I just want to make requests on the current thread). Should I use `Future::wait`?

PR #293, opened 2018-06-09
Better Content-Type and Content-Disposition handling for static files

axon-q:
* Compute and store Content-Type and Content-Disposition values when `NamedFile` is created
* Use structured Content-Disposition representation
* Add `set_content_type()` and `set_content_disposition()` builder methods to `NamedFile` to allow overriding the defaults
* Add `fs::file_extension_to_mime()` utility function

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -29,6 +29,14 @@ use param::FromParam;
 /// Env variable for default cpu pool size for `StaticFiles`
 const ENV_CPU_POOL_VAR: &str = "ACTIX_FS_POOL";
 
+/// Return the MIME type associated with a filename extension (case-insensitive).
+/// If `ext` is empty or no associated type for the extension was found, returns
+/// the type `application/octet-stream`.
+#[inline]
+pub fn file_extension_to_mime(ext: &str) -> mime::Mime {
+    get_mime_type(ext)
+}
+
 /// A file with an associated name; responds with the Content-Type based on the
 /// file extension.
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


src/fs.rs
```diff
@@ -54,15 +63,48 @@ impl NamedFile {
...
-        let file = File::open(path.as_ref())?;
-        let md = file.metadata()?;
+        use header::{ContentDisposition, DispositionType, DispositionParam};
         let path = path.as_ref().to_path_buf();
+
+        // Get the name of the file and use it to construct default Content-Type
+        // and Content-Disposition values
+        let content_type;
+        let content_disposition;
+        {
```
DoumanAsh: I would suggest to return directly instead of assigning:
```rust
let (content_type, content_disposition) = {
...
}```

File: src/fs.rs [hunk 2]
```diff
@@ -212,23 +275,10 @@ impl Responder for NamedFile {
...
-                    format!(
-                        "{inline_or_attachment}; filename={filename}",
-                        inline_or_attachment = inline_or_attachment,
-                        filename = file_name.to_string_lossy()
-                    ),
-                );
-            });
+
+            resp.set(header::ContentType(self.content_type.clone()));
+            resp.header("Content-Disposition", format!("{}", &self.content_disposition));
```
fafhrd91: maybe to use [HttpResponseBuilder::set()](https://github.com/actix/actix-web/blob/master/src/httpresponse.rs#L296)?

fafhrd91 (approved):
lgtm!

DoumanAsh (approved):
Just minor comment but otherwise good

axon-q:
@fafhrd91 seems it's not possible to use `set` for this header yet

fafhrd91:
@axon-q ok

fafhrd91:
thanks!

Issue #275, opened 2018-06-03
Lines payload stream


fafhrd91:
at the moment we have several method how we can read request payload, 
i.e. `HttpMessage::body()`, `HttpMessage::json()` etc. I think we need to support `readlines`

ghost:
how can I do this. Does this mean that HttpMessage::readlines() should return an iterator over strings

fafhrd91:
`::readline()` should return `Stream<Item=String, Error=ReadlinesError>`

fafhrd91:
implementation could be based on `String` extractor.

ghost:
I see that self is `Stream<Item = Bytes, Error = PayloadError>` so should I call map or fold or something like that to accumulate bytes until newline character than make a string and return it?

fafhrd91:
yes. just add config about line length. also you should handle encoding properly.

ghost:
ok. I am tired and this seems a little hard I'll try it tomorrow

ghost:
```rust
struct ReadLines<T>
where
    Self: Stream<Item = Bytes, Error = PayloadError> + Sized,
{
    from: T,
    buff: String,
}

impl<T> ReadLines<T>
where
    Self: Stream<Item = Bytes, Error = PayloadError> + Sized,
{
    fn new(from: T) -> Self {
        ReadLines {
            from,
            buff: String::new(),
        }
    }
}

impl<T> Stream for ReadLines<T>
where
    Self: Stream<Item = Bytes, Error = PayloadError> + Sized,
{
    type Item = String;
    type Error = ReadlinesError;
    
    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
        // do stuff
    }
}
```
is this the right way?

fafhrd91:
Mostly. 

1. I don’t think you need T
2. take stream ownership
3. You can not really use String as buffer because you can get bytes chunk splinted in the middle of Unicode codepoint. Use BytesMut instead.

ghost:
I got this now
```
struct ReadLines {
    from: ???,
    buff: BytesMut,
}

impl ReadLines {
    fn new(from: ???) -> Self {
        ReadLines {
            from,
            buff: BytesMut::new(),
        }
    }
}

impl Stream for ReadLines {
    type Item = String;
    type Error = ReadlinesError;
    
    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
        // do stuff
    }
}
```
how can I take ownership of the stream?

[...]

PR #312, opened 2018-06-13
Add HttpMessage::readlines()

ghost:
#275

### DIFF
--- a/src/httpmessage.rs
+++ b/src/httpmessage.rs
@@ -13,6 +13,7 @@ use std::str;
 
 use error::{
     ContentTypeError, HttpRangeError, ParseError, PayloadError, UrlencodedError,
+    Error, ErrorBadRequest
 };
 use header::Header;
 use json::JsonBody;
@@ -276,7 +277,7 @@ where
     T: HttpMessage + Stream<Item = Bytes, Error = PayloadError> + 'static,
 {
     req: T,
-    buff: Vec<u8>,
+    buff: BytesMut,
     limit: usize,
 }
 
@@ -288,12 +289,12 @@ where
     fn new(req: T) -> Self {
         Readlines {
             req,
-            buff: Vec::with_capacity(262_144),
+            buff: BytesMut::with_capacity(262_144),
             limit: 262_144,
         }
     }
     
-    /// Change max size of payload. By default max size is 256Kb
+    /// Change max line size. By default max size is 256Kb
     pub fn limit(mut self, limit: usize) -> Self {
         self.limit = limit;
         self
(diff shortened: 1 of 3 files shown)


src/httpmessage.rs
```diff
@@ -260,6 +260,101 @@ pub trait HttpMessage {
...
+{
+    type Item = String;
+    type Error = ReadlinesError;
+    
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        match self.req.poll() {
+            Ok(Async::Ready(Some(bytes))) => {
+                for b in bytes.iter() {
+                    if *b == '\n' as u8 {
+                        self.buff.push(*b);
```
fafhrd91: This is slooow 
Search for index then do extend_from_slice
Also you can try to use .split_to() and then extend

fafhrd91 (commented):
You also should check request charset. Check extractor for String

src/httpmessage.rs
ghost: Ok. fixing it

ghost: what should I do if encoding isn't utf8 should I return Result::Err(EncodingError) from ::new()
and ::readlines() method

ghost:
```Rust
fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
    // check encoding
    if !self.checked_encoding {
        let encoding = self.req.encoding()?;
        let enc: *const Encoding = encoding as *const Encoding;
        if enc != UTF_8 {
            return Err(ReadlinesError::EncodingError)
        }
        self.checked_encoding = true;
    }
    
    // check if there is a newline in the buffer
    let mut found: Option<usize> = None;
    for (ind, b) in self.buff.iter().enumerate() {
        if *b == '\n' as u8 {
            found = Some(ind);
            break;
        }
    }
    if let Some(ind) = found {
        // check if line is longer than limit
        if ind+1 > self.limit {
            return Err(ReadlinesError::LimitOverflow);
        }
        let line = str::from_utf8(&self.buff.split_to(ind+1))?.to_owned();
        return Ok(Async::Ready(Some(line)));
    }
    // poll req for more bytes
    match self.req.poll() {
        Ok(Async::Ready(Some(mut bytes))) => {
            // check if there is a newline in bytes
            let mut found: Option<usize> = None;
            for (ind, b) in bytes.iter().enumerate() {
                if *b == '\n' as u8 {
                    found = Some(ind);
                    break;
                }
            }
            if let Some(ind) = found {
                // check if line is longer than limit
                if ind+1 > self.limit {
                    return Err(ReadlinesError::LimitOverflow);
                }
                let line = str::from_utf8(&bytes.split_to(ind+1))?.to_owned();
                // extend buffer with rest of the bytes;
                self.buff.extend_from_slice(&bytes);
                return Ok(Async::Ready(Some(line)));
            }
            self.buff.extend_from_slice(&bytes);
            Ok(Async::NotReady)
        },
        Ok(Async::NotReady) => Ok(Async::NotReady),
        Ok(Async::Ready(None)) => {
            if self.buff.len() == 0 {
                return Ok(Async::Ready(None));
            }
            if self.buff.len() > self.limit {
                return Err(ReadlinesError::LimitOverflow);
            }
            let line = str::from_utf8(&self.buff)?.to_owned();
            self.buff.clear();
            return Ok(Async::Ready(Some(line)))
        },
        Err(e) => Err(ReadlinesError::from(e)),
    }
}
```
is this good?

fafhrd91: You need to decide string

https://github.com/actix/actix-web/blob/master/src/extractor.rs#L439

fafhrd91: You should commit your code, it is hard comment on it

You should decide encoding only when you split new string 
So first poll, then search new line in chunk, then extend, then split, then decide encod Ng

ghost: I search for a newline in the buffer first because there might be multiple newlines in chunk.
So I return a line if there is a newline in chunk. And extend the buffer with rest of it. Then I check
for newlines in the buffer new time. Is there a better way to do this.

fafhrd91: Ok, good.

Move ReadlinesError to errors module

ghost:
I can updates CHANGES.md and add test if this is good.

src/httpmessage.rs
```diff
@@ -260,6 +261,137 @@ pub trait HttpMessage {
...
+    T: HttpMessage + Stream<Item = Bytes, Error = PayloadError> + 'static,
+{
+    type Item = String;
+    type Error = ReadlinesError;
+    
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        let encoding = self.req.encoding()?;
+        // check if there is a newline in the buffer
+        let mut found: Option<usize> = None;
+        for (ind, b) in self.buff.iter().enumerate() {
```
fafhrd91: Last comment :)

You search buff for every poll even if you already search it before. I’d add attr for indication that buff does not contain new lines.

fafhrd91 (approved):
Please address my last comment, add test case and then you can merge

src/httpmessage.rs
ghost: ok. more comments are good actually I have never written something like this before. thank you for your patience.

ghost:
is this test sufficient?

PR #313, opened 2018-06-13
EtagHasher middleware

axon-q:
Hash-based ETag middleware implementation.


fafhrd91:
Seems you do not calculate hash of the response body?

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+            Ok(middleware::Response::Done(hr)) => hr,
+            _ => panic!(),
+        }
+    }
+
+    #[test]
+    fn test_default_create_etag() {
+        let mut eh = EtagHasher::new(DefaultHasher, DefaultFilter);
+        let mut req = TestRequest::default().finish();
+        let res = HttpResponse::Ok().body("test");
```
DoumanAsh: I'd suggest to to move `"test"` literal to constant like `TEST_ETAG`

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+pub struct DefaultFilter;
+impl<S> RequestFilter<S> for DefaultFilter {
+    fn filter(&self, req: &HttpRequest<S>, res: &HttpResponse) -> bool {
+        use http::{Method, StatusCode};
+        (*req.method() == Method::GET || *req.method() == Method::HEAD)
+            && res.status() == StatusCode::OK
+    }
+}
+
+/// The middleware struct. Contains a Hasher to compute ETag values for byte
```
DoumanAsh: Please put as a first line more descriptive name.
At least `ETag middleware` :)

Also please add that it works only on responses that it calculates ETag only for Responses with `Body::Binary`

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        let e = if let Body::Binary(b) = res.body() {
+            Some(EntityTag::strong(self.hasher.hash(b.as_ref())))
+        } else {
+            None
+        };
+
+        if let Some(etag) = e {
+            if !none_match(&etag, req) {
```
DoumanAsh: it seems like there is no need to return Option into `e`, just move this If section above where you check for `Body::Binary`?

axon-q:
@fafhrd91 The response body hash is calculated by this line:

```rust
            Some(EntityTag::strong(self.hasher.hash(b.as_ref())))
```

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// Defaults
+/// Computes an ETag value from a byte slice using a default cryptographic hash
+/// function.
+pub struct DefaultHasher;
+impl Hasher for DefaultHasher {
+    fn hash(&self, input: &[u8]) -> String {
+        use sha1;
+        let mut h = sha1::Sha1::new();
```
DoumanAsh: Wouldn't it be it better to just create `Sha1` once and store it hasher?
You'd just need to reset it on each call instead of creating always which would be cheaper

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/// Returns `true` when the request method is `GET` or `HEAD` and the
+/// original response status is `200 OK`, and `false` otherwise.
+pub struct DefaultFilter;
+impl<S> RequestFilter<S> for DefaultFilter {
+    fn filter(&self, req: &HttpRequest<S>, res: &HttpResponse) -> bool {
+        use http::{Method, StatusCode};
+        (*req.method() == Method::GET || *req.method() == Method::HEAD)
+            && res.status() == StatusCode::OK
```
DoumanAsh: Should it maybe ignore right here responses with empty bodies?
Or for that all bodies that cannot be calculated

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+        &mut self, req: &mut HttpRequest<S>, mut res: HttpResponse,
+    ) -> Result<middleware::Response> {
+        use header;
+        use Body;
+
+        if !self.filter.filter(req, &res) {
+            return Ok(middleware::Response::Done(res));
+        }
+
+        let e = if let Body::Binary(b) = res.body() {
```
DoumanAsh: Since `Body` enum, I'd suggest to use match, but it is up to you

src/middleware/etaghasher.rs
```diff
@@ -0,0 +1,318 @@
...
+        self(input)
+    }
+}
+impl<S, F: Fn(&HttpRequest<S>, &HttpResponse) -> bool> RequestFilter<S> for F {
+    fn filter(&self, req: &HttpRequest<S>, res: &HttpResponse) -> bool {
+        self(req, res)
+    }
+}
+
+// Defaults
```
DoumanAsh, line 111: This is idea I got looking at your code.

Wouldn't it be nice to create `ETag` trait and implement it for types that can be hashed?

For example user would benefit from being able to do it on its own for Streaming bodies (in case of files you could use even metadata)

But this is optional improvement if you're feeling up to it.
Feel free to ignore it

DoumanAsh (commented):
I input some comments, mostly minor

src/middleware/etaghasher.rs
axon-q: Yes!

src/middleware/etaghasher.rs
axon-q: Did this along with other doc improvements.

src/middleware/etaghasher.rs
axon-q: The code is structured this way so that the borrow taken when getting the body is dropped before we mutate the response.

src/middleware/etaghasher.rs
axon-q: Yes, done!

src/middleware/etaghasher.rs
axon-q: Yes, it's best to generate ETags directly for external resources. This is more for the use-case of dynamically generated responses. Added some comments in the docs about this.

axon-q:
@DoumanAsh Thanks for the detailed review, much appreciated!

src/middleware/etaghasher.rs
DoumanAsh: Ah I see, ok

DoumanAsh (approved):

DoumanAsh:
All good, but can you please also add tests to verify that default ETag hasher ignores non-GET requests and requests without Binary Body?

fafhrd91:
I am not sure if it should be part of main package.

DoumanAsh:
@fafhrd91 It doesn't seem to introduce new dependencies so I don't see much problem. But final decision is up to you.

axon-q:
@fafhrd91 The motivation for this is that it makes it possible for, e.g., a REST API to avoid sending data if the requester has cached it. This can save a lot of network overhead, and is quite a common middleware function for HTTP-based APIs. Of course it's up to you whether you want actix-web to support it. :)

fafhrd91:
@axon-q i think this middleware is too specific, and covers only basic case. let's move it to separate repo. you can create new repo in actix organization

axon-q:
@fafhrd91 @DoumanAsh After thinking about this more, I came up with a different design that should be better. I'll open a new PR so you can have a look.

PR #314, opened 2018-06-13
remove duplication of `App::with_state` in `App::new`

DJMcNab:
I don't think that this is a testable change, as the output of `App::new()` is entirely identical.

I tried moving `App::new()` into the other `impl` block, but for some reason the type checker didn't like the `Default::default` impl.

As this is such a trivial internal change, I haven't added a changelog entry (unless you think it is necessary, then I'll willingly add one).

Also, there are a couple of spelling mistakes in [CHANGES.md](https://github.com/actix/actix-web/blob/c8528e8920f35ada090a90c7c000db5ede237fcb/CHANGES.md):
line 83: `traling` should be `trailing`.
line 444: `Gracefull` should be `Graceful`.

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -192,20 +192,7 @@ impl App<()> {
     /// Create application with empty state. Application can
     /// be configured with a builder-like pattern.
     pub fn new() -> App<()> {
-        App {
-            parts: Some(ApplicationParts {
-                state: (),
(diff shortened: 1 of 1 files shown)


fafhrd91 (approved):

fafhrd91:
Thanks!

PR #324, opened 2018-06-18
ETag middleware (new version)

axon-q:
New design for ETag middleware. Supersedes #313.

This adds an `etag` module that exposes two middleware structs that can be used together or separately:

* `etag::Generator` is used to create ETags for `HttpResponse`s at the middleware level. It can be constructed from anything that implements the `etag::Generate` trait:
```rust
pub trait Generate<S> {
    fn generate(&mut self, req: &HttpRequest<S>, res: &HttpResponse) -> Option<EntityTag>;
}
```
The user can implement this trait to use any information from the `HttpRequest` and `HttpResponse` to produce an ETag. If `generate()` returns `None` or an ETag is already present, the response will not be changed. There is a `DefaultGenerator` that computes a hash-based ETag from the response body bytes.
* `etag::NotModified` simply checks the `If-None-Match` header of a request against the ETag of a response, and returns a `304 Not Modified` if there's a match.

This separation allows users to take advantage of `Not Modified` checking even if they implement ETags outside of middleware. It also lets them generate ETags in middleware without the `Not Modified` checking if desired.

@fafhrd91 This is a very simple and general design that's not tied to hashing or specific response body types. What do you think?

### DIFF
--- a/src/middleware/etag.rs
+++ b/src/middleware/etag.rs
@@ -1,0 +1,152 @@
+use error::Result;
+use header::{self, EntityTag};
+use httprequest::HttpRequest;
+use httpresponse::HttpResponse;
+use middleware;
+
+use std::marker::PhantomData;
+
+/// Trait for generating ETags from `(request, response)` pairs.
+pub trait Generate<S> {
+    fn generate(
+        &mut self, req: &HttpRequest<S>, res: &HttpResponse,
+    ) -> Option<EntityTag>;
+}
+
+/// The default generator computes an ETag as a cryptographic hash of the bytes of the
+/// response body.
+pub struct DefaultGenerator {
+    hashstate: ::sha1::Sha1,
+}
+impl DefaultGenerator {
+    /// Create a new instance.
+    pub fn new() -> Self {
+        DefaultGenerator {
+            hashstate: ::sha1::Sha1::new(),
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


DoumanAsh:
I took a quick look and it seems to be better as now it is less opinioned.
I'm not sure about `ETag` generation middleware, but `NotModified` middleware would be useful to have

src/middleware/etag.rs
```diff
@@ -0,0 +1,152 @@
...
+
+// Returns true if `req` doesn't have an `If-None-Match` header matching `req`.
+#[inline]
+fn none_match<S>(etag: &EntityTag, req: &HttpRequest<S>) -> bool {
+    use header::IfNoneMatch;
+    use httpmessage::HttpMessage;
+    match req.get_header::<IfNoneMatch>() {
+        Some(IfNoneMatch::Items(ref items)) => {
+            for item in items {
+                if item.weak_eq(etag) {
```
DoumanAsh, line 143: I could suggest `weak_eq` and `strong_eq` should be customization point for this middleware.

You could make trait parameter `EtagMatch` that would describe how to match etag with method `EtagMatch::if_none_match`
Then you could implement empty structs that implement each method with strong and weak match.
The empty struct could be trait parameter of your Middleware

fafhrd91:
i dont think this should be part of main crate, it is better to move this code to separate crate.

PR #325, opened 2018-06-18
Remove special casing for empty cookie list

sugar700:
The cookie crate already does that, so no need for actix-web to do it.

To be specific, the CookieJar looks like this.

```rust
#[derive(Default, Debug, Clone)]
pub struct CookieJar {
    original_cookies: HashSet<DeltaCookie>,
    delta_cookies: HashSet<DeltaCookie>,
}
```

And its default method calls `default` method. This means `HashSet::new` is called twice, and the languages guarantees no allocations for `HashSet::new()` in its documentation: "The hash set is initially created with a capacity of 0, so it will not allocate until it is first inserted into."

### DIFF
--- a/src/client/request.rs
+++ b/src/client/request.rs
@@ -131,7 +131,7 @@ impl ClientRequest {
         ClientRequestBuilder {
             request: Some(ClientRequest::default()),
             err: None,
-            cookies: None,
+            cookies: CookieJar::new(),
             default_headers: true,
         }
     }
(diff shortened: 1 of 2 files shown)


sugar700:
Although I suppose there is still a cost of generating a "random" state for `HashSet`. The cost appears to be minimal when it's not the first instantation of `HashSet`/`HashMap` in a given thread, requiring pretty much accessing a thread local and a non-atomic number increment. A random number generator is not involved when a random number was already generated in a given thread. Considering other parts of actix-web make use of `HashMap`, it's very likely the cost was already paid.

fafhrd91:
i am not sure about reason for this change. why it is better?

sugar700:
It's less code, pretty much that.

fafhrd91:
With this patch we would to construct cookie jar for each response even if we don’t use cookies

sugar700:
Okay, the cost isn't big (about 30 nanoseconds), slowing down code like `get("http://example.com").finish()` by 1%, but it exists.

PR #335, opened 2018-06-20
remove unnecessary use of unsafe in read_from_io

gnzlbg:


fafhrd91 (approved):
Good! But we need to bench this

fafhrd91:
I sent you invitation, you should be able to merge yourself

gnzlbg:
I have some time to upgrade the benchmarks tomorrow, so will send a PR there to update them, and then use the one you recommended on gitter to bench this. I'll update this with the results before merging.

fafhrd91:
Could you also modify this code?

https://github.com/actix/actix-web/blob/master/src/server/h1.rs#L439

fafhrd91:
@gnzlbg there is no reason for this function anymore, could you just inline code?

gnzlbg:
> Could you also modify this code?

Yes, I'll give it a shot. Will update this PR and ping you when its done :)

fafhrd91:
i played a bit more with this patch. this patch changes semantics and i think there is performance penalty.

i am going to revert it. at this point i am not sure it is worse of removing this particular unsafe

gnzlbg:
How did it change semantics? When I ran it all tests passed, so maybe I could add a test for the old semantics.

fafhrd91:
h1 parsers reads all data that available in socket. That is the reason for loop. Problem with this code is that it is very os specific

Issue #323, opened 2018-06-18
Implementation of `StaticFile` handler for `index_file` behavior may be incorrect.


garyhai:
For the nested subfolder without 'index.html' set by index_file, redirect url has duplicated 'tail'. Because the 

Code:
```
app.handler(
                "/",
                fs::StaticFiles::new(site.home.as_str())
                    .index_file("index.html"),
            )
```
Cases:
```
$ curl -ki https://localhost:8443/test
HTTP/2 302 
location: /testtest/index.html
date: Mon, 18 Jun 2018 06:55:48 GMT
content-length: 0

$ curl -ki https://localhost:8443/test/
HTTP/2 302 
location: /test/test/index.html
date: Mon, 18 Jun 2018 06:56:19 GMT
content-length: 0
```

When invoke HTTP GET method for "http://localhost/test", get the redirect URL as "http://localhost/testtest/index.html'.
When invoke HTTP GET method for "http://localhost/test/", get the redirect URL as "http://localhost/test/test/index.html'.

Fix in 'src/fs.rs'

```
            if path.is_dir() {
                if let Some(ref redir_index) = self.index {
                    // TODO: Don't redirect, just return the index content.
                    // TODO: It'd be nice if there were a good usable URL manipulation
                    // library
                    let mut new_path: String = req.path().to_owned();
                    // for el in relpath.iter() {
                    //     new_path.push_str(&el.to_string_lossy());
                    //     new_path.push('/');
                    // }
                    if !new_path.ends_with('/') {
                        new_path.push('/');
                    }
```

And the unit test cases 'test_redirect_to_index' and 'test_redirect_to_index_nested' should be fixed.

fafhrd91:
Would you provide PR?

garyhai:
Sorry. I am a newbie of rust programming. Actually, the code I pasted above is not the fix for resource pattern "{tail:.*}".

garyhai:
Thanks. My website works fine now.

PR #336, opened 2018-06-20
Fix duplicate tail of StaticFiles with index_file.

garyhai:
And fix the unit test cases by add request path. 

#323


garyhai:
Moved to branch '0.6'

Issue #254, opened 2018-05-28
Feature: Propagation of default_ressource() to underlying scopes.


lowlevl:
With this code,
```rust
fn v1<S: 'static>(scope: Scope<S>) -> Scope<S> {
  /* ... */
}

fn special<S: 'static>(scope: Scope<S>) -> Scope<S> {
  scope.default_resource(|r|
    r.f(|_|HttpResponse::Ok().body("Endpoint not found on /special"))
  )
  /* ... */
}

fn main() {
    App::new()
        .scope("/v1", v1) // Import the v1 API.
        .scope("/special", special) // Import special endpoints
        .default_resource(|r|
            /* Display a proper 404 json response */
            r.f(|_| HttpResponse::NotFound().body("Endpoint not found"))
        )
}
```
it works currently like that:

Request URI | Response Code | Response Body
--- | --- | ---
`/non-existent` | 404 | `Endpoint not found`
`/v1/non-existent` | 404 | ` `
`/special/non-existent` | 200 | `Endpoint not found on /special`

where I think this should bne like that:

Request URI | Response Code | Response Body
--- | --- | ---
`/non-existent` | 404 | `Endpoint not found`
`/v1/non-existent` | 404 | `Endpoint not found`
`/special/non-existent` | 200 | `Endpoint not found on /special`

with the root `default_resource` affecting the underlying `Scope` unless the `Scope` has his very `default_resource`, affecting each underlying scopes.

What do you think ? @fafhrd91

PR #339, opened 2018-06-21
Propagate scope default resource

joshleeb:
Propagate the application's default resources to any underlying scopes that haven't set their own default resources.

Closes: #254

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -29,7 +29,7 @@ pub struct HttpApplication<S = ()> {
 #[doc(hidden)]
 pub struct Inner<S> {
     prefix: usize,
-    default: ResourceHandler<S>,
+    default: Rc<RefCell<ResourceHandler<S>>>,
     encoding: ContentEncoding,
     resources: Vec<ResourceHandler<S>>,
     handlers: Vec<PrefixHandlerType<S>>,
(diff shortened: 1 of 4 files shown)


fafhrd91 (approved):
Thanks! Looks good.

Could you add entry to change log?

src/application.rs
```diff
@@ -29,7 +29,7 @@ pub struct HttpApplication<S = ()> {
 #[doc(hidden)]
 pub struct Inner<S> {
     prefix: usize,
-    default: ResourceHandler<S>,
+    default: Rc<RefCell<ResourceHandler<S>>>,
```
DoumanAsh, line 32: What is the reason for RefCell here and not Cell?
Is it possible that we can attempt to borrow &mut while there is already borrow?

DoumanAsh: On side note I wonder if instead of borrowing we could just clone the default resource to underlying scopes each time we create them?

fafhrd91: Clone would be unexpected. Problem is Handler trait uses mut self. Maybe we should remove it from Handler. Then we can remove all RefCells

DoumanAsh: Ah... then maybe yes it would be better to not make it `&mut`
In first place it should not be mutable during routing so maybe we should first make it immutable and then go without `RefCell` especially since it implies runtime cost

joshleeb:
> Could you add entry to change log?

@fafhrd91 updated

fafhrd91:
Thanks!

PR #346, opened 2018-06-25
set length of vector to max_bytes (closes #345)

### LINKED ISSUE #345: Error occurred during request handling: unexpected end of file
Here is simplified example which reproduces this error
https://github.com/max-frai/actix-issue

Async handler which returns NamedFile results in error:
`ERROR 2018-06-25T11:59:41Z: actix_web::pipeline: Error occurred during request handling: unexpected end of file ERROR 2018-06-25T11:59:41Z: actix_web::server::h1: Unhandled error: unexpected end of file`

ousado:


src/fs.rs
```diff
@@ -440,6 +440,8 @@ impl Stream for ChunkedReadFile {
                 let max_bytes: usize;
                 max_bytes = cmp::min(size.saturating_sub(counter), 65_536) as usize;
                 let mut buf = Vec::with_capacity(max_bytes);
+                // safe because memory is initialized/overwritten immediately
+                unsafe { buf.set_len(max_bytes); }
```
killercup, line 444: This sets the length of the buffer to `max_bytes`, but further down writes `nbytes` and `nbytes ≤ max_bytes`. So, in `Bytes::from(buf)` you return a buffer that may contain `max_bytes - nbytes` uninitialized bytes. Am I reading this correctly? (I think adding `.set_len(nbytes)` would fix this.)

ousado: You're right. I doubt this ever occurs in blocking file IO when not trying to read past the end of the file, but setting the length is the safe alternative anyway.

killercup: You could add `debug_assert!(nbytes <= max_bytes)` if you really wanted to make sure without any cost in release mode but it'd only test that `std::fs` is implemented correctly…

killercup (approved):
lgtm (but there might be a nicer option to write this without two unsafe blocks)

src/fs.rs
DoumanAsh: I think it would make a little difference to just initialize vec like
```rust
let mut buf = vec![0; max_bytes]
```
?

Though personally I don't see an issue, I wonder why you don't want just zero initialize it?

By the way I think in this case we could use `BytesMut` directly and then freeze it into `Bytes` directly if you're up for improvement of this code

ousado: The main reason I used set_len is that (I assume) this was the original intent of the code, with minimal changes.

I see no point in initializing to zero here, that's unnecessarily polluting the cache and using up memory bandwidth in a place where that might actually make a difference.

I considered using truncate instead of the second set_len, but after looking at [its O(n) implementation](https://doc.rust-lang.org/src/alloc/vec.rs.html#695-706) I decided against that. 

I'm not opposed to using BytesMut, but unless I missed something in the API, we can't avoid the unsafe set_len there, either (please correct me if I'm wrong).

DoumanAsh: Well I'm fine with it, btw `BytesMut` is mostly to avoid need for conversion from Vec to Bytes but I see your reasoning and it is fine by me

DoumanAsh (approved):

DoumanAsh:
Thanks

fafhrd91:
next time we need test for change

PR #355, opened 2018-06-30
Escape special characters in fs/directory_listing (i.e. the directory index page)

Gowee:
File path is escaped with *percent_encoding* (in case of characters like `\n` and `>`) and file name is escaped with *htmlescape* (for html entities such as `>`, `"` and `"`.


src/fs.rs
```diff
@@ -514,14 +519,14 @@ fn directory_listing<S>(
                         body,
                         "<li><a href=\"{}\">{}/</a></li>",
                         file_url,
-                        entry.file_name().to_string_lossy()
+                        file_name
```
DoumanAsh, line 522: I think we don't really need to encode file names in directory listing itself?

Gowee: Hi!
Supposing there is a file of which the name is `<p>a.txt` (valid on Linux, at least), then the body of the directory listing is `<body><h1>Index of /dir</h1><ul><li><a href="/dir/<p>a.txt"><p>a.txt</a></li></ul></body>`. 
And the browser will treat the `<p>` as a HTML element. As a result, the webpage is (the browser automatically complements with a closing `</p>`):
image
The correct one should be `<body><h1>Index of /dir</h1><ul><li><a href="/dir/%3Cp%3Ea.txt">&lt;p&gt;a.txt</a></li></ul></body>` so that the `<` and `>` would not be treated as the indication of HTML elements. Result:
image.

DoumanAsh: Ah I see, I didn't know you can create such file on Linux.
While it is certainly unlikely, it would be good if special characters would get in.

Is your screenshot from how it looks with your commit?

Gowee: Yes. I tested my commit with [https://github.com/svenstaro/miniserve/](https://github.com/svenstaro/miniserve/) and took that screenshot.

DoumanAsh (approved):

DoumanAsh:
Thank you for noticing this edgy case

Issue #286, opened 2018-06-07
 Directory custom 404 Not Found response


uncotion:
Hello,
When a file is not available in StaticFiles actix-web automatically shows 404 error without body, i want to show custom error but default_handler not working.
```
            .handler(
                "/static",
                actix_web::fs::StaticFiles::new("./static/")
                    .default_handler(p404)
                    .index_file("index.html"),
            )

```

```
fn p404(req: actix_web::HttpRequest) -> Result<actix_web::HttpResponse, actix_web::Error> {
    println!("{:?}lll", req);
    Ok(actix_web::HttpResponse::NotFound()
        .content_type("text/plain")
        .body("Not Found"))
}
```

vharavy:
I have tried the code that you provided and everything works fine: I clearly see the output in the standard out (including the `lll` that is printed after the request (in `p404`) as well as custom body `"Not Found"` in your case (I have modified it to make sure).

Looks like this has been solved already.

DJMcNab:
I can reproduce the bug that if a file is not present, a 404 page is given. The bug appears to be in:

https://github.com/actix/actix-web/blob/master/src/fs.rs#L697-L701

It appears that if the file cannot open, and error is returned, which I imagine will create an internal server error. The [`std::io::ErrorKind::NotFound`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.NotFound) variant should be handled in order to manage this, by calling `io::Error::kind` on the returned error when `NamedFile::open` is called. I don't have time to implement this myself, but it should be a trivial fix.

@VHaravy, if the static directory doesn't exist, then the default handler is used as it should be. I suspect that is what you were experiencing.

Maybe 'good first issue' would be suitable?

fafhrd91:
@DoumanAsh could you review this issue?

DoumanAsh:
@fafhrd91 Errors  are handled on `0.7` by default handler but not in `1.0`

fafhrd91:
open file error calls default handle in master

PR #357, opened 2018-07-01
Use StaticFile default handler when file is inaccessible

tessereth:
I also ran into #286 and I thought I'd have a go at fixing it. It seems that [canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) returns an error if the file doesn't exist. And `NamedFile::open` will return an error if you don't have permission to read the file. In both cases, the error was being propagated from `StaticFiles.handle()` rather than using the default handler.

I'm still very new to rust, let alone this framework, so I don't know if this is a _good_ way to solve this problem. So if you have any suggestions I'm all ears. I also don't know how to test the `NamedFile::open` error path in a platform-independent way so suggestions there are also welcome.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -629,6 +629,9 @@ impl From<UrlParseError> for UrlGenerationError {
 /// Errors which can occur when serving static files.
 #[derive(Fail, Debug, PartialEq)]
 pub enum StaticFileError {
+    /// Path is not a directory
+    #[fail(display = "Path is not a directory")]
+    IsNotDirectory,
     /// Cannot render directory
     #[fail(display = "Cannot render directory")]
     IsDirectory,
--- a/src/error.rs
+++ b/src/error.rs
@@ -626,6 +626,21 @@ impl From<UrlParseError> for UrlGenerationError {
     }
 }
 
+/// Errors which can occur when serving static files.
+#[derive(Fail, Debug, PartialEq)]
+pub enum StaticFileError {
+    /// Cannot render directory
+    #[fail(display = "Cannot render directory")]
+    IsDirectory,
+}
+
+/// Return `NotFound` for `StaticFileError`
+impl ResponseError for StaticFileError {
+    fn error_response(&self) -> HttpResponse {
+        HttpResponse::new(StatusCode::NOT_FOUND)
+    }
+}
+
 /// Helper type that can wrap any error and generate custom response.
 ///
 /// In following example any `io::Error` will be converted into "BAD REQUEST"
--- a/src/error.rs
+++ b/src/error.rs
@@ -630,10 +630,10 @@ impl From<UrlParseError> for UrlGenerationError {
 #[derive(Fail, Debug, PartialEq)]
 pub enum StaticFileError {
     /// Path is not a directory
-    #[fail(display = "Path is not a directory")]
+    #[fail(display = "Path is not a directory. Unable to serve static files")]
     IsNotDirectory,
     /// Cannot render directory
-    #[fail(display = "Cannot render directory")]
+    #[fail(display = "Unable to render directory without index file")]
     IsDirectory,
 }
 
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -555,7 +555,7 @@ fn directory_listing<S>(
 ///
 /// fn main() {
 ///     let app = App::new()
-///         .handler("/static", fs::StaticFiles::new("."))
+///         .handler("/static", fs::StaticFiles::new(".").unwrap())
 ///         .finish();
 /// }
 /// ```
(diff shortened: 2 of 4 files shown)


tessereth:
After thinking about it some more, here's version 2. I renamed `handle()` to `try_handle()` so I could make `handle()` catch all errors and call the default handler. This way the default handler will get called on every error path and we can easily log the error messages (which is I found convenient for debugging purposes). I believe this also removes the need for the `accessible` flag so I got rid of that too.

Let me know what you think!

fafhrd91:
sorry for delay. could you rebase against master?

src/fs.rs
```diff
@@ -657,54 +653,51 @@ impl<S: 'static> StaticFiles<S> {
         self.default = Box::new(WrapHandler::new(handler));
         self
     }
+
+    fn try_handle(&self, req: &HttpRequest<S>) -> Result<AsyncResult<HttpResponse>, Error> {
+        let tail: String = req.match_info().query("tail")?;
+        let relpath = PathBuf::from_param(tail.trim_left_matches('/'))?;
+
+        // full filepath
+        let path= self.directory.join(&relpath).canonicalize()?;
```
DoumanAsh: space before `=`

src/fs.rs
```diff
@@ -594,24 +593,21 @@ impl<S: 'static> StaticFiles<S> {
...
         let dir = dir.into();
 
-        let (dir, access) = match dir.canonicalize() {
+        let dir = match dir.canonicalize() {
             Ok(dir) => {
-                if dir.is_dir() {
-                    (dir, true)
-                } else {
+                if !dir.is_dir() {
                     warn!("Is not directory `{:?}`", dir);
```
DoumanAsh: Now I think about it, maybe we should refactor it to return error to user instead of keeping it as it is

@fafhrd91 You wouldn't mind to change this API?

fafhrd91: @DoumanAsh sure, api changes are fine

tessereth:
Rebased and missing whitespace added

src/fs.rs
tessereth: Do you want me to do that? Or should we leave this PR as-is?

DoumanAsh: @tessereth If it is not difficult then please, otherwise I can do it myself

src/fs.rs
```diff
@@ -652,56 +639,51 @@ impl<S: 'static> StaticFiles<S> {
...
+    }
 }
 
 impl<S: 'static> Handler<S> for StaticFiles<S> {
     type Result = Result<AsyncResult<HttpResponse>, Error>;
 
     fn handle(&self, req: &HttpRequest<S>) -> Self::Result {
-        if !self.accessible {
+        self.try_handle(req).or_else(|e| {
+            debug!("StaticFiles: Failed to handle {}: {}", req.path(), e);
```
DoumanAsh, line 684: I'm not sure should we log each failure to handle request?

There potentially can be lots of ill-formed requests, the only suspicious case I'd suggest to cover by such trace is `canonicalize()` but other variants such as non-existing file or `StaticFileError::IsDirectory` is not really a failure, but expected outcome when user requests unknown resource.

src/error.rs
```diff
@@ -626,6 +626,24 @@ impl From<UrlParseError> for UrlGenerationError {
...
 }
 
+/// Errors which can occur when serving static files.
+#[derive(Fail, Debug, PartialEq)]
+pub enum StaticFileError {
+    /// Path is not a directory
+    #[fail(display = "Path is not a directory")]
+    IsNotDirectory,
+    /// Cannot render directory
+    #[fail(display = "Cannot render directory")]
```
DoumanAsh: Please expand a bit error description: `Unable to render directory without index file`

src/error.rs
```diff
@@ -626,6 +626,24 @@ impl From<UrlParseError> for UrlGenerationError {
     }
 }
 
+/// Errors which can occur when serving static files.
+#[derive(Fail, Debug, PartialEq)]
+pub enum StaticFileError {
+    /// Path is not a directory
+    #[fail(display = "Path is not a directory")]
```
DoumanAsh: Please expand the error description: `Path is not a directory. Unable to serve static files`

DoumanAsh (commented):
Thank you!

I only have small concern about `debug` trace, I don't really think we should log each failure as the only possible suspicious case is `std::fs` method failure, but other cases are expected failures

src/fs.rs
tessereth: From the point of view of the library, I'd agree with you. Any errors logged here are expected behaviour. But as someone using the library, if I'm getting a 404 error, I'd like to know _why_. Is it because a file that's supposed to be there is missing? Is it because I forgot to set the index file? Is it because I don't have permission to open the file? That's why I put it at `debug!` level, to make it easier to debug things as a library user but so it wouldn't show normally. But if you'd rather I remove it then sure, I can do that.

DoumanAsh: I certainly understand your reasoning, I'm not strongly against it as it is possible to remove traces using compile time features of `log` crate.

@fafhrd91 Should we have debug trace here? Originally there was no such trace

DoumanAsh (approved):

src/fs.rs
tessereth: Just to explain where I'm coming from, I'm used to using rails which logs everything under the sun at debug level (including all web requests, all parameters passed, which controller action handles the request, all database requests, all templates rendered and stack traces on any error caught). It's a lot of stuff but it's also pretty useful. But it's definitely up to you guys what you want here.

fafhrd91: debug is fine

DoumanAsh:
@tessereth thank you.

tessereth:
Yay, thanks!

PR #360, opened 2018-07-03
implement FromParam for bool

marcelbuesing:
Support bool parameters e.g. `x/y/true/z`

Workaround:

```Rust
let flag: Option<bool> = req.match_info()
      .get("flag1")
      .and_then(|x| bool::from_str(x).ok());
```

After implementation:

```Rust
let flag: Result<bool,_> = req.match_info()
      .query("flag1");
```

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -240,6 +240,7 @@ macro_rules! FROM_STR {
     };
 }
 
+FROM_STR!(bool);
 FROM_STR!(u8);
 FROM_STR!(u16);
 FROM_STR!(u32);
@@ -294,6 +295,18 @@ mod tests {
         assert_eq!(
             PathBuf::from_param("/seg1/../seg2/"),
             Ok(PathBuf::from_iter(vec!["seg2"]))
+        );
+    }
+
+    #[test]
+    fn test_from_param() {
+        assert_eq!(
+            <bool as FromParam>::from_param("true").unwrap(),
+            true
+        );
+        assert_eq!(
+            <bool as FromParam>::from_param("false").unwrap(),
+            false
         );
     }
 }


src/param.rs
```diff
@@ -296,4 +297,16 @@ mod tests {
             Ok(PathBuf::from_iter(vec!["seg2"]))
         );
     }
+
+    #[test]
+    fn test_from_param() {
+        assert_eq!(
+            <bool as FromParam>::from_param("true").unwrap(),
```
DoumanAsh, line 304: Can you please also verify behavior of integers in test?

I.e. test that it cannot convert integers like `0` or `1` to bool

While currently we rely on `FromStr` it would be better to ensure that we wouldn't parse integers to bool in future

DoumanAsh (approved):
I'm bit unsure whether it is good idea as I just cannot yet imagine use case for such parameters.
But overall I don't see much problem.

@fafhrd91 You're ok with it?

fafhrd91:
I don’t see problem with impl. Question is, do we really need FromParam trait at all?

DoumanAsh:
@fafhrd91 Well we'd need to replace it with something for `Params::query` https://actix.rs/actix-web/actix_web/dev/struct.Params.html#method.query

I guess we could use `FromStr` but I'm not sure how wise it would be.

While convenience of extractors is nice, some may wish to work directly with raw request I think

marcelbuesing:
Actually initially I thought I can use `req.match_info().query("flag1")` to extract query parameter from the url so e.g. `abc?flag1=true.` Understanding that it was actually meant to be used for extracting the parameter in a different way, it is no longer really important to me to have the `FromParam` `bool` implementation. If you think this adds value, I will add the suggested 0/1 int tests. Otherwise I assume less code is better code ;).

fafhrd91:
let's close this pr, i dont think bool has much sense for path param, at least not in for of true/false

PR #368, opened 2018-07-06
Remove reimplementation of `LazyCell`

### LINKED ISSUE #366: Replace use of `UnsafeCell` with `LazyCell`
There exists a crate which is well tested and implements identical functionality:

http://indiv0.github.io/lazycell/lazycell/struct.LazyCell.html

I plan on opening a PR for this myself.

Diggsey:
Fixes #366

- `LazyCell` has no overhead over the current `UnsafeCell` and allows removing some extra uses of unsafe.
- The `mem::uninitialized()` call is unnecessary: experimentation in the playpen shows that the compiler generates byte-for-byte identical code if it is (safely) zero-initialized instead, since it can trivially determine that the entire array is overwritten.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ time = "0.1"
 encoding = "0.2"
 language-tags = "0.2"
 lazy_static = "1.0"
+lazycell = "1.0.0"
 parking_lot = "0.6"
 url = { version="1.7", features=["query_encoding"] }
 cookie = { version="0.10", features=["percent-encode"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,6 +107,7 @@ extern crate htmlescape;
 extern crate http as modhttp;
 extern crate httparse;
 extern crate language_tags;
+extern crate lazycell;
 extern crate mime;
 extern crate mime_guess;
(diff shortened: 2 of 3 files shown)


DoumanAsh:
Em... `LazyCell` uses `UnsafeCell` under hood so I'm not sure what you mean by overhead.

And not sure why we should use third party library for functionality of `UnsafeCell` or `Cell`

Diggsey:
> Em... LazyCell uses UnsafeCell under hood so I'm not sure what you mean by overhead.

I mean it does not add any additional runtime checks, or increase the size of the struct at all - it has no additional overhead compared to the current implementation.

> And not sure why we should use third party library for functionality of UnsafeCell or Cell

`Cell` can't be used here because it does not allow taking a reference to its interior. UnsafeCell requires unsafe code to access. Better to build on existing safe abstractions.

fafhrd91:
what code are you suggesting to replace with LazyCell?

fafhrd91:
Looks good. Let me test performance.

fafhrd91 (approved):

fafhrd91:
Great! thanks!

Issue #365, opened 2018-07-05
Comments on the WebSocket implementation


bluetech:
Some notes I noticed when skimming over the code. The presentation is jumbled but I thought it might be useful anyway.

---

This code can truncate on 32-bits architectures:

```rust
let len = NetworkEndian::read_uint(&buf[idx..], 8) as usize;
```

This repeats a couple times.

In general, `<u64> as usize` is suspicious, especially if read from the network.

---

```rust
LittleEndian::read_u32(mask);
```

I'm not sure why `LittleEndian` is used here - it looks a bit suspicious for big-endian architectures. Maybe `NativeEndian` should be used?

---

```rust
        // unmask
        if let Some(mask) = mask {
            let p: &mut [u8] = unsafe {
                let ptr: &[u8] = &data;
                &mut *(ptr as *const _ as *mut _)
            };
            apply_mask(p, mask);
        }
```

This seems quite dangerous, `data` is a `Bytes` but is mutated directly. If the invariant does indeed hold that `data` is the only reference to the underlying bytes, then the following should work (untested). If not, then the unsafe is probably incorrect, or at least requires justification.

```rust
        // unmask
        let mut data = data.try_mut().unwrap();
        if let Some(mask) = mask {
            apply_mask(&mut data, mask);
        }
```

---

There are a lot of calls like this:

```rust
 let raw_code = NetworkEndian::read_uint(payload.as_ref(), 2) as u16;
```

This can be simplified to:

```rust
 let raw_code = NetworkEndian::read_u16(payload.as_ref());
```

---

All addition operations on the payload length can overflow, e.g.

```rust
let mut buf = BytesMut::with_capacity(p_len + 10);
```

---

```rust
/// Mask/unmask a frame.
#[inline]
pub fn apply_mask(buf: &mut [u8], mask: u32) {
    unsafe { apply_mask_fast32(buf, mask) }
}

/// Faster version of `apply_mask()` which operates on 8-byte blocks.
///
/// unsafe because uses pointer math and bit operations for performance
#[inline]
#[cfg_attr(feature = "cargo-clippy", allow(cast_lossless))]
unsafe fn apply_mask_fast32(buf: &mut [u8], mask_u32: u32) { ... }
```

The usage of `unsafe` in `apply_mask` does not make sense - if it just forwards the input without any validations to an `unsafe fn`, then it means one of two things:

1. `apply_mask_fast32` should not be `unsafe`.
2. `apply_mask` should be `unsafe`.

[...]

PR #377, opened 2018-07-08
Refactor `apply_mask` implementation, removing dead code paths and re…

### LINKED ISSUE #376: `ws::mask::apply_mask` has several dead code paths
```rust
let head = min(len, (8 - (ptr as usize & 0x7)) & 0x3);
```

From this it looks like `head` can only ever be 0, 1, 2 or 3 (due to the bitwise and with 3) and yet later on there are blocks like `if head > 4 { ... }` which will never be hit.

Slightly worrying in an unsafe block. At the very least these blocks should be removed. Possibly the code can be simplified too.

Diggsey:
…ducing scope of unsafety

Fixes #376 

I believe there was also a bug in the previous implementation, where the middle section would sometimes be out of alignment by 4 bytes due to the `& 0x3` in the `head` calculation: this meant that the `head > 4` codepaths would never be hit, and so the pointer would never be adjusted by more than 3 bytes.

The new implementation fixes these issues, and uses the type system to contain the unsafety to the `align_buf` function. I ran several benchmarks of the two implementations, and couldn't detect any difference, but please run your own benchmarks to be sure.

### DIFF
--- a/src/ws/mask.rs
+++ b/src/ws/mask.rs
@@ -1,88 +1,106 @@
 //! This is code from [Tungstenite project](https://github.com/snapview/tungstenite-rs)
 #![cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
-use std::cmp::min;
-use std::mem::uninitialized;
+use std::slice;
 use std::ptr::copy_nonoverlapping;
 
+// Holds a slice guaranteed to be shorter than 8 bytes
+struct ShortSlice<'a>(&'a mut [u8]);
+
+impl<'a> ShortSlice<'a> {
+    unsafe fn new(slice: &'a mut [u8]) -> Self {
+        // Sanity check for debug builds
+        debug_assert!(slice.len() < 8);
+        ShortSlice(slice)
+    }
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+
 /// Faster version of `apply_mask()` which operates on 8-byte blocks.
-///
-/// unsafe because uses pointer math and bit operations for performance
 #[inline]
 #[cfg_attr(feature = "cargo-clippy", allow(cast_lossless))]
 pub(crate) fn apply_mask(buf: &mut [u8], mask_u32: u32) {
-    unsafe {
-        let mut ptr = buf.as_mut_ptr();
-        let mut len = buf.len();
+    // Extend the mask to 64 bits
+    let mut mask_u64 = ((mask_u32 as u64) << 32) | (mask_u32 as u64);
+    // Split the buffer into three segments
+    let (head, mid, tail) = align_buf(buf);
 
-        // Possible first unaligned block.
-        let head = min(len, (8 - (ptr as usize & 0x7)) & 0x3);
-        let mask_u32 = if head > 0 {
-            let n = if head > 4 { head - 4 } else { head };
-
-            let mask_u32 = if n > 0 {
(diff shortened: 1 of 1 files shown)


bluetech:
Looks very nice to me!

After reporting #365, I wrote a quickcheck test for the masking but didn't get a chance to send a PR. It would be nice to integrate here if you feel like it.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 4401e67..9fef7f3 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,6 +107,7 @@ tokio-openssl = { version="0.2", optional = true }
 [dev-dependencies]
 env_logger = "0.5"
 serde_derive = "1.0"
+quickcheck = "0.6"
 
 [build-dependencies]
 version_check = "0.1"
diff --git a/src/lib.rs b/src/lib.rs
index a301227..6449f12 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,6 +141,9 @@ extern crate actix as actix_inner;
 #[cfg(test)]
 #[macro_use]
 extern crate serde_derive;
+#[cfg(test)]
+#[macro_use]
+extern crate quickcheck;
 
 #[cfg(feature = "tls")]
 extern crate native_tls;
diff --git a/src/ws/mask.rs b/src/ws/mask.rs
index 16f0f6b..78a28c3 100644
--- a/src/ws/mask.rs
+++ b/src/ws/mask.rs
@@ -96,6 +96,22 @@ mod tests {
         }
     }
 
+    quickcheck! {
+        fn same_as_simple(buf: Vec<u8>, mask: (u8, u8, u8, u8)) -> bool {
+            let mask = [mask.0, mask.1, mask.2, mask.3];
+
+            let mut buf1 = buf.clone();
+            apply_mask_fallback(&mut buf1, &mask);
+
+            let mut buf2 = buf.clone();
+            // This is what the code does.
+            let mask_u32 = LittleEndian::read_u32(&mask);
+            apply_mask(&mut buf2, mask_u32);
+
+            buf1 == buf2
+        }
+    }
+
     #[test]
     fn test_apply_mask() {
         let mask = [0x6d, 0xb6, 0xb2, 0x80];
```

bluetech:
Although one obvious omission from the diff above is that it only tests aligned inputs (since malloced memory is probably aligned to 16 bytes or so), so it will have to be tweaked a bit in order to test these paths.

src/ws/mask.rs
```diff
@@ -1,89 +1,107 @@
...
-        while len >= 4 {
-            *(ptr as *mut u32) ^= mask_u32;
-            ptr = ptr.offset(4);
-            len -= 4;
-        }
+#[inline]
+// Unsafe: caller must ensure the buffer has the correct size and alignment
+unsafe fn cast_slice(buf: &mut [u8]) -> &mut [u64] {
+    // Assert correct size and alignment in debug builds
+    debug_assert!(buf.len() & 0x7 == 0);
```
DoumanAsh, line 70: Emm just to clarify does this mean that our buffer here must be at least aligned and with size `8` or any other that `len % 8 == 0`?

I feel like it would be better to have comment here for clarification

DoumanAsh (commented):
I did quick look at code and don't see any immediate problem. But I feel like we need to extend our test cases to  be sure that covered all possible paths in these functions

fafhrd91 (approved):

fafhrd91:
@Diggsey thanks!

PR #380, opened 2018-07-10
Fixed empty buffer being sent while there's still some inside

### LINKED ISSUE #352: http2 is broken because of h2 upgrade
I've created an issue yesterday at actix/examples#23 but I believe it is more appropriate to discuss it here. 

It appears the latest `h2` upgrade breaks `actix-web`'s HTTP2 because of the add of this check: https://github.com/carllerche/h2/blob/master/src/server.rs#L1048-L1053.

I tested locally with commenting out the if block and I am certain this is the part causing the issue. I am not familiar with HTTP2 protocol nor SSL nor actix-web's code. So unfortunately I could not come up with a fix for this. 

Thanks!

OtaK:
Could occur if bytes.len() was strictly double the size of buf.len()

Fixes #352

### DIFF
--- a/src/server/h2.rs
+++ b/src/server/h2.rs
@@ -405,9 +405,8 @@ impl<T: Read> Read for IoWrapper<T> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         if let Some(mut bytes) = self.unread.take() {
             let size = cmp::min(buf.len(), bytes.len());
-            buf[..size].copy_from_slice(&bytes[..size]);
-            if bytes.len() > size {
-                bytes.split_to(size);
+            buf[..size].copy_from_slice(&bytes.split_to(size));
+            if bytes.len() > 0 {
                 self.unread = Some(bytes);
             }
             Ok(size)


DoumanAsh:
@OtaK Could please elaborate why exactly it is so? Is this documented condition or what is the reason for this case?

OtaK:
Okay, so, after further review, seems I got hasty on my conclusions, and the existing code is actually working as intended. Furthermore my "fix" is functionally equivalent to the existing code.
Sorry for the disturbance then!

I think the modified code is also less efficient in terms of operations (forcing split_to compared to a simple slice lookup), so I think we can close this PR as not accepted.

OtaK:
@DoumanAsh While I'm still on the issue, I think I might have found the cause for the incompatibility mentioned in the issue.

The fixed code in h2 being this one, it checks if the buffer returns a read size of 0 and if it's the case, it considers the connection as closed.
https://github.com/carllerche/h2/blob/41aae14c6458ad1238debb18fe1e017ad45fef8d/src/server.rs#L1048-L1053

```rust
            if n == 0 {
                return Err(io::Error::new(
                    io::ErrorKind::ConnectionReset,
                    "connection closed unexpectedly",
                ).into());
            }
```

But here, we initialize the `Http2` struct with an empty buffer
https://github.com/actix/actix-web/blob/master/src/server/channel.rs#L50-L55

```rust
        if http2 {
            HttpChannel {
                node: None,
                proto: Some(HttpProtocol::H2(h2::Http2::new(
                    settings,
                    io,
                    peer,
                    Bytes::new(),
                ))),
            }
        } else {
```

A buffer of which the length is never checked until we try to ::read() on it.
See here: https://github.com/actix/actix-web/blob/master/src/server/h2.rs#L65

We pass it as `Some(buf)` without checking it being empty.

And thus, in the first pass in the `IoWrapper` here, it's still `Some(...)` while being empty, resulting in a read size of 0
https://github.com/actix/actix-web/blob/master/src/server/h2.rs#L406

Which is what we are returning to the h2 code.

Do you think it's the root of the problem? I can make another PR fixing this if my reasoning is correct. 

For example, we can fix this issue this way: 
```rust
impl<T: Read> Read for IoWrapper<T> {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        if let Some(mut bytes) = self.unread.take() {
            // This check here is important since it allows falling back on self.inner if bytes is Some while being actually empty
            if bytes.len() == 0 {
                return self.inner.read(buf);
            }
            let size = cmp::min(buf.len(), bytes.len());
            buf[..size].copy_from_slice(&bytes[..size]);
            if bytes.len() > size {
                bytes.split_to(size);
                self.unread = Some(bytes);
            }
            Ok(size)
        } else {
            self.inner.read(buf)
        }
    }
}
```

DoumanAsh:
@OtaK As I'm not very well familiar with h2 libarary I don't see immediate connection with what are you suggesting, I can take a look at it tomorrow. Meanwhile feel free to consider possible solution as I don't have immediate answer for you whether it is appropriate or not 😄

fafhrd91:
@OtaK thanks for the PR! i fixed h2 compatibility

PR #381, opened 2018-07-10
Add prefix aware RouteInfo::has_prefixed_route()

### LINKED ISSUE #354: has_route() doesn't know about prefixes.
has_route() currently does not have any understanding of routes containing a prefix. 

For example:
```
let app = App::new().scope("/scope", |s| s.resource("/sub", |r| r.method(Method::GET).f(index)) );
....
req.route().has_route("/scope/sub") // false
```

This prevents applications and middleware from inspecting routes containing a prefix. For example, one cannot apply the `NormalizePath` middleware against any prefix containing route. It will successfully *normalize* the path, but will fail to redirect when checking to see if the path exists before initiating a redirect.

---

This is documented on the has_route method, but after spending way too long trying to get the NormalizePath middleware to actually apply against prefixed routes it seemed pertinent to create an issue.

OtaK:
Fixes #354

### DIFF
--- a/src/httprequest.rs
+++ b/src/httprequest.rs
@@ -437,7 +437,9 @@ mod tests {
 
         let info = router.default_route_info();
         assert!(info.has_route("/user/test.html"));
+        assert!(info.has_prefixed_route("/user/test.html"));
         assert!(!info.has_route("/test/unknown"));
+        assert!(!info.has_prefixed_route("/test/unknown"));
 
         let req = TestRequest::with_header(header::HOST, "www.rust-lang.org")
             .finish_with_router(router);
@@ -467,7 +469,9 @@ mod tests {
         let mut info = router.default_route_info();
         info.set_prefix(7);
         assert!(info.has_route("/user/test.html"));
+        assert!(!info.has_prefixed_route("/user/test.html"));
         assert!(!info.has_route("/prefix/user/test.html"));
+        assert!(info.has_prefixed_route("/prefix/user/test.html"));
 
         let req = TestRequest::with_uri("/prefix/test")
             .prefix(7)
@@ -490,7 +494,9 @@ mod tests {
         let mut info = router.default_route_info();
         info.set_prefix(7);
         assert!(info.has_route("/index.html"));
+        assert!(!info.has_prefixed_route("/index.html"));
         assert!(!info.has_route("/prefix/index.html"));
+        assert!(info.has_prefixed_route("/prefix/index.html"));
 
         let req = TestRequest::with_uri("/prefix/test")
             .prefix(7)
@@ -513,6 +519,7 @@ mod tests {
 
         let info = router.default_route_info();
         assert!(!info.has_route("https://youtube.com/watch/unknown"));
+        assert!(!info.has_prefixed_route("https://youtube.com/watch/unknown"));
 
         let req = TestRequest::default().finish_with_router(router);
         let url = req.url_for("youtube", &["oHg5SJYRHA0"]);
--- a/src/router.rs
+++ b/src/router.rs
@@ -1,6 +1,7 @@
 use std::cmp::min;
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
+use std::path::Path;
 use std::rc::Rc;
 
 use regex::{escape, Regex};
@@ -144,6 +145,32 @@ impl ResourceInfo {
                 return true;
             }
         }
+        false
+    }
+
+    /// Check if application contains matching route.
+    ///
+    /// This method does take `prefix` into account
+    /// but behaves like `has_route` in case `prefix` is not set in the router.
+    ///
+    /// For example if prefix is `/test` and router contains route `/name`, the
+    /// following path would be recognizable `/test/name` and `has_prefixed_route()` call
+    /// would return `true`.
+    /// It will not match against prefix in case it's not given. For example for `/name`
+    /// with a `/test` prefix would return `false`
+    pub fn has_prefixed_route(&self, path: &str) -> bool {
+        if self.prefix == 0 {
+            return self.has_route(path);
+        }
+
+        let path_matcher = Path::new(if path.is_empty() { "/" } else { path });
+        let router_prefix = Path::new(&path[..(self.prefix as usize)]);
+        if let Ok(p) = path_matcher.strip_prefix(router_prefix) {
+            if let Some(p_str) = p.to_str() {
+                return self.has_route(&format!("/{}", p_str));
+            }
+        }
+
         false
     }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+* Add `.has_prefixed_route()` method to `router::RouteInfo` for route matching with prefix awareness
+
 * Add `HttpMessage::readlines()` for reading line by line.
 
 * Add `ClientRequestBuilder::form()` for sending `application/x-www-form-urlencoded` requests.


DoumanAsh:
@OtaK Would you please create unit test to verify your changes?

src/router.rs
```diff
@@ -100,11 +100,16 @@ impl RouteInfo {
...
     pub fn has_route(&self, path: &str) -> bool {
         let path = if path.is_empty() { "/" } else { path };
+        let ref prefix = self.router.prefix;
 
         for pattern in &self.router.patterns {
             if pattern.is_match(path) {
                 return true;
             }
+
+            if pattern.is_match(&format!("{}{}", prefix, path)) {
```
DoumanAsh: You match against path above, and then you match again... against prefix and path together...
Should it depend on whether there is prefix or not?

src/router.rs
```diff
@@ -100,11 +100,16 @@ impl RouteInfo {
     /// would return `false`.
```
DoumanAsh: If you are going to update the method to handle prefixes, you'd need to update documentation.

OtaK:
I instead added a new method that strips out given prefix, making it able to match against existing code. The motivation behind it is that at least there's no risk of breaking anything since quite a bit of code is relying on `has_route()`.

src/router.rs
```diff
@@ -108,6 +109,35 @@ impl RouteInfo {
...
         false
     }
+
+    /// Check if application contains matching route.
+    ///
+    /// This method does take `prefix` into account.
+    /// For example if prefix is `/test` and router contains route `/name`,
+    /// following path would be recognizable `/test/name` and `has_route_prefixed()` call
+    /// would return `true`.
+    pub fn has_route_prefixed(&self, path: &str) -> bool {
```
DoumanAsh: Suggest for name: `has_route_with_prefix`

src/httprequest.rs
```diff
@@ -470,7 +472,9 @@ mod tests {
         let (router, _) = Router::new("/prefix/", routes);
         let info = router.default_route_info();
         assert!(info.has_route("/user/test.html"));
+        assert!(info.has_route_prefixed("/user/test.html"));
```
DoumanAsh: But we have `/prefix`, why wouldn't we take it into account?
Shouldn't it return false if prefix is missing from path argument?

src/router.rs
```diff
@@ -108,6 +109,35 @@ impl RouteInfo {
...
+        let path = if path.is_empty() {
+            "/".into()
+        } else if !self.router.prefix.is_empty() {
+            let pb = PathBuf::from(path);
```
DoumanAsh: Use `Path::new()` instead, as argument is string slice, we can get `Path` directly without a need for allocation

src/router.rs
```diff
@@ -108,6 +109,35 @@ impl RouteInfo {
...
+            if let Ok(p) = pb.strip_prefix(Path::new(&self.router.prefix)) {
+                format!("/{}", p.to_str().unwrap_or(path))
+            } else {
+                path.into()
```
DoumanAsh: Shouldn't we return here false as prefix is missing?

DoumanAsh (changes requested):

src/router.rs
OtaK: That's what I named it initially, but that would semantically mean that it would accept a second argument (ie: `prefix: &str`). I can still use `has_route_with_prefix` if you'd like

DoumanAsh: hmm yeah maybe you right, but then it would be better to say `has_prefixed_route`

src/router.rs
OtaK: We'd still want to match the same way as `has_route` would, but being able to take `self.router.prefix` in account as an extra feature.

Also, I checked using this function in place of existing `has_route` and is causes breakage.

DoumanAsh: But how route can match, if prefix is missing?
Isn't the whole point of this function is to match against prefix and path together? 🤔

OtaK: Check out my updated code, the initial issue asked to be able to match both in a streamlined fashion, it should be clearer to reason about

src/httprequest.rs
```diff
@@ -470,7 +472,9 @@ mod tests {
         let (router, _) = Router::new("/prefix/", routes);
         let info = router.default_route_info();
         assert!(info.has_route("/user/test.html"));
+        assert!(info.has_prefixed_route("/user/test.html"));
```
DoumanAsh: In continuation from our discussion.
The current behavior makes method is confusing.

It matches both route with prefix or without prefix.

I think this behavior makes it confusing and it should match only in one fashion:
Match both prefix and path.
If prefix is missing from route, then match only against path.

If provides path missing prefix, then it cannot match route, this makes the function name `has_prefixed_route` corresponding to what it actually does.

It is not clear what is benefit matching path against route but ignoring prefix, if path doesn't contain it?

P.s. sorry that I'm being persistent about it, but I just feel this API would be confusing in future

If @fafhrd91 is ok with it I don't mind

DoumanAsh (approved):
Thank you.

Could you please add entry on new method in https://github.com/actix/actix-web/blob/master/CHANGES.md

OtaK:
Added and rebased on master as well.

fafhrd91:
@OtaK i refactored routing subsystem, could you rebase and review you PR again

OtaK:
Did it, I have some breakage because prefix doesn't exist anymore on router. I need to figure out how to get text prefix back and I'll commit the changes

fafhrd91:
ResourceInfo has prefix attribute

OtaK:
Pushed with rebase. Tests are passing on my side.

src/router.rs
```diff
@@ -146,6 +147,32 @@ impl ResourceInfo {
...
+    /// following path would be recognizable `/test/name` and `has_prefixed_route()` call
+    /// would return `true`.
+    /// It will not match against prefix in case it's not given. For example for `/name`
+    /// with a `/test` prefix would return `false`
+    pub fn has_prefixed_route(&self, path: &str) -> bool {
+        if self.prefix == 0 {
+            return self.has_route(path);
+        }
+
+        let path_matcher = Path::new(if path.is_empty() { "/" } else { path });
```
DoumanAsh, line 166: If there is prefix, then empty path cannot match, isn't it right?

src/router.rs
```diff
@@ -146,6 +147,32 @@ impl ResourceInfo {
...
+
+    /// Check if application contains matching route.
+    ///
+    /// This method does take `prefix` into account
+    /// but behaves like `has_route` in case `prefix` is not set in the router.
+    ///
+    /// For example if prefix is `/test` and router contains route `/name`, the
+    /// following path would be recognizable `/test/name` and `has_prefixed_route()` call
+    /// would return `true`.
+    /// It will not match against prefix in case it's not given. For example for `/name`
```
DoumanAsh, line 159: I'm not sure I follow this comment, what exactly you wanted to say here?
Did you mean that path `/name` would match route `/name` when prefix `/test` is omitted?

DoumanAsh (commented):
@fafhrd91 Should we replace old `has_route` with new one or leave it as separate API?

Note that `NormalizePath` uses `has_route` afaik

fafhrd91:
i think this method should be as simple as

```rust
self.has_route(&req.path[(self.prefix as usize)..])
```

@DoumanAsh we need both, component developer should use `has_route()`, but `has_prefixed_route()` is useful in middleware or when we need to manipulate path in general way.

PR #384, opened 2018-07-11
Customize websocket max size

kingxsp:


fafhrd91:
i dont want to add extra public api function just for one param. it only saves couple lines of code

kingxsp:
Ok! But is there any other way to set the max_size value?
```
fn ws_index(r: HttpRequest) -> Result<HttpResponse, Error> {
    ws::start(r, MyWebSocket)
}
```

fafhrd91:
you can manually do handshake and create WebsocketContext

kingxsp:
Ok！

CamJN:
@fafhrd91 Can you elaborate on how to manually do the handshake and create the WebsocketContext, especially in relation to the https://github.com/actix/examples/tree/master/websocket-chat example?

The docs at https://actix.rs/actix-web/actix_web/ws/fn.handshake.html are really minimal.

PR #387, opened 2018-07-11
fix missing content length

kingoflolz:
fix missing content length when no compression is used

### DIFF
--- a/src/client/writer.rs
+++ b/src/client/writer.rs
@@ -266,6 +266,10 @@ fn content_encoder(buf: BytesMut, req: &mut ClientRequest) -> Output {
             }
             #[cfg(not(any(feature = "flate2", feature = "brotli")))]
             {
+                let mut b = BytesMut::new();
+                let _ = write!(b, "{}", bytes.len());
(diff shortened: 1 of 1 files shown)


src/client/writer.rs
```diff
@@ -266,6 +266,10 @@ fn content_encoder(buf: BytesMut, req: &mut ClientRequest) -> Output {
             }
             #[cfg(not(any(feature = "flate2", feature = "brotli")))]
             {
+                let mut b = BytesMut::new();
+                let _ = write!(b, "{}", bytes.len());
+                req.headers_mut()
+                    .insert(CONTENT_LENGTH, HeaderValue::try_from(b.freeze()).unwrap());
```
DoumanAsh, line 272: Is there a reason not to use `bytes.len().to_string()`?

kingoflolz: not sure, I've just copied what was there previously

DoumanAsh: ah ok then

DoumanAsh (approved):

Issue #392, opened 2018-07-13
Cookie parser fails for certain characters


kamek-pf:
Hey,

I noticed some cookies always appear as `None` in `req.session()` (despite being in the Cookie header),  depending on what characters are being used apparently.
Quick example :

```rust
let payload = "hello";

let value = req.session().get::<String>("test");
println!("Value found in headers : {:?}", value); // Ok(Some("hello"))

let res = req.session().set("test", payload);
println!("Set result : {:?}", res); // Ok(())

// send response
```
Run this a couple times and everything works as expected.
Header looks like this :
```
actix-session=zsf1hIJu4CLUesyWHTQ/pedxPET4eqHjO5AQpvOawMU={"test":"\"hello\""}
```

Now if I change the payload, this is what happens :
```rust
let payload = "FJc%26continue_url%3Dhttp%253A%252F%252Fconnectivitycheck.gstatic.com%252Fgenerate_204"

let value = req.session().get::<String>("test"); 
println!("Value found in headers : {:?}", value); // Ok(None)

let res = req.session().set("test", payload);
println!("Set result : {:?}", res); // Ok(())
```
Header looks like this :
```
actix-session=dQxdyG08q5mLMu0Q6KAysLEtYbM/Uak/neM5L3QDj0Y={"test":"\"FJc%26continue_url%3Dhttp%253A%252F%252Fconnectivitycheck.gstatic.com%252Fgenerate_204\""}
```
It seems like the session middleware never picks up this value for some reason.
I assume this is an encoding issue. This happened while storing a URL, my current workaround is to base64 encode the payload and everything works fine.

Any idea what could be causing this ?

DoumanAsh:
Just to clarify is the second example fully percent encoded url?

And the second, can you please make short and fully functional example of your problem?

kamek-pf:
Yes, the url is percent encoded.
I sent a pull request with a test case based on what you did @DoumanAsh.
Thank you btw, I wasn't sure how to write a test for this, your sample helped a lot.

See comment in #395.

DoumanAsh:
@kamek-pf Can you please check your use case with my branch https://github.com/DoumanAsh/actix-web/tree/session_special_chars

Note that you should not percent encode yourself and instead leave it actix

Updated test case https://github.com/actix/actix-web/pull/393/files#diff-48d46aa5fa3906ed3a97ad3e4e6ab7dcR1005

kamek-pf:
I'll take a look later today.
[...]

PR #393, opened 2018-07-13
Session should write percent encoded cookies and add cookie middleware test

DoumanAsh:
I'll use CI to test it as I cannot build ring on my machine for some reason

Test case to verify #392 

Current issue: when secure cookie loads percent encoded content it fails to do so properly

PR #395, opened 2018-07-13
Session middleware test

kamek-pf:
Test case for #392

### DIFF
--- a/tests/test_middleware.rs
+++ b/tests/test_middleware.rs
@@ -993,3 +993,74 @@ fn test_resource_middleware_async_chain_with_error() {
     assert_eq!(num2.load(Ordering::Relaxed), 1);
     assert_eq!(num3.load(Ordering::Relaxed), 1);
 }
+
+#[cfg(feature = "session")]
+#[test]
+fn test_session_storage_middleware() {
+    use actix_web::middleware::session::{
+        CookieSessionBackend, RequestSession, SessionStorage,
+    };
+    use actix_web::HttpMessage;
+
+    const SIMPLE_NAME: &'static str = "simple";
+    const SIMPLE_PAYLOAD: &'static str = "kantan";
+    const COMPLEX_NAME: &'static str = "test";
+    const COMPLEX_PAYLOAD: &'static str = "FJc%26continue_url%3Dhttp%253A%252F%252Fconnectivitycheck.gstatic.com%252Fgenerate_204";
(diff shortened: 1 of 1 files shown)


kamek-pf:
So this test won't pass. Swap lines 1008 and 1009 and it does.

DoumanAsh:
Thanks I have difficulties to write these tests because I cannot use `session` feature on windows as of now.
I'll take a look at your test later on

tests/test_middleware.rs
```diff
@@ -993,3 +993,74 @@ fn test_resource_middleware_async_chain_with_error() {
...
+    });
+
+    let request_set = srv.get().uri(srv.url("/first")).finish().unwrap();
+    let response_set = srv.execute(request_set.send()).unwrap();
+
+    // we grab the "set-cookie" header from the response and use it as our "cookie" header for the next request
+    let cookie: &str = response_set.headers()["set-cookie"]
+        .to_str()
+        .unwrap()
+        .split(";") // cut off the last part ("; HttpOnly; Path=/"")
```
DoumanAsh: Should it be rsplit?

fafhrd91:
@DoumanAsh could you review and decide what to do with it

DoumanAsh:
@kamek-pf Currently test doesn't pass on https://github.com/actix/actix-web/pull/395/files#diff-48d46aa5fa3906ed3a97ad3e4e6ab7dcR1034
I don't see what is wrong with it and why session is not restored, so I'd need to review Session middleware code before to have proper comments

And my PR has the same issue 😄 
I wonder if we should either stick to your PR or mine #393
It is impossible, at the moment, for me to build with `session` feature due to `ring` issues so I wouldn't mind to close it up and help you to investigate issue

kamek-pf:
Oh sorry, I should have made this more clear :
It appears that if a single value contains problematic characters, then the session middleware is unable to restore anything at all.

The behavior you're seeing here is consistent with my observations.

DoumanAsh:
Ah thanks, that clarifies it as I was wondering what was wrong.

Then given than we both have the same PR, would you like me to continue work on issue in my PR and close your PR? Or vice versa?

kamek-pf:
They do look identical now, I don't mind at all which one gets merged, my goal was just to highlight the issue :p

DoumanAsh:
Ok, I'll close it in favor of my PR #393. hopefully I'll get to bottom of this soon enough

PR #399, opened 2018-07-15
Router refactoring

fafhrd91:


src/router.rs
```diff
@@ -111,104 +149,244 @@ impl RouteInfo {
 }
```
DoumanAsh, line 149: Since by default path is just empty instead of single slash, method above `has_route` doesn't need condition conditional set for empty path?

`let path = if path.is_empty() { "/" } else {path}`

src/router.rs
```diff
@@ -111,104 +149,244 @@ impl RouteInfo {
...
         }
     }
 
-    /// Query for matched resource
-    pub fn recognize(&self, req: &Request) -> Option<(usize, RouteInfo)> {
-        if self.0.prefix_len > req.path().len() {
-            return None;
+    pub(crate) fn register_external(&mut self, name: &str, rdef: ResourceDef) {
+        let inner = Rc::get_mut(&mut self.defs).unwrap();
+        if inner.named.contains_key(name) {
```
DoumanAsh: Should we use assert here?

src/router.rs
```diff
@@ -592,265 +764,226 @@ mod tests {
...
     fn test_recognizer_with_prefix() {
-        let routes = vec![
-            (ResourceDef::new("", "/name"), Some(Resource::default())),
-            (
-                ResourceDef::new("", "/name/{val}"),
-                Some(Resource::default()),
-            ),
-        ];
-        let (rec, _) = Router::new::<()>("/test", routes);
+        let mut router = Router::<()>::new();
```
DoumanAsh, line 849: Emmm... I'm not sure: where prefix `/test` is set?

src/scope.rs
```diff
@@ -286,20 +253,18 @@ impl<S: 'static> Scope<S> {
...
             false,
         );
-        Rc::get_mut(&mut self.resources)
-            .expect("Can not use after configuration")
-            .push((pattern, Rc::new(handler)));
+        let mut resource = Resource::new(pattern);
+        f(&mut resource);
 
+        Rc::get_mut(&mut self.router)
+            .unwrap()
```
DoumanAsh, line 266: I think it would be better to continue using `expect` with some message to easily identify problem, even if it is unlikely to happen.

src/scope.rs
```diff
@@ -979,9 +910,9 @@ mod tests {
             })
             .finish();
 
-        let req = TestRequest::with_uri("/app/t1").request();
-        let resp = app.run(req);
-        assert_eq!(resp.as_msg().status(), StatusCode::NOT_FOUND);
+        //let req = TestRequest::with_uri("/app/t1").request();
```
DoumanAsh: Is the PR yet to be finished? Or why this code is left commented out?

Also I see below that there is `TestRequest` with the same uri, but with different assert... which actually seems kinda strange.

src/router.rs
fafhrd91: i dont get it.

DoumanAsh: I mean check here https://github.com/actix/actix-web/blob/master/src/router.rs#L102 which is redundant now 

Github doesn't allow to place comments too far from modified code :(

src/router.rs
fafhrd91: `router.recognize(&req, &(), 5)`

there is no predefined prefix anymore

src/router.rs
DoumanAsh: Opps... I rechecked code, it is prefix who is by default `""` now, not path

fafhrd91: that is fine, we dont really need this check

fafhrd91: oh, i am wrong :)

src/scope.rs
fafhrd91: i can not imagine how this unwrap could panic

src/scope.rs
fafhrd91: good catch. difference is tail slash

src/router.rs
DoumanAsh: Sorry, it is just I'm confused with test cases below.

It adds two routes
`/name`
`/name/{val}`

Then attempts to match with following requests:
`/name` -> Default resource
`/test/name` -> 0 resource
`/test/name/value` -> 1 resource

But I do not see where prefix `/test` has been defined

src/scope.rs
DoumanAsh: This is true, but in case of any future modification that may lead to such situation...
Better to be easy to determine problem 😄 

But anyway it is minor thing so feel free to ignore

src/router.rs
DoumanAsh: I thought the point of test case is to verify correctness of prefix handling when it is set 😄 
But seeing as it is no longer the case, maybe test case should be renamed to something like `test_recognizer_with_scope`

fafhrd91: third param in `recognize()` call.

src/scope.rs
fafhrd91: you can add `except` if you'd like :)

DoumanAsh (approved):

PR #402, opened 2018-07-17
ws/context: Increase `write()` visibility to public

Turbo87:
I'm running a TCP socket to WebSocket gateway using actix and I noticed that the messages that I relay to the WebSocket clients are copied once per client as part of `ws::Frame::message()`. I would like to avoid that overhead by calling `ws::Frame::message()` only once, and passing the result to each of the WebSocket clients.

This PR changes the `WebsocketContext::write()` method to be public, so that users are able to pass in `Binary` directly including the framing.

### DIFF
--- a/src/ws/context.rs
+++ b/src/ws/context.rs
@@ -20,7 +20,7 @@ use context::{ActorHttpContext, Drain, Frame as ContextFrame};
 use error::{Error, ErrorInternalServerError, PayloadError};
 use httprequest::HttpRequest;
 
-use ws::frame::{Frame, FramedBinary};
+use ws::frame::{Frame, FramedMessage};
 use ws::proto::{CloseReason, OpCode};
 use ws::{Message, ProtocolError, WsStream, WsWriter};
 
@@ -138,7 +138,7 @@ where
     /// data you should prefer the `text()` or `binary()` convenience functions
     /// that handle the framing for you.
     #[inline]
-    pub fn write_raw(&mut self, data: FramedBinary) {
+    pub fn write_raw(&mut self, data: FramedMessage) {
         if !self.disconnected {
             if self.stream.is_none() {
                 self.stream = Some(SmallVec::new());
--- a/src/ws/context.rs
+++ b/src/ws/context.rs
@@ -133,7 +133,7 @@ where
 {
     /// Write payload
     #[inline]
-    fn write(&mut self, data: Binary) {
+    pub fn write(&mut self, data: Binary) {
         if !self.disconnected {
             if self.stream.is_none() {
                 self.stream = Some(SmallVec::new());
--- a/src/ws/context.rs
+++ b/src/ws/context.rs
@@ -138,7 +138,7 @@ where
     /// data you should prefer the `text()` or `binary()` convenience functions
     /// that handle the framing for you.
     #[inline]
-    pub fn write(&mut self, data: Binary) {
+    pub fn write_raw(&mut self, data: Binary) {
         if !self.disconnected {
             if self.stream.is_none() {
                 self.stream = Some(SmallVec::new());
@@ -172,19 +172,19 @@ where
     /// Send text frame
     #[inline]
     pub fn text<T: Into<Binary>>(&mut self, text: T) {
-        self.write(Frame::message(text.into(), OpCode::Text, true, false));
+        self.write_raw(Frame::message(text.into(), OpCode::Text, true, false));
     }
 
     /// Send binary frame
     #[inline]
     pub fn binary<B: Into<Binary>>(&mut self, data: B) {
-        self.write(Frame::message(data, OpCode::Binary, true, false));
+        self.write_raw(Frame::message(data, OpCode::Binary, true, false));
     }
 
     /// Send ping frame
     #[inline]
     pub fn ping(&mut self, message: &str) {
-        self.write(Frame::message(
+        self.write_raw(Frame::message(
             Vec::from(message),
             OpCode::Ping,
             true,
@@ -195,7 +195,7 @@ where
     /// Send pong frame
     #[inline]
     pub fn pong(&mut self, message: &str) {
-        self.write(Frame::message(
+        self.write_raw(Frame::message(
             Vec::from(message),
             OpCode::Pong,
             true,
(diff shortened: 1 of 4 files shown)


src/ws/context.rs
```diff
@@ -133,7 +133,7 @@ where
 {
     /// Write payload
```
DoumanAsh, line 134: Please describe in documentation to this method how user should create `Binary` as user should not manually initialize it.
and instead use proper constructor.

src/ws/context.rs
```diff
@@ -133,7 +133,7 @@ where
...
     #[inline]
-    fn write(&mut self, data: Binary) {
+    pub fn write(&mut self, data: Binary) {
```
DoumanAsh: Do you think it may be wise for us to introduce wrapper type for websocket frame, to avoid potential misuse?(e.g. user fill Binary himself)

Turbo87: I guess if `ws::Frame::message()` would return a dedicated type and `WebsocketContext::write()` would only accept that type it might be useful. I'm not sure though if `ws::Frame::message()` is used anywhere else though.

src/ws/context.rs
Turbo87: Docs added 👍

src/ws/context.rs
DoumanAsh: I think if we want to expose this API, it should be dedicated type to avoid potential misuse, as you imagine not all people would be careful enough when reading documentation :)

You could leave this `write` method for internal use and instead create new method `send_frame` or just `frame` that would accept wrapper type and pass it's Binary data into `write`

src/ws/context.rs
```diff
@@ -132,8 +132,13 @@ where
...
 {
     /// Write payload
+    ///
+    /// This is a low-level function that accepts framed messages that should
+    /// be created using `Frame::message()`. If you want to send text or binary
+    /// data you should prefer the `text()` or `binary()` convenience functions
+    /// that handle the framing for you.
     #[inline]
-    fn write(&mut self, data: Binary) {
+    pub fn write(&mut self, data: Binary) {
```
fafhrd91: maybe rename it to `write_raw()`?

Turbo87: good idea. renamed :)

src/ws/context.rs
Turbo87: I've introduced a `FramedBinary` type now for this purpose. This is technically a breaking change, but since master is using a newer `actix` version the next release would be a breaking release anyway 😉

src/ws/frame.rs
```diff
@@ -372,6 +374,10 @@ impl fmt::Display for Frame {
     }
 }
 
+/// A `Binary` representing a `WebSocket` message with framing.
+#[derive(Debug)]
+pub struct FramedBinary(pub Binary);
```
DoumanAsh: `pub` allows user to manually create the instance, I think we should prohibit it?

User should only correctly framed messages.

src/ws/context.rs
DoumanAsh: thanks!

src/ws/frame.rs
```diff
@@ -372,6 +374,10 @@ impl fmt::Display for Frame {
     }
 }
 
+/// A `Binary` representing a `WebSocket` message with framing.
```
DoumanAsh: I think it would be better to say just:

>`WebSocket` framed message

We don't really need to overload users :)

src/ws/frame.rs
DoumanAsh: Maybe name it `FramedMessage`?

Even `write_raw` says now 😄 

>This is a low-level function that accepts framed messages

DoumanAsh (commented):
I still have concern about `FramedBinary` being instantiated by user

src/ws/frame.rs
Turbo87: If I make it private or `pub(crate)` it complains about "can't leak private type" and won't compile

DoumanAsh: Oh what I mean is that its member `pub Binary` should be just `Binary` so that user couldn't instantiate it himself like `FramedBianry(Bytes::new().into())`

Turbo87: actually since the `frame` module is not `pub` and the struct is not exported it already is private and only accessible inside of `actix-web` so there is no change needed here

src/ws/frame.rs
Turbo87: updated

src/ws/frame.rs
Turbo87: done

DoumanAsh: In that case we should re-export it as public.
This `FramedBinary` will be leaked through `Frame::message` while it does prevents user fron instantiating it, it also prevents user from reading any documentation on it (you'll have blanked type `FramedBinary` without any information about it)
I think it would not be very nice for reader?

Turbo87: done

src/ws/frame.rs
```diff
@@ -372,6 +374,10 @@ impl fmt::Display for Frame {
     }
 }
 
+/// `WebSocket` message with framing.
+#[derive(Debug)]
+pub struct FramedMessage(pub Binary);
```
DoumanAsh: `pub Binary` -> `Binary` 😄

DoumanAsh (approved):

src/ws/frame.rs
Turbo87:
```
error[E0611]: field `0` of tuple-struct `ws::frame::FramedMessage` is private
   --> src/ws/client.rs:535:45
    |
535 |             let _ = inner.tx.unbounded_send(data.0.take());
    |                                             ^^^^^^

error[E0611]: field `0` of tuple-struct `ws::frame::FramedMessage` is private
   --> src/ws/context.rs:147:50
    |
147 |             stream.push(ContextFrame::Chunk(Some(data.0)));
    |                                                  ^^^^^^
```

Turbo87: I could make it `pub(crate)` though if that helps?

DoumanAsh: Oh I see, sorry, I didn't notice it is used outside of this module.
Yes, it should be public only within crate itself, we cannot let user to write non-framed data.

Turbo87: done

fafhrd91 (approved):

DoumanAsh:
@Turbo87 Thanks

PR #405, opened 2018-07-18
Initial config for static files

### LINKED ISSUE #351: StaticFiles's etag and last-modified header may not work in cluster
In out StaticFiles handler, etag and last-modified was set to check file changes and response with 304.
This work well in single node server.
But when it comes to scale out, the etag is annoying because it never be the same in different server.
So to have a better cache control, I think we should add the cache-control header, and have a option to disable etag and  last-modified header when user want to use actix-web in a cluster.
Any thoughts?

### LINKED ISSUE #401: Option to disable Content-Disposition for StaticFiles
Hi,
actix is really working well for a small website I built, thanks for that!

I noticed a change in `StaticFiles` starting with 0.6. It seems like `Content-Disposition: attachment; filename=<filename>` was added to the headers.
Unfortunately this does not work well in conjunction with `<object data="filename.pdf" type="application/pdf">` tags.
Firefox reacts like the file would not exist, chrome shows a "failed to load pdf document" error.
(Also see [this stackoverflow question](https://stackoverflow.com/questions/26611123/pdf-doesnt-load-in-a-object-tag))

It would be nice to have an option to switch this header off for `StaticFiles`.

DoumanAsh:
Closes #401
Closes #351

Initial draft for generic configuration of static files.

Currently allows to:
- Specify mapping for content disposition
- Disable/Enable Etag/LastModified
- Mapping of allowed methods for file service

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -6,6 +6,7 @@ use std::ops::{Deref, DerefMut};
 use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 use std::{cmp, io};
+use std::marker::PhantomData;
 
 #[cfg(unix)]
 use std::os::unix::fs::MetadataExt;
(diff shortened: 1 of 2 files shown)


src/fs.rs
```diff
@@ -27,6 +28,73 @@ use httprequest::HttpRequest;
...
+///
+/// impl StaticFileConfig for MyConfig {
+///     fn content_disposition_map(typ: mime::Name) -> DispositionType {
+///         DispositionType::Attachment
+///     }
+/// }
+///
+/// let file = NamedFile::open_with_config("foo.txt", MyConfig);
+///```
+pub trait StaticFileConfig: Default {
```
fafhrd91, line 59: why do you need `Default` if trait contains only static methods? that is actually more general question. maybe config should be stateful?

DoumanAsh: It is more or less for making it possible to pass config as argument so that in future we  could make it stateful if needed.
But I can remove it as currently I don't really see the need for it to be stateful

DoumanAsh: It is also more pretty than putting type parameter

fafhrd91 (approved):
Let’s rename CD to just T
Also you export ContentDisposition in this pr

DoumanAsh:
Rebased, but I renamed it to just `C`, similar how we use `S` to reflect on nature of the underlying type.
I'll merge once tests are passing

DoumanAsh:
Also added changelog entry

euank:
I'm not sure if this is the right place to ask, but I had a question about the choice of using traits here.
I'm happy to file a new issue or discuss this elsewhere if that's better.

Anyways, the question boils down to "why traits vs the builder pattern"?

Some other parts of actix use builders (e.g. `App::new().prefix("/foo").finish()`), and it seems like this could as well (e.g. `NamedFile::open_with_config("path", NamedFile::Config::new().etag(true).finish())`).

Maybe I'm missing something, or this wasn't an intentional choice, but rather just the first thing to come to mind... since I couldn't tell, I thought I'd ask.

cc @DoumanAsh for any insight you can provide. Thanks for contributing the feature in the first place btw, it's appreciated :)

DoumanAsh:
@euank I felt like builder is getting to big for its own good, but I kept old builder in place to avoid non backward compatible change.
Maybe I should move everything to trait itself for next major release.

The thing about trait it helps to separate configuration from `StaticFiles` and `NamedFile` itself and we can easily share the same configuration among both.
Considering that config is more or less static thing, I felt that it should be made as trait to be applied rather than always manually applying it

PR #406, opened 2018-07-18
Expose leaked private ContentDisposition

### LINKED ISSUE #404: Expose the method `header::Charset.name` as pub
1.
Currently, this is no easy way to get file names from a `multipart/form-data` request. Thus, getting file names has to be done manually by accessing `DispositionParam::Filename` and use some codec library like the crate *encoding*. To use the *encoding* crate, the label of Charset is needed to obtain the proper codec by using [`encoding::label::encoding_from_whatwg_label`](https://docs.rs/encoding/0.2.33/encoding/#overview)).
Although that can be achieved by writing a big `match` such as
```rust
// `cs` is of type `Charset`
let decoder = match cs {
  Charset::Iso-8859-1 => encoding::all::ISO_8859_6,
  Charset::Big5 => ...
}
```
But it is too complicated and unnecessary. `header::Charset.name` now actually returns the label of Charset (a list of charset and their labels can be found [here](https://encoding.spec.whatwg.org/#concept-encoding-get)). Therefore, I think it is reasonable to expose the `name` method of `header::Charset` as `pub fn` so that decoding of the file name can be done easily via something like `encoding::label::encoding_from_whatwg_label(charset.name()).decode(filename)`.

The link to that line: [https://github.com/actix/actix-web/blob/6b10e1eff6fb3455a4a56ea07367d18623198283/src/header/shared/charset.rs#L69](https://github.com/actix/actix-web/blob/6b10e1eff6fb3455a4a56ea07367d18623198283/src/header/shared/charset.rs#L69).

2.
I also notice that there is a typo  [https://github.com/actix/actix-web/blob/6b10e1eff6fb3455a4a56ea07367d18623198283/src/header/shared/charset.rs#L93](https://github.com/actix/actix-web/blob/6b10e1eff6fb3455a4a56ea07367d18623198283/src/header/shared/charset.rs#L93) (L93 and L132). The name of `Charset::Big5` is `"5"` which is obviously wrong and not valid label/name specified by [WHATWG](https://encoding.spec.whatwg.org/#concept-encoding-get). It needs correcting accordingly.
~If you accept my suggestions above, I also suggest that the function name `name` be changed to `label` to reflect what it is actually returning. `fn name` is used only internally for `impl Display for Charset`, so it won't introduce any breaking changes.~

DoumanAsh:
We expose it through https://actix.rs/actix-web/actix_web/multipart/struct.Field.html#method.content_disposition

So I think we should make all relevant types public too

Closes #404

### DIFF
--- a/src/header/shared/charset.rs
+++ b/src/header/shared/charset.rs
@@ -66,7 +66,7 @@ pub enum Charset {
 }
 
 impl Charset {
-    fn name(&self) -> &str {
+    fn label(&self) -> &str {
         match *self {
             Us_Ascii => "US-ASCII",
             Iso_8859_1 => "ISO-8859-1",
@@ -90,7 +90,7 @@ impl Charset {
             Iso_8859_8_E => "ISO-8859-8-E",
             Iso_8859_8_I => "ISO-8859-8-I",
             Gb2312 => "GB2312",
-            Big5 => "5",
+            Big5 => "big5",
             Koi8_R => "KOI8-R",
             Ext(ref s) => s,
         }
@@ -99,7 +99,7 @@ impl Charset {
 
 impl Display for Charset {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(self.name())
+        f.write_str(self.label())
     }
 }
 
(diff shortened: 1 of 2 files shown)


fafhrd91 (approved):
only question, should this be part of `http` or `http::header`?

DoumanAsh:
@fafhrd91 Not sure, I see that `ContentEncoding` is part of `http` so I just followed its example. 

**UPD:** I put it into header

DoumanAsh:
It seems it kinda hits coverage? Could be becuase of making method public?

But I don't really feel like it s good idea to just test text representation of enum

fafhrd91:
Strange that it is whole 1%

fafhrd91 (approved):

PR #416, opened 2018-07-23
Add FromRequest<S> implementation for Option<T> and Result<T> where T: FromRequest<S>

axos88:
This pull request adds an implementation of  `FromRequest<S>` for `Option<T>` and `Result<T>` where `T: FromRequest<S>`.

In case the `FromRequest<T>` extractor fails for any reason the `FromRequest<Option<T>>` returns `None`, the `FromRequest<Result<T>>` return the Error itself. They do not fail the request, so it is available and can be handled within the handler.

They can be useful for example when extracting `Authorization` / `Cookie` / `Session` / `Header` / `X` information that may or may not be present for the request.

### DIFF
--- a/src/extractor.rs
+++ b/src/extractor.rs
@@ -460,6 +460,126 @@ impl<S: 'static> FromRequest<S> for String {
     }
 }
 
+/// Optionally extract a field from the request
+///
+/// If the FromRequest for T fails, return None rather than returning an error response
+///
+/// ## Example
+///
+/// ```rust
+/// # extern crate actix_web;
+/// extern crate rand;
+/// #[macro_use] extern crate serde_derive;
+/// use actix_web::{http, App, Result, HttpRequest, Error, FromRequest};
+/// use actix_web::error::ErrorBadRequest;
+///
+/// #[derive(Debug, Deserialize)]
+/// struct Thing { name: String }
+///
+/// impl<S> FromRequest<S> for Thing {
+///     type Config = ();
+///     type Result = Result<Thing, Error>;
+///
+///     #[inline]
+///     fn from_request(req: &HttpRequest<S>, _cfg: &Self::Config) -> Self::Result {
+///         if rand::random() {
+///             Ok(Thing { name: "thingy".into() })
+///         } else {
+///             Err(ErrorBadRequest("no luck"))
+///         }
+///
+///     }
+/// }
+///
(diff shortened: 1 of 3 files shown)


src/extractor.rs
```diff
@@ -455,6 +460,126 @@ impl<S: 'static> FromRequest<S> for String {
...
+    type Config = T::Config;
+    type Result = Box<Future<Item = Option<T>, Error = Error>>;
+
+    #[inline]
+    fn from_request(req: &HttpRequest<S>, cfg: &Self::Config) -> Self::Result {
+        Box::new(T::from_request(req, cfg).into().then( |r| {
+            match r {
+                Ok(v) => future::ok(Some(v)),
+                Err(e) => {
+//                    if true { panic!("{:?}", e.as_response_error()); }
```
fafhrd91: remove this line

src/extractor.rs
```diff
@@ -792,4 +986,33 @@ mod tests {
         assert_eq!((res.1).0, "name");
         assert_eq!((res.1).1, "user1");
     }
+
+//    #[test]
```
fafhrd91: uncomment or remove

src/extractor.rs
```diff
@@ -16,7 +16,10 @@ use error::{Error, ErrorBadRequest, ErrorNotFound, UrlencodedError};
 use handler::{AsyncResult, FromRequest};
 use httpmessage::{HttpMessage, MessageBody, UrlEncoded};
 use httprequest::HttpRequest;
+use Result;
+use futures::future;
```
fafhrd91: move this to line 9

src/extractor.rs
```diff
@@ -16,7 +16,10 @@ use error::{Error, ErrorBadRequest, ErrorNotFound, UrlencodedError};
 use handler::{AsyncResult, FromRequest};
 use httpmessage::{HttpMessage, MessageBody, UrlEncoded};
 use httprequest::HttpRequest;
+use Result;
```
fafhrd91: remove this, use `Result<T, Error>` instead

fafhrd91 (changes requested):

axos88:
Rebased onto actix-web/master

fafhrd91 (approved):

fafhrd91:
very good! thanks!

PR #422, opened 2018-07-25
Update cookie dependency

mati865:
Updating `cookie` for security fix:
```
$ cargo audit 
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
    Scanning 191 crates for vulnerabilities (10 advisories in database)
error: Vulnerable crates found!

ID:	 RUSTSEC-2018-0001
Crate:	 untrusted
Version: 0.5.1
Date:	 2018-06-21
URL:	 https://github.com/briansmith/untrusted/pull/20
Title:	 An integer underflow could lead to panic
Solution: upgrade to: >= 0.6.2

error: 1 vulnerability found!
```

EDIT: Dependency chain is: `cookie` -> `ring` -> `untrusted`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ lazy_static = "1.0"
 lazycell = "1.0.0"
 parking_lot = "0.6"
 url = { version="1.7", features=["query_encoding"] }
-cookie = { version="0.10", features=["percent-encode"] }
+cookie = { version="0.11", features=["percent-encode"] }
 brotli2 = { version="^0.3.2", optional = true }
 flate2 = { version="1.0", optional = true, default-features = false }
 


DoumanAsh (approved):
@fafhrd91 it also brings ring to `0.13` which finally would fix problems on msvc

DoumanAsh:
Thanks for heads up, I was waiting for new cookie version

fafhrd91:
We should be careful with major upgrades within minor version changes. It could be braking change

DoumanAsh:
@fafhrd91 I checked commit history, they did update base64 dependency, but interface itself to `Cookie` remains the same afaik

fafhrd91:
I think that is fine for this case. But even for this update developers will need to modify cargo.toml file. 

This is fine, but in the future we need to be more careful with such changes

PR #433, opened 2018-07-31
feature: allow TestServer to open a websocket on any URL

### LINKED ISSUE #432: No way to specify a socket URI for TestServer.ws()
It appears that in `test.rs`:

```rust 
pub fn ws(
        &mut self,
    ) -> Result<(ws::ClientReader, ws::ClientWriter), ws::ClientError> {
        let url = self.url("/");
        self.rt
            .block_on(ws::Client::with_connector(url, self.conn.clone()).connect())
}
```
`url` is hard set to "/". 
I have several entry points for websockets for my app, and I would like to be able to specify the path. The app allows for this, but unfortunately, testing does not.

I'm not sure what the preferred approach might be, possibly a second fn that takes a path string?

Thanks!

jrconlin:
* added `TestServer::ws_at(uri_str)`
* modified `TestServer::ws()` to call `self.ws_at("/")` to preserve
behavior

Closes #432

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 * Support HTTP/2 with rustls #36
 
+* Allow TestServer to open a websocket on any URL # 433
+
 ### Fixed
 
 * Do not override HOST header for client request #428
@@ -22,7 +24,7 @@
 * Add implementation of `FromRequest<S>` for `Option<T>` and `Result<T, Error>`
 
 * Allow to handle application prefix, i.e. allow to handle `/app` path
-  for application with `/app` prefix. 
+  for application with `/app` prefix.
   Check [`App::prefix()`](https://actix.rs/actix-web/actix_web/struct.App.html#method.prefix)
   api doc.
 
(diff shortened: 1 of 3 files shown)


fafhrd91:
Looks good. Could you add test case and entry to changes.md

fafhrd91 (approved):

jrconlin:
Sure thing. Granted, I kinda figured that the test case for `ws()` should cover it, since it falls through to `ws_at("/")`, but I can create an explicit one.

fafhrd91:
Yes, I’d like to have explicit test case.

jrconlin:
By the way, we've been using [clog](https://github.com/clog-tool/clog-cli) as part of our release process. It automatically updates the CHANGES.md file from the submitted PRs, although it does require patch authors to properly tag their patches, but it's fairly straight forward. (Here's [an example](https://github.com/mozilla-services/autopush-rs/blob/master/CHANGELOG.md) of clog output.)

Not sure if it might help, but figured it might be useful.

fafhrd91:
thanks! I'll add test

fafhrd91:
oh! you added test, thanks

Issue #434, opened 2018-08-01
error[E0433]: failed to resolve. Could not find `_core` in `failure`


ensch:
_actix-web_ has a dependecy on _failure_ =0.1.1, but no dependecy on _failure_derive_ =0.1.1 

That leads to a combination of _failure_ 0.1.1 and _failure_derive_ 0.1.2 being selected by cargo, but this combination is not supported, leading to this compiler error:

```
   Compiling actix v0.7.3
error[E0433]: failed to resolve. Could not find `_core` in `failure`
  --> /Users/es/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-0.7.3/src/address/mod.rs:23:10
   |
23 | #[derive(Fail)]
   |          ^^^^ Could not find `_core` in `failure`

error[E0433]: failed to resolve. Could not find `_core` in `failure`
   --> /Users/es/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-0.7.3/src/actors/resolver.rs:133:10
    |
133 | #[derive(Fail, Debug)]
    |          ^^^^ Could not find `_core` in `failure`
```

Either the dependecy on _failure_ has to be changed from "=0.1.1" to e.g. "0.1.2" or a dependecy on the matching version of _failure_derive_ has to be added.

fafhrd91:
fixed in master

fafhrd91:
i will make new release later this week

PR #435, opened 2018-08-01
Add failure_derive version restriction so that it matches failure version

ensch:
Fix for https://github.com/actix/actix-web/issues/434


fafhrd91:
We need specific version of failure because of unsafe that we use 
https://github.com/actix/actix-web/blob/master/src/error.rs#L93

It needs to be fixed before we can relax dependency

fafhrd91:
fixed by  972b008

Issue #443, opened 2018-08-03
cause() is not implemented for InternalError's Fail trait


mitskevich:
As I understand, the best approach to throwing errors in `actix-web` handlers currently is something like this (simplified code):

```rust
fn authenticate_handler(req: &HttpRequest) -> Result<Json<AuthResult>, actix_web::Error> {
    let auth_data = get_auth_data_from_request(req).map_err(ErrorBadRequest)?;
    authenticate_or_fail(auth_data).map_err(ErrorUnauthorized)
}
```

Here I have two functions `get_auth_data_from_request` and `authenticate_or_fail` that return some `Result` with either correct data or my internal errors implementing `Fail` trait. Those errors get mapped into `InternalError` with helper functions `ErrorBadRequest` and `ErrorUnauthorized` which basically attach status code information to the resulting response.

The problem arise when I'm trying to output the error in the log (or in the email report, etc.). I want to display error with full cause chain, so I'm doing something like this in my error handler middleware (again, simplified):

```rust
impl<S> aw::middleware::Middleware<S> for MyErrorDisplayMiddleware {
    fn response(
        &self,
        req: &aw::HttpRequest<S>,
        mut resp: aw::HttpResponse,
    ) -> Result<aw::middleware::Response, aw::Error> {
        if let Some(err) = resp.error() {
            let err_str = format_fail(err.as_fail());
            warn!("API error for {}: {}", req.uri(), err_str);
        }
        Ok(aw::middleware::Response::Done(resp))
    }
}
```

Where `format_fail` is my function that displays any `Fail` error looping by `.cause()` calls. 

Doing that I have found that `Fail` implementation for `InternalError` doesn't contain `cause()` implementation. 

https://github.com/actix/actix-web/blob/e61ef7dee4a4e1017372783594b6f6bda6dc6f5c/src/error.rs#L685-L692

This means I can only display the top error in the chain. Was this done on purpose and is there any simple workaround for that issue? The only way I can think of now is to not use `InternalError` (and all helpers) at all, and use my own implementation instead...

fafhrd91:
Could you submit Pr with cause impl?

mitskevich:
Closing because there is no way to implement `.cause()` without changing public API (see PR).

PR #444, opened 2018-08-03
Implemented .cause() for InternalError's Fail trait.

mitskevich:
This PR fixes issue #443 at the cost of requiring `T: Fail` for all `InternalError`s. I understand that this is a breaking change for `actix-web` public APIs, but I don't see any other way to preserve error-cause chain information for internal errors without using unsafe code.

Please let me know if there is some more elegant approach I can use.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -238,7 +238,9 @@ where
         if self.fut.alive() {
             match self.fut.poll() {
                 Ok(Async::NotReady) | Ok(Async::Ready(())) => (),
-                Err(_) => return Err(ErrorInternalServerError("error")),
+                Err(_) => {
+                    return Err(ErrorInternalServerError(::error::FailMsg("error")))
+                }
             }
         }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -633,6 +633,35 @@ impl ResponseError for StaticFileError {
     }
 }
 
+/// Helper type for making `Fail` instance from `&'static str` without memory allocation.
+///
+/// In following example handler is returning error message with *BAD REQUEST* status code.
+///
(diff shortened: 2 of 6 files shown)


fafhrd91:
what about .downcast?

mitskevich:
That's actually what I intended to do initially, but we need to downcast to trait (not concrete type), and as far as I can tell, that's not possible (maybe only using unsafe code)...

Basically, I thought about doing something like this, but it doesn't compile (Fail is not sized):

```rust
    fn cause(&self) -> Option<&Fail> {
        self.cause.downcast_ref::<Fail>()
    }
```

This compiles, but always returns `None` (obviously, because we have some type, not ref in `self.cause`):

```rust
    fn cause(&self) -> Option<&Fail> {
        self.cause.downcast_ref::<&Fail>().map(|v| *v)
    }
```

I have tried boxing `self.cause` and use `Box::downcast()`, but again, this always returns `None`, because we are converting to trait, not to concrete type.

Please let me know if I'm missing something...

fafhrd91:
change is too intrusive, and I am not planing to release 0.8 in near future.
let's not change InternalError then. also this change prevent using it in form of BadRequestError("message"), which was original intent for it.

mitskevich:
Sure, no problem, I'm not 100% liking this approach too, especially because you won't be able to map `failure::Error` errors by `ErrorBadRequest` too (it doesn't implement `Fail`). I'll just use my own `InternalError` for my projects that require full error chain reporting.

I'm closing this PR, should I close the issue too?

fafhrd91:
yes, let's close issue too

PR #449, opened 2018-08-05
impl FromRequest for ()

erikdesjardins:
This allows writing `app.route("/foo", Method::GET, |()| foo)` when you don't care about the request, which is a bit more concise than `app.route("/foo", Method::GET, |_: HttpRequest| foo)`.

I didn't use the macro because it would've been a larger diff (need to change all the repeats from `+` to `*`) and the futures machinery from the macro isn't necessary for `()`.

### DIFF
--- a/src/extractor.rs
+++ b/src/extractor.rs
@@ -693,7 +693,7 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
 impl<S> FromRequest<S> for () {
     type Config = ();
     type Result = Self;
-    fn from_request(_req: &HttpRequest<S>, _cfg: &Self::Config) {}
+    fn from_request(_req: &HttpRequest<S>, _cfg: &Self::Config) -> Self::Result {}
 }
 
 tuple_from_req!(TupleFromRequest1, (0, A));
@@ -1013,6 +1013,6 @@ mod tests {
         assert_eq!((res.1).0, "name");
         assert_eq!((res.1).1, "user1");
 
-        <()>::extract(&req);
+        let () = <()>::extract(&req);
     }
 }
--- a/src/extractor.rs
+++ b/src/extractor.rs
@@ -690,6 +690,12 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
     }
 });
 
+impl<S> FromRequest<S> for () {
+    type Config = ();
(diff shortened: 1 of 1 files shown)


src/extractor.rs
```diff
@@ -1006,5 +1012,7 @@ mod tests {
         assert_eq!((res.0).1, "user1");
         assert_eq!((res.1).0, "name");
         assert_eq!((res.1).1, "user1");
+
+        <()>::extract(&req);
```
DoumanAsh: add `.except("To empty extract")`

src/extractor.rs
```diff
@@ -690,6 +690,12 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
     }
 });
 
+impl<S> FromRequest<S> for () {
+    type Config = ();
+    type Result = Self;
+    fn from_request(_req: &HttpRequest<S>, _cfg: &Self::Config) {}
```
DoumanAsh: It is unnecessary, but I kinda feel we better follow the original trait definition and add `-> Self::Result`
If you don't think so, you can ignore 😄

DoumanAsh (commented):
One minor comment.

DoumanAsh:
In case of closures it seems unnecessary, but it is nice idea for function handlers where we don't use request.

But are you sure it is good idea to not require at least `HttpRequest` as argument?

src/extractor.rs
erikdesjardins: Done

erikdesjardins: It's infallible, so it just returns `()`. I changed it to make that easier to see.

src/extractor.rs
erikdesjardins: Done.

erikdesjardins:
I find it more useful for closures since the type of `_` can't be inferred in `app.route("/foo", Method::GET, |_| foo)` so you're forced to annotate it. Doesn't really matter though, it's a minor papercut either way.

I'm not sure what you mean, it doesn't seem obviously bad not to take `HttpRequest`?

DoumanAsh:
Ah I see, it is generic parameter so it is ambiguous to compiler and user need to use explicit type argument.

Well maybe I'm overthinking and it is nice way to improve usability.

DoumanAsh (approved):

DoumanAsh:
thanks

PR #451, opened 2018-08-06
Write non-80 port in HOST of client's request

### LINKED ISSUE #448: Host header does not comply with RFC2616
Sending a request to a host on a port other then 80 should change the `host` header according to [RFC2616 14.23 Host](https://www.ietf.org/rfc/rfc2616.txt)
i.e.: Connecting to localhost:8081 should write the header `host: localhost:8081` and not `host: localhost`

DoumanAsh:
Closes #448

### DIFF
--- a/src/client/request.rs
+++ b/src/client/request.rs
@@ -629,7 +629,14 @@ impl ClientRequestBuilder {
             if let Some(parts) = parts(&mut self.request, &self.err) {
                 if let Some(host) = parts.uri.host() {
                     if !parts.headers.contains_key(header::HOST) {
-                        match host.try_into() {
+                        let mut wrt = BytesMut::with_capacity(host.len() + 5).writer();
+
+                        let _ = match parts.uri.port() {
+                            None | Some(80) | Some(443) => write!(wrt, "{}", host),
+                            Some(port) => write!(wrt, "{}:{}", host, port),
+                        };
+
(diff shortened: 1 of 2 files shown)


src/client/request.rs
```diff
@@ -629,7 +629,14 @@ impl ClientRequestBuilder {
             if let Some(parts) = parts(&mut self.request, &self.err) {
                 if let Some(host) = parts.uri.host() {
                     if !parts.headers.contains_key(header::HOST) {
-                        match host.try_into() {
+                        let mut wrt = BytesMut::with_capacity(host.len() + 5).writer();
+
+                        let _ = match parts.uri.port() {
+                            None | Some(80) => write!(wrt, "{}", host),
```
fafhrd91: I think we should support 443 as well

DoumanAsh: According to RFC:

>A "host" without any trailing port information implies the default port for the service requested (e.g., "80" for an HTTP URL)

It seems if we omit, 443, server would assume 80 rather than 443.
But I'm not sure about this one

P.s. or is HTTPs a separate service?

DoumanAsh: I checked how Firefox/Chrome sets it for HTTPs and they seems to omit, so let 443 to be omitted too

fafhrd91 (approved):

fafhrd91:
dont forget to include changes.md entry

PR #453, opened 2018-08-07
Reserve enough space for ServerError task to write status line

### LINKED ISSUE #452: Direct access to http://127.0.0.1:8080 can cause a thread to panic without routing being set
Direct access to http://127.0.0.1:8080 can cause a thread to crash without routing being set

DoumanAsh:
The same function is used in H1 writer https://github.com/actix/actix-web/blob/master/src/server/h1writer.rs#L167

But there we reserve buffer with enough space.

But I'm not sure about error task here and whether it is possible to predict required space.
So I put reserve in `ServerError` just for status line, in case it wouldn't be sufficient.

@fafhrd91 Do you think it is a problem that buffer is not sufficient for `ServerError` to write status line in first place?

Closes #452

### DIFF
--- a/src/server/error.rs
+++ b/src/server/error.rs
@@ -16,6 +16,9 @@ impl HttpHandlerTask for ServerError {
     fn poll_io(&mut self, io: &mut Writer) -> Poll<bool, Error> {
         {
             let bytes = io.buffer();
+            //Buffer should have sufficient capacity for status line
+            //and extra space
+            bytes.reserve(helpers::STATUS_LINE_BUF_SIZE + 1);
             helpers::write_status_line(self.0, self.1.as_u16(), bytes);
         }
         io.set_date();
--- a/src/server/helpers.rs
+++ b/src/server/helpers.rs
@@ -8,8 +8,10 @@ const DEC_DIGITS_LUT: &[u8] = b"0001020304050607080910111213141516171819\
       6061626364656667686970717273747576777879\
       8081828384858687888990919293949596979899";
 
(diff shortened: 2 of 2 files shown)


fafhrd91 (approved):

fafhrd91:
Did you test if it fixes #452

DoumanAsh:
@fafhrd91 Not yet, I'm still trying to write simple example to reproduce it

fafhrd91:
i think this id 32bit error

gmg137:
@DoumanAsh 
#### reproduce code
```
extern crate actix;
extern crate actix_web;
extern crate env_logger;

use actix_web::{middleware, server, App, http, HttpRequest};

fn index(_req: &HttpRequest) -> &'static str {
    "Hello world!"
}

fn main() {
    ::std::env::set_var("RUST_LOG", "actix_web=info");
    ::std::env::set_var("RUST_BACKTRACE", "1");
    env_logger::init();
    let sys = actix::System::new("xy");

    server::new(|| {
            App::new()
            // enable logger
            .middleware(middleware::Logger::default())
            .prefix("/api")
            .resource("/hello",|r| r.method(http::Method::GET).f(index))
    }).bind("127.0.0.1:8080")
        .unwrap()
        .shutdown_timeout(3)
        .start();

    println!("Started http server: 127.0.0.1:8080");
    let _ = sys.run();
}
```

####  Reappearance method: 
>  browsers visit three consecutive routes that do not exist.

aaa

fafhrd91:
@gmg137 could you try master

gmg137:
@fafhrd91 After testing, the use of master is no longer a problem.

Issue #457, opened 2018-08-08
Could the `HttpResponseBuilder` method `json` take paramater `value` by reference instead of value


davidMcneil:
The `HttpResponseBuilder` method [`json`](https://actix.rs/api/actix-web/stable/src/actix_web/httpresponse.rs.html#652-669) takes a parameter `value` of type T by value. Is this necessary? It looks like only a reference to value is used in the method.

DoumanAsh:
I think there should be no problem as we really need only reference to value, though we would need to wait for `0.8` for such change as it breaks compatibility of API.

fafhrd91:
We can add `.json2()` for 0.7

davidMcneil:
Great! I will put together a pull request.

DoumanAsh:
@fafhrd91 should we plan for it to become `json` in 0.8 then?

fafhrd91:
@DoumanAsh yes, but I don’t want to release 0.8

DoumanAsh:
Well Ok, I'll try to keep it mind for next major release once it happens in future.

fafhrd91:
Fixed by #458

PR #458, opened 2018-08-08
Add json2 HttpResponseBuilder method

davidMcneil:
Add `json2` method to `HttpResponseBuilder` which takes `value` by reference instead of value.  #457

### DIFF
--- a/src/httpresponse.rs
+++ b/src/httpresponse.rs
@@ -1193,6 +1193,30 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_json2() {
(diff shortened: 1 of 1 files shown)


fafhrd91:
Please add test

davidMcneil:
Ok, there already exists a `test_json` and `test_json_ct` because `json` simply calls `json2` and because for version 0.8 it seems like `json` will be replaced with `json2` I did not want to make redundant tests. But I can go ahead and make a `test_json2` and `test_json2_ct` for now.

fafhrd91 (approved):

fafhrd91:
Thanks!

PR #461, opened 2018-08-12
Rework Content-Disposition parsing totally

Gowee:
This PR is going to solve several bugs that exist in *content_disposition.rs* and improve the ergonomics of extracting data out of `Content-Disposition` header with *actix-web*.

The previous implementation:
- Cannot parse `Content-Disposition` if it contains special characters like " (double quote) and ; (semicolon) due to its implementation limits;
- Cannot parse `Content-Disposition` if it contains parameter values in UTF-8 as [`quoted-string`](https://tools.ietf.org/html/rfc2616#section-2.2) (e.g. `form-data; name=upload; filename="文件.webp"`). UTF-8 in `quoted-string` is not compliant with old RFCs,  however:
  1. [RFC 7578](https://tools.ietf.org/html/rfc7578) states clearly that:
  >  Some commonly deployed systems use multipart/form-data with file names directly encoded including octets outside the US-ASCII range. The encoding used for the file names is typically UTF-8, although HTML forms will use the charset associated with the form.
  2. Mainstream browsers like Chrome and Firefox (Gecko) follow this practice. **So the previous implementation cannot parse `Content-Disposition` at all from HTML forms submitted using Chrome and Firefox if non-ASCII characters are present in file name.**
  3. Generate `Content-Disposition` header without escaping some characters, resulting in unexpected file name truncation. For example, if a file called *quote"here.zip* is served with `fs::StaticFiles`, then browsers will see it as `quote` with all characters after the double quote stripped. 
  4. Other non-significant incompliance and problems.

Besides solving these problems listed, this commit also:
 - Tries to be compatible, even though not strictly compliant, with [RFC6266](https://www.ietf.org/rfc/rfc6266.txt) (`Content-Disposition` as response header used to indicate how to open files and file names to be saved as) and [RFC7578] (`Content-Disposition` generated for submitting HTML forms of `multipart/form-data`).
- Should be able to parse all `Content-Disposition` generated by implementations compliant with RFC6266 and RFC7578 under most circumstances. (But, it could also possibly accept some malformed headers.) 
- Adds plenty of test cases to ensure it works as expected.
- Tweaks related struct/enum to improve ergonomics and ensure conformance (e.g. the previous implementation treat `filename` and `filename*` equally, and users cannot distinguish between them, which is not wanted in [some cases](https://tools.ietf.org/html/rfc6266#section-4.3)).
- Adds a lot of useful methods to existing struct/enum to allow easier data extracting. For example:
[https://github.com/Gowee/actix-web/blob/c4e29eb1dc0bb3339207f0ca22cbcca48bb46012/src/header/common/content_disposition.rs#L240](https://github.com/Gowee/actix-web/blob/c4e29eb1dc0bb3339207f0ca22cbcca48bb46012/src/header/common/content_disposition.rs#L240)

### DIFF
--- a/src/header/common/content_disposition.rs
+++ b/src/header/common/content_disposition.rs
@@ -282,7 +282,7 @@ impl ContentDisposition {
 
         while left.len() > 0 {
             let (param_name, new_left) = split_once_and_trim(left, '=');
-            if param_name.len() == 0 {
+            if param_name.len() == 0 || param_name == "*" || new_left.len() == 0 {
                 return Err(::error::ParseError::Header);
             }
             left = new_left;
@@ -325,7 +325,7 @@ impl ContentDisposition {
                             }
                         }
                     }
-                    left = left.split_at(end.ok_or(::error::ParseError::Header)? + 1).1;
+                    left = &left[end.ok_or(::error::ParseError::Header)? + 1..];
                     left = split_once(left, ';').1.trim_left();
                     // In fact, it should not be Err if the above code is correct.
                     let quoted_string = String::from_utf8(quoted_string)
@@ -789,6 +789,31 @@ mod tests {
         };
         assert_eq!(a, b);
     }
+
+    #[test]
+    fn test_from_raw_param_value_missing() {
+        let a = HeaderValue::from_static("form-data; name=upload ; filename=");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+
+        let a = HeaderValue::from_static("attachment; dummy=; filename=invoice.pdf");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+
+        let a = HeaderValue::from_static("inline; filename=  ");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+    }
+
+    #[test]
+    fn test_from_raw_param_name_missing() {
+        let a = HeaderValue::from_static("inline; =\"test.txt\"");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+
+        let a = HeaderValue::from_static("inline; =diary.odt");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+
+        let a = HeaderValue::from_static("inline; =");
+        assert!(ContentDisposition::from_raw(&a).is_err());
+    }
+
 
     #[test]
     fn test_display_extended() {
--- a/src/header/common/content_disposition.rs
+++ b/src/header/common/content_disposition.rs
@@ -475,7 +475,7 @@ impl fmt::Display for DispositionType {
 
 impl fmt::Display for DispositionParam {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // All ASCII control charaters (0-30, 127) excepting horizental tab, double quote, and
+        // All ASCII control charaters (0-30, 127) excepting horizontal tab, double quote, and
         // backslash should be escaped in quoted-string (i.e. "foobar").
         // Ref: RFC6266 S4.1 -> RFC2616 S2.2; RFC 7578 S4.2 -> RFC2183 S2 -> ... .
         lazy_static! {
--- a/src/header/common/content_disposition.rs
+++ b/src/header/common/content_disposition.rs
@@ -9,6 +9,7 @@
 use header;
 use header::ExtendedValue;
 use header::{Header, IntoHeaderValue, Writer};
+use regex::Regex;
 
 use std::fmt::{self, Write};
 
@@ -47,11 +48,14 @@ pub enum DispositionType {
 
 impl<'a> From<&'a str> for DispositionType {
     fn from(origin: &'a str) -> DispositionType {
-        match origin.as_bytes().to_ascii_lowercase().as_slice() {
-            b"inline" => DispositionType::Inline,
-            b"attachment" => DispositionType::Attachment,
-            b"form-data" => DispositionType::FormData,
-            _ => DispositionType::Ext(origin.to_owned()),
+        if origin.eq_ignore_ascii_case("inline") {
+            DispositionType::Inline
+        } else if origin.eq_ignore_ascii_case("attachment") {
+            DispositionType::Attachment
+        } else if origin.eq_ignore_ascii_case("form-data") {
+            DispositionType::FormData
+        } else {
+            DispositionType::Ext(origin.to_owned())
         }
     }
 }
@@ -292,14 +296,13 @@ impl ContentDisposition {
                 let (ext_value, new_left) = split_once_and_trim(left, ';');
                 left = new_left;
                 let ext_value = header::parse_extended_value(ext_value)?;
-                cd.parameters
-                    .push(match param_name.to_ascii_lowercase().as_str() {
-                        "filename" => DispositionParam::FilenameExt(ext_value),
-                        _ => DispositionParam::UnknownExt(
-                            param_name.to_owned(),
-                            ext_value,
-                        ),
-                    });
+
+                let param = if param_name.eq_ignore_ascii_case("filename") {
+                    DispositionParam::FilenameExt(ext_value)
+                } else {
+                    DispositionParam::UnknownExt(param_name.to_owned(), ext_value)
+                };
+                cd.parameters.push(param);
             } else {
                 // regular parameters
                 let value = if left.starts_with('\"') {
@@ -313,10 +316,12 @@ impl ContentDisposition {
                             escaping = false;
                             quoted_string.push(c);
                         } else {
-                            if c == 0x5c // backslash
+                            if c == 0x5c
+                            // backslash
                             {
                                 escaping = true;
-                            } else if c == 0x22 // double quote
+                            } else if c == 0x22
+                            // double quote
                             {
                                 end = Some(i + 1); // cuz skipped 1 for the leading quote
                                 break;
@@ -340,12 +345,15 @@ impl ContentDisposition {
                 if value.len() == 0 {
                     return Err(::error::ParseError::Header);
                 }
-                cd.parameters
-                    .push(match param_name.to_ascii_lowercase().as_str() {
-                        "name" => DispositionParam::Name(value),
-                        "filename" => DispositionParam::Filename(value),
-                        _ => DispositionParam::Unknown(param_name.to_owned(), value),
-                    });
+
+                let param = if param_name.eq_ignore_ascii_case("name") {
+                    DispositionParam::Name(value)
+                } else if param_name.eq_ignore_ascii_case("filename") {
+                    DispositionParam::Filename(value)
+                } else {
+                    DispositionParam::Unknown(param_name.to_owned(), value)
+                };
+                cd.parameters.push(param);
             }
         }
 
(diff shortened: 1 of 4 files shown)


fafhrd91:
This is great!

@DoumanAsh could you review this pr.

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
+            parameters: Vec::new(),
+        };
 
-                let key = if let Some(key) = parts.next() {
-                    key.trim()
+        while left.len() > 0 {
+            let (param_name, new_left) = split_once_and_trim(left, '=');
+            if param_name.len() == 0 {
+                return Err(::error::ParseError::Header);
+            }
```
DoumanAsh: Should we check for `new_left != 0`?

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
+            left = new_left;
+            if param_name.ends_with('*') {
+                // extended parameters
+                let (param_name, _) = param_name.split_at(param_name.len() - 1); // trim asterisk
```
DoumanAsh: why `split_at` instead of slicing?

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
+                    let mut escaping = false;
+                    let mut quoted_string = vec![];
+                    let mut end = None;
+                    // search for closing quote
+                    for (i, &c) in left.as_bytes().iter().skip(1).enumerate() {
+                        if escaping {
+                            escaping = false;
+                            quoted_string.push(c);
+                        } else {
+                            if c == 0x5c // backslash
```
DoumanAsh: maybe it would be better to use char literals here instead of raw byte?

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
+                    });
+            } else {
+                // regular parameters
+                let value = if left.starts_with('\"') {
+                    // quoted-string: defined in RFC6266 -> RFC2616 Section 3.6
+                    let mut escaping = false;
+                    let mut quoted_string = vec![];
+                    let mut end = None;
+                    // search for closing quote
+                    for (i, &c) in left.as_bytes().iter().skip(1).enumerate() {
```
DoumanAsh, line 314: why use reference for `&c`?

DoumanAsh, line 314: Should we maybe use `.chars()` here?
This way you could just collect it into string, instead of going through parsing

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
+                                end = Some(i + 1); // cuz skipped 1 for the leading quote
+                                break;
+                            } else {
+                                quoted_string.push(c);
+                            }
+                        }
+                    }
+                    left = left.split_at(end.ok_or(::error::ParseError::Header)? + 1).1;
+                    left = split_once(left, ';').1.trim_left();
+                    // In fact, tt should not be Err if the above code is correct.
```
DoumanAsh: `tt` -> `it`

DoumanAsh (commented):
This is quite a work you did! Thank you.

I'm feeling that it would be better to use full fledged parser for `Content-Disposition` but since our parsing is quite lenient I suppose it is not really necessary.

One thing I'd like to ask:
How should we handle cases when param's value is missing?
I'd like to add such tests to be sure that we would react correctly

src/header/common/content_disposition.rs
Gowee: Yes but not necessary. Line 294 `header::parse_extended_value` will error out if `ext_value` is empty, which includes cases where `new_left.len() == 0`. The nested if block starting at L305 falls into the branch starting at L334, if `new_left` is empty, which accepts empty values. But check in L342 will ensure it error out.

src/header/common/content_disposition.rs
Gowee: Because I just forget to use...

src/header/common/content_disposition.rs
Gowee: See the following comments.

src/header/common/content_disposition.rs
Gowee: To dereference `&u8` into `u8`? I think `as_bytes().iter()` produces &u8 instead of u8, am I right?
I am not sure how to write that idiomatically in Rust.

Gowee: That is what I attempt at first, and then I find that the following `split_at` panics. Then I get to know `Char`, which is produced by the `.chars()` iterator, is different from `u8` in Rust. But `split_at` takes byte offset instead of Char offset. Seems that this is also not idiomatic... Any suggestions to improve?

src/header/common/content_disposition.rs
Gowee: Fix it

Gowee:
@DoumanAsh 
I agree. It would be more robust if a rigorous parser is used. But it will introduce new dependencies (if not writing a parser from the scratch). I am not sure what is your opinions on that. Besides, the new implementation should handle most cases and run a little faster.

Thanks for your patient review!

I assume cases where param's value is missing will only appear in `token` (e.g. `filename=sample.png`; L335) other than `quoted-string` (`filename="sample.pdf"`) and `ext-value` (`filename*=UTF-8''%c2%a3%20and%20%e2%82%ac%20rates''`). Because the latter both contains boundary characters and certainly will/should not be accepted if value is totally missing. 
In `token`, for param's value missing (e.g. `attachment; filename=` or `form-data; name=; filename=foobar`), it now just error out (see my first comment). I look through related statements in RFC7578 (-> RFC 2183 Section 2 -> RFC2045 Section 5.1) and RFC6266 (-> RFC2616 Section 3.2). They both require at least one character is present. So It should be syntax error if the value is empty as meets the current behavior. Negative test cases are welcome.

src/header/common/content_disposition.rs
DoumanAsh: Ah yeah, I see.
If I'm not mistaken though you can omit `&` here

DoumanAsh: Ah I see... Yes, in this case chars are nasty as they are themself four bytes and doesn't represent actual character's byte len

DoumanAsh:
@Gowee 
Yeah the implementation is already good as it is, 😄 

I see that code covers the case, but I didn't notice unit tests for that.
For example could you add test to verify handling of `attachment; filename=`?

src/header/common/content_disposition.rs
DoumanAsh: But if `new_left.len() == 0` it wouldn't fall into the line 289 branch.
I just feel like it would be better to leave earlier in bad case instead of checking it out.
Though I'm not sure should we just ignore such parameter or error....

src/header/common/content_disposition.rs
Gowee: After change `&c` to `c` in line 311:
```
error[E0277]: can't compare `&u8` with `{integer}`
   --> src/header/common/content_disposition.rs:316:34
    |
316 |                             if c == 0x5c // backslash
    |                                  ^^ no implementation for `&u8 == {integer}`
    |
    = help: the trait `std::cmp::PartialEq<{integer}>` is not implemented for `&u8`

error[E0277]: can't compare `&u8` with `{integer}`
   --> src/header/common/content_disposition.rs:319:41
    |
319 |                             } else if c == 0x22 // double quote
    |                                         ^^ no implementation for `&u8 == {integer}`
    |
    = help: the trait `std::cmp::PartialEq<{integer}>` is not implemented for `&u8`

error[E0308]: mismatched types
   --> src/header/common/content_disposition.rs:331:59
    |
331 |                     let quoted_string = String::from_utf8(quoted_string)
    |                                                           ^^^^^^^^^^^^^ expected u8, found &u8
    |
    = note: expected type `std::vec::Vec<u8>`
               found type `std::vec::Vec<&u8>`

```

DoumanAsh: Ah sorry, I was sure Rust would copy it as it is, but it seems you went for right syntax originally

src/header/common/content_disposition.rs
Gowee: Why wouldn't? Whether falling into L289 only dependent on the presence of trailing asterisk of `param_name`. 

Fine, I will add that check in a new commit.

I am also not sure and fail to find related statements in those standards or elsewhere. But I do not think it is really a problem in practice because that is generated by browsers and does not rely on end-users input. Besides, what the lenient parsing implementation wants to do is just ensuring whatever directly submitted by end-users (typically, file names) are parsed correctly as long as browsers produce headers correctly.

src/header/common/content_disposition.rs
Gowee: Aha, I am often messed up by those rules/implications too.

src/header/common/content_disposition.rs
DoumanAsh: If you think that the current version is ok as it is unlikely to happen then leave it as it is 😄

Gowee:
Several test cases are added and tweaks have been made accordingly.  : )

src/header/common/content_disposition.rs
```diff
@@ -21,27 +38,163 @@ pub enum DispositionType {
...
+    /// Used in *multipart/form-data* as defined in
+    /// [RFC7578](https://tools.ietf.org/html/rfc7578) to carry the field name and the file name.
+    FormData,
+    /// Extension type. Should be handled by recipients the same way as Attachment
+    Ext(String),
+}
+
+impl<'a> From<&'a str> for DispositionType {
+    fn from(origin: &'a str) -> DispositionType {
+        match origin.as_bytes().to_ascii_lowercase().as_slice() {
```
DoumanAsh: I know it makes it more elegant, but I think it would be better to use non-allocating comparison using `eq_ignore_ascii_case`

src/header/common/content_disposition.rs
```diff
@@ -65,88 +218,215 @@ pub enum DispositionParam {
...
-
-                let val = if let Some(val) = parts.next() {
-                    val.trim()
-                } else {
+                if value.len() == 0 {
                     return Err(::error::ParseError::Header);
-                };
+                }
+                cd.parameters
+                    .push(match param_name.to_ascii_lowercase().as_str() {
```
DoumanAsh: Can you please use `eq_ignore_ascii_case`? Know it is not so elegant but even more effectiveness 😄

src/header/common/content_disposition.rs
```diff
@@ -174,112 +454,423 @@ impl Header for ContentDisposition {
...
-                        try!(header::http_percent_encode(f, bytes))
-                    }
-                },
-                DispositionParam::Ext(ref k, ref v) => try!(write!(f, "; {}=\"{}\"", k, v)),
+    }
+}
+
+impl fmt::Display for DispositionParam {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // TODO: More special charaters in filename should be escaped.
```
DoumanAsh: Btw we already use regex for example, so if you think we'd need some more characters you could add it here too using regex crate I guess.

DoumanAsh (approved):
There are only two minor comments, but overall it is good :)

DoumanAsh:
@Gowee Could you please add [changes](https://github.com/actix/actix-web/blob/master/CHANGES.md) with mention of your more robust parser for `Content-Disposition`

DoumanAsh:
@fafhrd91 The commit makes breaking changes to `DispositionParam`, what do you think about it? Is it too early for such change?

src/header/common/content_disposition.rs
Gowee: yes. but I am not sure what is the idiomatic way to write that. is a long `if` block fine?

DoumanAsh: Yes, that's the only way sadly, you may wish to move if block outside of `cd.parameters.push` for a better read-ability

src/header/common/content_disposition.rs
```diff
@@ -467,17 +475,23 @@ impl fmt::Display for DispositionType {
 
 impl fmt::Display for DispositionParam {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // TODO: More special charaters in filename should be escaped.
+        // All ASCII control charaters (0-30, 127) excepting horizental tab, double quote, and
```
DoumanAsh: `horizental` -> `horizontal`

DoumanAsh (commented):
Just minor comment. We also need to consider breaking changes, if @fafhrd91 says it is a bit earlier, we might need to make `DispositionParam` without breaking changes.

Gowee:
https://github.com/actix/actix-web/pull/461/commits/f036c275e4f786d2bc58fd30b8ba3c15747de922
https://github.com/actix/actix-web/pull/461/commits/59ba1c84d1e0d0b4514a91fb087dce6ecedf7fa3
- Use regex to more rigorously escape characters in values like `filename` as suggested by @DoumanAsh.
While there is still a problem inherited from the previous code as well as the `http` crate: Some unprintable ASCII characters are allowed to be used *directly* in `Content-Disposition` as long as they are enclosed inside the quote pair and escaped. But all `HeaderValue::from_...` methods, including the one `actix-web` use to create CD header, do not accept characters in ASCII 0x1-0x31 and 0x127, and hence serving files whose name contains those special characters using `actix-web` with `Content-Disposition` set will fail. I have no idea how to solve this problem appropriately in view of the fact that `Content-Disposition` is so different than other headers. Anyway, it is not a significant problem because I suppose people do not often serve files with unprintable characters.
- All code using `to_ascii_lowercase` should have been modified to use `eq_ignore_ascii_case` if I did not miss some.
- Typo fixed.

fafhrd91:
I think breaking change is fine for this

Gowee:
If necessary, could you please help me write CHANGES.md instead? I might fall into a sleep soon  :sleeping:  and be a little busy tomorrow.

DoumanAsh:
@Gowee ahah, ok, I'll add changes myself. Thank you for your work

PR #477, opened 2018-08-19
add multipart for client

jeizsm:


DoumanAsh:
Hm... this is interesting, i see that this is your care and it already provides simple integration(I would suggest to make it through trait though)

But considering we already have own implementation of multipart for server side, it could make sense to have own implementation for client side.

@fafhrd91 thoughts on using existing client multipart form impl?

PR #478, opened 2018-08-20
Made extensions constructor public

fzgregor:
`Extensions` constructor has to be public for custom `IoStream` implementations that want to expose stream extensions.


DoumanAsh:
Oh...  true, you cannot really create `Extensions`

Where would you like to use it though? `IoStream` is not actually public

fzgregor:
Apart from `IoStream` I don't know any other place.

I understand it is not public in the sense of a "yet to be released" feature. But once it is and you want to implement your own `IoStream` it should, or?

I want to access the client's certificate in my API handlers to authorize accesses, which is what I solved with my own `Acceptor` and its own `IoStream` populating the request's `stream_extensions`: https://play.rust-lang.org/?gist=35d016c93f06f92e361760ed9c0163d6&version=stable&mode=debug&edition=2015

DoumanAsh:
Should we make `IoStream` public then too?

fafhrd91:
@DoumanAsh I’ll take care of this Pr

fafhrd91 (approved):

fafhrd91:
thanks!

PR #482, opened 2018-08-24
Fix server startup log message

0x1793d1:
**Actix web 0.7.4** broke the server startup log message.

The message I had:
Starting server on http://Some(V4(0.0.0.0:8080))

I fixed this with an **unwrap** and the display placeholder, because the address should be correct at this stage.


fafhrd91 (approved):

fafhrd91:
thanks

PR #491, opened 2018-08-29
Add ability to pass a custom TlsConnector

untitaker:
I want to be able to configure all the options on
[`native_tls::TlsConnectorBuilder`](https://docs.rs/native-tls/0.2.1/native_tls/struct.TlsConnectorBuilder.html)
like I do with the other SSL backends.

In the process I refactored a bunch of code, with the main purpose of
not copy-pasting the default field values everwhere.

If you agree with this approach, I will add docs for the native-tls
variant of with_connector

### DIFF
--- a/src/client/connector.rs
+++ b/src/client/connector.rs
@@ -17,14 +17,16 @@ use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_timer::Delay;
 
 #[cfg(feature = "alpn")]
-use openssl::ssl::{Error as OpensslError, SslConnector, SslMethod};
-#[cfg(feature = "alpn")]
-use tokio_openssl::SslConnectorExt;
+use {
+    openssl::ssl::{Error as SslError, SslConnector, SslMethod},
+    tokio_openssl::SslConnectorExt
+};
 
 #[cfg(all(feature = "tls", not(feature = "alpn")))]
-use native_tls::{Error as TlsError, TlsConnector as NativeTlsConnector};
-#[cfg(all(feature = "tls", not(feature = "alpn")))]
-use tokio_tls::{TlsConnector};
+use {
+    native_tls::{Error as SslError, TlsConnector as NativeTlsConnector},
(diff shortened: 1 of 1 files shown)


src/client/connector.rs
```diff
@@ -293,71 +253,32 @@ impl SystemService for ClientConnector {}
...
-                limit_per_host: 0,
-                acquired: 0,
-                acquired_per_host: HashMap::new(),
-                available: HashMap::new(),
-                to_close: Vec::new(),
-                waiters: Some(HashMap::new()),
-                wait_timeout: None,
-                paused: Paused::No,
+            #[cfg(all(feature = "tls", not(feature = "alpn")))]
+            { NativeTlsConnector::builder().build().unwrap().into() }
```
DoumanAsh, line 261: It is a bit not nice looking, I would suggest to not inline {} with expression like that

DoumanAsh:
Thanks! I think it is a good idea to let user customize it, ~but in your current commit I don't see how user can provide configuration yet...~

Sorry, it is a bit messy with all this refactoring so I didn't noticed it is alised as `SslConnector`

@fafhrd91 you're ok with it, too?

untitaker:
I aliased all the various connector types to `SslConnector` s.t. I wouldn't have that many branches.

Even for the case where no SSL support at all is compiled, there is a `SslConnector = ()`. This saves quite a few `#[cfg(..)]` and should not have any impact on runtime.

mitsuhiko (approved):
Looks fine to me

src/client/connector.rs
untitaker: that's fair, I am generally not too happy with the syntax around `#[cfg(..)]`. I wanted to omit the braces but apparently I can't do that for exprs

DoumanAsh: yeah, sadly this is limitation of `cfg` I usually prefer to make it clear by creating visible scope
```
#[cfg]
{
    <expr>
}
```

fafhrd91 (approved):

PR #496, opened 2018-08-31
Fix issue with 'Connection: close' in ClientRequest

Neopallium:
To fix issue #495

### DIFF
--- a/src/client/parser.rs
+++ b/src/client/parser.rs
@@ -39,6 +39,7 @@ impl HttpResponseParser {
         T: IoStream,
     {
         loop {
+            // Read some more data into the buffer for the parser.
             match io.read_available(buf) {
                 Ok(Async::Ready((false, true))) => {
                     return Err(HttpResponseParserError::Disconnect)
@@ -49,6 +50,8 @@ impl HttpResponseParser {
                     return Err(HttpResponseParserError::Error(err.into()))
                 }
             }
+
+            // Call HTTP response parser.
             match HttpResponseParser::parse_message(buf)
                 .map_err(HttpResponseParserError::Error)?
             {
@@ -60,6 +63,7 @@ impl HttpResponseParser {
                     if buf.capacity() >= MAX_BUFFER_SIZE {
                         return Err(HttpResponseParserError::Error(ParseError::TooLarge));
                     }
+                    // Parser needs more data.  Loop and read more data.
                 }
             }
         }
--- a/src/client/parser.rs
+++ b/src/client/parser.rs
@@ -41,7 +41,8 @@ impl HttpResponseParser {
         // if buf is empty parse_message will always return NotReady, let's avoid that
         if buf.is_empty() {
             match io.read_available(buf) {
-                Ok(Async::Ready((_, true))) => {
+                Ok(Async::Ready((true, true))) => (),
+                Ok(Async::Ready((false, true))) => {
                     return Err(HttpResponseParserError::Disconnect)
                 }
                 Ok(Async::Ready((_, false))) => (),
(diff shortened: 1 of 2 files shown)


fafhrd91:
Could you add basic test

Neopallium:
I added test `test_connection_close`.  I checked that it fails without the change from this PR.

src/client/parser.rs
```diff
@@ -41,7 +41,8 @@ impl HttpResponseParser {
         // if buf is empty parse_message will always return NotReady, let's avoid that
         if buf.is_empty() {
             match io.read_available(buf) {
-                Ok(Async::Ready((_, true))) => {
+                Ok(Async::Ready((true, true))) => (),
+                Ok(Async::Ready((false, true))) => {
```
untitaker: This can be merged with a branch further below

untitaker (commented):
Thank you very much for this patch! Do you know of a way to test your bug? I'd imagine sending a large number of requests could do the trick as the server will eventually hit some limit.

src/client/parser.rs
Neopallium: merged with which branch?

untitaker: Sorry, wrong line. I mean that this:

```rust
match io.read_available() {
    Ok(Async::Ready((true, true))) => (),
    Ok(Async::Ready((false, true))) => <err>,
    Ok(Async::Ready((_, false))) => ()
}
```

can be written as:

```rust
match io.read_available() {
    Ok(Async::Ready((false, true))) => <err>,
    Ok(Async::Ready(_)) => ()
}
```

it's not important
}
    Ok(Async::Ready((_, false

Neopallium:
@untitaker I merged those two branches and simplified the code.  I didn't like the duplicate code for handling the call to `read_available()`

untitaker:
I am not sure but I think you're breaking something by doing this (see the comment you removed)

Neopallium:
I known what the comment meant.  The important thing is to try to read data into the buffer before calling the parser.  The `if buf.is_empty()` block is only needed at the start of the request.

Previous logic:
1. if buffer is empty. read some data.
2. parse data in buffer
3. if parser is not ready (needs more data).  read some more data and loop to step 2.

New logic:
1. read some data.
2. parse data in buffer.
3. if parser is not ready (needs more data).  loop to step 1.

The `poll()` method should always try reading data into the buffer before calling the parser.  Since the previous call to the `poll()` method didn't have a complete http response.  So it is a waste to run the parser again just to have it request more data.

Neopallium:
I have added some comments to the code.

fafhrd91:
> The poll() method should always try reading data into the buffer before calling the parser. Since the previous call to the poll() method didn't have a complete http response. So it is a waste to run the parser again just to have it request more data.
> 
That is not true for pipelined request. It is possible that to have multiple complete requests in buf.

Neopallium:
ah.  Ok.  I know how to change it.

src/client/parser.rs
```diff
@@ -38,42 +38,35 @@ impl HttpResponseParser {
...
+                    Async::NotReady => {
+                        if buf.capacity() >= MAX_BUFFER_SIZE {
+                            return Err(HttpResponseParserError::Error(ParseError::TooLarge));
                         }
+                        // Parser needs more data.
                     }
                 }
             }
+            // Read some more data into the buffer for the parser.
+            match io.read_available(buf) {
```
untitaker, line 60: I feel like this kind of thing could very easily go into a busy loop that always reads zero bytes

Neopallium: if `read_available()` isn't able to ready any data that will be handled by the `NotReady` branch which will return from the `poll()` method.

untitaker: Cool, I just tested it with our code and it seems to work fine.

fafhrd91 (approved):

fafhrd91:
thanks!

PR #498, opened 2018-09-01
Fix Issue #490

### LINKED ISSUE #490: missing status text "Not Found" in response status line
```
extern crate actix_web;
use actix_web::{server, App, HttpResponse};

fn main() {
    server::new(|| {
        App::new()
        .prefix("/app")
        .resource("", |r| r.f(|_| HttpResponse::Ok()))  // <- handle `/app` path
        .resource("/", |r| r.f(|_| HttpResponse::Ok())) // <- handle `/app/` path
        .resource("/test", |r| {
            r.get().f(|_| HttpResponse::Ok());
            r.head().f(|_| HttpResponse::MethodNotAllowed());
        })
    }).bind("localhost:3000")
        .unwrap()
        .run();
}
```
```
$ curl -I http://localhost:3000/app/z
HTTP/1.1 404 Not Found  // <= it works here

$ curl -I http://localhost:3000/app
HTTP/1.1 200 OK

$ curl -I http://localhost:3000/ap
HTTP/1.1 404  // <= missing `Not Found` here
```

Neopallium:
Includes a testcase to check for the reason text in an HTTP 404 response.

The router has a default handler for 404 responses that correctly sends the "Not Found" reason in the response, because it uses the normal HTTP Response writing code.

If no handler is found (i.e. no routes), then the HTTP1 handler creates a ServerError task to handle the response writing.  ServerError wasn't including the HTTP status reason text.

Closes #490

### DIFF
--- a/src/server/error.rs
+++ b/src/server/error.rs
@@ -21,7 +21,12 @@ impl HttpHandlerTask for ServerError {
             bytes.reserve(helpers::STATUS_LINE_BUF_SIZE + 1);
             helpers::write_status_line(self.0, self.1.as_u16(), bytes);
         }
+        // Convert Status Code to Reason.
+        let reason = self.1.canonical_reason().unwrap_or("<unknown status code>");
+        io.buffer().extend_from_slice(reason.as_bytes());
+        // No response body.
         io.buffer().extend_from_slice(b"\r\ncontent-length: 0\r\n");
+        // date header
         io.set_date();
         Ok(Async::Ready(true))
     }
--- a/src/server/error.rs
+++ b/src/server/error.rs
@@ -22,7 +22,7 @@ impl HttpHandlerTask for ServerError {
             helpers::write_status_line(self.0, self.1.as_u16(), bytes);
         }
         // Convert Status Code to Reason.
-        let reason = self.1.canonical_reason().unwrap_or("<unknown status code>");
+        let reason = self.1.canonical_reason().unwrap_or("");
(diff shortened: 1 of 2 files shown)


src/server/error.rs
```diff
@@ -21,7 +21,12 @@ impl HttpHandlerTask for ServerError {
             bytes.reserve(helpers::STATUS_LINE_BUF_SIZE + 1);
             helpers::write_status_line(self.0, self.1.as_u16(), bytes);
         }
+        // Convert Status Code to Reason.
+        let reason = self.1.canonical_reason().unwrap_or("<unknown status code>");
```
DoumanAsh: Should we just leave out unknown status code?

tests/test_server.rs
```diff
@@ -932,6 +932,29 @@ fn test_application() {
     assert!(response.status().is_success());
 }
 
+#[test]
+fn test_issue_490() {
```
DoumanAsh: well, I guess you could just name it `test_default_handler_response` 😄

DoumanAsh (approved):

Neopallium:
I checked what nginx uses for unknown status codes.  They don't send a reason text.  So I have updated the PR to do the same.

Neopallium:
FYI, `src/httpresponse` can return `<unknown status code>` for unknown codes.  Maybe that should be changed too.

DoumanAsh:
@Neopallium Thanks, it would be good then to change it there too I guess, but not necessary for this PR 😄

PR #502, opened 2018-09-05
Store visit and login timestamp as well as peer address in the identity cookie

uzytkownik:
Currently if cookie is leaked the only method of revoking it is to change the key (`max-age` is settable but it is still possible to get cookie from browser cache). This change allows the developer to specify the deadlines after which the cookie is no longer accepted

### DIFF
--- a/src/middleware/identity.rs
+++ b/src/middleware/identity.rs
@@ -49,10 +49,12 @@
 //! ```
 use std::cell::RefCell;
 use std::rc::Rc;
+use std::time::SystemTime;
 
 use actix_service::{Service, Transform};
 use futures::future::{ok, Either, FutureResult};
 use futures::{Future, IntoFuture, Poll};
+use serde::{Deserialize, Serialize};
 use time::Duration;
 
 use crate::cookie::{Cookie, CookieJar, Key, SameSite};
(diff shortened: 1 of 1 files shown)


fafhrd91:
Looks good. Could you add test case and changes.md entry

uzytkownik:
Sorry - it seems that I pushed old version for some reason.

fafhrd91:
Could you make it backward compatible with existing cookie format.

uzytkownik:
@fafhrd91 I run into problems with my system, so there is an unexpected delay.

uzytkownik:
@fafhrd91 Done & rebased on top of the tree.

uzytkownik:
@fafhrd91 There seems to be a frequent merge conflicts. Could you review the change?

src/middleware/identity.rs
```diff
@@ -284,84 +286,133 @@ where
...
+struct CookieIdentityExtention {
+    login_timestamp: Option<SystemTime>
 }
 
 impl CookieIdentityInner {
     fn new(key: &[u8]) -> CookieIdentityInner {
+        let key_v2: Vec<u8> = key.iter().chain([1, 0, 0, 0].iter()).map(|e| *e).collect();
         CookieIdentityInner {
             key: Key::from_master(key),
+            key_v2: Key::from_master(&key_v2),
```
fafhrd91, line 319: why do we need second key?

uzytkownik: The old value was essentially unstructured so I did not want to depend on 'does it parse as JSON'. That way it is backward compatible even if username happened to be parsable as JSON (I asked on [crypto.SE](https://crypto.stackexchange.com/questions/62108/is-using-derived-keys-for-hkdf-aes-gcm-safe) before writing the code).

fafhrd91: Can we just enable new identity and handle backward compatibility via parse error. I’d like to drop old identity in 3-4 versions.

uzytkownik: > Can we just enable new identity and handle backward compatibility via parse error.

At that point this depend on how conservative we want to be with what could have been stored in identity. Before this was essentially a free-form UTF-8 string so users might've store JSON.

I'm usually prefer to err on side of caution but it can be handled by parse error as well. It would just produce very strange results when user rollbacks the actix-web to previous version.

Note that current design does not preclude removing second key - it just requires that pre-processing of key is kept.

> I’d like to drop old identity in 3-4 versions.

How would you like to handle the transition? Currently you can rolllback the update of `actix-web` and, as long as you don't use new features, the cookies keep working. Assuming we have two types of keys you can either want to switch v2 cookies immediately or:

 - Introduce v2 cookies for new functionality in version X keep legacy cookies for plain type
 - Switch to v2 cookies everywhere and keep legacy cookies for parsing only
 - Remove legacy cookies

fafhrd91:
thanks!

PR #505, opened 2018-09-09
Fix issue with HttpChannel linked list.

### LINKED ISSUE #504: Linked List of HttpChannels is broken
https://github.com/actix/actix-web/blob/cdb57b840e138a60b6b733648008ca877a916e2b/src/server/channel.rs#L80

should be: `if self.node.is_none() {`

Otherwise the `HttpChannel` object are never added to the `Node` linked list.

Also the `traverse` method should either be renamed `shutdown` or take a closure to call on each element.

I found these issues when trying to dump the status of all active connections (to debug the leaking connections issue).

Neopallium:
Closes #504.

### DIFF
--- a/src/server/channel.rs
+++ b/src/server/channel.rs
@@ -77,7 +77,7 @@ where
     type Error = ();
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        if self.node.is_some() {
+        if self.node.is_none() {
             let el = self as *mut _;
             self.node = Some(Node::new(el));
             let _ = match self.proto {
--- a/src/server/channel.rs
+++ b/src/server/channel.rs
@@ -55,7 +55,7 @@ where
         }
     }
 
-    fn shutdown(&mut self) {
+    pub(crate) fn shutdown(&mut self) {
         match self.proto {
             Some(HttpProtocol::H1(ref mut h1)) => {
                 let io = h1.io();
@@ -232,7 +232,7 @@ impl Node<()> {
         }
     }
 
-    pub(crate) fn traverse<T, H>(&self)
+    pub(crate) fn traverse<T, H, F: Fn(&mut HttpChannel<T, H>)>(&self, f: F)
     where
         T: IoStream,
         H: HttpHandler + 'static,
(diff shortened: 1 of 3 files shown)


fafhrd91:
Could you rebase Pr against master

Neopallium:
rebased

Neopallium:
I just realized that the "update branch" button on github does a merge not a rebase.  Should I do a real rebase with git before sending PR requests?

Normally I don't rebase a branch if it would require a "force update" to push.  But it does make for nicer merges.

DoumanAsh:
Hmmm strangely enough when I try to view all changes it show too few fils and changes...
Could you maybe try to rebase your branch?

Neopallium:
trying to rebase right now.

Neopallium:
Ok.  Finally got a clean rebase.  Now the PR only has 2 commits and 2 files changed.

src/server/http.rs
```diff
@@ -637,7 +637,7 @@ where
 
     fn shutdown(&self, force: bool) {
         if force {
-            self.settings.head().traverse::<TcpStream, H>();
+            self.settings.head().traverse(|ch: &mut HttpChannel<TcpStream, H>| ch.shutdown());
```
DoumanAsh, line 640: Is argument type here necessary?

src/server/channel.rs
```diff
@@ -77,7 +77,7 @@ where
     type Error = ();
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        if self.node.is_some() {
+        if self.node.is_none() {
```
DoumanAsh, line 80: Hm really strange how exactly it supposed to work in past 😄

Neopallium: The linked list seems to only be used for graceful shutdown.

src/server/http.rs
Neopallium: I only added the type because it wouldn't compile without it.  I am new to rust, so maybe I am missing something in the definition of the traverse method?

DoumanAsh: Ah sorry then, I was expecting that Rust would be able to see the type here(since it should see it from `self.settings.head()`), but if not. My  bad, I mistook it.

DoumanAsh (approved):

DoumanAsh:
@Neopallium Could  you please prepare entry in changelog?

Neopallium:
@DoumanAsh Done.  Is the format of the entry correct?

Neopallium:
Also rebased.

DoumanAsh:
@Neopallium Yes, Thank you

DoumanAsh:
@Neopallium I think you have rights to merge, so once CI is passed, feel free to 😄

PR #506, opened 2018-09-10
Don't ignore errors in std::fmt::Debug implementations

uzytkownik:


src/client/response.rs
```diff
@@ -95,12 +95,12 @@ impl ClientResponse {
 
 impl fmt::Debug for ClientResponse {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let res = writeln!(f, "\nClientResponse {:?} {}", self.version(), self.status());
-        let _ = writeln!(f, "  headers:");
+        writeln!(f, "\nClientResponse {:?} {}", self.version(), self.status())?
```
DoumanAsh: Please add `;`

DoumanAsh (changes requested):

src/client/response.rs
uzytkownik: Sorry. It looks like I pushed prematurely.

DoumanAsh (approved):

PR #514, opened 2018-09-18
Try to separate HTTP/1 read & write disconnect handling, to fix #511.

### LINKED ISSUE #511: Websocket server finished() isn't called if client disconnects
I was trying to do the websocket tutorial/examples and implement a chat, and found that my server code wasn't cleaning up clients when they disconnected. Even when running the websocket example, the StreamHandler's finished() method doesn't ever get called. That is, if I add a finished() method to examples/websocket/src/main.rs, similar to the client one, and run:

```bash
cd examples/websocket
cargo run --bin server
```

and also

```bash
cd examples/websocket
cargo run --bin client
```

and then hit CTRL-C on the client, then the server doesn't notice. Whereas if I hit CTRL-C on the server, the client does get a "Server disconnected" message within a few seconds, meaning finished() was called.

(Also, I noticed that the server is using WebsocketContext, while the client is using a regular Context. Shouldn't the client also use WebsocketContext?)

netstat shows that the client closes its half of the socket for writing / sends a FIN (FIN_WAIT2 state) while the server is keeping its half open (CLOSE_WAIT state).

This also happens with current actix-web/master / 0.7.8 and actix 0.7.0. I also tried actix master, but then the example doesn't compile anymore.

sapir:
I'm guessing this change isn't (entirely) correct, but I'm rather confused as to how the read & write streams are handled here, and what exactly is supposed to happen when the client finishes writing (sends a FIN / a "read disconnect") happens before the server is done writing.

Anyway, it seems to fix #511.

### DIFF
--- a/src/server/h1.rs
+++ b/src/server/h1.rs
@@ -136,7 +136,7 @@ where
         }
     }
 
-    fn notify_disconnect(&mut self) {
+    fn write_disconnected(&mut self) {
         self.flags.insert(Flags::WRITE_DISCONNECTED);
 
         // notify all tasks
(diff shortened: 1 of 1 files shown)


DoumanAsh (approved):
It looks good?

But I'm not sure on order of `self.keepalive_timer.take();`
I'd like someone else to look at it too

fafhrd91:
Let’s wait with this pr, I’d like to look closer

fafhrd91:
thanks!

PR #515, opened 2018-09-20
Send response to inform client of error

### LINKED ISSUE #512: server should return 400 or 301 on malformed request URL
Currently, if a request is made where the url path is malformed, the server silently closes the connection.  This causes undesired behavior/errors at the downstream (eg, nginx rev proxy)

Instead, actix should return a 400 or 301 as per RFC 7230 (https://tools.ietf.org/html/rfc7230#section-3.1.1) in this case.

"Recipients of an invalid request-line SHOULD respond with either a 400 (Bad Request) error or a 301"

DoumanAsh:
Closes #512

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## [0.7.9] - 2018-09-x
+
+### Fixed
+
+* HTTP1 decoding errors are reported to the client. #512
+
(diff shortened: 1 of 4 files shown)


fafhrd91:
Changes?

fafhrd91 (approved):

DoumanAsh:
Opps forgot, I'll update changelog

PR #516, opened 2018-09-21
add FromParam for bool

mwarning:
Allow false/true as url parameter:
```
let is_valid : bool = req.match_info().query("is_valid");
```

### DIFF
--- a/src/param.rs
+++ b/src/param.rs
@@ -244,6 +244,7 @@ macro_rules! FROM_STR {
     };
 }
 
+FROM_STR!(bool);
 FROM_STR!(u8);
 FROM_STR!(u16);
 FROM_STR!(u32);


DoumanAsh:
@mwarning Could you please add simple test to check its usage?

mwarning:
hm, I do not know how. This did not compile:
```
assert_eq!(
  Params::with_url(&Url::new("/foo/bar?value=true".parse::<http::Uri>().unwrap())).query("value"),
  Ok(true)
);
```

DoumanAsh:
See example of tests in the same file https://github.com/actix/actix-web/blob/master/src/param.rs#L274-L277

DoumanAsh:
Well it is not necessary now I think about it.
Can you please rebase your changes

DoumanAsh (approved):

DoumanAsh (changes requested):
I remembered one thing. In past we had similar PR but Nikolay decided against it.
https://github.com/actix/actix-web/pull/360

@fafhrd91 Are you still of the same opinion?

mwarning:
I use a GET request to call a method that accepts a bool as parameter. That is why I like to parse a bool.
Right now I have a workaround:
```
let info = req.match_info();
if let (Ok(count), Ok(closed)) = (info.query("count"), info.query("closed")) {
  let closed : u32 = closed;
  add_line(count, closed != 0);
}
```

fafhrd91:
Path extractor seems easier. In you css you can just do

`let data = Path::<(u32, bool)>::extract(&request)`

And if you use handler with extractor it is even easier

`fn index(data: Path<(u32, bool)>) -> HttpResponse `

DoumanAsh:
@fafhrd91 emm so I'm not sure do you mean this PR is not necessary or...?

fafhrd91:
I’d prefer to kill FromParam

mwarning:
@fafhrd91 I do not see how I can apply `extract` to my problem. Does this parse `foo.bar?123&true`?

fafhrd91:
match_info does not contain query, only patch match information. Query is extractor for req query

mwarning:
I can't use a path, this might conflict with existing files.
This might might be a way to get what I need: https://github.com/actix/actix-web/blob/master/src/extractor.rs#L962
But writing a struct seems to be a bit of a hassle. Also, how do I import the Deserialize macro?

fafhrd91:
example with count and closed

```rust
#[derive(Deserialize)]
struct Params {
    count: u32,
    closed: bool,
}
```

and use it like this:

```rust
fn index(params: Query<Params>) -> HttpResponse {
   params.count, params.closed
}
```

fafhrd91:
i will remove FromParams trait in 0.8

mwarning:
ok

I need to match `index.html?count=42&close=true`, but this did not seem to match:
```
if let Ok(p) = Query::<Params>::from_request(&req, &()) {
	println!("count: {}, close: {}", p.count, p.close);
}
```

fafhrd91:
it should, you need to enable logging. also in my example i use `closed`.

https://actix.rs/actix-web/actix_web/struct.Query.html

mwarning:
This is more of my code:
```
#[derive(Deserialize)]
struct Params {
	count: u32,
	close: bool,
}

fn web_handler(req: &HttpRequest) -> HttpResponse {
  if *req.method() != http::Method::GET {
    return HttpResponse::MethodNotAllowed().finish();
  }

  println!("{:?}", req.query()); // prints: {"uid": "1537985931152", "close": "0", "count": "3"}

  if let Ok(p) = Query::<Params>::from_request(&req, &()) {
    println!("count: {}, close: {}", p.count, p.close);
  }
  return HttpResponse::Ok().finish();
}
```
Hm, maybe I have to also parse&ignore that uid my browser is sending.. EDIT: nope, that did not change. The order also does not seem to be important. actix-web 0.7.7

fafhrd91:
What error do you get? Also try to change `close` to `usize`

mwarning:
I changed close to usize and that fixed the problem. Thanks a lot!

mwarning:
Sorry for asking here, what kind of query does `Query::<(String, String)>::from_request(&req, &())` match?
(is there some chat that does not require a login?)

fafhrd91:
I don’t think you can do (String, String) for query

PR #520, opened 2018-09-26
Reduce code covered by `unsafe` blocks

expenses:
Feel free to reject this if its overly pedantic, but I noticed that some of the `unsafe` blocks in the code covered more than they needed to, so I've tried to split them up and reduce the total lines covered by `unsafe` blocks. This means that there are more `unsafe` blocks than there were before however, but I don't think thats an especially good way of measuring code soundness.


fafhrd91:
Looks good! Thanks

DoumanAsh (approved):
Thanks, Looks good

PR #530, opened 2018-10-03
Add support for WebSocket fragmentation.

surban:
This adds support for fragmentation in the WebSocket server and client.
See https://tools.ietf.org/html/rfc6455#section-5.4 for specifications.

### DIFF
--- a/src/ws/client.rs
+++ b/src/ws/client.rs
@@ -273,10 +273,21 @@ impl Client {
     }
 }
 
+enum ContinuationOpCode {
+    Binary,
+    Text
+}
+
+struct Continuation {
+    opcode: ContinuationOpCode,
+    buffer: Vec<u8>,
+} 
+
 struct Inner {
     tx: UnboundedSender<Bytes>,
     rx: PayloadBuffer<Box<Pipeline>>,
     closed: bool,
+    continuation: Option<Continuation>,    
 }
 
 /// Future that implementes client websocket handshake process.
@@ -433,6 +444,7 @@ impl Future for ClientHandshake {
             tx: self.tx.take().unwrap(),
             rx: PayloadBuffer::new(resp.payload()),
             closed: false,
+            continuation: None,
         };
 
         let inner = Rc::new(RefCell::new(inner));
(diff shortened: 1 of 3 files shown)


fafhrd91:
Could you add tests.

fafhrd91:
i think we need some setting for limiting size of the buffer

surban:
I've added an option to limit the continuation buffer size and added the requested tests.

src/ws/mod.rs
```diff
@@ -88,9 +88,9 @@ pub enum ProtocolError {
     /// A payload reached size limit.
     #[fail(display = "A payload reached size limit.")]
     Overflow,
-    /// Continuation is not supported
-    #[fail(display = "Continuation is not supported.")]
-    NoContinuation,
+    /// Bad continuation frame sequence or overflow of continuation buffer.
+    #[fail(display = "Bad continuation frame sequence or overflow of continuation buffer.")]
+    BadContinuation,
```
fafhrd91, line 93: could you add info to MIGRATION.md about `NoContinuation` and `BadContinuation`

fafhrd91 (approved):

src/ws/client.rs
```diff
@@ -192,6 +194,14 @@ impl Client {
         self
     }
 
+    /// Set max continuation size
```
DoumanAsh, line 197: Could you please add in doc comment how we use this size? (i.e. user need to be aware that this is size of buffer we're going to use to store frames until it is finished)

src/ws/client.rs
```diff
@@ -145,6 +146,7 @@ impl Client {
             origin: None,
             protocols: None,
             max_size: 65_536,
+            max_continuation_size: 1_048_576,
```
DoumanAsh, line 149: Why continuation frame has such big size?
I'm not very well familiar with websocket protocol, but I'd expect it to be few times more than default `max_size` but 1mb seems quite big as default.

Though I'm not sure what is common practice, but maybe something around `256kb` would be better as default?
But I don't feel strongly about it, so I leave it to you since Nikolay is ok with it

src/ws/client.rs
```diff
@@ -541,15 +635,68 @@ impl ClientWriter {
     /// Send text frame
     #[inline]
     pub fn text<T: Into<Binary>>(&mut self, text: T) {
+        assert!(
```
DoumanAsh, line 638: Is it likely to happen for user?
Maybe it would be better to have `debug_assert` and print warning in release mode instead of sending frame?

I'm ok with panic in `*_part` methods though as it is mis-use

src/ws/client.rs
```diff
@@ -541,15 +635,68 @@ impl ClientWriter {
...
+            self.continuation == ClientWriterContinuation::Inactive,
+            "Cannot send text frame while continuation is active."
+        );
         self.write(Frame::message(text.into(), OpCode::Text, true, true));
     }
 
+    /// Send a part of a text frame.
+    /// The receiver will concatenate all parts when receiving the final
+    /// frame with `finished == true`.
```
DoumanAsh, line 647: Please add to comment that method will panic when in the middle of Binary Continuation

src/ws/client.rs
```diff
@@ -541,15 +635,68 @@ impl ClientWriter {
...
+        self.continuation = if !finished {
+            ClientWriterContinuation::Text
+        } else {
+            ClientWriterContinuation::Inactive
+        };
+    }
+
     /// Send binary frame
     #[inline]
     pub fn binary<B: Into<Binary>>(&mut self, data: B) {
```
DoumanAsh, line 671: The same comment as for `text` method

DoumanAsh (commented):
I have few questions and suggestion, but overall looks good

DoumanAsh:
@surban ping?

fafhrd91:
Implemented

PR #548, opened 2018-10-15
add missing breaking change

prasannavl:
Looks like the `server::Server` has been moved into actix-net as of the latest version. This is a public api which is a breaking change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@
 
 ### Changed
 
+* [BREAKING] server::Server struct moved into actix-net
+
 * Set min version for actix
 
 * Set min version for actix-net
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,8 +13,6 @@
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


fafhrd91:
Server wasn’t public. All server/acceptor api was hidden

prasannavl:
@fafhrd91 - I made this change, after all my builds broke with the recent update. So, I'm rather certain it was - now, switched to `actix-net::server::Server` and it's fine. 

PS: The `start` returns `Addr<Server>`

prasannavl:
https://github.com/actix/actix-web/blob/65e9201b4d586df303b7d0870cdb55a107d0327b/src/server/http.rs#L473

https://github.com/actix/actix-web/blob/52195bbf167618039ef5b39c9e83c06643052e0b/src/server/server.rs#L61

prasannavl:
That being said. I think it's the wrong version I added the change to. It seems this breaking change was introduced quite a while ago. Looking for the right version now.

fafhrd91:
`Server` wasn't mean to be public

PR #550, opened 2018-10-15
add a way to configure error treatment for Query extractor

mockersf:
The goal is to be able to reply with an error explaining the issue to help the caller fix her call

```rust
#[derive(Debug, Serialize, Deserialize)]
struct MyObj {
    name: String,
    number: i32,
}

fn extract_item(item: Query<MyObj>) -> HttpResponse {
    println!("model: {:?}", &item);
    HttpResponse::Ok().json("ok")
}

fn main() {
    server::new(|| {
        App::new()
            .resource("/qp", |r| r.method(http::Method::GET).with(extract_item))
            .resource("/qp2", |r| {
                r.method(http::Method::GET)
                    .with_config(extract_item, |cfg| {
                        cfg.0.error_handler(|e, _| {
                            error::InternalError::from_response(
                                "missing query parameter",
                                HttpResponse::build(StatusCode::BAD_REQUEST).body(format!("{}", e)),
                            ).into()
                        });
                    })
            })
    }).bind("127.0.0.1:8080")
    .unwrap()
    .run();
}
```

```
> curl -i localhost:8080/qp\?name=1
HTTP/1.1 400 Bad Request
content-length: 0
date: Mon, 15 Oct 2018 17:57:38 GMT

> curl -i localhost:8080/qp2\?name=1
HTTP/1.1 400 Bad Request
content-length: 22
date: Mon, 15 Oct 2018 17:57:32 GMT

missing field `number`
```

### DIFF
--- a/src/extractor.rs
+++ b/src/extractor.rs
@@ -200,14 +200,66 @@ impl<T, S> FromRequest<S> for Query<T>
 where
     T: de::DeserializeOwned,
 {
-    type Config = ();
+    type Config = QueryConfig<S>;
     type Result = Result<Self, Error>;
 
     #[inline]
-    fn from_request(req: &HttpRequest<S>, _: &Self::Config) -> Self::Result {
+    fn from_request(req: &HttpRequest<S>, cfg: &Self::Config) -> Self::Result {
+        let req2 = req.clone();
+        let err = Rc::clone(&cfg.ehandler);
         serde_urlencoded::from_str::<T>(req.query_string())
-            .map_err(|e| e.into())
+            .map_err(move |e| (*err)(e, &req2))
             .map(Query)
+    }
+}
+
+/// Query extractor configuration
+///
+/// ```rust
+/// # extern crate actix_web;
+/// #[macro_use] extern crate serde_derive;
+/// use actix_web::{error, http, App, HttpResponse, Query, Result};
+///
+/// #[derive(Deserialize)]
+/// struct Info {
+///     username: String,
+/// }
+///
+/// /// deserialize `Info` from request's body, max payload size is 4kb
+/// fn index(info: Query<Info>) -> Result<String> {
+///     Ok(format!("Welcome {}!", info.username))
+/// }
+///
+/// fn main() {
+///     let app = App::new().resource("/index.html", |r| {
+///         r.method(http::Method::GET).with_config(index, |cfg| {
+///             cfg.0.error_handler(|err, req| {
+///                 // <- create custom error response
+///                 error::InternalError::from_response(err, HttpResponse::Conflict().finish()).into()
+///             });
+///         })
+///     });
+/// }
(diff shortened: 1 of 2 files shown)


src/extractor.rs
```diff
@@ -200,17 +200,69 @@ impl<T, S> FromRequest<S> for Query<T>
...
+///             cfg.0.error_handler(|err, req| {
+///                 // <- create custom error response
+///                 error::InternalError::from_response(err, HttpResponse::Conflict().finish()).into()
+///             });
+///         })
+///     });
+/// }
+/// ```
+pub struct QueryConfig<S> {
+    ehandler: Rc<Fn(serde_urlencoded::de::Error, &HttpRequest<S>) -> Error>,
```
DoumanAsh, line 291: `ehandler` -> `error_handler`

DoumanAsh, line 291: Hmm just to clarify is `Rc` necessary?

DoumanAsh (commented):
If we would start adding such configuration to `Query` maybe we should do it for all other extractors too?

mockersf:
there is already one for `Json`, which I pretty much copied for `Query` : https://actix.rs/api/actix-web/stable/actix_web/dev/struct.JsonConfig.html

src/extractor.rs
mockersf: copied from https://github.com/actix/actix-web/blob/master/src/json.rs#L186, do you think I should change it there too ?

mockersf: same, copied from the same place, and I don't trust myself enough around `Rc` to remove it

DoumanAsh:
Ah I see, then most likely it would need to be cloned. Not sure if it is needed for function pointer, but let's leave it as it is for now.

What I'm thinking do we need to extend it for other extractors aside from `Json` and `Query`?

DoumanAsh (approved):

DoumanAsh:
@mockersf Btw could you please add entry to CHANGELOG?

mockersf:
Looking at https://actix.rs/api/actix-web/stable/actix_web/trait.FromRequest.html#implementors, `Form` and `Json` can already be configured, leaving `Session`, `State`, `HttpRequest` and `Path`

I think it would make sense for `Path` to be configurable in case of error (you may want to return a specific error if matching with a `u32` in the path and you got something else, the default error handler returns a 404).
I don't see a use case for the three remaining (`Session`, `State` and `HttpRequest`) and looking at the code they don't even have error cases.

I'll add one for `Path` later and add both in the CHANGELOG

DoumanAsh:
Sorry about it, but I just recently merged something too and there are conflicts in CHANGES.md, could you please rebase and resolve conflict?

mockersf:
sure !

DoumanAsh:
Thanks

Issue #560, opened 2018-10-24
Extremely slow StaticFiles peformance


transkatgirl:
When connecting to a HTTP server built with actix-web, if the server uses the builtin fs::StaticFiles method for serving content, there is a severe performance impact in newer versions of actix-web (> 0.7.8).

Here's a simple example server to show the issue:
```rust
extern crate actix_web;
use actix_web::{fs::StaticFiles, server, App};

fn main() {
    server::new(|| {
        App::new()
	        .handler("/",StaticFiles::new("./static/").unwrap().index_file("index.html"))
    }).bind("[::]:8080")
        .unwrap()
        .run();
}
```
(Note that the "index.html" file used for testing is 1024 bytes. When serving large files, the performance impact is much lower.)

Performance on actix-web v0.7.13: 
```
Running 10s test @ http://localhost:8080/index.html
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    41.77ms    8.92ms  47.70ms   95.54%
    Req/Sec   383.49     41.16   540.00     69.00%
  7643 requests in 10.01s, 9.37MB read
Requests/sec:    763.67
Transfer/sec:      0.94MB
```

Performance on actix-web v0.7.8: 
```
Running 10s test @ http://localhost:8080/index.html
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     0.92ms    0.89ms  28.22ms   91.59%
    Req/Sec    19.48k     1.13k   22.51k    71.50%
  387887 requests in 10.01s, 474.98MB read
Requests/sec:  38765.56
Transfer/sec:     47.47MB
```

DoumanAsh:
Hm... interesting, I don't think there should such serious change to StaticFiles.
Thanks for noticing it

fafhrd91:
I enabled timeouts by default in latest versions of actix-web. that might affect default performance, but it should not be that dramatic

transkatgirl:
After a bit of testing, it looks like ChunkedReadFile is what's slowing down StaticFiles so much.


Also, something completely unrelated I noticed: Why is ChunkedReadFile public, but all of it's fields are private? Making it's fields public too would probably be a good idea.

DoumanAsh:
Just a note: there are no significant changes to the way we read static files.

@kittyhacker101 So I'm curious do you see a similar problem with your regular handlers?

transkatgirl:
[...]

PR #571, opened 2018-10-30
Improve StaticFiles performance, and make ChunkedReadFile public

transkatgirl:
This fixes issue #560, and also addresses some concerns brought up about ChunkedReadFile.

Sorry about the weird commit names, I was using my phone to do this.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -351,7 +351,7 @@ impl<C: StaticFileConfig> Responder for NamedFile<C> {
             if let Some(current_encoding) = self.encoding {
                 resp.content_encoding(current_encoding);
             }
-            let reader = ChunkedReadFile {	
+            let reader = ChunkedReadFile {
                 size: self.md.len(),
                 offset: 0,
                 cpu_pool: self.cpu_pool.unwrap_or_else(|| req.cpu_pool().clone()),
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -351,7 +351,8 @@ impl<C: StaticFileConfig> Responder for NamedFile<C> {
             if let Some(current_encoding) = self.encoding {
                 resp.content_encoding(current_encoding);
             }
-            let reader = ChunkedReadFile {
+            let reader = 
+            {
                 size: self.md.len(),
                 offset: 0,
                 cpu_pool: self.cpu_pool.unwrap_or_else(|| req.cpu_pool().clone()),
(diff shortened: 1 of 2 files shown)


src/server/service.rs
```diff
@@ -88,7 +88,8 @@ where
         Ok(Async::Ready(()))
     }
 
-    fn call(&mut self, req: Self::Request) -> Self::Future {
+    fn call(&mut self, mut req: Self::Request) -> Self::Future {
+        let _ = req.set_nodelay(true);
```
DoumanAsh, line 92: @fafhrd91 it is been removed in c674ea912691d86379d610a4794a50cef4b2feac

Do you remember why? Or was it mistake?

fafhrd91: There is separate service for that

src/fs.rs
```diff
@@ -475,12 +475,12 @@ impl<C: StaticFileConfig> Responder for NamedFile<C> {
 /// A helper created from a `std::fs::File` which reads the file
 /// chunk-by-chunk on a `CpuPool`.
 pub struct ChunkedReadFile {
-    size: u64,
-    offset: u64,
-    cpu_pool: CpuPool,
-    file: Option<File>,
-    fut: Option<CpuFuture<(File, Bytes), io::Error>>,
-    counter: u64,
+    pub size: u64,
```
fafhrd91, line 478: What is the reason for this change?

transkatgirl: ChunkedReadFile is public and shows up in the documentation, but it isn't useable in handlers, because its fields are private.

src/server/service.rs
DoumanAsh: You mean `StreamConfigurationService`?

src/fs.rs
transkatgirl: So, are you okay with this change being included in the pull request?

It makes sense for ChunkedReadFile's fields to be public, as it's not useable when the fields are private. Why have ChunkedReadFile show up in the docs if you don't want people to use it?

src/server/service.rs
transkatgirl: Well, this is the simplest solution I found to the issue. If you want to solve the issue in a different way, I could modify the pull request.

kittyhacker101: Well, this is the simplest solution I found to the issue. If you want to solve the issue in a different way, I could modify the pull request.

fafhrd91:
i dont think it make sense to expose `ChunkedReadFile` atm, it is too specific to actix-web. i hid it for now. also added nodelay in master

Issue #556, opened 2018-10-19
URL encoded '+' character (%2B) not decoded by path extractor


davechallis:
When decoding using the `Path` extractor, if the path contains `%2B`, then it's extracted as is, instead of decoding to a `+` character.

E.g. using a minimal example server:

```
extern crate actix;
extern crate actix_web;
use actix_web::{Path, server, App, FromRequest, HttpRequest};

fn index(req: &HttpRequest) -> String { Path::<String>::extract(req).unwrap().into_inner() }

fn main() {
    let sys = actix::System::new("foo");
    server::new(|| App::new() .resource("/{path}", |r| r.f(index)))
        .bind("127.0.0.1:8080").unwrap().start();
    let _ = sys.run();
}
```

Then querying with e.g. the following URL encoded characters `@£$%^&+=` (`%40%C2%A3%24%25%5E%26%2B%3D`) results in all of the except for `%2B` being decoded by actix:

```
$ curl '127.0.0.1:8080/%40%C2%A3%24%25%5E%26%2B%3D'
@£$%^&%2B=
```

Expected: `@£$%^&+=`

mockersf:
This is related to https://github.com/actix/actix-web/blob/master/src/uri.rs#L35:
```rust
lazy_static! {
    static ref DEFAULT_QUOTER: Quoter = { Quoter::new(b"@:", b"/+") };
}
```
`+` is a protected character when decoding `%`-encoded strings.

It was introduced in #182

davechallis:
Was this an intentional change or a bug introduced? If deliberate, it's definitely one area where actix-web differs from almost all other web frameworks I've used, where all percent encoded strings in paths are decoded.

Currently it means that there's no way to differentiate between a `%`-encoded `+`, and a `%`-encoded `%` followed by the characters `2B`, e.g. using the server example above, both of these queries give an identical response from path decoding:

    $ curl '127.0.0.1:8080/%2B'
    %2B

    $ curl '127.0.0.1:8080/%252B'
    %2B

So manually adding an additional decoding step would be needed to get the correct path from the 1st example, but then it would give the wrong result for the 2nd example.

DoumanAsh:
@mockersf Do you think there are risks if we would stop retaining encoded `+`?

davechallis:
Looking at the code linked to, I guess the same issue applies to quoting of `/` too, e.g. there's no way to differentiate between:

    $ curl 'localhost:8080/%2F'
    %2F

[...]

PR #577, opened 2018-11-04
decode reserved characters when extracting path with configuration

mockersf:
concerning #556

I limited default decoding to unreserved characters as per [RFC3986](
https://tools.ietf.org/html/rfc3986#section-2.4) those can be decoded anytime.

I then added an option to path extractor (enabled by default) to also decoded the reserved characters.

I'm open to discussion if you think anything can be improved

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -16,6 +16,20 @@ macro_rules! unsupported_type {
     };
 }
 
+macro_rules! percent_decode_if_needed {
+    ($value:expr, $decode:expr) => {
+        if $decode {
+            if let Some(ref value) = RESERVED_QUOTER.requote($value.as_bytes()) {
+                Rc::make_mut(&mut value.clone()).parse()
+            } else {
+                $value.parse()
+            }
+        } else {
+            $value.parse()
+        }
+    }
+}
+
 macro_rules! parse_single_value {
     ($trait_fn:ident, $visit_fn:ident, $tp:tt) => {
         fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
@@ -26,19 +40,10 @@ macro_rules! parse_single_value {
                     format!("wrong number of parameters: {} expected 1",
                             self.req.match_info().len()).as_str()))
             } else {
-                let v: &str = &self.req.match_info()[0];
-                let v_parsed = if self.decode {
-                    let decoded = RESERVED_QUOTER.requote(v.as_bytes());
-                    if let Some(ref value) = decoded {
-                        Rc::make_mut(&mut value.clone()).parse()
-                    } else {
-                        v.parse()
-                    }
-                } else {
-                    v.parse()
-                }.map_err(|_| de::value::Error::custom(
-                    format!("can not parse {:?} to a {}", &self.req.match_info()[0], $tp)
-                ))?;
+                let v_parsed = percent_decode_if_needed!(&self.req.match_info()[0], self.decode)
+                    .map_err(|_| de::value::Error::custom(
+                        format!("can not parse {:?} to a {}", &self.req.match_info()[0], $tp)
+                    ))?;
                 visitor.$visit_fn(v_parsed)
             }
         }
@@ -256,19 +261,10 @@ macro_rules! parse_value {
         fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
             where V: Visitor<'de>
         {
-            let v: &str = &self.value;
-            let v_parsed = if self.decode {
-                let decoded = RESERVED_QUOTER.requote(v.as_bytes());
-                if let Some(ref value) = decoded {
-                    Rc::make_mut(&mut value.clone()).parse()
-                } else {
-                    v.parse()
-                }
-            } else {
-                v.parse()
-            }.map_err(|_| de::value::Error::custom(
-                format!("can not parse {:?} to a {}", &self.value, $tp)
-            ))?;
+            let v_parsed = percent_decode_if_needed!(&self.value, self.decode)
+                .map_err(|_| de::value::Error::custom(
+                    format!("can not parse {:?} to a {}", &self.value, $tp)
+                ))?;
             visitor.$visit_fn(v_parsed)
         }
     }
--- a/src/de.rs
+++ b/src/de.rs
@@ -1,7 +1,10 @@
+use std::rc::Rc;
+
 use serde::de::{self, Deserializer, Error as DeError, Visitor};
 
 use httprequest::HttpRequest;
 use param::ParamsIter;
+use uri::RESERVED_QUOTER;
 
 macro_rules! unsupported_type {
     ($trait_fn:ident, $name:expr) => {
@@ -23,11 +26,20 @@ macro_rules! parse_single_value {
                     format!("wrong number of parameters: {} expected 1",
                             self.req.match_info().len()).as_str()))
             } else {
-                let v = self.req.match_info()[0].parse().map_err(
-                    |_| de::value::Error::custom(
-                        format!("can not parse {:?} to a {}",
-                                &self.req.match_info()[0], $tp)))?;
-                visitor.$visit_fn(v)
+                let v: &str = &self.req.match_info()[0];
+                let v_parsed = if self.decode {
+                    let decoded = RESERVED_QUOTER.requote(v.as_bytes());
+                    if let Some(ref value) = decoded {
+                        Rc::make_mut(&mut value.clone()).parse()
+                    } else {
+                        v.parse()
+                    }
+                } else {
+                    v.parse()
+                }.map_err(|_| de::value::Error::custom(
+                    format!("can not parse {:?} to a {}", &self.req.match_info()[0], $tp)
+                ))?;
+                visitor.$visit_fn(v_parsed)
             }
         }
     }
@@ -35,11 +47,12 @@ macro_rules! parse_single_value {
 
 pub struct PathDeserializer<'de, S: 'de> {
     req: &'de HttpRequest<S>,
+    decode: bool,
 }
 
 impl<'de, S: 'de> PathDeserializer<'de, S> {
-    pub fn new(req: &'de HttpRequest<S>) -> Self {
-        PathDeserializer { req }
+    pub fn new(req: &'de HttpRequest<S>, decode: bool) -> Self {
+        PathDeserializer { req, decode }
     }
 }
 
@@ -53,6 +66,7 @@ impl<'de, S: 'de> Deserializer<'de> for PathDeserializer<'de, S> {
         visitor.visit_map(ParamsDeserializer {
             params: self.req.match_info().iter(),
             current: None,
+            decode: self.decode,
         })
     }
 
@@ -107,6 +121,7 @@ impl<'de, S: 'de> Deserializer<'de> for PathDeserializer<'de, S> {
         } else {
             visitor.visit_seq(ParamsSeq {
                 params: self.req.match_info().iter(),
+                decode: self.decode,
             })
         }
     }
@@ -128,6 +143,7 @@ impl<'de, S: 'de> Deserializer<'de> for PathDeserializer<'de, S> {
         } else {
             visitor.visit_seq(ParamsSeq {
                 params: self.req.match_info().iter(),
+                decode: self.decode,
             })
         }
     }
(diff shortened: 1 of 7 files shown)


mockersf:
a few issue when looking at RFC:
* Some characters are neither in reserved or unreserved sets: ``< > " \ ` { } |``. I decided to decode them on the second pass
* Non ascii characters are not mentioned in the RFC. They get decoded on the first pass to help match routes

src/de.rs
```diff
@@ -252,16 +256,27 @@ macro_rules! parse_value {
         fn $trait_fn<V>(self, visitor: V) -> Result<V::Value, Self::Error>
             where V: Visitor<'de>
         {
-            let v = self.value.parse().map_err(
-                |_| de::value::Error::custom(
-                    format!("can not parse {:?} to a {}", self.value, $tp)))?;
-            visitor.$visit_fn(v)
+            let v: &str = &self.value;
+            let v_parsed = if self.decode {
+                let decoded = RESERVED_QUOTER.requote(v.as_bytes());
```
DoumanAsh: It seems the code should be re-used here, could you please put it into a separate function

src/extractor.rs
```diff
@@ -159,12 +160,20 @@ impl<S> PathConfig<S> {
         self.ehandler = Rc::new(f);
         self
     }
+
+    /// Disable decoding
```
DoumanAsh: Please describe in more details what kind of decoding is done

src/uri.rs
```diff
@@ -7,18 +7,24 @@ const GEN_DELIMS: &[u8] = b":/?#[]@";
 const SUB_DELIMS_WITHOUT_QS: &[u8] = b"!$'()*,";
 #[allow(dead_code)]
 const SUB_DELIMS: &[u8] = b"!$'()*,+?=;";
+
+// https://tools.ietf.org/html/rfc3986#section-2.2
 #[allow(dead_code)]
-const RESERVED: &[u8] = b":/?#[]@!$'()*,+?=;";
+const RESERVED: &[u8] = b":/?#[]@!$&'()*,+?;=";
+#[allow(dead_code)]
```
DoumanAsh: `allow(dead_code)` is not needed

src/extractor.rs
```diff
@@ -1090,6 +1099,50 @@ mod tests {
         assert_eq!(*Path::<i8>::from_request(&req, &&PathConfig::default()).unwrap(), 32);
     }
 
+    #[test]
+    fn test_extract_path_decode() {
```
DoumanAsh, line 1104: Could you maybe add few test cases from the relevant issue like:
decoding `file%2F%2Fvar%2Flog%2Fsyslog`

src/uri.rs
```diff
@@ -7,18 +7,24 @@ const GEN_DELIMS: &[u8] = b":/?#[]@";
 const SUB_DELIMS_WITHOUT_QS: &[u8] = b"!$'()*,";
 #[allow(dead_code)]
 const SUB_DELIMS: &[u8] = b"!$'()*,+?=;";
+
+// https://tools.ietf.org/html/rfc3986#section-2.2
 #[allow(dead_code)]
```
DoumanAsh: `allow(dead_code)` is not needed?

src/uri.rs
```diff
@@ -7,18 +7,24 @@ const GEN_DELIMS: &[u8] = b":/?#[]@";
...
-const RESERVED: &[u8] = b":/?#[]@!$'()*,+?=;";
+const RESERVED: &[u8] = b":/?#[]@!$&'()*,+?;=";
+#[allow(dead_code)]
+const RESERVED_PLUS_EXTRA: &[u8] = b":/?#[]@!$&'()*,+?;=%^ <>\"\\`{}|";
+
+// https://tools.ietf.org/html/rfc3986#section-2.3
+#[allow(dead_code)]
```
DoumanAsh: `allow(dead_code)` is not needed

src/uri.rs
```diff
@@ -7,18 +7,24 @@ const GEN_DELIMS: &[u8] = b":/?#[]@";
...
-    ABCDEFGHIJKLMNOPQRSTUVWXYZ
-    1234567890
-    -._~";
 const ALLOWED: &[u8] = b"abcdefghijklmnopqrstuvwxyz
     ABCDEFGHIJKLMNOPQRSTUVWXYZ
     1234567890
     -._~
     !$'()*,";
+#[allow(dead_code)]
 const QS: &[u8] = b"+&=;b";
```
DoumanAsh: This is no longer needed right?
I think you can remove it

src/uri.rs
```diff
@@ -126,8 +124,12 @@ impl Quoter {
                                 idx += 1;
                                 continue;
                             }
+
+                            buf.extend_from_slice(&pct);
```
DoumanAsh, line 95: Is this code supposed to be reached?

DoumanAsh (changes requested):

DoumanAsh:
Thank you, I have concern that we might consider it as breaking changes though.

@fafhrd91 If you have time, I'd like you to take a look too

DoumanAsh:
>Non ascii characters are not mentioned in the RFC. They get decoded on the first pass to help match routes

Most browsers would encode non-ascii characters also

src/de.rs
mockersf: best I could do was move it to another macro, function was complicated because of types

src/uri.rs
mockersf: yes it can be reached, it's the default case when we don't want to decode the character. I removed all the dead code, it should be clearer

src/extractor.rs
mockersf: I added several cases from the issue

mockersf:
I agree it is a breaking change, the following characters are not decoded on first pass anymore: `'@:!$'()* ,`
it should not break for anyone not doing anything too strange, but still...

tests/test_handlers.rs
```diff
@@ -672,6 +672,6 @@ fn test_unsafe_path_route() {
     let bytes = srv.execute(response.body()).unwrap();
     assert_eq!(
         bytes,
-        Bytes::from_static(b"success: http:%2F%2Fexample.com")
```
fafhrd91, line 675: i think `.match_info()` should contain unquoted values

fafhrd91:
i think breaking change is not that big, just add to entry to MIGRATION.md

src/de.rs
DoumanAsh: that's fine, I just prefer to have less code repetition

tests/test_handlers.rs
DoumanAsh: Wouldn't it be better for `Params` to have method that can unquote me so that user would be able to unquote on demand?

DoumanAsh (approved):
Looks good, could you please fill CHANGELOG.md and MIGRATION.md with information about what will be changed

fafhrd91:
@mockersf ping

mockersf:
After testing, the "breaking" is even more limited, only the `'  '` character is changing because pretty much no one is respecting the RFC and the other characters are not automatically encoded by most http client

DoumanAsh:
@mockersf Hm... would it  be ok for us to leave `' '` un-decoded? I kinda think it might be inconvenient too much for users

mockersf:
That a choice of convenience vs rfc, but it's possible and easy to change.

I just tested two other http frameworks I use:
* Flask in Python is decoding spaces
* Play in Scala is not even compiling with a space in a route

So... that's your choice, I'm OK with both

DoumanAsh:
I'm not sure to be honest, I suspect many users would prefer to specify routes with plain spaces,
@fafhrd91 thoughts?

fafhrd91:
i'd split path into segments first, then url decode each of segments and then match against resources

davechallis:
I'd agree with @fafhrd91 , that sounds like the most correct (and unsurprising) way of handling it.

In terms of convenience vs. RFC, I'd side in favour of the RFC (again for least surprise when starting to use actix-web, or when frequently switching between languages). Though providing a convenience function or type could be an option, allowing users to switch between providing raw/encoded routes (but at least it would be explicit then).

Also, are spaces in routes common enough to warrant convenience over the RFC? I've always found them to be very rare in most sites/APIs I've come across (spaces seem to be much more common in dynamic path components, e.g. usernames, file uploads, document names, etc.).

DoumanAsh:
I don't really feel strongly about as I do not use spaces in routes, so I'm fine with either way.

DoumanAsh:
@mockersf Let's follow up with @fafhrd91 suggestion?

mockersf:
This is pretty much what is happening with this PR except for two class of characters:
* Characters in the unreserved set from the RFC (`abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-._~`). This lets actix match both `/a` and `/%61` to route `/a`, in the unlikely case where a client would encode every character in the url, which it is allowed to do
* Characters outside of ASCII are decoded before splitting into path segment. They are not mentioned in the RFC, I think it's better to be able to write them unencoded in routes, and Rust let you put `"😀"` which is more readable than `"%F0%9F%98%80"`

DoumanAsh:
@mockersf Oh so we're ready with it?

mockersf:
I updated the PR with the 0.7.14 release

yes, what this changes is:
* some characters that shouldn't have been decoded before matching routes are not anymore : `"!$'()*,"` and `" "`, space being more notable as it's the only one most http client will automatically encode
* every characters are now decoded when extracting path parameters by default, and this can be disabled. previously there was no mean to decode the remaining characters

DoumanAsh:
Yeah, it looks good.

Just a clarification for @fafhrd91 query, what about [Params](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.Params.html) ?

mockersf:
first decoding is done when creating the [`uri::Url`](https://github.com/actix/actix-web/blob/master/src/uri.rs#L45). Raw `http::Uri` is stored in `uri`, decoded `String` is stored in `path`
`Params` are built from the [`path`](https://github.com/actix/actix-web/blob/master/src/param.rs#L158), so they are have the decoded allowed characters
Then `Path` extractor takes those `Params` and decode the remaining characters

DoumanAsh:
@mockersf Hmm What I think is that `Params` itself should be able to have method to decode self, in case user wouldn't use `Path` extractor

mockersf:
@DoumanAsh I tried that today, but I quickly get stuck
`Params` is working on `&str` with no copy, but decoding the string implies making a copy of it, and I lose the lifetime and can't go back to a `&str` that lives long enough...

DoumanAsh:
@mockersf Yes, since it is `&str` we need method like `fn decode(&self) -> String`, so that user would call it by himself, if needed. But otherwise we do not need to do it automatically

mockersf:
I added a [`fn get_decoded(&self, key: &str) -> Option<String>`](https://github.com/actix/actix-web/pull/577/files#diff-e89b073ce3beca3ee7465aea55b9067eR107) method to `Params` mirroring [`fn get(&self, key: &str) -> Option<&str>`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.Params.html#method.get)

But I think `Params` are more often used with [`params.iter()`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.Params.html#method.iter), [`params[0]`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.Params.html#impl-Index%3Cusize%3E) or [`params["key"]`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.Params.html#impl-Index%3C%26%27a%20str%3E). I can add a `decoded_iter()` method that would give an iterator that would decode everything, but `Index` impls can't be overloaded to return either raw or decoded. So I think leaving those to raw path components is better

DoumanAsh:
@mockersf Yeah I think it is good to be like that

@fafhrd91 I'd like your approve on review

src/param.rs
```diff
@@ -103,6 +103,17 @@ impl Params {
         }
     }
 
+    /// Get URL-decoded matched parameter by name without type conversion
+    pub fn get_decoded(&self, key: &str) -> Option<String> {
+        self.get(key).map(|value| {
+            if let Some(ref mut value) = RESERVED_QUOTER.requote(value.as_bytes()) {
```
DoumanAsh, line 109: I think match would work here better, but it is only style question, I leave it up to you

DoumanAsh (approved):
Thanks for your work on it, looks great

src/param.rs
mockersf: [Clippy disagrees...](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#single_match_else)

DoumanAsh: Ah yeah, here clippy is bad, I totally disagree with it.
But let's leave it at that.

fafhrd91 (approved):

DoumanAsh:
@mockersf Sorry about it, but could you resolve conflict in `CHANGES.md`?

mockersf:
sure !

DoumanAsh:
Thank you

PR #578, opened 2018-11-05
fix keep-alive bug

legokichi:
see https://github.com/actix/actix-web/issues/439#issuecomment-435802115


fafhrd91 (commented):
This is wrong solution

legokichi:
sure

PR #584, opened 2018-11-08
Change htmlescape in favor of askama_escape

zzau13:
I have adjusted the lifetime of `file_name` and `file_url` because they did not need beyond the if.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -11,10 +11,10 @@ use std::{cmp, io};
 #[cfg(unix)]
 use std::os::unix::fs::MetadataExt;
 
+use askama_escape::{escape as escape_html_entity};
 use bytes::Bytes;
 use futures::{Async, Future, Poll, Stream};
 use futures_cpupool::{CpuFuture, CpuPool};
-use htmlescape::encode_minimal as escape_html_entity;
 use mime;
 use mime_guess::{get_mime_type, guess_mime_type};
 use percent_encoding::{utf8_percent_encode, DEFAULT_ENCODE_SET};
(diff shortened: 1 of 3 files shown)


src/fs.rs
```diff
@@ -561,6 +561,20 @@ impl Directory {
     }
 }
 
+// show file url as relative to static path
+macro_rules! encode_file_url {
```
DoumanAsh, line 565: Why macros is necessary?

DoumanAsh (commented):
Just clarification: is there any benefit for us to use `askama_escape`?

src/fs.rs
zzau13: It is a macro sanitize, so as not to repeat myself. It is necessary to limit the lifetime of the strings within the scope of the `if`. It has been embedded in the commit because it is in the same logic.

zzau13:
The main motivation is performance, I have to do the comparison but I already announce that the increase is better than 50%. On the other hand, the current library, https://github.com/veddan/rust-htmlescape, is not maintained years ago.

src/fs.rs
DoumanAsh: Couldn't it be done as inline function?

zzau13: With this implementation I still have problems in the life time of the borrowed values
```
// show file url as relative to static path
#[inline]
fn encode_file_url<'a>(path: Cow<'a, str>) -> PercentEncode<'a, DEFAULT_ENCODE_SET> {
    utf8_percent_encode(&path, DEFAULT_ENCODE_SET)
}

// " -- &quot;  & -- &amp;  ' -- &#x27;  < -- &lt;  > -- &gt;  / -- &#x2f;
#[inline]
fn encode_file_name<'a>(name: Cow<'a, str>) -> Escaped<'a> {
    escape_html_entity(&name)
}
```

DoumanAsh: Does it complain because of of intermediate function there?

zzau13: Exactly, in the borrow path and borrow name with the typical error:
```
escape_html_entity(&name)                                                     
                   ^^^^ borrowed value does not live long enough  
```

DoumanAsh: Oh, ok, let's leave it as it is then

DoumanAsh (approved):

src/fs.rs
zzau13: Ok thanks. My first commit to actix-web, it will not be the last one :smiley:

fafhrd91:
thanks!

PR #591, opened 2018-11-18
Complete error helper functions.

wayeast:
Just adding helper methods for the remaining 4xx and 5xx http error codes.


DoumanAsh (approved):
Thank you

PR #618, opened 2018-12-08
add impl FromRequest for Either<A,B> 

axos88:
This PR adds an extractor for Either<A,B>.
The code got a wee bit ugly, so I'd appreciate any input on how to make it nicer.

There is also a huge caveat that holds the same for Tuple<> extractors, so we may want to document it there as well?


src/extractor.rs
```diff
@@ -634,6 +635,151 @@ where
...
+///     });
+/// }
+/// ```
+impl<A: 'static, B: 'static, S: 'static> FromRequest<S> for Either<A,B> where A: FromRequest<S>, B: FromRequest<S> {
+    type Config = EitherConfig<A,B,S>;
+    type Result = AsyncResult<Either<A,B>>;
+
+    #[inline]
+    fn from_request(req: &HttpRequest<S>, cfg: &Self::Config) -> Self::Result {
+        let a = A::from_request(&req.clone(), &cfg.a).into().map(|a| Either::A(a));
```
DoumanAsh, line 715: To be honest I'm not sure if this approach overall is good idea. And neither have I idea on how to make it better, because due to `&HttpRequest` we need to start extracting it immediately

What would be motivation for using such extractor at all?

axos88: What do you mean by we need to start extracting it immediately?

My motivation for this is that I have multiple ways to authenticate a request (api key, oauth2, etc).

So I'd want to extract the auth info using Either<ApiKeyAuth, OAuth2Auth>. the request body can then decide if it cares only that the request is authenticated, or it needs to do further authz checks.

This wouldn't be that much different from the Option<A> or Result<A> extractors.

axos88: And I think that your concerns hold for the tuple extractors as well, although I'm not 100% what you are referring to.

DoumanAsh: >What do you mean by we need to start extracting it immediately?

I mean that current way(extracting both at once) cannot be avoided as we only have `&HttpRequest` and you cannot carry it into combinators

>And I think that your concerns hold for the tuple extractors as well, 

In case of `Either` though you only need single value, while tuple should have both, no?

>This wouldn't be that much different from the Option or Result extractors.

This is actually because you may do twice of work while Result/Option would extract only single one.

I personally don't see it as a _big_ problem(we only need to document this behavior), but whether it is good idea to be in actix itself I'm not sure.

fafhrd91:
@DoumanAsh let’s onclude this extractor. 

But I agree behavior should be documented.

DoumanAsh:
Ok, looking at code, I see that docs are already pretty detailed

DoumanAsh (approved):

DoumanAsh:
Ah yes, @axos88 can you please add note to `CHNAGES.md` that `Either` can be extractor now

fafhrd91:
@axos88 @DoumanAsh we need to port this to 1.0

axos88:
I haven't been following changes closely. Is there a doc about migrating / list of changes available?

fafhrd91:
migration list is not available. but extractor api is almost the same

axos88:
I might have some time next week to take a look. When is the 1.0 release planned?

fafhrd91:
let's wait a bit. i am thinking about slightly different approach

PR #619, opened 2018-12-10
impl ResponseError for SendError when possible

### LINKED ISSUE #528: Improve ergonomics of SendError conversion
As things stand right now bubbling up the SendError returned by `try_send` in a function returning `actix_web::Result` or `Box<Future<Item = HttpResponse, Error = actix_web::Error>>` does not work with just the `?` operator because there is no automatic conversion glue from `SendError` to `actix_web::Error`. Bubbling up a `SendError` seems like a common task to want to do, so it might be nice to have such conversion glue in place. I managed to get this to work with the following two patches.

Patch to actix:
```
diff --git a/src/address/mod.rs b/src/address/mod.rs
index 11c11d7..2c3d821 100644
--- a/src/address/mod.rs
+++ b/src/address/mod.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use failure;
 
 pub(crate) mod channel;
 mod envelope;
@@ -55,6 +56,10 @@ impl<T> fmt::Display for SendError<T> {
     }
 }
 
+impl<T> failure::Fail for SendError<T>
+where T: Send + Sync + 'static {
+}
+
 impl fmt::Debug for MailboxError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         write!(fmt, "MailboxError({})", self)
```

Patch to actix-web:
```
diff --git a/src/error.rs b/src/error.rs
index 76c8e79..28b6c38 100644
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,7 +5,7 @@ use std::string::FromUtf8Error;
 use std::sync::Mutex;
 use std::{fmt, io, result};
 
-use actix::MailboxError;
+use actix::{MailboxError, SendError};
 use cookie;
 use failure::{self, Backtrace, Fail};
 use futures::Canceled;
@@ -136,6 +136,10 @@ pub trait ResponseError: Fail + InternalResponseErrorAsFail {
     }
 }
 
+impl<T> ResponseError for SendError<T>
+where T: Send + Sync + 'static {
+}
+
 impl fmt::Display for Error {
         fmt::Display::fmt(&self.cause, f)
```

It won't work for messages that aren't `Send + Sync + 'static`, but it might be nice to have for simple cases.

I could have the API design wrong here. Maybe it's too surprising for a `addr.try_send(msg)?` to suddenly stop working once `msg` is no longer `'static`.

ethanpailes:
#528 explains the motivations for this change.

I took the liberty of update the CHANGES.md to include a note on this change, cause I saw some chatter about doing so in other PRs. If I should let the maintenance team do that, or format it in a different way, just let me know.

Closes #528

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Changes
 
+* `SendError` now implements `ResponseError`.
+
 ## [0.7.15] - 2018-12-05
 
 ## Changed
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,7 +5,7 @@ use std::string::FromUtf8Error;
 use std::sync::Mutex;
 use std::{fmt, io, result};
 
-use actix::MailboxError;
+use actix::{MailboxError, SendError};
 use cookie;
 use failure::{self, Backtrace, Fail};
 use futures::Canceled;
@@ -134,6 +134,10 @@ pub trait ResponseError: Fail + InternalResponseErrorAsFail {
     fn error_response(&self) -> HttpResponse {
         HttpResponse::new(StatusCode::INTERNAL_SERVER_ERROR)
     }
+}
+
+impl<T> ResponseError for SendError<T>
+where T: Send + Sync + 'static {
 }
 
 impl fmt::Display for Error {


CHANGES.md
```diff
@@ -1,5 +1,7 @@
 # Changes
 
+* `SendError` now implements `ResponseError`.
```
DoumanAsh: It is more suited for `Added` section rather than `Changes`.

Since we did add `ResponseError` impl for `actix::SendError`

DoumanAsh (approved):
Thank you, small comment

CHANGES.md
ethanpailes: Ah, I missed that section. I'll fix it when I get a chance.

DoumanAsh: I'll correct it myself

PR #622, opened 2018-12-11
Prepare release 0.7.16

DoumanAsh:
@fafhrd91 Would we like to not re-export full `actix` at all?


src/extractor.rs
```diff
@@ -764,6 +764,9 @@ impl Default for EitherCollisionStrategy {
     }
 }
 
+///Determines Either extractor configuration
+///
+///By default uses default `EitherCollisionStrategy`
```
axos88:
```suggestion
///By default uses default `EitherCollisionStrategy::FastestSuccessful`
```

DoumanAsh: Oh damn, I forgot to add variant itself, thx for noticing it

src/extractor.rs
```diff
@@ -764,6 +763,9 @@ impl Default for EitherCollisionStrategy {
     }
 }
 
+///Determines Either extractor configuration
+///
+///By default uses default `EitherCollisionStrategy::FastestSuccessful`
```
axos88:
```suggestion
///By default uses `EitherCollisionStrategy::FastestSuccessful`
```

axos88: Sorry, just noticed this now :/

PR #632, opened 2018-12-16
H1 decoded should ignore header cases

### LINKED ISSUE #631: keepalives not working
I'm testing keepalive requests with the following code but when using Apache bench but I don't get keepalives. 

```
extern crate actix_web;
use actix_web::{server, App, HttpRequest};

fn hello_world(_req: &HttpRequest) -> &'static str {
    "Hello World!"
}

fn main() {
    server::new(|| App::new().resource("/", |r| r.f(hello_world)))
        .keep_alive(server::KeepAlive::Tcp(75))
        .bind("0.0.0.0:8080")
        .expect("Can not bind to port 8080")
        .run();
}
```

```
ab -k -n 300000 -c 10 http://10.1.11.206:8080/
...
Concurrency Level:      10
Time taken for tests:   17.192 seconds
Complete requests:      62671
Failed requests:        0
Write errors:           0
Keep-Alive requests:    0
Total transferred:      8084817 bytes
HTML transferred:       752076 bytes
Requests per second:    3645.36 [#/sec] (mean)
Time per request:       2.743 [ms] (mean)
Time per request:       0.274 [ms] (mean, across all concurrent requests)
Transfer rate:          459.24 [Kbytes/sec] received
```

This is showing no requests where Keep-Alive and the performance is pretty bad. This is on Centos 7.6, rustc 1.31.0 and  setting actix-web = "0.7". Am I missing something in the docs that i need to set ?

DoumanAsh:
Fixes #631 

P.s. I think `http` lib doesn't normalize it since we're performing unchecked initialization

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## [0.7.17] - 2018-xx-xx
+
+### Fixed
+
+* HTTP1 decoder should perform case-insentive comparison for client requests (e.g. `Keep-Alive`). #631
+
 ## [0.7.16] - 2018-12-11
 
 ### Added
--- a/src/server/h1.rs
+++ b/src/server/h1.rs
@@ -942,6 +942,14 @@ mod tests {
         let req = parse_ready!(&mut buf);
 
         assert!(!req.keep_alive());
+
+        let mut buf = BytesMut::from(
+            "GET /test HTTP/1.1\r\n\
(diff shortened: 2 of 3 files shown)


fafhrd91 (approved):

fafhrd91:
Could you submit same patch to actix-http, I am planing to replace http parser with one from actix-http

DoumanAsh:
kk

PR #637, opened 2018-12-18
Support custom content types in JsonConfig

### LINKED ISSUE #635: Parse JSON body when content type is text/plain
To support legacy clients of an API, we have a [requirement](https://github.com/mozilla-services/syncstorage-rs/issues/113) to extract JSON bodies when `Content-Type` is `text/plain`. Unless I'm missing something, it seems like this isn't possible out of the box due to the implementation of `actix_web::dev::JsonBody`.

Is there any appetite for a way to work round this, possibly via extending `JsonConfig`? I'd be happy to work on a PR if so.

philbooth:
Fixes #635.

It's my first time poking around in this codebase so I'm not sure what you'll think about the approach. Apologies if I've gone about it the wrong way, I'm happy to address any feedback you have.

r?

### DIFF
--- a/src/httpmessage.rs
+++ b/src/httpmessage.rs
@@ -200,7 +200,7 @@ pub trait HttpMessage: Sized {
     /// # fn main() {}
     /// ```
     fn json<T: DeserializeOwned>(&self) -> JsonBody<Self, T> {
-        JsonBody::new(self)
+        JsonBody::new::<()>(self, None)
     }
 
     /// Return stream to http payload processes as multipart.
--- a/src/json.rs
+++ b/src/json.rs
@@ -143,7 +143,7 @@ where
         let req2 = req.clone();
         let err = Rc::clone(&cfg.ehandler);
         Box::new(
-            JsonBody::new(req)
+            JsonBody::new(req, Some(cfg))
                 .limit(cfg.limit)
                 .map_err(move |e| (*err)(e, &req2))
                 .map(Json),
@@ -155,6 +155,7 @@ where
 ///
 /// ```rust
 /// # extern crate actix_web;
+/// extern crate mime;
 /// #[macro_use] extern crate serde_derive;
 /// use actix_web::{error, http, App, HttpResponse, Json, Result};
 ///
@@ -173,6 +174,9 @@ where
 ///         r.method(http::Method::POST)
 ///               .with_config(index, |cfg| {
 ///                   cfg.0.limit(4096)   // <- change json extractor configuration
+///                      .content_type(|mime| {  // <- accept text/plain content type
+///                          mime.type_() == mime::TEXT && mime.subtype() == mime::PLAIN
+///                      })
 ///                      .error_handler(|err, req| {  // <- create custom error response
 ///                          error::InternalError::from_response(
 ///                              err, HttpResponse::Conflict().finish()).into()
@@ -184,6 +188,7 @@ where
 pub struct JsonConfig<S> {
     limit: usize,
     ehandler: Rc<Fn(JsonPayloadError, &HttpRequest<S>) -> Error>,
+    content_type: Option<Box<Fn(mime::Mime) -> bool>>,
 }
 
 impl<S> JsonConfig<S> {
@@ -201,6 +206,15 @@ impl<S> JsonConfig<S> {
         self.ehandler = Rc::new(f);
         self
     }
+
+    /// Set predicate for allowed content types
+    pub fn content_type<F>(&mut self, predicate: F) -> &mut Self
+    where
+        F: Fn(mime::Mime) -> bool + 'static,
+    {
+        self.content_type = Some(Box::new(predicate));
+        self
+    }
 }
 
 impl<S> Default for JsonConfig<S> {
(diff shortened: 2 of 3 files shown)


src/httpmessage.rs
```diff
@@ -200,7 +200,7 @@ pub trait HttpMessage: Sized {
     /// # fn main() {}
     /// ```
     fn json<T: DeserializeOwned>(&self) -> JsonBody<Self, T> {
-        JsonBody::new(self)
+        JsonBody::new::<()>(self, None)
```
DoumanAsh, line 203: is adding `::<()>` necessary?

src/json.rs
```diff
@@ -201,13 +207,44 @@ impl<S> JsonConfig<S> {
...
+    pub fn content_type(&mut self, content_type: ContentTypeFilter) -> &mut Self {
+        self.content_types.push(content_type);
+        self
+    }
 }
 
 impl<S> Default for JsonConfig<S> {
     fn default() -> Self {
+        let mut content_types = Vec::new();
+        content_types.push(ContentTypeFilter::new(None, Some(mime::JSON), None));
```
DoumanAsh: Json should be always default mime type, do not add it by default

src/json.rs
```diff
@@ -253,10 +290,18 @@ pub struct JsonBody<T: HttpMessage, U: DeserializeOwned> {
...
         let json = if let Ok(Some(mime)) = req.mime_type() {
-            mime.subtype() == mime::JSON || mime.suffix() == Some(mime::JSON)
+            if let Some(cfg) = cfg {
+                cfg.content_types.iter().any(|ct| {
+                    (ct.type_.is_none() || ct.type_.unwrap() == mime.type_()) &&
+                        (ct.subtype.is_none() || ct.subtype.unwrap() == mime.subtype()) &&
+                        (ct.suffix.is_none() || ct.suffix == mime.suffix())
+                })
+            } else {
+                mime.subtype() == mime::JSON || mime.suffix() == Some(mime::JSON)
```
DoumanAsh: This condition should be checked first before going into `if let Some(cfg) = cfg`

I think it would be better to treat extra content types as extra.

src/json.rs
```diff
@@ -201,13 +207,44 @@ impl<S> JsonConfig<S> {
...
             limit: 262_144,
             ehandler: Rc::new(|e, _| e.into()),
+            content_types,
+        }
+    }
+}
+
+/// Filter type for allowed `Content-Type` header values.
+pub struct ContentTypeFilter {
+    type_: Option<mime::Name<'static>>,
```
DoumanAsh: I don't think it is convenient to use struct with lots of optional fields like that.

I think it might be better if user would be able to supply `mime::Mime` rather than allowing that accurate checks.
On side note it might be better if user could provide function that takes `mime::Mime` and returns whether it should be treated as json or not.

DoumanAsh (changes requested):
Thank you, but I feel like approach with storing Vec of such filters is not the best. I don't think it would be convenient to use and it seems a bit wasted memory wise.

What I think is that it might be better for user to supply just `mime::Mime` instead
Or maybe function that accepts `mime::Mime` and returns `bool` that indicates whether this mime should be treated as json

philbooth:
> I feel like approach with storing Vec of such filters is not the best

I've updated it to use a predicate instead of a `Vec`.

src/httpmessage.rs
philbooth: Without it I get:

```
   --> src/httpmessage.rs:203:9
    |
203 |         JsonBody::new(self, None)
    |         ^^^^^^^^^^^^^ cannot infer type for `S`
```

Is there something else I can do to help the compiler out there?

DoumanAsh: Ah I see, now since we carry state in config we get it into `new` too...
There is nothing to do about it

src/json.rs
```diff
@@ -217,6 +232,7 @@ impl<S> Default for JsonConfig<S> {
 /// Returns error:
 ///
 /// * content type is not `application/json`
+///   (unless specified otherwise in [`JsonConfig`](dev/struct.JsonConfig.html))
```
DoumanAsh: I think you can omit `otherwise`

src/json.rs
```diff
@@ -440,4 +459,41 @@ mod tests {
...
+        ).header(
+            header::CONTENT_LENGTH,
+            header::HeaderValue::from_static("16"),
+        ).set_payload(Bytes::from_static(b"{\"name\": \"test\"}"))
+        .finish();
+        assert!(handler.handle(&req).as_err().is_none())
+    }
+
+    #[test]
+    fn test_with_json_and_bad_content_type() {
```
DoumanAsh: Please add test case to verify that `cfg.content_type(...)` works correctly.
For example, it maps text/plain as json, but pass request with html

DoumanAsh (approved):

philbooth:
(apologies if you were reviewing, I just force-pushed again to fix a couple of comments)

philbooth:
Thanks for your help @DoumanAsh! I've pushed fixes for your last two comments.

DoumanAsh:
@philbooth Sorry, I forgot to mention, could you please add mention to [CHANGES](https://github.com/actix/actix-web/blob/master/CHANGES.md) about new thingy you made?

DoumanAsh:
Sorry, I didn't notice you pushed changes (force push doesn't send me notifications)

Thank you

Issue #598, opened 2018-11-21
Don't use `.readline()` for multipart boundaries


abonander:
The trailing boundary in a multipart request isn't required to be followed by CRLF, as specified in [RFC 2046 Section 5.1.1](https://tools.ietf.org/html/rfc2046#section-5.1.1):

```bnf
multipart-body := [preamble CRLF]
                  dash-boundary transport-padding CRLF
                  body-part *encapsulation
                  close-delimiter transport-padding
                  [CRLF epilogue]
```

Because it's not required, the client API of my `multipart` crate isn't sending the trailing CRLF and so is causing problems with Actix: abonander/multipart#113

This is due to the use of `PayloadBuffer::readline()` throughout the `multipart.rs` module to read multipart boundaries. It may work fine with some clients but since it's not strictly conformant there may be other client implementations it might fail with, thus I don't think changing `multipart`'s behavior is the correct solution here.

The solution I ended up going with in `multipart`/`multipart-async` was to search for the boundary bytes themselves; `twoway` makes this simple and fast. I just ended up prepending the two leading hyphens to the search text since they always appear and then verifying the presence of the leading CRLF and trailing hyphen-pair/CRLF.

abonander:
@DoumanAsh I would tag this as a correctness issue because it's throwing errors on in-spec request bodies.

fafhrd91:
this kind of should be fixed. but i am not sure. in any case, i am not planing to do any further changes.

abonander:
I don't see the wisdom in closing a correctness bug as wontfix unless you're certain that it's not causing issues in the wild.

fafhrd91:
i dont have enough resources. as i understand browsers work, curl work.

PR #640, opened 2018-12-20
Create wrapper classes around pollable streams for better composability

mzhang28:
I noticed that many of the methods in the `PayloadBuffer` API actually returns stream-like `Poll` objects, so I figured that it might be a good idea to create wrappers for them to reduce the boilerplate (for ex. `Async::NotReady`).

This probably won't affect public API, since `PayloadBuffer` is mainly used internally, but I was gonna try to tackle #598 and figured it would be good if these APIs were composable.

If this is a good idea I'll continue for the rest of the methods, otherwise I guess I'll just drop it.

### DIFF
--- a/src/payload.rs
+++ b/src/payload.rs
@@ -284,9 +284,9 @@ impl Inner {
 
 /// Payload buffer
 pub struct PayloadBuffer<S> {
-    len: usize,
-    items: VecDeque<Bytes>,
-    stream: S,
+    pub(crate) len: usize,
+    pub(crate) items: VecDeque<Bytes>,
+    pub(crate) stream: S,
 }
 
 impl<S> PayloadBuffer<S>
@@ -308,7 +308,7 @@ where
     }
 
     #[inline]
-    fn poll_stream(&mut self) -> Poll<bool, PayloadError> {
+    pub(crate) fn poll_stream(&mut self) -> Poll<bool, PayloadError> {
         self.stream.poll().map(|res| match res {
             Async::Ready(Some(data)) => {
                 self.len += data.len();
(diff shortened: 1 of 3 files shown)


src/payload.rs
```diff
@@ -420,42 +382,164 @@ where
...
+            match self.inner.poll_stream()? {
+                Async::Ready(true) => self.poll(),
+                Async::Ready(false) => Ok(Async::Ready(None)),
+                Async::NotReady => Ok(Async::NotReady),
+            }
+        }
+    }
+}
+
+pub struct Copy<'a, S> {
```
DoumanAsh: Since it is the name of std trait I think it would be better to rename it something like `CopyBuffer`

DoumanAsh (commented):
To be honest I'm not sure how it makes things better? It only creates extra structs that implements traits with `poll` method

src/payload.rs
mzhang28: Oh you're right, I'll change that

mzhang28:
The motivation behind this was mainly because I wanted to try to fix #598 but the way the code is written right now is very cumbersome with manual matching over `Poll`, and that's because of the way `PayloadBuffer`'s API is written.

The change this would make is instead of doing something like this:

```rs
match something.readline()? {
    Ok(Async::NotReady) => Ok(Async::NotReady),
    Ok(Async::Ready(None)) => ...,
    Ok(Async::Ready(Some(x)) => { the meat of the code },
}
```

you can do something like

```rs
something.readline()?.map(|x| { the meat of the code })
```

and leave the rest to be passed along.

DoumanAsh:
Hmmm well it does make some sense for future

@fafhrd91 Will 0.8 rewrite affect it?

mzhang28:
Oh shoot didn't realize there was a branch for the upcoming 0.8 release, let me rework this

DoumanAsh:
@iptq no, 0.8 is for breaking changes right now.

To clarfiy new changes @fafhrd91 is working on, would make lots of differences to internals as part of code will be moved to `actix-net` and `actix-http`

fafhrd91:
i dont think this is relevant anymore

PR #644, opened 2018-12-23
Only return a single Origin value

### LINKED ISSUE #603: Cors, Returning one allowed origin header when multiple allowed origin headers have been defined
Hi,

Is there a way to return only one allowed origin from a whitelist? I know that I can send a comma separated list of allowed origins, but chrome complains about this. [Sounds like the recommended way to handle this is to check for the origin in the whitelist and send back only the origin](https://stackoverflow.com/a/1850482).

Can I do this with the CORS middleware without resorting to setting the `origins` to `AllOrSome::All`?

Thanks

bluec0re:
Browsers do not accept a list of origins in the A-C-A-O header [1] [2] [3]. This pull request returns only a single origin if a match was found.

fixes #603.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMultipleAllowOriginNotAllowed
[2] https://www.w3.org/TR/cors/#access-control-allow-origin-response-header
[3] https://cs.chromium.org/chromium/src/services/network/cross_origin_read_blocking.cc?sq=package:chromium&dr=C&g=0&l=245-266

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 * HTTP1 decoder should perform case-insentive comparison for client requests (e.g. `Keep-Alive`). #631
+* Access-Control-Allow-Origin header should only a return a single, matching origin. #603
 
 ## [0.7.16] - 2018-12-11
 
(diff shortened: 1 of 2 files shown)


fafhrd91:
looks good! could you add entry ti changes.md

fafhrd91 (approved):

bluec0re:
Done.

DoumanAsh:
Thank you

PR #647, opened 2018-12-27
Add with-cookie init-method for TestRequest

winkidney:
Thanks for your great works on actix-web, it's really awesome.
I've made a PR for actix-web's test suite to meet my needs, I'll appreciate that if you could help to review.

# What
+ Add the ability to allow us to add cookie to `test::TestRequest` object

# Why
While writing tests, I want to customized my request cookie to test if the handler can read proper values and compare them to some other value in headers, just like: 

```rust
fn validate_request_token<T>(request: &HttpRequest<T>) -> bool {
    let token_in_cookie = match request.cookie("some-cookie") {
        None => return false,
        Some(c) => c.value().to_owned(),
    };
    // do something with token_in_cookie.
    true
}
```

But in TestRequest ( which located in test.rs), I found that there is no way for us to set cookies for TestRequest and a private method named "set_cookies" seems working for actix-web itself.

So I add a method to help us to add cookie to our TestRequest.

**If test for this feature is required, please let me know, thanks**

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -507,6 +507,10 @@ impl TestRequest<()> {
     {
         TestRequest::default().header(key, value)
     }
+    /// Create TestRequest and set request cookie
+    pub fn with_cookie(cookie: Cookie<'static>) -> TestRequest<()> {
+        TestRequest::default().cookie(cookie)
+    }
 }
 
 impl<S: 'static> TestRequest<S> {
@@ -540,6 +544,27 @@ impl<S: 'static> TestRequest<S> {
     /// Set HTTP Uri of this request
     pub fn uri(mut self, path: &str) -> Self {
         self.uri = Uri::from_str(path).unwrap();
+        self
+    }
+
+    /// set cookie of this request
+    pub fn cookie(mut self, cookie: Cookie<'static>) -> Self {
+        let mut should_insert = true;
+        let mut cookies = match self.cookies {
+            Some(old_cookies) => {
+                for old_cookie in &old_cookies {
+                    if old_cookie == &cookie {
+                        should_insert = false
+                    };
+                }
+                old_cookies
+            }
+            None => { Vec::<Cookie>::new() }
+        };
+        if should_insert {
+            cookies.push(cookie)
+        };
+        self.cookies = Some(cookies);
         self
     }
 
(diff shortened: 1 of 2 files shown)


src/test.rs
```diff
@@ -543,6 +547,27 @@ impl<S: 'static> TestRequest<S> {
         self
     }
 
+    /// set cookie of this request
+    pub fn cookie(mut self, cookie: Cookie<'static>) -> Self {
+        let mut should_insert = true;
+        let mut cookies = match self.cookies {
```
DoumanAsh: I think it would be better to match by mut reference and insert inside `Some(ref mut cookies)` instead of moving out and moving in again

winkidney: Thanks for you advice, all done.

src/test.rs
```diff
@@ -543,6 +547,25 @@ impl<S: 'static> TestRequest<S> {
...
+                for old_cookie in old_cookies.iter() {
+                    if old_cookie == &cookie {
+                        should_insert = false
+                    };
+                };
+                if should_insert {
+                    old_cookies.push(cookie);
+                };
+            }
+            None => {}
```
DoumanAsh: But we need to insert cookie, if there is `None` in `self.cookies` right?

winkidney: Will be redo.

winkidney: @DoumanAsh Would you please review again ? It's done.

src/test.rs
```diff
@@ -507,6 +507,10 @@ impl TestRequest<()> {
     {
         TestRequest::default().header(key, value)
     }
+    /// Create TestRequest and set request cookie
```
DoumanAsh, line 511: Add empty line before this doc string

DoumanAsh (approved):
Thanks, please update CHANGES.md with mention of this new function

src/test.rs
winkidney: Got it. 👍 
Maybe we need a contribution guide for creating a pull request or a feature so that we can do all the things we need at one time, FYI.

winkidney: @DoumanAsh All done. And, should I add a tag named "v0.7.18"?

DoumanAsh (approved):

DoumanAsh:
Thanks

PR #650, opened 2018-12-29
show index if index file doesn't exist

esjeon:
Currently, `show_index` is completely ignored if `index_file` is set. It's saner to show index if the designated index file doesn't exist.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -749,6 +749,11 @@ impl<C: StaticFileConfig> StaticFiles<C> {
         if path.is_dir() {
             if let Some(redir_index) = C::index_file() {
                 path.push(redir_index);
+                if !path.exists() {
+                    path.pop();
+                    let dir = Directory::new(self.directory.clone(), path);
+                    return Ok(C::directory_listing(&dir, &req)?.into())
+                }
             } else if C::show_index() {
                 let dir = Directory::new(self.directory.clone(), path);
                 return Ok(C::directory_listing(&dir, &req)?.into())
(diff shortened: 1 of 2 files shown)


DoumanAsh:
I'm not sure if it is actually saner?

If user specifies `index_file` then he expects to have file, if there is no file it might be actually error?
But I'm not sure to be honest, maybe it does make sense

esjeon:
Well, at least, [that's how Apache works](https://wiki.apache.org/httpd/DirectoryListings), and this is useful during development. With this, you can load up application normally, and then poke around files freely using directory listing.

Or perhaps, this would better be provided by 3rd-party middleware, because Apache-like behaviour would  not be a main goal of http framework. Still, it's just few lines of code. (Note: Apache does have it in a module called _mod_autoindex_)

src/fs.rs
```diff
@@ -749,6 +749,11 @@ impl<C: StaticFileConfig> StaticFiles<C> {
         if path.is_dir() {
             if let Some(redir_index) = C::index_file() {
                 path.push(redir_index);
+                if !path.exists() {
```
DoumanAsh: We should show index only if `C::show_index()` returns true!

DoumanAsh (changes requested):

DoumanAsh (approved):

DoumanAsh:
Thank you, please add note on new change in [CHANGES.md](https://github.com/actix/actix-web/blob/master/CHANGES.md)

DoumanAsh:
Thanks

PR #651, opened 2019-01-03
Improve change askama_escape in favor of v_htmlescape

zzau13:
Improves the performance losses of false positives (all no html escapable between ascii 34 and 62) and adds support for SIMD. [Benches on Travis](https://travis-ci.org/rust-iendo/v_htmlescape/jobs/472791580#L688)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ cell = ["actix-net/cell"]
 actix = "0.7.9"
 actix-net = "0.2.6"
 
-askama_escape = "0.1.0"
+v_htmlescape = "0.3.2"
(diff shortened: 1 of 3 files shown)


DoumanAsh (approved):
Well why not I guess if it is better

fafhrd91:
@DoumanAsh could you check how v_htmleacape code

DoumanAsh:
@fafhrd91 I did take a quick look, there is some macro hell, but it is more or less well tested. Though I doubt there would be much benefit from any further optimizations

fafhrd91:
Thanks

Issue #371, opened 2018-07-07
serde query extractor fails on simple enum types


mqudsi:
It seems that something in how the contents of a query string are passed to serde is causing breakage with simple enum types.

For the enum

```rust
// #[serde(untagged)]
// #[serde(rename_all = "camelCase")]
#[derive(Deserialize)]
pub enum ResponseType {
    Token,
    Code
}

#[derive(Deserialize)]
pub struct AuthRequest {
    client_id: String,
    redirect_uri: String,
    response_type: ResponseType,
    scope: String,
}
```

and with the handler

```rust
fn authorize(params: Query<AuthRequest>) -> impl Responder {
    format!("Do you authorize this application to sync files on your behalf?")
}
```

a request to `/authorize?...&response_type=code` (or `Code` or `ResponseType::Code` or `1` or any of a dozen other permutations I tried) I get back an error along the lines of the following:

```
 DEBUG actix_web::pipeline    > Error { err: "invalid type: string \"code\", expected enum ResponseType" }


 INFO  actix_web::middleware::logger > 127.0.0.1:52722 [06/Jul/2018:22:54:00 -0500] "GET /authorize?client_id=ebb82a88-8197-11e8-85a6-fb14d7e72e94&redirect_uri=msn.com&response_type=code HTTP/1.1" 400 0 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.99 Safari/537.36" 0.000427
```

As you can see from the commented-out code, I also tried playing around with different serde enum representations and formatting options.

mqudsi:
I suppose it might actually be this: https://github.com/nox/serde_urlencoded/issues/35

fafhrd91:
I don’t think we can do anything. This need bug fix from upstream.

@Dowwie am I right?

Dowwie:
@fafhrd91 it seems that serde_urlencoded is no longer being maintained.. there's been a pending PR since Feb 2018 that upgrades to support enums: https://github.com/nox/serde_urlencoded/pull/30

I pinged the author weeks ago but haven't gotten any response

fafhrd91:
that is unfortunate. so we need fork it or we need replacement

DoumanAsh:
@Dowwie I'll ping nox on IRC if he is still around there I'll ask him about crate's state

DoumanAsh:
[...]

PR #664, opened 2019-01-13
Issue: sequence in query and form extractors do not work

AnderEnder:
Common multiple params usage in query and form can not be used with actix-web:
- example url: `/index.html?first_name=John&last_name=Doe&emails=john%40example.com&emails=doe%40example.com`
- example form body: `first_name=John&last_name=Doe&emails=john%40example.com&emails=doe%40example.com`

These are two test cases to represent the issue.

### DIFF
--- a/tests/test_handlers.rs
+++ b/tests/test_handlers.rs
@@ -144,6 +144,50 @@ fn test_query_enum_extractor() {
     assert_eq!(response.status(), StatusCode::BAD_REQUEST);
 }
 
+#[derive(Debug, Deserialize)]
+pub struct Customer {
+    first_name: String,
+    last_name: String,
+    emails: Vec<String>,
+}
+
+#[test]
+fn test_query_sequence_extractor() {
+    let mut srv = test::TestServer::new(|app| {
+        app.resource("/index.html", |r| {
+            r.with(|p: Query<Customer>| format!("{:?}", p))
+        });
+    });
+
+    // client request
+    let request = srv
+        .get()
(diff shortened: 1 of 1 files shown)


DoumanAsh:
There is not much meaning with test case alone, but if I remember correctly we use `serde_urlencoded` to extract from query, so the functionality can be only added there

AnderEnder:
Sure, these test case only represents the issues, the behaviour which is required, but not provided by `actix-web`. My expectation that web framework should support it. Currently I can not implement simple API and this stops the development.

AFAIK, `serde_urlencoded` does not want to support anything more, so the question to `actix-web` how to support it.

The issue is quite similar to https://github.com/actix/actix-web/issues/371, so the private  `serde_urlencoded` was created as result.

DoumanAsh:
>AFAIK, serde_urlencoded does not want to support anything more, so the question to actix-web how to support it.

the author of library did accept PR and we removed private fork. I think first step would be for us to write PR that implements this feature

AnderEnder:
The author did not accept required feature. Related PR: https://github.com/nox/serde_urlencoded/pull/42

fafhrd91:
We could switch to different library. 

@AnderEnder you should understand, this is open source software and maintainers may not have free time to implement every feature. PR is always welcomed

AnderEnder:
@fafhrd91, there is no similar library. And this is a common case(not feature) when a form submit multiple values.

@DoumanAsh and @fafhrd91
What do you suggest? Create a private fork here?

DoumanAsh:
@AnderEnder There is [serde_qs](https://crates.io/crates/serde_qs) that was created to address this limitation it seems

PR #670, opened 2019-01-16
Use NamedFile with an existing File

### LINKED ISSUE #659: Use NamedFile with an existing std::fs::File
I have an application written in actix-web that generates a temporary file upon a client request, streams this file to the client and deletes it afterwards. To create the file I am using [`tempfile`](https://docs.rs/tempfile/3.0.5/tempfile/fn.tempfile.html). I would like a convenient way to stream this `File` (with a pseudo filename) to the client, but from what I can tell this is not possible with the current API. I have looked at `NamedFile`, but I can not provide the `File` to it myself, just the path.

My proposal would be to add a function `NamedFile::from_file<P: AsRef<Path>>(file: File, path: Path`. It would be largely equivalent to calling `NamedFile::open`, except that the path is used only to guess mime type and set the content disposition, it does not need to actually exist.

Example usage:
```rust
fn handle_request(_req: &HttpRequest) -> Result<NamedFile> {
    let file = tempfile()?;
    
    // write some data into the file
    // ...

    NamedFile::from_file(file, "data.tar.gz")
}
```

I'm more than willing to whip up a pull request for this in case you're interested :wink:

wildarch:
Resolves #659.

This adds `NamedFile::from_file` and `NamedFile::from_file_with_config` next to the usual `NamedFile::open` and `NamedFile::open_with_config`.

I have included documentation and an example/doctest for both methods.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,6 +129,7 @@ tokio-uds = { version="0.2", optional = true }
 [dev-dependencies]
 env_logger = "0.6"
 serde_derive = "1.0"
+tempfile = "3.0"
 
 [build-dependencies]
 version_check = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,7 +129,6 @@ tokio-uds = { version="0.2", optional = true }
 [dev-dependencies]
 env_logger = "0.6"
 serde_derive = "1.0"
-tempfile = "3.0"
 
 [build-dependencies]
 version_check = "0.1"
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -128,17 +128,30 @@ impl NamedFile {
     /// # Examples
     ///
     /// ```rust
-    /// extern crate tempfile;
     /// extern crate actix_web;
     ///
     /// use actix_web::fs::NamedFile;
-    /// use tempfile::tempfile;
     /// use std::io::{self, Write};
+    /// use std::env;
+    /// # use std::fs::{self, File};
+    /// # use std::path::PathBuf;
+    /// #
+    /// # fn path() -> PathBuf {
+    /// #     let mut path = env::temp_dir();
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -129,6 +129,7 @@ tokio-uds = { version="0.2", optional = true }
 [dev-dependencies]
 env_logger = "0.6"
 serde_derive = "1.0"
+tempfile = "3.0"
```
DoumanAsh: It would be great if we could avoid using extra dependencies

Mostly because actix-web takes long time to build tests already 😄

src/fs.rs
```diff
@@ -187,6 +225,18 @@ impl<C: StaticFileConfig> NamedFile<C> {
             _cd_map: PhantomData,
         })
     }
```
DoumanAsh, line 227: Add empty line after

wildarch: Sure!

Cargo.toml
wildarch: I can also do it with a regular `std::fs::File`. The only problem would be that in some obscure corner cases, it will leave a in the user's temp directory. I guess that is an acceptable tradeoff?

wildarch:
In the latest commit I have removed the dependency on `tempfile` while preserving the same example code.

Cargo.toml
DoumanAsh: Yeah it is fine, I'd like to not increase compile time as much as possible.

Well, to avoid creating empty files, just tag all examples as `no_run` so that `cargo test` would only compile them

DoumanAsh (approved):
Please add note about these changes to changelog and tag all examples as `no_run` to avoid running it on cargo test

DoumanAsh:
Thanks

PR #676, opened 2019-01-24
Serve static file directly instead of redirecting

utterstep:
Resolved TODO in `fs::StaticFiles::try_handle`: now index file is served directly, without extra redirect (saving one network RT at least).

_PS this is my first contribution to this project, let me know if I should do something extra, like updating the CHANGES.md (assuming that my changes are OK :))_

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -1458,6 +1458,17 @@ mod tests {
             resp.headers().get(header::CONTENT_DISPOSITION).unwrap(),
             "attachment; filename=\"test.binary\""
         );
+
+        // nonexistent index file
+        let req = TestRequest::default().uri("/tests/unknown").finish();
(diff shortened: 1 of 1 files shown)


DoumanAsh:
It is actually solved on 0.8 branch already, but let me check your PR later on

utterstep:
Got it, will be waiting for your response.
In a meantime, added commit with some test tidying up: NOT_FOUND cases + updated names (redirect -> serve) — maybe tests will be useful in case if main fixes are redundant

DoumanAsh (approved):

DoumanAsh:
Thanks

fafhrd91:
@DoumanAsh I think we should move static files support to separate crate. What do you think?

DoumanAsh:
Yeah, I think it wouldn't be a bad idea to move it into own crate for 0.8 version

PR #681, opened 2019-01-31
Implement failure::Fail for actix_web::Error.

marmistrz:
This is done in order to allow using actix_web::Error with crates such as failure.

As of `actix-web` 0.7.18 is it impossible to return `failure::Error` from functions using `actix-web` internally:
```
    let mut sys = System::new("test");
    let fut = get_future();
    sys.block_on(fut).map(|e| e.into())
```
which fails with
```
269 |     sys.block_on(fut).map_err(|e| e.into()) //.map_err(|e| format_err!("{}", e))
    |                                     ^^^^ the trait `std::error::Error` is not implemented for `actix_web::error::Error`

```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -44,41 +44,33 @@ pub type Result<T, E = Error> = result::Result<T, E>;
 /// for it that can be used to create an http response from it this means that
 /// if you have access to an actix `Error` you can always get a
 /// `ResponseError` reference from it.
+///
+/// This type supports conversion to the `failure::Error` type. To use this
+/// type as `std::error::Error`, you can use the `compat` method from `failure`.
 pub struct Error {
     cause: Box<ResponseError>,
     backtrace: Option<Backtrace>,
 }
 
-impl Error {
-    /// Deprecated way to reference the underlying response error.
-    #[deprecated(
-        since = "0.6.0",
-        note = "please use `Error::as_response_error()` instead"
-    )]
-    pub fn cause(&self) -> &ResponseError {
-        self.cause.as_ref()
-    }
-
-    /// Returns a reference to the underlying cause of this `Error` as `Fail`
-    pub fn as_fail(&self) -> &Fail {
-        self.cause.as_fail()
-    }
-
-    /// Returns the reference to the underlying `ResponseError`.
-    pub fn as_response_error(&self) -> &ResponseError {
-        self.cause.as_ref()
-    }
-
+impl failure::Fail for Error {
     /// Returns a reference to the Backtrace carried by this error, if it
     /// carries one.
     ///
     /// This uses the same `Backtrace` type that `failure` uses.
-    pub fn backtrace(&self) -> &Backtrace {
-        if let Some(bt) = self.cause.backtrace() {
+    fn backtrace(&self) -> Option<&Backtrace> {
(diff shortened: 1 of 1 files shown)


fafhrd91:
could you add test

mehcode:
This would be massively awesome. Currently my integration tests return `actix_web::Result` and have to `unwrap` non client request errors.

marmistrz:
I've rethought this change and we definitely don't want to just `impl std::error::Error` - we'll just throw away all the context information.

I've just properly implemented `failure::Fail` for `actix_web::Error` and this should do the trick. Probably the test should rather be a doc-test, but I'm not sure how to rewrite it so that it's suited for doc.

mehcode:
@marmistrz This PR removes `as_fail` and `cause`. Can those be put back in so this won't be a breaking change?

marmistrz:
> @marmistrz This PR removes as_fail and cause. Can those be put back in so this won't be a breaking change?

@mehcode They should be implemented automatically, IIRC. You may need to import some traits though. I'll recheck tomorrow

mehcode:
@marmistrz Still makes it a breaking change as libraries ( such as sentry-actix ) are depending on them being inherent methods.

marmistrz:
@mehcode this change has to be breaking, because to properly implement `Fail` we need to return `Option<&Backtrace>` and not `&Backtrace` as actix-web currently does it. 

I'd like to hear what the opinion of the maintainers is, though.

fafhrd91:
just fyi, I am dropping failure in 0.8, failure will be supported but actin_web::Error won't be based on Fail anymore.

fafhrd91:
@marmistrz could you re-evaluate this PR for 1.0? 1.0 is in master now

marmistrz:
Ok. I'll try migrating my code to 1.0 and see if things work as expected or if there are any changes needed.

marmistrz:
This PR still looks relevant, because even though current master drops the `failure` crate, the error types still don't seem to implement `std::error::Error`. I'll wait with updating the PR to target 1.0 until the migration guide is ready.

fafhrd91:
@marmistrz could you update it then. but i dont want to enforce `Error` trait on `ResponseError`

marmistrz:
Ok, so I'll wait until I can port my app to 1.0 and then see which types actually need the implementation of `std::error::Error`.

May I ask why you don't want `std::error::Error` for `ResponseError`?

fafhrd91:
`std::error::Error` adds a lot of boilerplate code. ie i usually just create error enum, something like

```rust
#[derive(From, Display)]
enum MyError {
  Err1,
  Err2,
  Err3(OtherError),
}
```

and `Error` trait would cause a lot of boilerplate unless we introduce some derive macro. i use derive_more extensively for `From` and `Display` macros

DoumanAsh:
@fafhrd91 Current `Error` is cheap because it depends on `Display` now. i.e. it requires no implementation since it uses `Display` to show error
But yeah, I think there is no need to require `std::error::Error` since it serves a little purpose, but we can make our own errors to impl the `Error` trait

fafhrd91:
yes, `actix_http::Error` can implement `std::Error`

fafhrd91:
`std::error::Error` impl added to `Error`. It is not possible to implement `Fail` for Error because failure requires `Send + Sync` for errors

PR #687, opened 2019-02-06
Fix typo

medwards:
The original
> This method requires to run within properly configured Actix system.

was not correct, so merged it with the previous statement about panicking (since it panics for me if I try to launch it /wo an actix system).

Could alternatively be
> This method requires a properly configured Actix system.

or a number of similar variants if the behaviour when run without an Actix system must remain undefined.

### DIFF
--- a/src/server/http.rs
+++ b/src/server/http.rs
@@ -451,9 +451,8 @@ impl<H: IntoHttpHandler, F: Fn() -> H + Send + Clone> HttpServer<H, F> {
     /// For each address this method starts separate thread which does
     /// `accept()` in a loop.
     ///
-    /// This methods panics if no socket addresses get bound.
-    ///
-    /// This method requires to run within properly configured `Actix` system.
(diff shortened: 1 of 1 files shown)


src/server/http.rs
```diff
@@ -451,9 +451,8 @@ impl<H: IntoHttpHandler, F: Fn() -> H + Send + Clone> HttpServer<H, F> {
     /// For each address this method starts separate thread which does
     /// `accept()` in a loop.
     ///
-    /// This methods panics if no socket addresses get bound.
-    ///
-    /// This method requires to run within properly configured `Actix` system.
+    /// This methods panics if no socket addresses get bound or it is not run
```
DoumanAsh: >no socket addresses get bound

Maybe say `no socket address can be bound

DoumanAsh: >or it is not run

Or `Actix` system is not yet configured?

DoumanAsh (approved):
Thanks, please note that reviewers are not notified when you perform force push, so feel free to ping via comment

PR #688, opened 2019-02-06
Make interface of Router public

popzxc:
This PR does not change any logic, but makes interface of the `Router` structure public.

Motivation: Router provides interfaces for (suprisingly) routing management, but currently its interface is almost completely hidden under `pub(crate)`. Making it public will allow users to implement non-trivial routing rules, such as ability to enable/disable some parts of application in the runtime by creating a multiple routers on their side.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -417,7 +417,7 @@ pub trait RouteHandler<S>: 'static {
     fn finish(&mut self) {}
 }
 
-/// Route handler wrapper for Handler
+/// Route handler wrapper for Handler.
 pub struct WrapHandler<S, H, R>
 where
     H: Handler<S, Result = R>,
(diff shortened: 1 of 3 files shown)


DoumanAsh:
@fafhrd91 Do you think it is good idea?

popzxc:
A bit of excuses: I understand, that in normal conditions nobody should use that. But when I've asked about routing management in runtime on gitter, I was suggested to implement routing manually. This PR allows me (and everyone who will ever be in similar situation) to implement manual routing comfortably.

Since `Router` is located at the `dev` module, it's pretty obvious (as for me) that people won't misuse that feature by accident.

fafhrd91:
here is partial port https://github.com/actix/actix-net/tree/master/router
i am planing to move it to actix-web repo and switch actix-web to new router.

fafhrd91:
@popzxc we can change router's api to improve re-usability

Issue #680, opened 2019-01-30
If-Modified-Since is not ignored if If-None-Match is specified


bbigras:
> A recipient must ignore If-Modified-Since if the request contains an If-None-Match header field; the condition in If-None-Match is considered to be a more accurate replacement for the condition in If-Modified-Since, and the two are only combined for the sake of interoperating with older intermediaries that might not implement If-None-Match.

https://greenbytes.de/tech/webdav/rfc7232.html

I tested with curl and I had a 304. `if-modified-since` matched but I used a bogus `if-none-match`.

DoumanAsh:
afaik I think we currently ignore non-matching `If-None-Match` and try to match `If-Modified-Since`

DoumanAsh:
We should take out [If-Modified-Since](https://github.com/actix/actix-web/blob/master/src/fs.rs#L444) only if `If-None-Match` is not present regardless of whether it matches

bbigras:
> We should take out If-Modified-Since only if If-None-Match is not present regardless of whether it matches

Sorry if I'm not understanding correctly. Are you suggesting that both `If-Modified-Since` and `If-None-Match` should match to result in a 304?

My problem is that when building docker image with NixOS's [dockerTools.buildLayeredImage](https://nixos.org/nixpkgs/manual/#ssec-pkgs-dockerTools-buildLayeredImage), the modified date on the file are all set to 1970-01-01 00:00:01Z. I reported the issue (https://github.com/NixOS/nixpkgs/issues/54955) but it could be a limitation of the way NixOS handles things.

So when I update my static files I get new etags but with the same modified date which cause 304 problems with actix-web.

DoumanAsh:
I meant that we should ignore presence of `If-Modified-Since` if there is `If-None_match` as per RFC

You can disable `Last-Modified` check via custom config btw https://actix.rs/actix-web/actix_web/fs/trait.StaticFileConfig.html

bbigras:
> I meant that we should ignore presence of If-Modified-Since if there is If-None_match as per RFC

:+1: 

> You can disable Last-Modified check via custom config btw https://actix.rs/actix-web/actix_web/fs/trait.StaticFileConfig.html

oh thanks!!

wilhg:
Hi @DoumanAsh , if it is possible, I would like to try to fix this bug for starting my contribution. Is it OK?

DoumanAsh:
@cuebyte Sure, contributions are appreciated

PR #692, opened 2019-02-07
Ignored the If-Modified-Since if If-None-Match is specified (#680)

wilhg:
Fixed the issue #680 , and added a unit test for the case.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -968,7 +968,25 @@ mod tests {
     }
 
     #[test]
-    fn test_if_modified_since_ignored() {
+    fn test_if_modified_since_without_if_none_match() {
+        let mut file = NamedFile::open("Cargo.toml")
+            .unwrap()
+            .set_cpu_pool(CpuPool::new(1));
+        let since = header::HttpDate::from(
+            SystemTime::now().add(Duration::from_secs(60)));
+
+        let req = TestRequest::default()
+            .header(header::IF_MODIFIED_SINCE, since)
+            .finish();
+        let resp = file.respond_to(&req).unwrap();
+        assert_eq!(
+            resp.status(),
+            StatusCode::NOT_MODIFIED
+        );
+    }
+
+    #[test]
+    fn test_if_modified_since_with_if_none_match() {
         let mut file = NamedFile::open("Cargo.toml")
             .unwrap()
             .set_cpu_pool(CpuPool::new(1));
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -441,6 +441,8 @@ impl<C: StaticFileConfig> Responder for NamedFile<C> {
         // check last modified
         let not_modified = if !none_match(etag.as_ref(), req) {
             true
+        } else if req.headers().contains_key(header::IF_NONE_MATCH) {
+            false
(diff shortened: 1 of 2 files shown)


src/fs.rs
```diff
@@ -963,6 +967,33 @@ mod tests {
         assert_eq!(m, mime::APPLICATION_OCTET_STREAM);
     }
 
+    #[test]
+    fn test_if_modified_since_ignored() {
+        assert!(NamedFile::open("test--").is_err());
```
DoumanAsh: Is this assert part of test actually?

src/fs.rs
```diff
@@ -963,6 +967,33 @@ mod tests {
...
+        let mut file = NamedFile::open("Cargo.toml")
+            .unwrap()
+            .set_cpu_pool(CpuPool::new(1));
+        {
```
DoumanAsh: What is purpose of this block? and block below with assignment of file

DoumanAsh (commented):
Please add mention of this change to CHANGES.md

src/fs.rs
wilhg: I don't know. I copied the code from another unit test, and I can see in the other unit tests which related to static file, they all contain these 2 blocks. If you think it is meaningless, I can try to remove all these blocks in each unit test.

What's more, I also don't understand the purpose of `.set_cpu_pool(CpuPool::new(1))`, could you please explain to me? Thank you.

src/fs.rs
wilhg: I will remove it :)

src/fs.rs
DoumanAsh: Yeah, I think you can remove it as this test only wants to test your fix anyway.

As for setting custom cpu pool, I'm not sure, but I see no harm it anyway as you gonna avoid using implicit global one

wilhg:
@DoumanAsh please check again.

src/fs.rs
```diff
@@ -963,6 +967,25 @@ mod tests {
         assert_eq!(m, mime::APPLICATION_OCTET_STREAM);
     }
 
+    #[test]
+    fn test_if_modified_since_ignored() {
+        let mut file = NamedFile::open("Cargo.toml")
+            .unwrap()
+            .set_cpu_pool(CpuPool::new(1));
+        let since = header::HttpDate::from(
```
DoumanAsh, line 993: Sorry, I forgot one thing: could you please check if we have positive test case for `If-Modified-Since` if not, I'd appreicate if you could add it.

i.e. the same test but with no `IF_NONE_MATCH` header

DoumanAsh (approved):

src/fs.rs
wilhg: No problem, I added a positive test case for the `If-Modified-Since`, pls check

DoumanAsh: Thanks, but it would be nice if the `If-None-Match` test would verify that we're actually ignoring `If-Modified-Since`
Now that you have both tests you should see that both results in the same outcome, could you please adjust `since` value so that it could result in sending back file and having `If-None-Match` to prevent that? 😄 

P.s. most likely you need to take epoch time

DoumanAsh:
Thanks

PR #697, opened 2019-02-13
httpresponse: add constructor for HttpResponseBuilder

lucab:
This adds a constructor for `HttpResponseBuilder`, parametrized on
status code. It allows to directly build a response builder in a
conditional way. The usecase is for consumers where the response status
code is not statically known but received as an external parameter
instead.

### DIFF
--- a/src/httpresponse.rs
+++ b/src/httpresponse.rs
@@ -48,10 +48,10 @@ impl HttpResponse {
         self.0.as_mut()
     }
 
-    /// Create http response builder with specific status.
+    /// Create a new HTTP response builder with specific status.
     #[inline]
     pub fn build(status: StatusCode) -> HttpResponseBuilder {
-        HttpResponsePool::get(status)
+        HttpResponseBuilder::new(status)
     }
 
     /// Create http response builder
@@ -346,6 +346,12 @@ pub struct HttpResponseBuilder {
 }
 
 impl HttpResponseBuilder {
+    /// Create a new HTTP response builder with specific status.
+    #[inline]
+    pub fn new(status: StatusCode) -> HttpResponseBuilder {
+        HttpResponsePool::get(status)
+    }
+
     /// Set HTTP status code of this response.
     #[inline]
     pub fn status(&mut self, status: StatusCode) -> &mut Self {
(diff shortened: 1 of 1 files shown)


DoumanAsh:
I think it would be better to add such method to `HttpResponse` rather than builder itself

lucab:
Ah, there is actually already such a constructor there: https://docs.rs/actix-web/0.7.18/actix_web/struct.HttpResponse.html#method.build.

I would say that as a consumer, this API surprises me quite a bit. I'm used to having a constructor on the type that I want to build, not having to look for one somewhere else (outside of the module). Also, `build` is usually the method to go from a builder to the final type, not viceversa.

I'm sorry if I ended up bikeshedding on API organization, but this caught me by surprise. What do you want me to do with this PR?

DoumanAsh:
@lucab Rust doesn't have concept of constructors at all. The static methods are just free functions after all.
The point here is that you have `Type::build() -> TypeBuilder`, something like that I guess

Since the method is already present I don't see much for this...
But there is no harm in having it, but please take the same approach as `HttpResponse::build()` in this function, and `HttpResponse::build()` then could just call this new function

lucab:
@DoumanAsh ack. I've aligned the two methods to go through the same logic and rebased on current master. PTAL.

As a side note, I'm aware the concept of constructors does not exist in Rust. At the same time, I would say that `new` is the idiomatic method name for constructing a new type instance: https://doc.rust-lang.org/stable/std/index.html?search=new. Sticking to that means fewer surprises to users.

Similarly, I think the usual naming choice is `Type::builder() -> TypeBuilder` and `TypeBuilder::build() -> Type` (quick examples from [hyper](https://docs.rs/hyper/0.12.24/hyper/?search=build) and [diesel](https://docs.rs/diesel/1.4.1/diesel/?search=build)), but actix-web has them reversed. If you are planning an API break soon, I'd be happy if actix-web can align to usual idioms.

lucab:
Gentle ping for a review. I rebased to latest master.

DoumanAsh (approved):

DoumanAsh:
Sorry, it seems I missed notification? I suspect forced pushes don't notify

fafhrd91:
@lucab @DoumanAsh could you check actix-http

lucab:
@fafhrd91 ack, I can open a PR with this there too.

fafhrd91:
That would be good, thanks

lucab:
actix-http `ResponseBuilder` already has a constructor like this: https://github.com/actix/actix-http/blob/00b7dc7887540249560f0f195f2e59df51e10f61/src/response.rs#L300. No further cherry-picking needed.

PR #700, opened 2019-02-16
Implement Responder for () converting it into OK/200.

boxdot:
This allows to easily write handlers returning an empty body on success:

```rust
fn delete_user(path: Path<String>) -> Result<()> {
    Ok(())
}
```

or even just (in case a handler can't fail)

```rust
fn delete_user(path: Path<String>) {}
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### Added
 
 * Add `from_file` and `from_file_with_config` to `NamedFile` to allow sending files without a known path. #670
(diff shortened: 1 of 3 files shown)


fafhrd91:
Could you add test and entry to changes.md

boxdot:
It looks like appveyor CI is flaky.

DoumanAsh (approved):

fafhrd91:
implemented in 1.0

Issue #705, opened 2019-02-25
Replacing instead of appending headers in HttpResponseBuilder


kornelski:
Behavior of `HttpResponseBuilder`'s `header()` method is surprising to me, because it appends a new copy of a header instead of setting/replacing previous values.

I have code that is similar to an HTTP proxy, so I wanted to copy headers from the upstream, and then override some of them with my own values. Because my own headers are set conditionally, it's simpler (and feels more robust) to copy all headers and then replace the ones that I want to set, rather than duplicate conditional logic to avoid copying some headers.

There are special-case methods `content_encoding`, `content_type`, `content_length` that replace the respective headers, but there are no methods that replace arbitrary headers.

Would you accept a PR that adds a way to replace headers?

What should it be? `set_header()`, `insert_header()`? `insert()`?

DoumanAsh:
I think it would be useful since you have need.
The name I'd suggest is `set_header`

@fafhrd91 cc You might consider having it in next major version too?

kornelski:
I haven't noticed it before, but there's also `.set()` method that appends. Having `.set()` and `.header()` that appends, and `.set_header()` that replaces would be inconsistent.

In that case I'll add `remove()` which doesn't interfere with the appending methods, and can be used as `.remove(h).header(h, v)` to replace headers.

fafhrd91:
@DoumanAsh could you review 1.0 and close issue if it solved

DoumanAsh:
`.remove()` is missing I guess, but we have way to append and replace headers.

@kornelski Do you still care about having `.remove()`? If so you could add it in https://github.com/actix/actix-http/blob/master/src/response.rs#L403
But otherwise I think we're good?

kornelski:
OK, I'll add it

PR #707, opened 2019-02-26
Add .remove() to response builder

kornelski:
Partial solution to #705

### DIFF
--- a/src/httpresponse.rs
+++ b/src/httpresponse.rs
@@ -426,6 +426,38 @@ impl HttpResponseBuilder {
         }
         self
     }
+    /// Set or replace a header with a single value.
+    ///
+    /// ```rust
+    /// # extern crate actix_web;
+    /// use actix_web::{http, HttpRequest, HttpResponse};
+    ///
+    /// fn index(req: HttpRequest) -> HttpResponse {
+    ///     HttpResponse::Ok()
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Why not to have `insert` method with similar semantics to `HeaderMap::insert`?

kornelski:
That could be added as well. 

The builder has `set` and `header` taking headers in two different ways. Should there be two versions of insert too?

DoumanAsh:
`set` is hidden for now, so I don't think we need to consider it as it is not part of public API let's say

kornelski:
OK, added `insert` too

DoumanAsh (approved):

DoumanAsh:
Thanks

DoumanAsh:
P.s. Damn I forgot, I should update CHANGES.md

fafhrd91:
@DoumanAsh could you check if we have same api in 1.0

DoumanAsh:
@kornelski Could you please map the same method to response's builder in [actix-http](https://github.com/actix/actix-http) which will be used by actix-web 1.0

File `src/response.rs`

kornelski:
actix-http is archived now, so I can't add this.

fafhrd91:
actix-http is part of actix-web repo

PR #712, opened 2019-03-02
Ensure that Content-Length zero is specified in empty request

### LINKED ISSUE #710: empty body in post ClientRequest does not populate Content-Length
Whist implementing a google recaptcha backend I was getting HTTP 411 (Length Required) error responses from Google using roughly this code.

```rust
ClientRequest::post(recaptcha_url)
  .finish()
  .unwrap()
  .send();
```

and even adding a content-length manually still produced a 411. Eg.

```rust
ClientRequest::post(recaptcha_url)
  .content_length(0)
  .finish()
  .unwrap()
  .send();
```

A debug output of the request showed the content-length in the header section but a Wireshark dump of the outgoing request showed no such header.

<table>
<tr>
<td width="50%">Req/Res Debug Output</td>
<td width="50%">Wireshark Log</td>
</tr>
<tr>
<td width="50%">2019-02-28_11 07 03-6756a110</td>
<td width="50%">2019-02-28_11 08 38-2be2739e</td>
</tr>
</table>

This is the code I had to use:

```rust
ClientRequest::post(recaptcha_url)
  .body("")
  .unwrap()
  .send();
```

Obviously a fine workaround but this shows that a default content-length of 0 wasnt getting set and that the debug output was mis-reporting it.

```
actix = "0.7"
actix-web = { version = "0.7", features = ["ssl"] }
```

DoumanAsh:
Closes #710

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 
 * Ignored the `If-Modified-Since` if `If-None-Match` is specified. #680
 
+* Do not remove `Content-Length` on `Body::Empty` and insert zero value if it is missing for `POST` and `PUT` methods.
+
 ## [0.7.18] - 2019-01-10
 
 ### Added
--- a/src/client/writer.rs
+++ b/src/client/writer.rs
@@ -16,9 +16,9 @@ use flate2::write::{GzEncoder, ZlibEncoder};
 use flate2::Compression;
 use futures::{Async, Poll};
 use http::header::{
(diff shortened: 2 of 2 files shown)


DoumanAsh:
@fafhrd91 Should we send 0 content length always for HTTP requests? What do you think? Though I guess we could do it only for requests with content(POST/PUT)

fafhrd91 (approved):
We need test and we have to check if actix-http is ok

DoumanAsh:
@fafhrd91 I've added some simple tests for `content_encoder` function in case of empty body.

~I'll check `actix-http` later on~

`actix-http` adds `Content-Length: 0` in case of empty body so should be fine

PR #713, opened 2019-03-02
Support to set header names of `ClientRequest` as Camel-Case

PeterDing:
This is the case for supporting to request for servers which don't perfectly implement the `RFC 7230`. It is important for an app which uses `ClientRequest` as core part.

Because `http` crate will have no plans to support the Camel-Case headers, ([see](https://github.com/hyperium/http/issues/228#issuecomment-404996750)), We need to add an option to support it, as `hyper` [did](https://github.com/hyperium/http/issues/228#issuecomment-404996755).

So, I implement this case, as following.

* Add field `upper_camel_case_headers` to `ClientRequest`.

* Add function `set_upper_camel_case_headers` to `ClientRequest`
  and `ClientRequestBuilder` to set field `upper_camel_case_headers`.
  `true` is for turning on Camel-Case headers.
  `false` to disable it.
  Default is `false`.

* Add trait `client::writer::UpperCamelCaseHeader` for
  `http::header::HeaderName`, let it can be converted to Camel-Case
  then writed to buffer.

* Add test `test_client::test_upper_camel_case_headers`.


fafhrd91:
Let’s remove upper, I think camel case is enough

fafhrd91:
this PR should be moved to actix-http repo

fafhrd91:
@PeterDing could you update PR for current master

actix-http/src/message.rs
```diff
@@ -40,6 +40,7 @@ pub struct RequestHead {
     pub method: Method,
     pub version: Version,
     pub headers: HeaderMap,
+    pub upper_camel_case_headers: bool,
```
fafhrd91: Let's use Flags instead

PeterDing:
@fafhrd91 I only add code to support the upper camel case for `RequestHead`.

fafhrd91:
RequestHead is fine

fafhrd91:
thanks!

PR #718, opened 2019-03-07
feat: enable use of patch as request method

jnnkky:
When I was trying out actix-web I noticed that I am not able to register handlers for the PATCH HTTP method.

```rust
app.resource("/users/{userId}", |r| {
        r.get().f(get_user);
        r.patch().f(patch_user); // can't do this
        r.delete().f(delete_user);
    });
```

I dove into the code and added the capabilities where I could figure out additions are needed. I saw it come up only once [here](https://github.com/actix/actix-web/blob/ceca96da281ec0b7d3bdce202f1eb84d2447ce15/src/pred.rs#L161). Is there a reason PATCH is not supported yet?

If I am completely wrong feel free to close this PR or point me in the right direction 👍 

Edit: I was not sure where to add tests for this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 * Add client HTTP Authentication methods `.basic_auth()`  and `.bearer_auth()`. #540
 
+* Add support for PATCH HTTP method
+
 ### Fixed
 
 * Ignored the `If-Modified-Since` if `If-None-Match` is specified. #680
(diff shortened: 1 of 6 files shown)


DoumanAsh:
Could you please add note to CHANGELOG?
In addition to that please add test case with patch requests

jnnkky:
@DoumanAsh Kindly check again, did not get around to adjust the PR until today.

DoumanAsh (approved):

DoumanAsh:
Thanks

PR #731, opened 2019-03-25
Remove StaticFilesConfig

### LINKED ISSUE #730: actix-fs: Tidy up configuration capability of StaticFiles/NamedFiles
Subj.

Since trait based configuration makes things more complicated we're going to remove it completely

As a side note, we can also check out extra features people might need

DoumanAsh:
Closes #730


actix-files/src/lib.rs
```diff
@@ -289,20 +282,44 @@ impl<S: 'static, C: StaticFileConfig> Files<S, C> {
         self
     }
 
+    /// Specifies mime override callback
+    pub fn mime_override<F>(mut self, f: F) -> Self where for<'r, 's> F: Fn(&mime::Name) -> DispositionType + 'static {
```
fafhrd91: do you really need this `for<>`?

actix-files/src/lib.rs
```diff
@@ -317,40 +334,31 @@ where
...
     type Response = ServiceResponse;
     type Error = Error;
-    type Service = FilesService<P, C>;
+    type Service = Self;
     type InitError = ();
     type Future = FutureResult<Self::Service, Self::InitError>;
 
     fn new_service(&self, _: &()) -> Self::Future {
-        ok(FilesService {
+        ok(Files {
```
fafhrd91: should `Files` just implement Clone trait?

actix-files/src/named.rs
```diff
@@ -11,17 +10,31 @@ use std::os::unix::fs::MetadataExt;
...
+use actix_web::http::header::{self, DispositionType, ContentDisposition, DispositionParam};
 use actix_web::http::{ContentEncoding, Method, StatusCode};
 use actix_web::{Error, HttpMessage, HttpRequest, HttpResponse, Responder};
 
-use crate::config::{DefaultConfig, StaticFileConfig};
 use crate::range::HttpRange;
 use crate::ChunkedReadFile;
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct Flags {
```
fafhrd91: lets use bitflags crate instead

fafhrd91 (changes requested):

actix-files/src/lib.rs
DoumanAsh: Right

actix-files/src/lib.rs
DoumanAsh: Yeah about that, it turned out that for some reason I cannot properly clone it.

I think it might be because of generic `P` without `Clone` bounds, I'm not sure if I can add it there?

To clairfy for some reason just Files::clone() returns `&Files`

fafhrd91: You need manually implement Clone trait

DoumanAsh: Done

fafhrd91 (approved):

Issue #709, opened 2019-02-27
Empty responses are not closed in HTTP/2


fanzeyi:
This might be related to #439.

When the server is started with `.bind_ssl` and the handler returns an empty string, the connection won't be closed. (Update: I noticed it's probably a bug related to HTTP/2)

Here is a minimal code to reproduce this problem:

```rust
use actix_web::{http, server, App, Path, Responder};
use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};

fn index(_: Path<(u32, String)>) -> impl Responder {
    ""
}

fn main() {
    let mut builder = SslAcceptor::mozilla_intermediate(SslMethod::tls()).unwrap();
    builder
        .set_private_key_file("key.pem", SslFiletype::PEM)
        .unwrap();
    builder.set_certificate_chain_file("cert.pem").unwrap();

    server::new(|| App::new().route("/{id}/{name}/index.html", http::Method::GET, index))
        .bind_ssl("127.0.0.1:8080", builder)
        .unwrap()
        .run();
}
```

Cargo.toml:
```rust
[package]
name = "actix-empty-ssl"
version = "0.1.0"
authors = ["Zeyi Fan <zeyi@fb.com>"]
edition = "2018"

[dependencies]
actix-web = { version = "0.7.18", features = ["ssl"] }
openssl = "0.10.18"
```

```bash
$ curl -k -i https://localhost:8080/123/abc/index.html
HTTP/2 200
content-type: text/plain; charset=utf-8
date: Wed, 27 Feb 2019 22:30:04 GMT
content-length: 0

# cURL hangs here forever.
```

Update:

If I pass `--http1.1` to curl, the connection would finish properly:

```bash
$ curl -k --http1.1 -i https://localhost:8080/123/abc/index.html
HTTP/1.1 200 OK
content-length: 0
content-type: text/plain; charset=utf-8
date: Wed, 27 Feb 2019 22:36:46 GMT
```

fanzeyi:
I tried to debug this issue and I think the bug might be inside the `poll_completed` function of H2Writer. The H2Writer is being polled forever with the flags `STARTED` and `EOF`, and because the response is empty (so the buffer is empty). The `reserve_capacity` function is never called. This causes poll_capacity keeps returning `Async::NotReady`. 

I attempted to fix this but I don't think this is correct, but I'm posting it here anyways:

```diff
diff --git a/src/server/h2writer.rs b/src/server/h2writer.rs
[...]

PR #737, opened 2019-03-29
[0.7] Fix never-ending HTTP2 empty response

fanzeyi:
This is the fix I mentioned in #709. I ended up with testing this change with our code base, and it seems to work fine.

I tested with sending 5000 requests to the server via curl sequentially as well as sending the requests simultaneously (with empty responses mixed).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 
 * Fix preflight CORS header compliance; refactor previous patch (#603). #717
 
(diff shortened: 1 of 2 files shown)


fafhrd91 (approved):

fafhrd91:
Could you update changes.md

fanzeyi:
Done! I also tested a few POST requests and it works fine.

fafhrd91:
thanks! we will release 0.7.19 later today

PR #739, opened 2019-03-29
fix the example in README.md

joekyo:
Remove the trailing `;` and add a missing `)` so the example works actually.


fafhrd91:
thanks!

PR #743, opened 2019-03-30
test-server: Request functions should accept path

DoumanAsh:
As name implies, I changed request shortcuts to accept paths.

Since we're still alpha I think breaking changes are ok.
And it seems more convenient than having hardcoded `/`

### DIFF
--- a/test-server/src/lib.rs
+++ b/test-server/src/lib.rs
@@ -159,33 +159,33 @@ impl TestServerRuntime {
     }
 
     /// Create `GET` request
-    pub fn get(&self) -> ClientRequest {
-        self.client.get(self.url("/").as_str())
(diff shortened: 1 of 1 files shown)


fafhrd91 (approved):

DoumanAsh:
I had to change the way we specify dev dependency for `test-server` https://github.com/actix/actix-web/pull/743/files#diff-80398c5faae3c069e4e6aa2ed11b28c0L104

Also added it as member of workspace

fafhrd91:
Don’t merge it for now. We will merge it before next release

fafhrd91:
Or rather, let’s merge but only changes to test-server. We’ll change tests after new test server release

DoumanAsh:
Left only `test-server` changes. I'll wait for CI and then merge if it is ok

Issue #734, opened 2019-03-26
CookieIdentityPolicy should raise a warning when a private cookie has been tampered with


ghost:
Hello
I have a question about the identity middleware and `CookieIdentityPolicy`.

let's say I manage my session cookies with

```rust
CookieIdentityPolicy::new(private_key).name("my_cookie_name")
```

The way it's managed, session cookies are signed and encrypted with `private_key`, so that attackers and clients can't see/modify them without the server knowing.

`CookieIdentityPolicy` retrieves the cookies from the request with https://github.com/actix/actix-web/blob/86a21c956c247ab098a7631bc18ae60032ef5eac/src/middleware/identity.rs#L352
but only returns `None` (i.e. user is logged out) if `req.cookies()` returns an error.

Such errors are returned if:

1. the cookie isn't valid utf-8 https://github.com/actix/actix-http/blob/c5c7b244be264154ad9b60335a88722b0139206b/src/httpmessage.rs#L119
2. the cookie doesn't have a name/`=` https://github.com/alexcrichton/cookie-rs/blob/9edf0ee7037440bf8244722c50804a278524811c/src/parse.rs#L108
3. the urldecoded value isn't valid utf-8 https://github.com/alexcrichton/cookie-rs/blob/9edf0ee7037440bf8244722c50804a278524811c/src/parse.rs#L82

The last two errors are from `Cookie::parse_encoded`.

The actual question:
If `private_key` is leaked, cookies can be altered in such a way that it raises error 1 or 3, and these errors are ignored by `CookieIdentityPolicy` (which just returns as if the user is logged out).

TLDR: Shouldn't there be a warning if the session cookie is at least correctly signed/encrypted but isn't valid (urlencoded) utf-8? for example

```
Cookie: my_cookie_name=[random bytes signed and encrypted with `private_key`]
```

fafhrd91:
this is good point. would you provide PR?

ghost:
Sure, though it doesn't seem that cookie-rs has support for cookies that are not utf-8

ghost:
You forked cookie-rs, I can patch the fork and solve the breaking changes in actix-web?

fafhrd91:
sure

PR #746, opened 2019-03-30
Warn when an unsealed private cookie isn't valid UTF-8

ghost:
Fixes https://github.com/actix/actix-web/issues/734

For now, actix-http builds (tests also build, but not all pass)

I don't know what to do with the "encoded" methods, as it's not in the rfcs...
Also there are some chars that are not allowed in cookie names/cookie values, and the code doesn't check for them now.

Memos:
https://tools.ietf.org/html/rfc6265
https://tools.ietf.org/html/rfc2616#section-2.2

### DIFF
--- a/actix-http/src/cookie/secure/private.rs
+++ b/actix-http/src/cookie/secure/private.rs
@@ -1,3 +1,4 @@
+use log::error;
 use ring::aead::{open_in_place, seal_in_place, Aad, Algorithm, Nonce, AES_256_GCM};
 use ring::aead::{OpeningKey, SealingKey};
 use ring::rand::{SecureRandom, SystemRandom};
@@ -57,9 +58,14 @@ impl<'a> PrivateJar<'a> {
         let unsealed = open_in_place(&key, nonce, ad, 0, sealed)
             .map_err(|_| "invalid key/nonce/value: bad seal")?;
 
-        ::std::str::from_utf8(unsealed)
-            .map(|s| s.to_string())
-            .map_err(|_| "bad unsealed utf8")
+        if let Ok(unsealed_utf8) = ::std::str::from_utf8(unsealed) {
+            Ok(unsealed_utf8.to_string())
+        } else {
+            error!("Private cookie does not have utf8 content!");
+            error!("It is likely the secret key used to encrypt them has been leaked.");
+            error!("Please change it as soon as possible.");
+            Err("bad unsealed utf8")
+        }
     }
 
     /// Returns a reference to the `Cookie` inside this jar with the name `name`
--- a/actix-http/src/cookie/secure/private.rs
+++ b/actix-http/src/cookie/secure/private.rs
@@ -1,6 +1,6 @@
 use std::str;
 
-use log::error;
+use log::warn;
 use ring::aead::{open_in_place, seal_in_place, Aad, Algorithm, Nonce, AES_256_GCM};
 use ring::aead::{OpeningKey, SealingKey};
 use ring::rand::{SecureRandom, SystemRandom};
@@ -63,7 +63,7 @@ impl<'a> PrivateJar<'a> {
         if let Ok(unsealed_utf8) = str::from_utf8(unsealed) {
             Ok(unsealed_utf8.to_string())
         } else {
-            error!("Private cookie does not have utf8 content!
+            warn!("Private cookie does not have utf8 content!
 It is likely the secret key used to encrypt them has been leaked.
 Please change it as soon as possible.");
             Err("bad unsealed utf8")
--- a/actix-http/src/cookie/secure/private.rs
+++ b/actix-http/src/cookie/secure/private.rs
@@ -1,3 +1,5 @@
+use std::str;
+
 use log::error;
 use ring::aead::{open_in_place, seal_in_place, Aad, Algorithm, Nonce, AES_256_GCM};
 use ring::aead::{OpeningKey, SealingKey};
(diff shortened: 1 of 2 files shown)


fafhrd91:
@Llaurence could #392 be fixed by this change as well?

fafhrd91:
@DoumanAsh ?

ghost:
Ah, it seems the links i posted are outdated, i'm updating them now.

If we enforce restrictions on input (cookie name/value/domain/etc.) and instead of percent-encoding them then #392 might be solved. But i need to check the new rfc to be sure

ghost:
`cargo test --all-features` pass by adding `path = "actix-http"` in /Cargo.toml and removing tests/test_{http,}server.rs and examples/client.rs . . .

I haven't tested with a real app yet.

todo
1. update doc examples
2. change the Display impl/Cookie::to_string to Cookie::to_vec
3. make Cookie::parse, Cookie::named and Cookie::new validate inputs
4. Make `CookieIdentityInner::set_cookie` use `SignedJar` to sign cookies (?)

DoumanAsh:
@fafhrd91 I'll take care of reviewing it

@Llaurence Will the final goal be to support all range of UTF-8 characters?

actix-http/src/cookie/mod.rs
```diff
@@ -737,10 +750,14 @@ impl<'c> Cookie<'c> {
         }
 
         if let Some(path) = self.path() {
+            // TODO think of something else than this
```
DoumanAsh: Where `fmt_parameters` is used?

If we do not have encoding information we either cannot do anything about it, or we need to infer encoding.
Both sounds bad...

actix-http/src/cookie/parse.rs
```diff
@@ -75,37 +75,68 @@ fn indexes_of(needle: &str, haystack: &str) -> Option<(usize, usize)> {
...
+    let name = CookieStr::Concrete(Cow::Owned(decoded_name.into_owned().into_bytes()));
+    let val = CookieStr::Concrete(Cow::Owned(decoded_value.into_owned().into_bytes()));
 
     Ok((name, val))
 }
 
+/// Trims SP (0x20) characters from the given slice.
+///
+/// Used for cookies.
+fn trim_slice(s: &[u8]) -> &[u8] {
```
DoumanAsh: Please add some simple tests for function itself in this file

DoumanAsh (commented):
We need to resolve the major problem: How to represent non-utf8 data.
Since we do not have encoding information we'd need to try to infer it using library such as https://docs.rs/encoding_rs/0.8.17/encoding_rs/

This brings us to question: do we need to support non-utf8? Or should we support all range of utf-8 characters instead?

ghost:
> do we need to support non-utf8? Or should we support all range of utf-8 characters instead?

This PR is meant to be a fix for https://github.com/actix/actix-web/issues/734, I thought cookie-rs should be able to parse any kind of data and then CookieIdentityPolicy warn the user when the decrypted data isn't utf8.

But the warning can just be printed by PrivateJar::unseal, there's indeed no need to change the whole cookie api... 

https://github.com/actix/actix-web/blob/d846328f36573a4b28ce354db73618049c0658fd/actix-http/src/cookie/secure/private.rs#L46

actix-http/src/cookie/mod.rs
ghost: It was used by `impl Display for Cookie`

ghost:
Old work is at https://github.com/Llaurence/actix-web/tree/cookie-utf8-old though it's not really useful anymore

DoumanAsh:
Yeah, if we can go away without switching to arbitary bytes in `cookie` I think that would be better...
I think we should support only UTF-8 and warn on non-UTF-8 input

actix-http/src/cookie/secure/private.rs
```diff
@@ -57,9 +58,14 @@ impl<'a> PrivateJar<'a> {
         let unsealed = open_in_place(&key, nonce, ad, 0, sealed)
             .map_err(|_| "invalid key/nonce/value: bad seal")?;
 
-        ::std::str::from_utf8(unsealed)
-            .map(|s| s.to_string())
-            .map_err(|_| "bad unsealed utf8")
+        if let Ok(unsealed_utf8) = ::std::str::from_utf8(unsealed) {
```
DoumanAsh: `::std::str::from_utf8` -> `str::from_utf8`

actix-http/src/cookie/secure/private.rs
```diff
@@ -57,9 +58,14 @@ impl<'a> PrivateJar<'a> {
...
+            Ok(unsealed_utf8.to_string())
+        } else {
+            error!("Private cookie does not have utf8 content!");
```
DoumanAsh: Please use single invocation of `error` macro.
There is no need to flow with multiple invocations (remember that logs have cost)

DoumanAsh (changes requested):

DoumanAsh:
Please consider adding test case for error branch

actix-http/src/cookie/secure/private.rs
```diff
@@ -57,9 +60,14 @@ impl<'a> PrivateJar<'a> {
         let unsealed = open_in_place(&key, nonce, ad, 0, sealed)
             .map_err(|_| "invalid key/nonce/value: bad seal")?;
 
-        ::std::str::from_utf8(unsealed)
-            .map(|s| s.to_string())
-            .map_err(|_| "bad unsealed utf8")
+        if let Ok(unsealed_utf8) = str::from_utf8(unsealed) {
+            Ok(unsealed_utf8.to_string())
+        } else {
+            error!("Private cookie does not have utf8 content!
```
DoumanAsh: Looks good, but I think it should be a `warn` macro rather than error.
It doesn't per se would indicate immediate error, but more or less a warning as original issue asks

DoumanAsh (approved):
Looks good, just minor comment

DoumanAsh:
Thanks

PR #748, opened 2019-03-31
Expand codegen to allow specify guards and async

DoumanAsh:
Route's attributes:

- `"path"` - Raw literal string with path for which to register handler
- `async` - Attribute to indicate that registered function is asynchronous.
- `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`

If resource type is not specified, attempts to infer it based on return type (if it fails, assume regular function)

### DIFF
--- a/actix-web-codegen/Cargo.toml
+++ b/actix-web-codegen/Cargo.toml
@@ -19,3 +19,4 @@ syn = { version = "0.15", features = ["full", "parsing"] }
 actix-web = { version = "1.0.0-alpha.2" }
 actix-http = { version = "0.1.0-alpha.2", features=["ssl"] }
 actix-http-test = { version = "0.1.0-alpha.2", features=["ssl"] }
+futures = { version = "0.1" }
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -1,118 +1,94 @@
 #![recursion_limit = "512"]
+//! Actix-web codegen module
+//!
+//! Generators for routes and scopes
+//!
+//! ## Route
+//!
+//! Macros:
+//!
+//! - [get](attr.get.html)
+//! - [post](attr.post.html)
+//! - [put](attr.put.html)
+//! - [delete](attr.delete.html)
+//!
+//! ### Attributes:
+//!
+//! - `"path"` - Raw literal string with path for which to register handle. Mandatory.
+//! - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
+//!
+//! ## Notes
+//!
+//! Function name can be specified as any expression that is going to be accessible to the generate
+//! code (e.g `my_guard` or `my_module::my_guard`)
+//!
+//! ## Example:
+//!
+//! ```rust
+//! use actix_web::HttpResponse;
+//! use actix_web_codegen::get;
+//! use futures::{future, Future};
+//!
+//! #[get("/test")]
+//! fn async_test() -> impl Future<Item=HttpResponse, Error=actix_web::Error> {
+//!     future::ok(HttpResponse::Ok().finish())
+//! }
+//! ```
 
 extern crate proc_macro;
 
+mod route;
+
 use proc_macro::TokenStream;
-use quote::quote;
 use syn::parse_macro_input;
 
-/// #[get("path")] attribute
+/// Creates route handler with `GET` method guard.
+///
+/// Syntax: `#[get("path"[, attributes])]`
+///
+/// ## Attributes:
+///
+/// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
+/// - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
 #[proc_macro_attribute]
 pub fn get(args: TokenStream, input: TokenStream) -> TokenStream {
     let args = parse_macro_input!(args as syn::AttributeArgs);
-    if args.is_empty() {
-        panic!("invalid server definition, expected: #[get(\"some path\")]");
-    }
-
-    // path
-    let path = match args[0] {
-        syn::NestedMeta::Literal(syn::Lit::Str(ref fname)) => {
-            let fname = quote!(#fname).to_string();
-            fname.as_str()[1..fname.len() - 1].to_owned()
-        }
-        _ => panic!("resource path"),
-    };
-
-    let ast: syn::ItemFn = syn::parse(input).unwrap();
-    let name = ast.ident.clone();
-
-    (quote! {
-        #[allow(non_camel_case_types)]
-        struct #name;
-
-        impl<P: 'static> actix_web::dev::HttpServiceFactory<P> for #name {
-            fn register(self, config: &mut actix_web::dev::ServiceConfig<P>) {
-                #ast
-                actix_web::dev::HttpServiceFactory::register(
-                    actix_web::Resource::new(#path)
-                        .guard(actix_web::guard::Get())
-                        .to(#name), config);
-            }
-        }
-    })
-    .into()
+    let gen = route::Args::new(&args, input, route::GuardType::Get);
+    gen.generate()
 }
 
-/// #[post("path")] attribute
+/// Creates route handler with `POST` method guard.
+///
+/// Syntax: `#[post("path"[, attributes])]`
+///
+/// Attributes are the same as in [get](attr.get.html)
 #[proc_macro_attribute]
 pub fn post(args: TokenStream, input: TokenStream) -> TokenStream {
     let args = parse_macro_input!(args as syn::AttributeArgs);
-    if args.is_empty() {
-        panic!("invalid server definition, expected: #[post(\"some path\")]");
-    }
-
-    // path
-    let path = match args[0] {
-        syn::NestedMeta::Literal(syn::Lit::Str(ref fname)) => {
-            let fname = quote!(#fname).to_string();
-            fname.as_str()[1..fname.len() - 1].to_owned()
-        }
-        _ => panic!("resource path"),
-    };
-
-    let ast: syn::ItemFn = syn::parse(input).unwrap();
-    let name = ast.ident.clone();
-
-    (quote! {
-        #[allow(non_camel_case_types)]
-        struct #name;
-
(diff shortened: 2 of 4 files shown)


actix-web-codegen/src/lib.rs
```diff
@@ -1,118 +1,171 @@
...
-/// #[get("path")] attribute
+/// Creates route handler without attached method guard.
+///
+/// Syntax: `#[handler("path"[, attributes])]`
+///
+/// ## Attributes:
+///
+/// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
+/// - `async` - Attribute to indicate that registered function is asynchronous.
+/// - `sync` - Attribute to indicate that registered as regular function.
```
fafhrd91: cant you auto detect async/sync functions by `impl Future`?

actix-web-codegen/src/lib.rs
```diff
@@ -1,118 +1,171 @@
...
+/// - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
```
fafhrd91, line 52: we probably should add some predefined guards, something like `content_type = "text/plain"`

fafhrd91:
regarding scope. first of all syntax is ugly.
much better to have resource registration, i.e:

```rust
#[service]
struct MyResource {
   ...
}

impl MyResource {
   #[get]
   fn index(&self) -> impl Responder {
       ...
   }

   #[post]
   fn update(&self) -> impl Responder {
       ...
   }

   #[get("/sub-path/index.html")]
   fn sub_index(&self) -> impl Responder {
       ...
   }
}
```

then you can register it with normal `App::new().service(MyResource)`
or within scope `App::new().service(web::scope("/prefix").service(MyResource)`

also what would be also good is async service initialization with custom parameters.
so developer could encapsulate resource initialization within service, something like redis or db connection

actix-web-codegen/src/lib.rs
DoumanAsh: I do auto detect it, if you take a look at tests, some of them don't mark as async/sync
These markers are mostly for cases when user want to override it

DoumanAsh:
@fafhrd91 regarding that, I thought about it. 

But the problem will be is that we will not be able to properly implement it as we already have macros for routes that generate own structs inside these `impl` so it is going to break stuff.
Unless it is possible to use `get` like macros in both outside of `Service` and within, I'm not sure about it.
Also `Service` derive will need to be over `impl MyService` rather than over struct definition itself, as we'll need to analyze methods
Anyway I'll think about it

actix-web-codegen/src/lib.rs
fafhrd91: If you auto detect then remove attributes.

DoumanAsh: You cannot auto-detect everything, user might have `MyBoxedAlias` type and there is no way to reliable detect whether it is future or not

DoumanAsh: I'll remove `sync` since it is default

actix-web-codegen/src/lib.rs
DoumanAsh: Do you mean it also as having these guards in `actix_web::guard`?

actix-web-codegen/src/lib.rs
fafhrd91: we dont need to support 100% of use cases,  80% is more than enough, but those 80% should be easy to use.

actix-web-codegen/src/lib.rs
fafhrd91: no, just for proc macros.

DoumanAsh:
Remove `scope` macro

actix-web-codegen/src/lib.rs
DoumanAsh: Remove `sync` but `async` remains. Though it is optional for use

DoumanAsh:
>also what would be also good is async service initialization with custom parameters.
so developer could encapsulate resource initialization within service, something like redis or db connection

Does HttpServiceFactory allows for async initialization?

fafhrd91:
It does

DoumanAsh:
@fafhrd91 I have some troubles right now to come up with proper macro for service you want.
I'm a bit unsure how to handle routing there.
Since if we're going to make completely new service we cannot really use `Route` or `Scope`.
Which would make me write lots of routing related code in generated code

But I give another go to a scope macro and came up with something like that:

```rust
#[derive(Scope)]
pub struct MyScope {
    #[service]
    api: Api
    #[guard]
    guards: actix_web::guard::AllGuard
}

[#scope]
impl MyScope {
    [#get("/")]
    fn index() -> impl Future<Item=HttpResponse, Error=actix_web::Error> {
        future::ok(HttpResponse::Ok().finish())
     }
}
```

fafhrd91:
Idea is interesting, but I don’t see much value in macros for scope. I’d rather have struct as a service impl. Yes, that would require to create completely new service and router. But if you want to give users to use &mut self in theirs method then you have to build new routing, existing functionality does not provide any of it

DoumanAsh:
@fafhrd91 Scope macro is mostly to segregate parts of your routes into separate components instead of writing boilerplate yourself

fafhrd91:
It is worth of solving this by macro. Something like mass service registration is enough, ie .services([route1,route2,.. ])

I am fine with simple macros that we have now.

DoumanAsh:
@fafhrd91 Well, if we don't need scope macros then let's merge it as it is.
I think macros for service implementation might be overly complex to include in this PR

fafhrd91:
Ok, makes sense. I’ll review later today

DoumanAsh:
Ok, I'll publish my other macros as library on crates.io later on.
Then i'll think about services

actix-web-codegen/src/lib.rs
```diff
@@ -1,118 +1,109 @@
 #![recursion_limit = "512"]
+//! Actix-web codegen module
+//!
+//! Generators for routes and scopes
+//!
+//! ## Route
+//!
+//! Macros:
+//!
+//! - [handler](attr.handler.html)
```
fafhrd91: remove this

actix-web-codegen/src/lib.rs
```diff
@@ -1,118 +1,109 @@
...
+//! - [get](attr.get.html)
+//! - [post](attr.post.html)
+//! - [put](attr.put.html)
+//! - [delete](attr.delete.html)
+//!
+//! ### Attributes:
+//!
+//! - `"path"` - Raw literal string with path for which to register handle. Mandatory.
+//! - `async` - Attribute to indicate that registered function is asynchronous.
```
fafhrd91: remove this too

fafhrd91 (changes requested):

actix-web-codegen/src/lib.rs
DoumanAsh: You mean to remove it from API?

actix-web-codegen/src/lib.rs
DoumanAsh: Ehm... I think it might be useful to specify when you cannot auto-detect asyncness... Are you sure?

fafhrd91: Yes

actix-web-codegen/src/lib.rs
fafhrd91: Yes

actix-web-codegen/src/lib.rs
DoumanAsh: Done

actix-web-codegen/src/lib.rs
DoumanAsh: Done

fafhrd91:
@DoumanAsh can we make something like:

```rust
#[get("/")]
#[guard="function"]
fn handler(...) -> {
   ...
}
```

DoumanAsh:
I think it might be possible... but aside from flooding namespace, I'm not sure about order of evaluation for macros.
I suspect it would first come, first to be expand.

I'd say we should avoid creating too many attributes that are not part of Derive, but if you want I can try to check it out

DoumanAsh:
@fafhrd91 I checked whether order of attributes matter.
If `#[get]` comes after `#[guard]` we will not be able to see `#[guard]` from within `#[get]`
I guess it could be possible to work it around, but that going to be a bit complicated if we're going to have multiple attributes like `#[guard]`

actix-web-codegen/src/route.rs
```diff
@@ -0,0 +1,165 @@
+extern crate proc_macro;
+
+use std::fmt;
+
+use proc_macro::TokenStream;
+use quote::{quote};
+
+enum ResourceType {
```
fafhrd91, line 8: i dont think we need this

actix-web-codegen/src/route.rs
```diff
@@ -0,0 +1,165 @@
...
+        match self {
+            &ResourceType::Async => write!(f, "to_async"),
+            &ResourceType::Sync => write!(f, "to"),
+        }
+    }
+}
+
+#[derive(PartialEq)]
+pub enum GuardType {
+    None,
```
fafhrd91: do we use this?

actix-web-codegen/Cargo.toml
```diff
@@ -13,9 +13,11 @@ proc-macro = true
 
 [dependencies]
 quote = "0.6"
+proc-macro2 = "0.4"
```
fafhrd91: do you use this?

actix-web-codegen/src/route.rs
DoumanAsh: I could use bool, but own enum makes it easier to just impl `Display` for it and use it in formatter

actix-web-codegen/src/route.rs
DoumanAsh: Nope, removing it

actix-web-codegen/Cargo.toml
DoumanAsh: Nope, removing it

actix-web-codegen/src/route.rs
fafhrd91: Right. It is ok. I misread it

fafhrd91 (approved):

PR #750, opened 2019-04-01
added docs for wrap and wrap_fn

Dowwie:


DoumanAsh (approved):
I think there is also scope middleware to do.

In addition to that you might want to have simple example for `wrap_fn` too

src/app.rs
```diff
@@ -112,7 +112,26 @@ where
         self
     }
 
-    /// Register a middleware.
+    /// Registers heavyweight Application-level middleware, in the form of a
```
fafhrd91: i am not sure about `heavyweight` and `lightweight`. should we maybe use extra info for clarification. something like "in form of reusable components", and "in-place request/response modification".
i expect at some point `wrap` and `wrap_fn` will be merged together, but at the moment rust's type inference is not powerful enough for that.

Dowwie:
I've updated the wording and added docs for scope-level middleware.

fafhrd91 (approved):

Firstyear:
@Dowwie Thanks for adding these docs, they look really great! One tiny future nitpick is we probably want to avoid emphasis (*word*) style I think? Anyway, thanks for the contribution!

PR #755, opened 2019-04-05
Fix clippy warning

k-nasa:
## Fixed according to warning of clippy
I fixed what has become a clippy deny.But there are still many warnings.

I do not know if clippy is perfect but I think it makes sense to work

### DIFF
--- a/actix-http/src/body.rs
+++ b/actix-http/src/body.rs
@@ -359,7 +359,7 @@ where
     }
 
     fn poll_next(&mut self) -> Poll<Option<Bytes>, Error> {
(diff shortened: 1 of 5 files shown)


actix-http/src/service/mod.rs
```diff
@@ -1,5 +1,5 @@
+mod http_service;
```
fafhrd91: do not rename this module

DoumanAsh (approved):

actix-http/src/service/mod.rs
k-nasa: I fixed it

PR #760, opened 2019-04-08
Fix body propagation in Response::from_error.

boxdot:
Fix error body propagation in Response::from_error.

Also:

* Fix typos.
* Remove period from error messages.
* Add test for custom response in json error handler.

### DIFF
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -1,7 +1,7 @@
 //! Http response
 use std::cell::{Ref, RefMut};
 use std::io::Write;
-use std::{fmt, io, str};
+use std::{fmt, str};
 
 use bytes::{BufMut, Bytes, BytesMut};
 use futures::future::{ok, FutureResult, IntoFuture};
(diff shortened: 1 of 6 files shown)


fafhrd91:
I think with this change stops error message propagation

boxdot:
> I think with this change stops error message propagation

I don't quite get it. `error_response` actually uses the response if it `InternalError::Response`.

Right now, it is not possible to return an error from a handler which will contain
some JSON information (or a different content type to text). Many APIs actually return a structured error message instead of plain text.

```
let mut buf = BytesMut::new();
let _ = write!(Writer(&mut buf), "{}", error);
resp.headers_mut()
    .insert(header::CONTENT_TYPE, HeaderValue::from_static("text/plain"));
resp.error = Some(error);
resp.set_body(Body::from(buf))
```

Dokuro-YH:
I have encountered the same problem and my custom json error doesn't work.

fafhrd91:
fixed in master

boxdot:
It would have been possible to commit on top of my changes. But I guess whatever works. Thanks for merging.

PR #783, opened 2019-04-17
Add Normalization middleware for in place

### LINKED ISSUE #636: Ignore double slashes in routes
Hi!

I need routes like routes like `/v1/something` to be matched with `POST /v1//something/` despite it's an client error.

I know there is a `NormalizePath` but it doesn't work with `POST` requests.
It's also impossible to fix this in the `Middleware`, at least as far as I know.

Any ideas on how this can be solved?

DoumanAsh:
Closes  #636

@fafhrd91 I'm not sure if there is actually any reason to use redirect when we can modify in place?
I think in place modification is much better approach performance-wise

### DIFF
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -1,0 +1,109 @@
+//! `Middleware` to normalize request's URI
+
+use regex::Regex;
+use actix_service::{Service, Transform};
+use futures::future::{self, FutureResult};
+
+use crate::service::{ServiceRequest, ServiceResponse};
+
+#[derive(Default, Clone, Copy)]
+/// `Middleware` to normalize request's URI in place
+///
+/// Performs following:
+///
+/// - Merges multiple slashes into one.
+pub struct NormalizePath;
+
+impl<S> Transform<S> for NormalizePath
+where
+    S: Service<Request = ServiceRequest, Response = ServiceResponse>,
+{
+    type Request = ServiceRequest;
+    type Response = ServiceResponse;
+    type Error = S::Error;
+    type InitError = ();
+    type Transform = NormalizePathNormalization<S>;
+    type Future = FutureResult<Self::Transform, Self::InitError>;
+
+    fn new_transform(&self, service: S) -> Self::Future {
+        future::ok(NormalizePathNormalization {
+            service,
+            merge_slash: Regex::new("//+").unwrap()
+        })
+    }
+}
+
+pub struct NormalizePathNormalization<S> {
+    service: S,
+    merge_slash: Regex,
+}
+
+impl<S> Service for NormalizePathNormalization<S>
+where
+    S: Service<Request = ServiceRequest, Response = ServiceResponse>,
+{
+    type Request = ServiceRequest;
+    type Response = ServiceResponse;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(&mut self) -> futures::Poll<(), Self::Error> {
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, mut req: ServiceRequest) -> Self::Future {
+        let head = req.head_mut();
+
+        let path = head.uri.path();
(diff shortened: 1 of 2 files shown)


src/middleware/normalize.rs
```diff
@@ -0,0 +1,93 @@
+//! `Middleware` to normalize request's URI
+
+use regex::Regex;
+use actix_service::{Service, Transform};
+use futures::future::{self, FutureResult};
+
+use crate::service::{ServiceRequest, ServiceResponse};
+
+lazy_static::lazy_static! {
+    static ref MERGE_SLASH: Regex = Regex::new("//+").unwrap();
```
fafhrd91: remove this, just store in middleware

src/middleware/normalize.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+#[derive(Default, Clone, Copy)]
+/// `Middleware` to normalize request's URI in place
+///
+/// Performs following:
+///
+/// - Merges double slashes into one.
+/// - Removes trailing slash.
+pub struct InPlace;
```
fafhrd91: i dont like name, let's name it `Normalize` or something like this

src/middleware/normalize.rs
```diff
@@ -0,0 +1,93 @@
...
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, mut req: ServiceRequest) -> Self::Future {
+        let head = req.head_mut();
+
+        let path = head.uri.path();
+        let original_len = path.len();
+        let path = MERGE_SLASH.replace_all(path, "/");
+        let path = path.trim_end_matches('/');
```
fafhrd91: i am not sure about this. what case does this cover?

fafhrd91 (changes requested):

fafhrd91:
redirect keeps urls clean, but lets not implement redirect for now

src/middleware/normalize.rs
DoumanAsh: To trim trailing slash

src/middleware/normalize.rs
DoumanAsh: Ok

src/middleware/normalize.rs
DoumanAsh: Ok, but I'll name it `NormalizePath`

src/middleware/normalize.rs
fafhrd91: So, that means I wouldn’t be able to have resources with trailing slash?

DoumanAsh: Yes, would you like to make it optional?

General idea is that it would equally normalize all paths so that they would have the same style.
After all, is there a difference between `/path` and `/path/`?

fafhrd91: That is no option. This should not happen. Remove it. And add more tests

DoumanAsh: Ok

src/middleware/normalize.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn test_in_place_normalization() {
+        let srv = FnService::new(|req: ServiceRequest| {
+            assert_eq!("/v1/something/", req.path());
+            req.into_response(HttpResponse::Ok().finish())
+        });
+
+        let mut normalize = block_on(NormalizePath.new_transform(srv)).unwrap();
+
+        let req = TestRequest::with_uri("/v1//something////").to_srv_request();
+        let _ = block_on(normalize.call(req)).unwrap();
```
fafhrd91: add assert for response.status().is_success(). same for next test

fafhrd91 (changes requested):

src/middleware/normalize.rs
DoumanAsh: Done

fafhrd91 (approved):

Issue #785, opened 2019-04-17
Should `time` be replaced with `std::time` for actix_web::cookie::Cookie max_age?


twharmon:
It seems the time crate isn't maintained anymore.

DoumanAsh:
`time` has some parsing functionality so you cannot replace it directly with `std::time`

fafhrd91:
@twharmon Pr is welcomed

fafhrd91:
we can migrate to chrono

Kilerd:
Ummm, I'm facing this issue too, when setting Cookie's max_age, I need to add `time` to my dependence which actually we don't use to parse time instead of using `chrono`.
I think there are two choices, one is exposing `time:Duration` as public so that user does not need to add `time` to the dependency list. Or for the long term, replace it with `chrono`

fafhrd91:
I think we should use `chrono`

Kilerd:
okay, I can help

fafhrd91:
i think `max_age` should accept number of seconds in `usize`. and `max_age_time` should accept chrono time

Kilerd:
I'm working on it, but I don't think that `max_age` method should accept `usize` as the parameter. because  for cookies , it can accept `0` and `-1`.
 - `0` means delete the cookies right not
- `-1` means delete the cookies when the browser is closed

I perfer to use `isize`

fafhrd91:
`max_age` now accepts value in seconds, max_age_time accepts time::Duration

PR #789, opened 2019-04-19
replate `time::Duration` with `chrono::Duration` and add `max_age_time` method

Kilerd:
as #785 says, I replate it and add two more tests for cookies' max-age

### DIFF
--- a/actix-http/src/cookie/builder.rs
+++ b/actix-http/src/cookie/builder.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 
-use time::{Duration, Tm};
+use time::Tm;
+use chrono::Duration;
 
 use super::{Cookie, SameSite};
 
@@ -16,7 +17,7 @@ use super::{Cookie, SameSite};
 ///
 /// ```rust
 /// use actix_http::cookie::Cookie;
-/// use time::Duration;
+/// use chrono::Duration;
 ///
 /// # fn main() {
 /// let cookie: Cookie = Cookie::build("name", "value")
@@ -200,7 +201,7 @@ impl CookieBuilder {
     ///
     /// ```rust
     /// use actix_http::cookie::Cookie;
-    /// use time::Duration;
+    /// use chrono::Duration;
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ license = "MIT/Apache-2.0"
 exclude = [".gitignore", ".travis.yml", ".cargo/config", "appveyor.yml"]
 edition = "2018"
 
+[package.metadata.docs.rs]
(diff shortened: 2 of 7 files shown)


fafhrd91:
i'd like to replace time in `actix-http::cookie` as well. do you think you can do that?

Kilerd:
sure, pls wait for a second

Kilerd:
`actix-http` also use `time::Tm` to represent the time when the cookie expires, I think I need more time to replace it.

Kilerd:
but looking through the code, I don't think we need to change the code using `Tm`

DoumanAsh:
If we're going to use `chrono` we need to use it for parsing too.
There is no benefit in having `chrono::Duration` while retaining `time`

In addition to that I think `std::time::Duration` should `chono::Duration`

Kilerd:
it means we need to maintain another copy of cookies repo, and I also do not know why `actix-http` need to reimplement cookies(compare these two repos, I can't see a lot of differences) rather than using `cookie`.

DoumanAsh:
@Kilerd we need own `cookie` because the existing one is no longer being maintained and in future, due to `ring` policy, it will be no longer possible to properly compile it as `ring 0.13.x` will be yanked completely.
We already have it so there is nothing bad about modifying it, since old `cookie` is unlikely to get updated anyway

Dowwie:
@Kilerd What is wrong with using std::time::Duration?  What is the benefit of this change?

Dowwie:
> i'd like to replace time in `actix-http::cookie` as well. do you think you can do that?

what is wrong with std::time crate?

DoumanAsh:
@Dowwie we need parsing of time strings

Dowwie:
> @Dowwie we need parsing of time strings

ah.  understood.  thanks

actix-http/src/cookie/builder.rs
```diff
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 
-use time::{Duration, Tm};
+use time::Tm;
```
fafhrd91, line 3: we should drop `time`, there is no reason to keep both.

fafhrd91:
@Kilerd could you also update actix-session/src/cookie.rs same as you did for identity

fafhrd91 (approved):

fafhrd91:
thanks

fafhrd91:
i had to remove chrono.

reasons:
* it can not parse rfc822 format
* it requires custom format string for rfc822
* internally it uses time crate

there is no reason to use it. i kept max_age changes though

Kilerd:
sorry for the late responding

> it can not parse rfc822 format

yep when I try to use chrono parsing the cookie expire time, I also found this issue.

Kilerd:
and why not just expose the `time:Duration` as public?

fafhrd91:
it would introduce confusion and it is used only for cookies.

PR #799, opened 2019-04-29
added clarification to docs regarding middleware processing sequence, added delete method to TestRequest

Dowwie:


fafhrd91 (approved):
good!

PR #825, opened 2019-05-07
Make `App::configure` take an `FnOnce`

nvzqz:
I noticed that this function is only ever called once, so it would be nice to have it be more flexible on what arguments can be passed.


fafhrd91:
thanks!

PR #830, opened 2019-05-08
Make JsonConfig send

### LINKED ISSUE #827: Make JsonConfig Send
JsonConfig is not Send:

```rust
// ...

pub struct JsonConfig {
    limit: usize,
    ehandler: Option<Rc<Fn(JsonPayloadError, &HttpRequest) -> Error>>, // <- Rc is not Send
}

// ...
```

This does not allow to share the config instance between workers causing each worker to have its own copy of the JsonConfig instance.

e.g. the following snippet does not compile:

```rust
// ...

pub fn instance_json_config() -> JsonConfig {
  // ...
}

pub fn main() -> io::Result<()> {
    let server_address = env::var("SERVER_ADDRESS").unwrap();
    let server_workers = env::var("SERVER_WORKERS").unwrap().parse().unwrap();
    let state = AppState::new();

    let json_config = Arc::new(instance_json_config());

    HttpServer::new(move || {
        App::new()
            .data(state.clone())
            .data(json_config.clone())
            .configure(app_router)
    })
    .workers(server_workers)
    .bind(server_address)?
    .run()
}
```

So I ended up writing something like that:

```rust
// ...

pub fn instance_json_config() -> JsonConfig {
  // ...
}

pub fn main() -> io::Result<()> {
    let server_address = env::var("SERVER_ADDRESS").unwrap();
    let server_workers = env::var("SERVER_WORKERS").unwrap().parse().unwrap();
    let state = AppState::new();

    HttpServer::new(move || {
        App::new()
            .data(state.clone())
            .data(instance_json_config())
            .configure(app_router)
    })
    .workers(server_workers)
    .bind(server_address)?
    .run()
}
```

which in turns will create a json config for each worker.

Is there another way to achieve that or am I missing something?

daddinuz:
Fixes #827 

Note that `ehandler` now must be Send + Sync, I don't know if this may cause compatibility issues:

```rust
// ...

pub struct JsonConfig {
    limit: usize,
    ehandler: Option<Arc<Fn(JsonPayloadError, &HttpRequest) -> Error + Send + Sync>>, // <--
}

// ...
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ### Changes
 
 * `App::configure` take an `FnOnce` instead of `Fn`
-
+* `JsonConfig` is now `Send + Sync`, this implies that `error_handler` must be `Send + Sync` too.
 
 ## [1.0.0-beta.3] - 2019-05-04
 
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -236,7 +236,7 @@ where
 #[derive(Clone)]
 pub struct JsonConfig {
     limit: usize,
-    ehandler: Option<Arc<Fn(JsonPayloadError, &HttpRequest) -> Error>>,
+    ehandler: Option<Arc<dyn Fn(JsonPayloadError, &HttpRequest) -> Error + Send>>,
 }
 
 impl JsonConfig {
@@ -249,7 +249,7 @@ impl JsonConfig {
     /// Set custom error handler
     pub fn error_handler<F>(mut self, f: F) -> Self
     where
-        F: Fn(JsonPayloadError, &HttpRequest) -> Error + 'static,
+        F: Fn(JsonPayloadError, &HttpRequest) -> Error + Send + 'static,
     {
         self.ehandler = Some(Arc::new(f));
         self
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -236,7 +236,7 @@ where
 #[derive(Clone)]
 pub struct JsonConfig {
     limit: usize,
-    ehandler: Option<Arc<dyn Fn(JsonPayloadError, &HttpRequest) -> Error + Send>>,
+    ehandler: Option<Arc<dyn Fn(JsonPayloadError, &HttpRequest) -> Error + Send + Sync>>,
 }
 
 impl JsonConfig {
@@ -249,7 +249,7 @@ impl JsonConfig {
     /// Set custom error handler
     pub fn error_handler<F>(mut self, f: F) -> Self
     where
-        F: Fn(JsonPayloadError, &HttpRequest) -> Error + Send + 'static,
+        F: Fn(JsonPayloadError, &HttpRequest) -> Error + Send + Sync + 'static,
     {
         self.ehandler = Some(Arc::new(f));
         self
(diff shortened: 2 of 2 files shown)


daddinuz:
Just discovered that there is an issue when registering an `Arc<JsonConfig>` with `App::data`.

`FromRequest` is implemented for `Json<T>` with associated `type Config = JsonConfig`, so the call to  `app_data::<JsonConfig>` to retrieve the configs inside the method `FromRequest::from_request` will return a None option because type `JsonConfig` != `Arc<JsonConfig>`.

Any suggestions?

daddinuz:
I've found a solution that I have implemented in this [branch](https://github.com/actix/actix-web/compare/master...daddinuz:register_raw_app_data):

Since `Data` internally wraps the value in `Arc`, we could wrap `JsonConfig` in `Data` instead of `Arc`, avoiding also a double indirection.

I noticed that by now there is only `App::data` which takes an arbitrary object and internally wraps it into `Data`, so I have just added the method `register_data` which takes an instance of `Data` directly and store it in the `App`.

So, instead of writing (as in the issue example):

```rust
// ...

pub fn instance_json_config() -> JsonConfig {
  // ...
}

pub fn main() -> io::Result<()> {
    let server_address = env::var("SERVER_ADDRESS").unwrap();
    let server_workers = env::var("SERVER_WORKERS").unwrap().parse().unwrap();

    let state = AppState::new();
    let json_config = Arc::new(instance_json_config());

    HttpServer::new(move || {
        App::new()
            .data(state.clone())
            .data(json_config.clone())
            .configure(app_router)
    })
    .workers(server_workers)
    .bind(server_address)?
    .run()
}
```

I would write:

```rust
use actix_web::web::Data;

// ...

pub fn instance_json_config() -> JsonConfig {
  // ...
}

pub fn main() -> io::Result<()> {
    let server_address = env::var("SERVER_ADDRESS").unwrap();
    let server_workers = env::var("SERVER_WORKERS").unwrap().parse().unwrap();

    let state = AppState::new();
    let json_config = Data::new(instance_json_config());

    HttpServer::new(move || {
        App::new()
            .data(state.clone())
            .register_data(json_config.clone())
            .configure(app_router)
    })
    .workers(server_workers)
    .bind(server_address)?
    .run()
}
```

solving [this](https://github.com/actix/actix-web/pull/830#issuecomment-490676508).

fafhrd91:
@daddinuz could you add entry to changes.md

you dont change anything, just use Data<JsonConfig>, `Into<Data<T>>` for `Data<T>` returns self

daddinuz:
@fafhrd91 sure, I'm going to update changes.md as soon as possible.

Regarding `Into<Data<T>> for Data<T>` this is will not work because there would be conflicting implementations since `From` trait is already implemented for `Data<T>`

fafhrd91:
let's add register_data in separate PR

fafhrd91:
actually, let's wait with register_data. for now `JsonConfig` could implement `Clone`

daddinuz:
Ok I've added entires to CHANGES.md and updated the branch to the latest master

fafhrd91 (approved):

fafhrd91:
thanks!

Issue #821, opened 2019-05-06
Feature request: Websocket actor support for responding to Sec-WebSocket-Protocol header


jairinhohw:
Using the websocket actor from the actix-web-actors crate, i ran into a problem with trying to create an actor for handling graphql requests from Apollo using the [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws/) package, in that it works well in firefox, but the websocket requests don't complete on chrome due to a missing Sec-WebSocket-Protocol header on the response.
Seeing the [code (line 43)](https://github.com/actix/actix-web/blob/master/actix-web-actors/src/ws.rs) for the handshake, in the commented out part of the documentation, it appears as that at some point it was allowed to pass a list of supported protocols, but i could not find it implemented in this repository.

fafhrd91:
you can this header yourself.

therealprof:
That worked. It is a bit tedious to copy over all code to implement a custom handshake and also my code won't profit from any fixed to it but at least it can be done. 😅

fafhrd91:
Pr is welcomed

jairinhohw:
I did make it work by copying and altering the handshake to respond to any protocol request, but i don't know how to make it possible to create an websocket actor passing an optional list of protocols, otherwise it will not be possible to have the old behavior of not accepting protocol request.

fafhrd91:
pr is merged

PR #835, opened 2019-05-10
add support for specifying protocols on websocket handshake

jairinhohw:
issue #821

### DIFF
--- a/actix-web-actors/src/ws.rs
+++ b/actix-web-actors/src/ws.rs
@@ -31,7 +31,7 @@ where
     A: Actor<Context = WebsocketContext<A>> + StreamHandler<Message, ProtocolError>,
     T: Stream<Item = Bytes, Error = PayloadError> + 'static,
 {
-    let mut res = handshake(req, &vec![])?;
+    let mut res = handshake(req)?;
     Ok(res.streaming(WebsocketContext::create(actor, stream)))
 }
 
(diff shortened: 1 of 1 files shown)


actix-web-actors/src/ws.rs
```diff
@@ -31,7 +31,24 @@ where
...
+    let mut res = handshake(req, &vec![])?;
+    Ok(res.streaming(WebsocketContext::create(actor, stream)))
+}
+
+/// Do websocket handshake and start ws actor.
+///
+/// `protocols` is a sequence of known protocols.
+pub fn start_with_protocols<A, T>(
+    actor: A,
+    protocols: &Vec<&str>,
```
fafhrd91: let's change this to `&[str]`

actix-web-actors/src/ws.rs
```diff
@@ -40,10 +57,13 @@ where
...
-// /// `protocols` is a sequence of known protocols. On successful handshake,
-// /// the returned response headers contain the first protocol in this list
-// /// which the server also knows.
-pub fn handshake(req: &HttpRequest) -> Result<HttpResponseBuilder, HandshakeError> {
+/// `protocols` is a sequence of known protocols. On successful handshake,
+/// the returned response headers contain the first protocol in this list
+/// which the server also knows.
+pub fn handshake(
+    req: &HttpRequest,
+    protocols: &Vec<&str>,
```
fafhrd91: this is breaking change, let's use new function

fafhrd91: also, should this be `Option<&[str]>`?

fafhrd91 (changes requested):

actix-web-actors/src/ws.rs
jairinhohw: it would make no sense to pass a `Some([])`, so i though to pass the empty vec to not use a protocol

fafhrd91:
thanks

PR #844, opened 2019-05-14
0.7 - Having IoStream implementation for SslStream to include peer certificate

fanzeyi:
Hello!

A little background: we have a half-production internal service that uses actix-web 0.7, and we'd like to get some information of the client certificate the user is using for security measure. Currently we are able to enforce the peer certificate validation in OpenSSL but we'd like to do more than just validation. This involves having some sort of access to the x509 identity in the actix-web part of the code. 

I dug in a little and discovered `stream_extensions` in the codebase. It seems no one is actually using this extension and I wonder if it is okay to modify the openssl bit of code that adds the client certificate to this extension.

Example code of using this:

```rust
use actix_web::{HttpRequest, Responder};
use openssl::hash::MessageDigest;
use openssl::x509::X509;

fn index(request: HttpRequest) -> impl Responder {
    if let Some(stream_extensions) = (*request).stream_extensions() {
        if let Some(cert) = (*stream_extensions).get::<X509>() {
            println!(
                "got peer certificate: {:?}",
                cert.digest(MessageDigest::sha256())
            );
        }
    } else {
        println!("no extension")
    }

    ""
}
```

We'd like very much to not maintain an internal fork of actix-web, but we do understand it is not actix-web's responsibility to provide this feature. Although it does make some sense to have the SslStream provide some information about the SSL connection. :P

Please let me know what do you think about this, and if there is a better way to gain the access to the certificate (or just `openssl::ssl::SslStream` inside actix-web). Thanks!

### DIFF
--- a/src/server/ssl/openssl.rs
+++ b/src/server/ssl/openssl.rs
@@ -1,4 +1,5 @@
 use std::net::{Shutdown, SocketAddr};
+use std::rc::Rc;
 use std::{io, time};
 
 use actix_net::ssl;
@@ -6,6 +7,7 @@ use openssl::ssl::{AlpnError, SslAcceptor, SslAcceptorBuilder};
 use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_openssl::SslStream;
 
+use extensions::Extensions;
 use server::{IoStream, ServerFlags};
 
 /// Support `SSL` connections via openssl package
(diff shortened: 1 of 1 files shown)


fafhrd91 (approved):

fafhrd91:
I'll merge this PR but I am not planing to make new 0.7 release

PR #846, opened 2019-05-16
Support building Query<T> from a query string

### LINKED ISSUE #843: 1.0.0 - Can ServiceRequest::from_parts be made public?
First, thanks for all your hard work, and moving to 1.0 is, for me overall, a good experience. 

I'm writing some middleware which needs access to the `HttpRequest` and the only way I see to accomplish this is `service_request.into_parts()`.

Then to finish the middleware `call` I need to use the `ServiceRequest` for `into_response` or pass to `srv.call(req)`. I was hoping to reconstruct the service request with `ServiceRequest::from_parts` however it turns out it's private. 

Can it be made public, or is there a better way to get `HttpRequest` from this?

milesgranger:
Will close #843 

Had a bit of a struggle going with the original suggestions, but thought that since one can't clone `ServiceRequest` and so long as they could only get a reference to the underlying parts, the approach here might be alright, since that'd keep `ServiceRequest::from_parts` private. But I suspect I'm wrong here. 

Appreciate the help. :+1:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ### Add
 
+* Add `Query<T>::from_query()` to extract parameters from a query string. #846
 * `QueryConfig`, similar to `JsonConfig` for customizing error handling of query extractors.
 
 ### Changes
--- a/src/types/query.rs
+++ b/src/types/query.rs
@@ -52,6 +52,16 @@ impl<T> Query<T> {
     pub fn into_inner(self) -> T {
         self.0
     }
+
+    /// Get query parameters from the path
+    pub fn from_query(query_str: &str) -> Result<Self, QueryPayloadError>
+    where
+        T: de::DeserializeOwned,
+    {
+        serde_urlencoded::from_str::<T>(query_str)
+            .map(|val| Ok(Query(val)))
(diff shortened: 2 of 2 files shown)


src/service.rs
```diff
@@ -61,6 +61,16 @@ impl ServiceRequest {
         ServiceRequest { req, payload }
     }
 
+    /// Get a mutable reference to parts
+    pub fn parts_mut(&mut self) -> (&mut HttpRequest, &mut Payload) {
```
fafhrd91: This should not be allowed, otherwise request could be cloned and you would get panic trying to modify request head

src/service.rs
```diff
@@ -61,6 +61,16 @@ impl ServiceRequest {
...
+        (&mut self.req, &mut self.payload)
+    }
+
+    /// Get a reference to parts
+    pub fn parts(&self) -> (&HttpRequest, &Payload) {
```
fafhrd91: Same here

fafhrd91:
Query should provide method that accepts path and returns extracted value

fafhrd91 (changes requested):
Let’s name function `from_query`.
Could you also add entry to changes.md

fafhrd91 (approved):
thanks!

Issue #824, opened 2019-05-07
Feature request: json convenience method for actix_web::test::TestRequest


boustrophedon:
Something like a `set_json_payload` for TestRequest would be nice.

The implementation would be pretty much the same as `actix_web::client::ClientRequest::send_json` - set the header to `application/json` and call `serde_json::to_string` on the body parameter. The only problem is that it would have to return a Result, which none of the other methods on TestRequest currently do.

I'd be willing to work on this if you think it's a good idea.

fafhrd91:
@boustrophedon do you plan to work on this ticket?

boustrophedon:
Sure!

What do you think about the return type? Should it return a result or panic if the data cannot be serialized?

fafhrd91:
panic is fine for TestRequest

fafhrd91:
merged

PR #851, opened 2019-05-18
Add `set_json_payload` method to TestRequest

boustrophedon:
Implements #824 - Add a `TestRequest::set_json_payload` convenience method.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -2,7 +2,7 @@
 use std::cell::RefCell;
 use std::rc::Rc;
 
-use actix_http::http::header::{Header, HeaderName, IntoHeaderValue};
+use actix_http::http::header::{ContentType, Header, HeaderName, IntoHeaderValue};
 use actix_http::http::{HttpTryFrom, Method, StatusCode, Uri, Version};
 use actix_http::test::TestRequest as HttpTestRequest;
 use actix_http::{cookie::Cookie, Extensions, Request};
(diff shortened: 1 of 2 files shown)


src/test.rs
```diff
@@ -477,6 +478,14 @@ impl TestRequest {
         self
     }
 
+    /// Set request payload serialized to JSON
+    pub fn set_json_payload<T: Serialize>(mut self, data: &T) -> Self {
```
fafhrd91: let's name it `set_json`

fafhrd91 (approved):
could you also add entry to changes

boustrophedon:
I updated CHANGES.md, added a line in the comment about setting the Content-Type header, and added a line to the test to assert the content type is application/json.

boustrophedon:
The travis build failed only on the beta build with a spurious error:
```
After this operation, 38.1 MB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  libssl-dev libssl1.0.0 openssl
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
The command "sudo apt-get install -y openssl libssl-dev libelf-dev libdw-dev cmake gcc binutils-dev libiberty-dev" failed and exited with 100 during .
```

Can you rerun?

fafhrd91 (approved):

fafhrd91:
thanks!

PR #870, opened 2019-05-28
Don't DISCONNECT from stream when reader is empty

### LINKED ISSUE #868: Fail to get Chunked Payload
It seems that when parsing a chunked Payload, it sometimes returns 0-bytes from the request.
To be a bit more specific, here's how to reproduce it:

I have this Actix web-server:
```rust
use actix_web::{error, middleware, web, App, Error, HttpResponse, HttpServer};
use futures::{Future, Stream};

pub fn save_payload(body: web::Payload) -> impl Future<Item = i64, Error = Error> {
    body
        .map_err(|e| {
            println!("save_file failed, {:?}", e);
            error::ErrorInternalServerError(e)
        })
        .fold(0i64, move |mut acc, bytes| -> Result<i64, Error> {
            acc += bytes.len() as i64;
            Ok(acc)
        })
        .map_err(|e| {
            println!("save_file failed, {:?}", e);
            error::ErrorInternalServerError(e)
        })
}

pub fn upload(
    body: web::Payload,
) -> impl Future<Item = HttpResponse, Error = Error> {
    save_payload(body)
        .map(|size| {
            println!("Received {} bytes of data", size);
            HttpResponse::Ok().body("OK")
        })
        .map_err(|e| {
            println!("failed: {}", e);
            e
        })
}

fn main() -> std::io::Result<()> {
    std::env::set_var("RUST_LOG", "actix_server=info,actix_web=info");
    env_logger::init();

    HttpServer::new(|| {
        App::new()
            .route("/{route_path}", web::post().to_async(upload))
    })
    .bind("127.0.0.1:8080")?
    .run()
}
```

It listens to any POST request on `8080` and prints the total length.

Now, using this `ffmpeg` command (it creates chunks of video streams that are POSTed to the output URL):
```bash
$> ffmpeg -t 30 -i http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 -codec copy -f dash http://127.0.0.1:8080/playlist.mpd
```

The server prints _sometimes_ (not always, but quite frequently) `Received 0 bytes of data`.
If you want to verify the result, you can have the same server in Node:
```javascript
const http = require('http');

http.createServer((req, res) => {
    if (req.method === 'POST') {
        let body = '';
        req.on('data', chunk => {
            body += chunk.toString();
        });
        req.on('end', () => {
            console.log("Received at", req.url, body.length, "bytes");
            res.end("OK");
        });
    } 
}).listen(8080);
```

I might do something wrong, let me know if you need more details.

PS : I tried to reproduce it with `curl` but I couldn't, so it might something specific to FFmpeg...

ngotchac:
This should fix #868 

The issue seemed to be that `read_available` would return `true` when nothing has been read from the stream, which disconnects the connection. However, for chunked transfers, it could be that nothing is read from the stream at some point (?)

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -799,7 +799,7 @@ where
         match read {
             Ok(n) => {
                 if n == 0 {
-                    return Ok(Some(true));
+                    return Ok(Some(false));
                 } else {
                     read_some = true;
                     unsafe {
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -693,18 +693,20 @@ where
                     }
                 } else {
                     // read socket into a buf
-                    if !inner.flags.contains(Flags::READ_DISCONNECT) {
-                        if let Some(true) =
-                            read_available(&mut inner.io, &mut inner.read_buf)?
-                        {
-                            inner.flags.insert(Flags::READ_DISCONNECT);
-                            if let Some(mut payload) = inner.payload.take() {
-                                payload.feed_eof();
-                            }
-                        }
-                    }
+                    let should_disconnect = if !inner.flags.contains(Flags::READ_DISCONNECT) {
+                        read_available(&mut inner.io, &mut inner.read_buf)?
+                    } else {
+                        None
+                    };
 
                     inner.poll_request()?;
+                    if let Some(true) = should_disconnect {
+                        inner.flags.insert(Flags::READ_DISCONNECT);
+                        if let Some(mut payload) = inner.payload.take() {
+                            payload.feed_eof();
+                        }
+                    };
+
                     loop {
                         if inner.write_buf.remaining_mut() < LW_BUFFER_SIZE {
                             inner.write_buf.reserve(HW_BUFFER_SIZE);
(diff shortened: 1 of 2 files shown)


fafhrd91 (commented):
`n == 0` Is disconnect

ngotchac:
To be honest, I don't know that much about HTTP requests, but it seems that this fixes the issue I had (#868), pass all the tests and doesn't introduce any other issue. Plus, reading from [the docs](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read):

> If n is 0, then it can indicate one of two scenarios:
> 
> 1. This reader has reached its "end of file" and will likely no longer be able to produce bytes. **Note that this does not mean that the reader will always no longer be able to produce bytes.**
> 2. The buffer specified was 0 bytes in length.

fafhrd91:
This is very sensitive part of code, I need test case or reproducible example

ngotchac:
I added an example to reproduce the issue in #868 , would that be sufficient?

ngotchac:
Hey @fafhrd91, I actually found the real issue: the stream was actually closed before the server would send a response, thus ending the `io::Read` stream and setting the `Flags::READ_DISCONNECT` flag. `fn poll_request` would do nothing then, since `fn can_read()` returned `false` because of the disconnection flag. The last chunk of the request would never make it to the decoder, and the body would thus be missing at least a chunk.

I added a test to shows this issue (first run is without the fix, second is with the fix):

image

fafhrd91 (approved):

fafhrd91:
thanks!

PR #880, opened 2019-05-31
Scope configuration

denysvitali:
As discussed on Gitter, this PR implements scope configuration, using the same syntax of the App configuration.

I'll improve the doc and add more tests, but for now it seems to work as expected


src/scope.rs
```diff
@@ -83,6 +84,62 @@ impl Scope {
...
+        where
+            F: FnOnce(&mut ScopeConfig),
+    {
+        let mut cfg = ScopeConfig::new();
+        f(&mut cfg);
+        self.services.extend(cfg.services);
+        self.guards.extend(cfg.guards);
+        //self.data.extend(cfg.data);
+
+        if cfg.data.is_some() {
```
fafhrd91: What is the reason?

src/config.rs
```diff
@@ -239,6 +240,58 @@ impl ServiceConfig {
     }
 }
 
+pub struct ScopeConfig {
```
fafhrd91: Let’s use ServiceConfig

fafhrd91 (changes requested):

fafhrd91:
Let’s use ServiceConfig, in that case we could call config function on app level or service level

denysvitali:
Okay, I'll do that :rocket:

src/scope.rs
denysvitali: Because of this:

https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/scope.rs#L63-L66

`data` is an `Option<Extensions>` in `Scope`, but a `Vec<Box<DataFactory>>` in `App`, therefore the configure function must handle the case where the `Option` is `None`, and merge accordingly.

Am I wrong?

https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/app.rs#L34

denysvitali: For reference:
https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/app.rs#L29-L38

vs:
https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/scope.rs#L63-L71

denysvitali: Whilst we can do this in `App`:
https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/app.rs#L129-L139

We can't do it in `Scope` because of the `Option` type:
https://github.com/actix/actix-web/blob/ddce1725725bfc89a1e385b1a366488b1c2d79e5/src/scope.rs#L118-L143

fafhrd91:
Could you complete it today, I’d like to make release. Or I can implement configure method

denysvitali:
It should be okay now, can you please check so that we can add it to 1.0.0?

src/scope.rs
```diff
@@ -83,6 +84,59 @@ impl Scope {
...
+    pub fn configure<F>(mut self, f: F) -> Self
+        where
+            F: FnOnce(&mut ServiceConfig),
+    {
+        let mut cfg = ServiceConfig::new();
+        f(&mut cfg);
+        self.services.extend(cfg.services);
+
+        if !cfg.data.is_empty() {
+            if (&self).data.is_some() {
```
fafhrd91: If it is none create one

actix-http/src/extensions.rs
```diff
@@ -2,6 +2,8 @@ use std::any::{Any, TypeId};
 use std::fmt;
 
 use hashbrown::HashMap;
+use hashbrown::hash_map::{Keys, Values};
```
fafhrd91: Remove changes to extensions.rs

fafhrd91 (changes requested):

denysvitali:
If you give me a couple of hours, I'll make the requested changes.

Do you think we are still on time to add this to the 1.0.0 release?

fafhrd91:
sure

fafhrd91:
remove all changes to actix-http

denysvitali:
Done :)

fafhrd91 (approved):

fafhrd91:
Thanks!

denysvitali:
You're welcome 👍

PR #890, opened 2019-06-05
Mark default enabled package features in the docs

senden9:
I find it quite useful to see the default features right in the documentation. So i don't need to dig into `Cargo.toml`.


fafhrd91 (approved):
thanks!

PR #895, opened 2019-06-05
Remove ring subsubdependency when using actix-multipart.

simlay:
This is related to #741. 

Setting the default features to false seems like an anti-pattern but it doesn't seem to be a feature to [disable single features in cargo](rust-lang/cargo#3126) yet.

PR #900, opened 2019-06-06
Remove ring subsubdependency when using actix-files.

simlay:
This is basically the same PR as #895 but for actix-files rather than actix-multipart.


fafhrd91 (approved):
thanks!

PR #908, opened 2019-06-11
get_identity from HttpMessage

### LINKED ISSUE #904: Extract Session from ServiceRequest
I create a middleware that checks permissions with `Session`. so I call `into_parts()` in `ServiceRequest` to get `HttpRequest` and use this to get `Session`. but there is no way to put them back or create a new `ServiceRequest` to pass to next middleware.

```
    fn call(&mut self, sreq: ServiceRequest) -> Self::Future {
        let (req, _) = sreq.into_parts();

        Session::extract(&req).map(|session| {
            // ...
        })
    }

```

nottxy:
Add `middleware::identity::RequestIdentity` trait to `get_identity` from `HttpMessage`

This PR will fix https://github.com/actix/actix-web/issues/904

### DIFF
--- a/src/middleware/identity.rs
+++ b/src/middleware/identity.rs
@@ -61,7 +61,10 @@ use crate::cookie::{Cookie, CookieJar, Key, SameSite};
 use crate::error::{Error, Result};
 use crate::http::header::{self, HeaderValue};
 use crate::service::{ServiceRequest, ServiceResponse};
-use crate::{dev::Payload, FromRequest, HttpMessage, HttpRequest};
+use crate::{
+    dev::{Extensions, Payload},
+    FromRequest, HttpMessage, HttpRequest,
+};
 
 /// The extractor type to obtain your identity from a request.
 ///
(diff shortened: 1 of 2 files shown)


src/middleware/identity.rs
```diff
@@ -119,13 +118,35 @@ impl Identity {
...
+        }
+    }
 }
 
 struct IdentityItem {
     id: Option<String>,
     changed: bool,
 }
 
+/// Helper trait that allows to get Identity
```
fafhrd91: need more docs. add note that it could be used in middleware but identity policy must be set before any other middleware that needs identity

max-frai:
I think we also need get_session to get session from `actix-session` inside middlewares, not only identity?

nottxy:
It is already supported, https://docs.rs/actix-session/0.1.1/actix_session/trait.UserSession.html

```
use actix_session::UserSession;

let session = http_request.get_session();
let session = service_request.get_session();
```

fafhrd91 (approved):

fafhrd91:
thanks!

PR #912, opened 2019-06-12
updated actix-session to support login and logout functionality

Dowwie:
To see the impact of these changes, see [main.rs in this repo](https://github.com/Dowwie/actix_session_test/blob/master/src/main.rs).

Specifically:
- L38:  invoke session.renew() at successful login
- L45:  session.purge() at logout


Dowwie:
Note that I currently cannot get a Session from within the scope of a ``from_request`` of an Extractor using UserSession trait because ``from_request`` receives an &HttpRequest where as a &mut HttpRequest is required for the UserSession trait's get_session to work.

In other words, I'm trying to get a Session from within an extractor and haven't found an alternative to cloning HttpRequest to do so.  One possible solution is to simply make actix-session/session.rs/SessionInner public and then I can replicate the logic in get_session to get it from extensions.  I don't need mutability in the case where Session already exists in request state and I *don't* want to create it if it it isn't.

fafhrd91:
you can always clone http request. also maybe get_session should take readonly ref

Dowwie:
Yeah, currently I am cloning req.  What do you think about renaming "get_session" to "get_or_create_session", to make it more inline with the underlying behavior, and add another public method, get_session, that simply returns Option<Session> ?

Dowwie:
@fafhrd91 looks good now?

actix-session/src/cookie.rs
```diff
@@ -308,7 +308,7 @@ where
         Session::set_session(state.into_iter(), &mut req);
 
         Box::new(self.service.call(req).map(move |mut res| {
-            if let Some(state) = Session::get_changes(&mut res) {
+            if let (_status, Some(state)) = Session::get_changes(&mut res) {
```
fafhrd91: we should update cookies only if session has changes, otherwise we should not set cookie. i think this is the reason for #923

fafhrd91 (changes requested):
looks good, except cookie session

Dowwie:
> looks good, except cookie session

@fafhrd91 I added a commit that I believe is what you were asking for -- confirm?  https://github.com/actix/actix-web/pull/912/commits/0e05b37082fa57444858b46e1a8b57b731777f1c

@max-frai -- Can you run this PR through your integration tests and confirm that this change addresses your issues in [923](https://github.com/actix/actix-web/issues/923)?  I'm not using cookie sessions but rather redis sessions so I haven't got integration tests for this side.

Dowwie:
One important note about this new approach is that when a programmer wants to create a session with the first request to a web server , the only way to create the session is to set state.  Currently, a new, empty session is created upon first request.  The way that the current session logic accomplishes the session creation is to create a session entry with empty state -- no changes.  I'm leaning towards replicating this to keep behavior consistent.  It's not hard work to add support for initializing a session upon first request, even if its empty and has no changes.  

Update:  I've addressed this in a subsequent commit

actix-session/src/cookie.rs
```diff
@@ -308,10 +308,10 @@ where
         Session::set_session(state.into_iter(), &mut req);
 
         Box::new(self.service.call(req).map(move |mut res| {
-            if let Some(state) = Session::get_changes(&mut res) {
-                res.checked_expr(|res| inner.set_cookie(res, state))
-            } else {
-                res
+            match Session::get_changes(&mut res) {
+                (SessionStatus::Changed, Some(state)) =>
```
fafhrd91: i think we should update cookie on `change` and `renew`, and delete cookie on `purge`

fafhrd91:
i am fine with new state. i added comment for cookie session

actix-session/src/cookie.rs
Dowwie: the latest commit should reflect this

Dowwie:
@fafhrd91 the latest commit addresses unchanged update to cookie and adds support for purging cookie -- should be ready.

actix-session/src/cookie.rs
Dowwie: done

fafhrd91 (approved):

fafhrd91:
@Dowwie could you also update changes.md for actix-session

Dowwie:
> @Dowwie could you also update changes.md for actix-session

done.

Dowwie:
> @Dowwie could you also update changes.md for actix-session

I updated actix-session CHANGES and actix-web CHANGES.  This release introduces SessionState and new Session behaviors -- renew and purge.  I thought this is sufficient to bump actix-session from 0.1.x to 0.2 and so updated Cargo.toml and CHANGES respectively.

fafhrd91 (approved):

PR #922, opened 2019-06-18
Use encoding_rs crate instead of unmaintained encoding crate

messense:
https://docs.rs/encoding_rs/0.8.17/encoding_rs/

https://hsivonen.fi/encoding_rs/


fafhrd91:
could you add entries to changes.md to actix-http and actix-web

fafhrd91 (approved):

messense:
> could you add entries to changes.md to actix-http and actix-web

Done.

PR #944, opened 2019-06-26
removed changed from SessionInner actix/actix-web#943

### LINKED ISSUE #943: Is ``changed`` unnecessary in SessionInner?
I'm currently reimplementing cookie sessions (to gain a better understanding of actix-web and its services) and I noticed that the field ``changed`` is set to true 3 times but apart from that the field is never used. Is it used somewhere else and I missed something?

https://github.com/actix/actix-web/blob/c0c71f82c00fdac964bcf588c2ea49c4c18d5de7/actix-session/src/lib.rs#L104

https://github.com/actix/actix-web/blob/c0c71f82c00fdac964bcf588c2ea49c4c18d5de7/actix-session/src/lib.rs#L120

https://github.com/actix/actix-web/blob/c0c71f82c00fdac964bcf588c2ea49c4c18d5de7/actix-session/src/lib.rs#L130

https://github.com/actix/actix-web/blob/c0c71f82c00fdac964bcf588c2ea49c4c18d5de7/actix-session/src/lib.rs#L137

jensmetzner:
close #943

### DIFF
--- a/actix-session/src/lib.rs
+++ b/actix-session/src/lib.rs
@@ -101,7 +101,6 @@ impl UserSession for ServiceRequest {
 #[derive(Default)]
 struct SessionInner {
     state: HashMap<String, String>,
-    changed: bool,
 }
 
 impl Session {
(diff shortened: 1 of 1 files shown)


fafhrd91:
proper fix is in #912

Issue #277, opened 2018-06-03
Responder for tuples


fafhrd91:
here is idea. we can use tuples for response generation, something like:

```rust
fn index(path: Path<..>) -> impl Responder {
    (StatusCode::OK, Json::new(model))
}
```

or 

```rust
fn index(path: Path<..>) -> impl Responder {
    (StatusCode::OK, (http::CONTENT_TYPE, "plain/json"), Json<MyStruct>)
}
```
order should not matter

```rust
fn index(path: Path<..>) -> impl Responder {
    ("Welcome!", StatusCode::OK, (http::CONTENT_TYPE, "plain/text"))
}
```

@mitsuhiko @DoumanAsh thoughts?

DoumanAsh:
One thing that bothers me is that it might be quite difficult to make all components of tuple generic.

What I can think as mandatory arguments are:
- Status code
- Body(or no body)

Everything else can be described as generic `HttpRequestPartBuilder` which would let's say have method that takes `&mut HttpRequest` and modifies it accordingly

The problem is that can we make all elements of tuple completely generic and therefore optional? 
e.g. can we omit things like `StatusCode`?
I actually don't want to allow it since it is clearly error prone as normally user would want to respond with proper code

joshleeb:
The first example could be useful, as it seems like that is a common style of response, but for the most part I agree with @DoumanAsh. 

Looking at the examples above I am also required to think a bit more to determine what each element of the tuple is. 

I do think this makes sense in a limited capacity, such as when defining multiple headers:

```rust
fn index(req: HttpRequest) -> HttpResponse {
    HttpResponse::Ok()
        .headers(vec![
             ("X-TEST", "value"),
             (http::header::CONTENT_TYPE, "application/json"),
        ])
        .finish()
}
```

Or for building common responses, such as a response with a JSON body and status code.

But for this header example, it's already pretty similar to what we can currently do:

```rust
fn index(req: HttpRequest) -> HttpResponse {
    HttpResponse::Ok()
        .header("X-TEST", "value")
        .header(http::header::CONTENT_TYPE, "application/json")
        .finish()
}
```

fafhrd91:
this method is just simple extension to what have already in form of Responder's, i.e. `Json<T>`
i see this as a simple form for returning `Json` with custom status code.

something like `Ok((StatusCode::CREATED, Json<MyStruct>))`

fafhrd91:
[...]

PR #954, opened 2019-07-04
Impl Responder for (T, StatusCode) where T: Responder

messense:
It's quite common to respond with custom status code, for example in Python Flask web framework you can do:

```python
@app.errorhandler(404)
def not_found(error):
    return render_template('error.html'), 404
```

http://flask.pocoo.org/docs/1.0/quickstart/#about-responses

Adding this impl would make life easier when you want to return different status code in different situation. For example, return `201` when a record is created, `200` when a record is updated, consider that you may also need to handle errors, so the  route function might be defined as:

```rust
fn upsert() -> Result<web::Json<T>, Error> {
}
```

without this impl you have to return `impl Responder` or `HttpResponse` type, returning `HttpResponse` type isn't very convenient and with `impl Responder` you might have to do something like this:

```rust
fn upsert() -> impl Responder {
    let res: Result<(T, StatusCode), Error> = {
        // do something, return different status code
    };
    res.map(|(x, s)| web::Json(x).with_status(s))
}
```

which is better but not as clear as returning `(T, StatusCode)`

Ref #277

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## [1.0.4] - TBD
 
+### Added
+
+* Add `Responder` impl for `(T, StatusCode) where T: Responder`
+
 ### Changed
 
 * Upgrade `rand` dependency version to 0.7
--- a/src/responder.rs
+++ b/src/responder.rs
@@ -137,6 +137,22 @@ impl Responder for () {
     }
 }
 
+impl<T> Responder for (T, StatusCode)
+where
+    T: Responder,
+{
+    type Error = T::Error;
+    type Future = CustomResponderFut<T>;
+
+    fn respond_to(self, req: &HttpRequest) -> Self::Future {
+        CustomResponderFut {
+            fut: self.0.respond_to(req).into_future(),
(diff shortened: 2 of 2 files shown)


fafhrd91:
You can use extra Responder methods https://docs.rs/actix-web/1.0.3/actix_web/trait.Responder.html#method.with_status

messense:
@fafhrd91 As I mentioned in the PR description, `with_status` works but it's not very ergonomic when dealing with `Result` type and different status code. Because it requires `impl Responder` as return type, it doesn't work nicely with `?` operator.

fafhrd91:
you can use .map()

fafhrd91:
`Ok(res?.with_status(400))` also works

messense:
Yeah, all that works but it usually requires adding type annotation to a local variable (`Result<T, E>`) which we tends to omit at first, then you get a `cannot infer type` compile error.

image

fafhrd91 (approved):

fafhrd91:
Would be also good to add text to the book

fafhrd91:
thanks

PR #956, opened 2019-07-04
Update the copyless version to 0.1.4

jrmuizel:
< 0.1.4 failed to check for null when doing allocations which could lead to null dereferences.


fafhrd91 (approved):
Thanks

PR #984, opened 2019-07-17
Log error results in Logger middleware (closes #938)

### LINKED ISSUE #938: Error reporting in 1.0
I am still quite new to actix and rust in general, so I apologise if I have missed something obvious in the text below.

In actix-web 0.7, if a handler returned an actix Error, the error message was logged:

```
ERROR 2019-06-24T07:29:59Z: actix_web::pipeline: Error occurred during request handling, status: 500 Internal Server Error test
DEBUG 2019-06-24T07:29:59Z: actix_web::pipeline: Custom { kind: Other, error: StringError("test") }
INFO 2019-06-24T07:30:00Z: actix_web::middleware::logger: 127.0.0.1:56449 "GET /error HTTP/1.1" 500 0 "-" ...
```

In 1.0, errors no longer appear to be logged. If the logger middleware is enabled, it is possible to see that a handler has returned a 500 response, but actix gives no indication of what the error actually was in the logs.

```
 INFO 2019-06-24T08:49:54Z: actix_web::middleware::logger: 127.0.0.1:59661 "GET /error HTTP/1.1" 500 4 "-" ...
```

I gather that I can implement a custom error type in order to log the error, and optionally display something else in the http response. The following seems to work:

```
#[derive(Debug)]
struct UserError {
    err: failure::Error
}

impl actix_web::error::ResponseError for UserError {
    fn error_response(&self) -> HttpResponse {
        error!("caught internal server error: {:?}", self);
        HttpResponse::new(
            http::StatusCode::INTERNAL_SERVER_ERROR)
    }
}

impl fmt::Display for UserError {
        write!(f, "hidden error")
    }
}

impl<T: Into<failure::Error>> From<T> for UserError {
    fn from(t: T) -> UserError { UserError { err: t.into() } }
}

#[get("/error")]
fn errortest() -> impl Future<Item = String, Error = UserError> {
    futures::future::err(format_err!("test fail")).from_err()
}
```

But that took me quite some time to figure out as someone new to the ecosystem. The other web frameworks I have used in the past all hide error messages from end users by default, and show them only to the developer, so the fact that actix does the opposite caught me by surprise. While custom error types may be the best solution in the long run, for the benefit of people starting out, might it make more sense if actix's Error type behaved differently to how it does now?

Thanks for all your work on actix :-)

vmalloc:


src/middleware/logger.rs
```diff
@@ -202,6 +203,10 @@ where
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let res = futures::try_ready!(self.fut.poll());
 
+        if let Some(error) = res.response().error() {
+            debug!("Error in response: {:?}", error);
```
fafhrd91: You should filter 500.

fafhrd91 (changes requested):
Could you add entry to changes

src/middleware/logger.rs
vmalloc: done

vmalloc:
Added to changelog

fafhrd91 (approved):

fafhrd91:
thanks!

PR #989, opened 2019-07-18
Re-implement Host predicate

antonok-edm:
There used to be a Host predicate to allow matching requests based on the host. For some reason, this got commented out and abandoned. I've re-implemented it and added some additional tests.


fafhrd91 (approved):

fafhrd91:
I commented it out because of connection_info, but your version is good. Thanks!

PR #988, opened 2019-07-18
Add the start_with_addr() function

naerbnic:
This function allows a client to obtain an address to the target websocket actor on construction.

PR #991, opened 2019-07-19
make Query payload public

### LINKED ISSUE #987: Consider making Query payload public
At the moment it is possible to destructure the `web::Json` and `web::Path` types inside the handler signature, which is very useful e.g. instead of
```
fn handler(wrapped: Json<MyData>) {
    let my_data = wrapped.into_inner();
    // ...
}
```
we do
```
fn handler(Json(my_data): Json<MyData>) {
    // ...
}
```
However this is not possible with `Query` as the contents are not public. I understand the logic, it stops users from constructing a `Query` with invalid contents. However, this is already possible with `Json` (no guarantee that the contents will be deserializable) and it is not possible to return a `Query` as a `Responder` anyway so I don't think there is really an issue.

jesskfullwood:
Closes #987

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 * Re-implement Host predicate (#989)
 
+### Changed
+
+* `Query` payload made `pub`. Allows user to pattern-match the payload.
+
 
 ## [1.0.5] - 2019-07-18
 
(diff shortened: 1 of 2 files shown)


fafhrd91 (approved):
Could you add entry to changes

jesskfullwood:
~~I can - it would require adding a new header for `1.0.6`, which might result in merge conflicts if #988 does the same~~ Ignore me, I hadn't pulled since yesterday

fafhrd91 (approved):

fafhrd91:
Thanks

PR #992, opened 2019-07-19
add note about Query decoding

jesskfullwood:
I got caught out trying to use `Query<i64>` and having it fail at runtime, so I thought this note might help others.


fafhrd91 (approved):

fafhrd91:
Thanks!

PR #1004, opened 2019-07-22
actix-files: "Specified path is not a directory" error should include the path

erikdesjardins:
This is useful if you have more than one `Files` instance, so you can tell which path doesn't exist.


fafhrd91 (approved):

fafhrd91:
Thanks!

PR #1014, opened 2019-07-28
Allow HeaderMap to be cloned

svenstaro:
This was possible in the previous actix-web and I quite liked that. I don't see any reason this shouldn't be cloneable.


fafhrd91 (approved):

fafhrd91:
thanks!

PR #1015, opened 2019-07-28
Fix two dyn warnings

svenstaro:
Quite trivial as it only gets rid of the warnings by doing what the compiler suggests.


fafhrd91 (approved):

fafhrd91:
thanks

PR #1040, opened 2019-08-15
chore(readme): fix copy paste error

robertohuertasm:
It seems that the header and some badges were copy-pasted from the `actix-identity` crate. I've just changed the text and the badges links.


JohnTitor (approved):

JohnTitor:
Thanks!

Issue #1038, opened 2019-08-15
actix-multipart: Incomplete Error if body does not end with CRLF


jefshe:
Im currently trying to migrate an application server (previously using apache) which parses multipart form data to actix-web.
When sending multipart requests from my client to the server however I get a MultipartError::Incomplete

The client's request looks something like this:
```
Content-Type: multipart/form-data; boundary=---------------------------8d2c716897da0e3                
Content-Length: 558                                                                                   
Accept-Encoding: gzip, deflate                                                                        
Connection: close                                                                                     
                                                                                                                                                                           
-----------------------------8d2c716897da0e3                                                          
Content-Disposition: form-data; name="width"                                                          
                                                                                                      
8                                                                                                     
-----------------------------8d2c716897da0e3                                                          
Content-Disposition: form-data; name="height"                                                         
                                                                                                      
7.097222                                                                                              
-----------------------------8d2c716897da0e3--
```

Ive traced the problem down to the following line: https://github.com/actix/actix-web/blob/979c4d44f4626e4346714753106d2e94c9b845d1/actix-multipart/src/server.rs#L779

Because the last line of the request does not end with CRLF actix throws the error. I propose that the framework be more flexible about this.

Since this was previously being parsed on apache, it would be nice for actix-multipart to handle it as well rather than needing to hack around the problem

fafhrd91:
fixed in master

PR #1042, opened 2019-08-16
actix-multipart: Correctly parse multipart body which does not end in CRLF

jefshe:
relates to #1038 
my reading of https://tools.ietf.org/html/rfc2046 AND https://tools.ietf.org/html/rfc7578 suggest to me that the server implementation should handle the case where no CRLF is present at the end of the multipart form. Here are the supporting paragraphs:

 ```
4.1.  "Boundary" Parameter of multipart/form-data

   As with other multipart types, the parts are delimited with a
   boundary delimiter, constructed using CRLF, "--", and the value of
   the "boundary" parameter.  The boundary is supplied as a "boundary"
   parameter to the multipart/form-data type.  As noted in Section 5.1
   of [RFC2046], the boundary delimiter MUST NOT appear inside any of
   the encapsulated parts, and it is often necessary to enclose the
   "boundary" parameter values in quotes in the Content-Type header
   field.
```
A "boundary" is considered to start with CRLF

```
     delimiter := CRLF dash-boundary

     close-delimiter := delimiter "--"

     preamble := discard-text

     epilogue := discard-text
```

```

     multipart-body := [preamble CRLF]
                       dash-boundary transport-padding CRLF
                       body-part *encapsulation
                       close-delimiter transport-padding
                       [CRLF epilogue]
```
The ending CRLF is considered part of the epilogue which can be ignored

### DIFF
--- a/actix-multipart/src/server.rs
+++ b/actix-multipart/src/server.rs
@@ -781,7 +781,7 @@ impl PayloadBuffer {
     /// Read bytes until new line delimiter or eof
     pub fn readline_or_eof(&mut self) -> Result<Option<Bytes>, MultipartError> {
         match self.readline() {
-            Err(MultipartError::Incomplete) => Ok(Some(self.buf.take().freeze())),
+            Err(MultipartError::Incomplete) if self.eof => Ok(Some(self.buf.take().freeze())),
             line => line
         }
     }
--- a/actix-multipart/src/server.rs
+++ b/actix-multipart/src/server.rs
@@ -167,7 +167,7 @@ impl InnerMultipart {
         boundary: &str,
     ) -> Result<Option<bool>, MultipartError> {
         // TODO: need to read epilogue
-        match payload.readline()? {
+        match payload.readline_or_eof()? {
             None => {
                 if payload.eof {
                     Ok(Some(true))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by fafhrd91 — CHANGES_REQUESTED** [review]
Could you add simple unit test and update changes.md

**Review by neoeinstein — CHANGES_REQUESTED** [review]
Could you add another test that doesn't include the final `CRLF` in the `tests` module?

File: actix-multipart/src/server.rs [hunk 1]
```diff
@@ -768,7 +767,7 @@ impl PayloadBuffer {
             .map(|idx| self.buf.split_to(idx + line.len()).freeze());
 
         if res.is_none() && self.eof {
-            Err(MultipartError::Incomplete)
+            Ok(self.buf.take().freeze())
```
JohnTitor: Mismatched types error here, probably `Ok(Some(self.buf.take().freeze()))`.

fafhrd91: i dont think this is right code. this could lead to partial reads reported as valid

jefshe: Yep good call, ive moved this out into a new function `readline_or_eof` which is called by `read_boundary`, a partial read should then result in a `BoundaryError`

jefshe:
> 
> 
> Could you add another test that doesn't include the final `CRLF` in the `tests` module?

> 
> 
> Could you add simple unit test and update changes.md

Added a unit test now

leizzer (approved):

fafhrd91:
thanks!

Issue #978, opened 2019-07-15
Register `Data` without wrapping


0ndorio:
It seems like the `App` doesn't provide a way to register a `Data` element without wrapping it's content with an additional layer of `Arc`. This is true for `App::data()`, `App::register_data()` and `App::data_factory()`. The documentation refers to `web::Data::new` for this use case but this is just a plain reexport and will also wrap.

@fafhrd91 Mentioned in gitter that it should be possible to come up with a custom `Data` implementation instead but this also doesn't work as the only way to setup `Data` is via `Data::new` and the `DataFactory` trait is only public inside the crate itself.

--

So the open questions are: 

 - Is one of both ways expected to work and if so, which one is the one that need a fix? (I would be happy to provide a merge request fixing it.)
 - Or am I expected to change my internal data type to avoid the second layer?

jesskfullwood:
I have hit this problem before.Suppose one has a type that does it's own synchronization
```
struct Counter(Arc<AtomicU32>)
```
Ideally we use this with a factory function like
```
let counter = Counter::new();
App::new().data_factory(move || counter.clone());
```
There is no way to do this without adding a second `Arc` - this is clear from the definition of `Data`:
```
pub struct Data<T>(Arc<T>);
```
We would have to make a breaking change to the definition of `Data` to make this work.

jesskfullwood:
I think the way we solve this is with a new extractor, something like
```
#[derive(Clone)]
struct SyncData<T: Clone + Send + Sync>(T)
```
With the explanation that `T` will be cloned each time the extractor is used.

fafhrd91:
Added `App::app_data()` and `HttpRequest::app_data()` methods.

`Data<T>` extractor could be implemented with this methods

PR #1052, opened 2019-08-23
Data::to_arc

Ploppz:
I find this function to be convenient, because sometimes you would like to share the Arc<T> with for example a function that takes Arc<T>.
Otherwise, all my functions who needed some shared state from the actix web server needed to depend on the Data type.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 * Form immplements Responder, returning a `application/x-www-form-urlencoded` response
 
+* Add `into_inner` to `Data`
+
 ### Changed
 
 * `Query` payload made `pub`. Allows user to pattern-match the payload.
--- a/src/data.rs
+++ b/src/data.rs
@@ -77,6 +77,11 @@ impl<T> Data<T> {
     pub fn get_ref(&self) -> &T {
         self.0.as_ref()
     }
+
+    /// Convert to the internal Arc<T>
+    pub fn into_inner(self) -> Arc<T> {
+        self.0
+    }
 }
 
 impl<T> Deref for Data<T> {


neoeinstein:
Seems reasonable, but it does now expose the internals of the `Data` type, and more or less locks us into using an `Arc<T>` internally. I don't see that has a big problem right now.

`Path`, `Query`, and `Json` use `into_inner()` to get access to the internal representation `T`. We could use the same naming, but in this case, we would return `Arc<T>`. That would make it easy to remember: Unwrap `Path`, `Query`, `Json`, and `Data`, by using `into_inner()`, but in the case of `Data`,  you get a shared pointer when you unwrap.

Ploppz:
I considered it, but had a bit of doubt because the "inner" value we really care about is T. And we also have `Deref<T>` to reflect this fact.
I am not against `into_inner()` though. You think that would be better?
edit: on second thought maybe into_inner makes more sense since the Arc is the hidden "inner" and not otherwise accessible.

jesskfullwood:
I agree with @neoeinstein - this is a useful feature but am a bit reluctant to commit to the internal representation of `Data` until we have resolved #978 one way or another.

fafhrd91 (approved):

fafhrd91:
@Ploppz please add entry to changes.md

Ploppz:
Just one more consideration:
On second thought maybe it makes more sense with `fn inner(&self) -> &Arc<T>` because then one can `data.inner().clone()` rather than cloning the Data before getting the Arc if one wants to reuse Data.
Maybe both `into_inner` and `inner`?
(I already did, I can revert if unwanted)

fafhrd91 (approved):

jesskfullwood:
> On second thought maybe it makes more sense with fn inner(&self) -> &Arc<T> because then one can data.inner().clone() rather than cloning the Data before getting the Arc if one wants to reuse Data.

Personally I think it is confusing to have both methods. I cannot think of any circumstances in which you might want to reuse the `Data` object. I would prefer just the `into_inner` method like the other extractors.

Ploppz:
You are right; Data is not really needed inside the function. I changed it (rebased and force-pushed)

Ploppz:
I hate to be oscillating but on third thought, isn't it kinda more ergonomic to not have to say `let mydata = mydata.to_inner();` at the start of the function but rather wherever you need it: `mydata.inner().clone()`? Or at least the option to? Just a thought, I don't really mind, I'll let whoever decide.

jesskfullwood:
Well, the alternative is simply `data.clone().into_inner()`. I don't see any difference.

Ploppz:
oh, you are right. For some reason I though it could lead to more Arc cloning than necessary.

JohnTitor (approved):
I think the current form is good.

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 * Form immplements Responder, returning a `application/x-www-form-urlencoded` response
 
+* Add `inner` and `into_inner` to `Data`
```
jesskfullwood: Out of date

Ploppz: Ups. Fixed.

Issue #1062, opened 2019-08-28
Request Extensions leak


pjenvey:
Request extension values can leak in between requests. This can be especially bad when e.g. storing pooled database connections in extensions, this leak can quickly exhaust such a pool.

actix-http's message module cache clears certain values when reclaiming req/resp "head" types, but actix-web's request module cache does not.

fafhrd91:
1.0.7 is released

fafhrd91:
thanks for reporting!

PR #1063, opened 2019-08-28
clear extensions before reclaiming HttpRequests in their pool

pjenvey:
Issue #1062


pjenvey:
We may want to clear more than extensions like actix-http's Head::clear (which isn't public) does

fafhrd91 (approved):

PR #1064, opened 2019-08-28
Added FrozenClientRequest for easier retrying HTTP calls

dpypin:
Added `FrozenClientRequest` that represents immutable `ClientRequest`.

In scenarios where you need to retry an HTTP call, using current API, you must create a new `ClientRequest`, because it's `send_*` group of methods consume the request. `FrozenClientRequest` has non-consuming `send_*` methods, so you can retry failed call using the original frozen request.
Also `send_*` methods of a `FrozenClientRequest` have an additional parameter - `extra_headers: Option<HeaderMap>` in case you need to add or override some headers.

A real life scenario where I need these changes is calling Cosmos DB REST API. Cosmos DB has two authentication keys - primary and secondary. To support easier rollover of secrets my Cosmos DB client accepts both keys and if an HTTP call using a primary key returned 401 HTTP code (Unauthorized), then my client retries the call using a secondary key.


fafhrd91 (approved):
good work!

leizzer (approved):

JohnTitor (approved):

PR #1070, opened 2019-09-04
Ensure that awc::ws::WebsocketsRequest sets the Host header

eugene-bulkin:
Before connecting, this sets the `Host` header to the `uri` host value. Not doing this will cause 400 Bad Request errors to be returned from a WebSocket server.

### DIFF
--- a/awc/src/ws.rs
+++ b/awc/src/ws.rs
@@ -233,7 +233,9 @@ impl WebsocketsRequest {
             return Either::A(err(InvalidUrl::UnknownScheme.into()));
         }
 
-        self.head.headers.insert(header::HOST, HeaderValue::from_str(uri.host().unwrap()).unwrap());
+        if !self.head.headers.contains_key(header::HOST) {
+            self.head.headers.insert(header::HOST, HeaderValue::from_str(uri.host().unwrap()).unwrap());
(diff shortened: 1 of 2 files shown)


awc/src/ws.rs
```diff
@@ -233,6 +233,8 @@ impl WebsocketsRequest {
             return Either::A(err(InvalidUrl::UnknownScheme.into()));
         }
 
+        self.head.headers.insert(header::HOST, HeaderValue::from_str(uri.host().unwrap()).unwrap());
```
fafhrd91: you need to check if header is set already

eugene-bulkin: Updated.

JohnTitor (approved):

fafhrd91:
@eugene-bulkin could you add entry to changes.md

eugene-bulkin:
> @eugene-bulkin could you add entry to changes.md

Should I put it in a new `0.2.5` section?

fafhrd91:
yes, create 0.2.5 section

eugene-bulkin:
> yes, create 0.2.5 section

Done!

fafhrd91:
thanks!

PR #1077, opened 2019-09-06
Change Ping and Pong payloads to Bytes

### LINKED ISSUE #1049: Websockets: Ping and Pong should have binary data
The "Ping" and "Pong" messages defined in  actix-web/actix-http/src/ws/codec.rs take a String as parameter. This means that only valid utf8 can be used in these messages. But I think the [websocket spec](https://tools.ietf.org/html/rfc6455) does not actually require this.

The main problem with this is that a client that does not limit itself to utf8 ping messages will not receive correct pong messages because the data was mangled by from_utf8_lossy. This happens for example when using [this](https://websockets.readthedocs.io) python websocket library.

pathim:
Fixes #1049

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Changes
 
+## Unreleased
+
+### Fixed
+
+* Websockets: Ping and Pong should have binary data (#1049)
+
+### Changed
+
+* Make `ping` and `pong` methods of `WebsocketContext` take an `Into<Bytes>` argument
+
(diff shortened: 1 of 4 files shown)


JohnTitor:
Could you also update CHANGES.md?

fafhrd91:
Let’s wait with this Pr. I am not sure about breaking change atm

pathim:
PR #1081 contains the same fix but also fixes other things and puts the breaking change behind a feature flag. Therefore this PR is no longer useful

PR #1080, opened 2019-09-09
fix h2 not using error response

ronaldslc:
There is a discrepancy in error response thrown from ServiceResponse errors between h1 and h2 connections. 

h2 requests will always end up with a code 500 error with no body.

The behaviour in h1 should be the preferred behaviour instead since `actx_http::ResponseError` maybe returned by a Service with specific client error responses.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@
 
 *  Add `middleware::Conditon` that conditionally enables another middleware
 
+### Fixed
+
+* h2 will use error response #1080
 
 ## [1.0.7] - 2019-08-29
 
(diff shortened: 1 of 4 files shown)


fafhrd91 (approved):

ronaldslc:
done, is that ok?

fafhrd91:
lgtm, thanks!

fafhrd91:
you have to fix tests, and add one for your change

ronaldslc:
You want a test to compare equal h1 and h2 service error? That would just be duplication of test_h1_service_error and test_h2_service_error with an extra assert right?

I'm not familiar with the test helpers, is it possible to get a single `TestServerRuntime` and specify the client protocol?

fafhrd91:
Thanks! Test that you fixed is enough

PR #1081, opened 2019-09-10
autobahn tests compliance

Licenser:
[Running](https://github.com/Licenser/actix-autobahn) the [autobahn tests](https://github.com/crossbario/autobahn-testsuite) for websocket standard compliance I noticed that the Ping/Pong message should be binary not text. I updated this.

* Fix ping/pong to be binary not text.
* RSV, if set, on the default Codec (w/o extension) should cause an error.
* Add support for fragmented messages
* Improve WS Close handling to make it more compliant


fafhrd91:
dont use bytes for Ping/Pong, it is breaking change.

Licenser:
Hi @fafhrd91,
I know it's a breaking change, however the current behaviour [violates the RFC](https://tools.ietf.org/html/rfc6455#page-37) perhaps we could include the change in a version change?

fafhrd91:
just skip it. otherwise pr would need to be on hold for some time

Licenser:
I put it behind a feature and made it default that allows to slowly migrate to correct handling. And allows correct behaviour for those that need it.

* keep current behaviour
* (next version) deprecate behaviour (make it non default)
* (next next version) remove incorrect behaviour

Licenser:
Sorry @pathim I didn't saw you already hat a PR open, I started looking on autobahn for actix this morning and kind of went through it on autopilote.

svenstaro:
Needs rebase.

Licenser:
Done

fafhrd91:
could you remove  `actix-web-actors` changes and remove `invalid-ping-payload` feature

actix-http/src/ws/codec.rs
```diff
@@ -68,6 +86,27 @@ impl Codec {
         self.server = false;
         self
     }
+
+    fn combine_payload(&mut self, payload: Option<BytesMut>) -> Option<BytesMut> {
```
fafhrd91, line 90: i think it would be better to bubble up frames with continuation to application layer, application can decide if it wants to accumulate payload or do something with partial payload

Licenser:
> could you remove `actix-web-actors` changes and remove `invalid-ping-payload` feature

Removing those changes would defeat the purpose of the pull request as it would return to non standard conform behaviour.

fafhrd91:
This pr is not compatible with master

PR #1085, opened 2019-09-12
Feature uds: Add listen_uds to ServerBuilder

karlri:
DEPENDS on forked actix-net. Please amend before submission.
Allows directly passing an Unix Listener instead of a path. Useful
for when running as a daemon under systemd with the systemd crate.


JohnTitor:
It depends on an external crate for now so I modified the title to avoid merging.

karlri:
Awaiting merge of approved PR in actix-net, and new actix-net version: https://github.com/actix/actix-net/pull/43
Will amend as soon as that happens. :)

JohnTitor:
actix/actix-net#43 has been merged. we need to update this.

karlri:
Do I refer to the actix-net official git in Cargo.toml then, or wait for them to update actix-server Cargo.toml to 0.6.1 or similar? I don't know actix tagging workflows.

JohnTitor:
Ah yes, we should probably wait for new release.

@fafhrd91 any thoughts?

fafhrd91:
actix-server is released

karlri:
Will force-push update of Cargo.toml, rebase and add some change log entry

karlri:
I updated my fork so now we could hopefully remove the "[DO NOT MERGE] ". Error handling is improved and a change log entry has been added. I also formatted the code with cargo fmt.

Note that there is some duplicated code between listen_uds and bind_uds. In theory, bind-uds could convert path to UnixListener and just call self.listen_uds, but we have to deal with existing paths etc. The code to deal with existing paths currently resides in actix-server:

```
        if let Err(e) = std::fs::remove_file(addr.as_ref()) {
            // NotFound is expected and not an issue. Anything else is.
            if e.kind() != std::io::ErrorKind::NotFound {
                return Err(e);
            }
        }

        let lst = UnixListener::bind(addr)?;
```

To me, it seemed that it is better to have duplicated code in the same crate in functions next to each other rather than across crates?

fafhrd91 (approved):

PR #1089, opened 2019-09-13
Make UrlencodedError::Overflow more informative

vandenoever:
I got bitten by the fact that the error reported a limit of 256k while the actual default limit was far smaller.

With this patch, the actual limit is reported as well as the size of the oversized payload.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,8 +32,12 @@ pub enum UrlencodedError {
     #[display(fmt = "Can not decode chunked transfer encoding")]
     Chunked,
     /// Payload size is bigger than allowed. (default: 256kB)
-    #[display(fmt = "Urlencoded payload size is bigger than allowed (default: 256kB)")]
-    Overflow,
+    #[display(
+        fmt = "Urlencoded payload size is bigger ({} bytes) than allowed (default: {} bytes)",
+        size,
+        limit
+    )]
+    Overflow { size: usize, limit: usize },
     /// Payload size is now known
     #[display(fmt = "Payload size is now known")]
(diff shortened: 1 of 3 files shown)


JohnTitor:
Did you test this patch on your local? It seems we need to work more.

vandenoever:
I neglected to update the tests. I've done so now.

fafhrd91 (approved):
would be good to update all Overflow errors. also could you add entry to changes.md

vandenoever:
@fafhrd91 I've added to CHANGES.md. All use of UrlencodedError::Overflow is has been adapted so that `cargo build --all` and `cargo test` work.

fafhrd91 (approved):

src/error.rs
```diff
@@ -32,8 +32,12 @@ pub enum UrlencodedError {
...
     Chunked,
     /// Payload size is bigger than allowed. (default: 256kB)
-    #[display(fmt = "Urlencoded payload size is bigger than allowed (default: 256kB)")]
-    Overflow,
+    #[display(
+        fmt = "Urlencoded payload size is bigger ({} bytes) than allowed (default: {} bytes)",
+        size,
+        limit
+    )]
+    Overflow { limit: usize, size: usize },
```
JohnTitor: IMO: In other places its ordering is `size, limit` but `limit, size` here. I think the order should be unified.

vandenoever:
@JohnTitor: I've swapped limit and size in the definition.

JohnTitor (approved):
Thanks!

PR #1090, opened 2019-09-14
Remove `mem::uninitialized()`

JohnTitor:
`mem::uninitialized()` will be deprecated in 1.39 ([docs.rs](https://doc.rust-lang.org/nightly/std/mem/fn.uninitialized.html)). This replaces it with `MaybeUninit`.

### DIFF
--- a/actix-http/src/client/connector.rs
+++ b/actix-http/src/client/connector.rs
@@ -212,7 +212,7 @@ where
     pub fn finish(
         self,
     ) -> impl Service<Request = Connect, Response = impl Connection, Error = ConnectError>
-                 + Clone {
+           + Clone {
         #[cfg(not(any(feature = "ssl", feature = "rust-tls")))]
         {
             let connector = TimeoutService::new(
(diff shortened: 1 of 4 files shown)


fafhrd91 (approved):

bluetech:
FYI, I think these changes use `MaybeUninit` incorrectly. Quoting from the [`assume_uninit()` docstring](https://doc.rust-lang.org/stable/core/mem/union.MaybeUninit.html#safety):

> It is up to the caller to guarantee that the `MaybeUninit<T>` really is in an initialized state. Calling this when the content is not yet fully initialized causes undefined behavior.

Thus, the following code and the others cause undefined behavior:

```rust
let mut buf: [u8; 41] = unsafe { mem::MaybeUninit::uninit().assume_init() };
```

JohnTitor:
@bluetech Hmm, but `mem::uninitialized` also causes UB in this situation? I can't come up with the solution for this at now so pull request is welcomed.

fafhrd91:
@bluetech could you show UB in this code?

PR #1093, opened 2019-09-16
Add ability to use `Infallible` as `HttpResponse` error type

nWacky:
Sometimes there is 0 chance of having errors when returning user defined type as `HttpResponse`. 
Consider the following code:
```rust
pub struct MyResponse(String);

impl Responder for MyResponse {
    type Error = Infallible;
    type Future = Box<dyn futures::Future<Item = HttpResponse, Error = Infallible>>;

    fn respond_to(self, _: &HttpRequest) -> Self::Future {
        Box::new(futures::future::ok(
             HttpResponse::build(StatusCode::OK)
                .body(self.0)
        ))
    }
}

// `MyResponse` will be returned from `actix_web` handler directly somewhere later

fn main() {
    println!("It compiled!");
}
```

It does not compile with this error:

```console
error[E0277]: the trait bound `std::convert::Infallible: actix_http::error::ResponseError` is not satisfied
   --> src/main.rs:16:18
    |
 16 | impl Responder for MyResponse {
    |      ^^^^^^^^^ the trait `actix_http::error::ResponseError` is not implemented for `std::convert::Infallible`
    |
    = note: required because of the requirements on the impl of `std::convert::From<std::convert::Infallible>` for `actix_http::error::Error`
    = note: required because of the requirements on the impl of `std::convert::Into<actix_http::error::Error>` for `std::convert::Infallible`
```
 
In this case it would be nice to be able to use [`Infallible`](https://doc.rust-lang.org/beta/std/convert/enum.Infallible.html) (will become primitive type `never` in future) as error type to indicate that the result can never be `Err`.

P.S. Also please don't mind a lot of commits generated by me, I was trying to check this solution on existing codebase which didn't recompile crate

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,6 @@
 
 * Allow to re-construct `ServiceRequest` from `HttpRequest` and `Payload`
 
-* Allow to use `std::convert::Infallible` as `actix_http::error::Error`
-
 
 ## [1.0.7] - 2019-08-29
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 * Allow to re-construct `ServiceRequest` from `HttpRequest` and `Payload`
 
+* Allow to use `std::convert::Infallible` as `actix_http::error::Error`
+
 
 ## [1.0.7] - 2019-08-29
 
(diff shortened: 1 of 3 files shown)


fafhrd91 (approved):

nWacky:
@fafhrd91, 

Yes, updated changes.md

fafhrd91 (approved):

fafhrd91:
you should update actix-http/changes.md

JohnTitor (approved):

fafhrd91:
Thanks!

PR #1094, opened 2019-09-17
Implement `register_data` method on `Resource` and `Scope`.

jimblandy:
Without these methods, it seems impossible to avoid introducing double-`Arc` structures when application data needs to be overridden on individual `Scope`s or `Resource`s.


src/scope.rs
```diff
@@ -1066,12 +1075,14 @@ mod tests {
 
     #[test]
     fn test_override_data() {
```
fafhrd91: could you keep this test and add another for register_data

fafhrd91:

fafhrd91 (approved):

fafhrd91:
thanks!

PR #1097, opened 2019-09-19
Fix peer addr string

### LINKED ISSUE #1095: ConnectionInfo.remote() returns remote socket address not IP
The documentation in https://docs.rs/actix-web/1.0.7/actix_web/dev/struct.ConnectionInfo.html#method.remote says that it returns the IP of the peer. However, if it's obtained directly from the peer socket address, it returns also the peer socket address (e.g. `127.0.0.1:51286` and not `127.0.0.1`). This may cause real problems if this returned str value is used as is for rate limiting for example.

Chronophagus:
Fixes #1095

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -118,7 +118,7 @@ impl ConnectionInfo {
             }
             if remote.is_none() {
                 // get peeraddr from socketaddr
-                peer = req.peer_addr.map(|addr| format!("{}", addr));
+                peer = req.peer_addr.map(|addr| format!("{}", addr.ip()));
             }
         }
 
--- a/src/info.rs
+++ b/src/info.rs
@@ -151,6 +151,7 @@ impl ConnectionInfo {
     /// - Host
     /// - Uri
     /// - Server hostname
+    #[inline]
     pub fn host(&self) -> &str {
         &self.host
     }


fafhrd91:
I don’t have strong opinion, but this change drops some information.

@actix/contributors ?

cetra3:
Is there a way to get the connecting port out if this change is made? I don't think losing that information is worth it as you can just split on the colon char if you want to filter to the ip.

Chronophagus:
Well, you won't get connecting port if remote addr was resolved via headers anyway. But now I realize that someone may already depend on this behavior(that there may be a port in the string), and make use of it somehow, so this can be considered as breaking change. Anyway the docs should mention that method returns "[IP]" when remote resolved via headers, or "[IP]:[PORT]" when remote resolved via socket, because now it's unclear.

robjtede:
Yeah I'd be worried about changing this behaviour; I've, personally, written code that parses this field and discovered the inconsistency with port numbers. Any reasonable code will try to parse it both ways but this is technically breaking in cases where port numbers are expected to always exist and be parsed out. I think a better approach could involve a new method on the impl ConnectionInfo that's guaranteed to only return the address portion.

svenstaro:
Yeah, let's not change this. At least not in 1.0.x.

fafhrd91:
closing after discussion. we should update docs instead

PR #1110, opened 2019-09-30
Add From<Payload> for crate::dev::Payload

popzxc:
`actix_http::Payload` is visible through `dev` module, but users aren't enable to obtain it, because `types::Payload` wrapper doesn't have a conversion method.

However, `FromRequest::from_request` (which is also public) requires `dev::Payload`, so as for now users are unable to call this method.

So, an example to explain it better:

Some framework provides an interface to build API, and uses actix as a backend. Types of the processed data can differ, so framework has to parse request manually for each endpoint. For some endpoints, it's just `Query` parsed from query string. For others it's Json. 

Now, even given both `HttpRequest` and `Payload`, framework is unable to construct a `Json` object, because it can't access the `dev::Payload`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## [1.0.9] - 2019-xx-xx
+
+### Added
+
+* Add `Payload::into_inner` method and make stored `def::Payload` public. (#1110)
+
 ## [1.0.8] - 2019-09-25
 
 ### Added
--- a/src/types/payload.rs
+++ b/src/types/payload.rs
@@ -45,6 +45,12 @@ use crate::request::HttpRequest;
 /// ```
 pub struct Payload(crate::dev::Payload);
 
+impl From<Payload> for crate::dev::Payload {
+    fn from(payload: Payload) -> Self {
+        payload.0
+    }
+}
+
 impl Stream for Payload {
     type Item = Bytes;
     type Error = PayloadError;
--- a/src/types/payload.rs
+++ b/src/types/payload.rs
@@ -43,11 +43,12 @@ use crate::request::HttpRequest;
 ///     );
 /// }
 /// ```
-pub struct Payload(crate::dev::Payload);
+pub struct Payload(pub crate::dev::Payload);
 
-impl From<Payload> for crate::dev::Payload {
-    fn from(payload: Payload) -> Self {
-        payload.0
+impl Payload {
+    /// Deconstruct to a inner value
+    pub fn into_inner(self) -> crate::dev::Payload {
+        self.0
     }
 }
 


fafhrd91:
You definitely can construct json manually, https://github.com/actix/examples/blob/master/json/src/main.rs#L35

Do you have other example? I  fine with this pr, but description does not show real need for payload access

fafhrd91:
Let’s use same approach as for Query. Use pub for payload and add into_inner

popzxc:
Well, yes, but transforming `Payload` into `BytesMut` and then calling `serde_json` manually seems too low-level (and not really `actix`-based) approach to me, when you can do just:

```rust
let index = move |request: HttpRequest, mut payload: Payload| -> impl Future<Item = HttpResponse, Error = Error> {
    let future = Json::from_request(&request, &mut payload.into())
        .map(Json::into_inner)
        .and_then(whatever_you_want_to_do)
```

popzxc:
>Let’s use same approach as for Query. Use pub for payload and add into_inner

Well, using the same approach as for Query will require updating the trait signature, or implementing an additional method for every structure that implements `FromRequest` and uses a payload (e.g. `Form`).

In the first case the change eventually will become breaking, and the second approach is less flexible, as I think.

Or am I wrong?

fafhrd91:
I don’t understand problem with Json. You use Json if you have to deserialize incoming payload or use Payload if you have to access stream of incoming bytes.

popzxc:
OK, let's be more concrete then.

We have an actix backend in exonum. And sources are already pretty complicated, because we want to keep things as generic as possible:

https://github.com/exonum/exonum/blob/master/exonum/src/api/backends/actix.rs#L166

Now we're going to migrate from 0.7 to 1.00.

With this PR merge we'll be able to just write:

```rust
let index = move |request: HttpRequest, payload: mut Payload| -> FutureResponse {
    let handler = handler.clone();
    let context = request.state().clone();
    let future = Json::from_request(&request, &mut payload.into())
        .map(Json::into_inner)
        .from_err()
        .and_then(move |query: Q| {
            handler(&context, query)
                .map(|value| HttpResponse::Ok().json(value))
                .map_err(From::from)
        })
        .into_future()
};
```

However, without it this function will look like:

```rust
let index = move |request: HttpRequest, payload: mut Payload| -> FutureResponse {
    let handler = handler.clone();
    let context = request.state().clone();
    let future = payload
        .from_err()
        .fold(BytesMut::new(), move |mut body, chunk| {
            if (body.len() + chunk.len()) > MAX_SIZE {
                Err(error::ErrorBadRequest("overflow"))
            } else {
                body.extend_from_slice(&chunk);
                Ok(body)
            }
        })
        .and_then(move |query: Q| {
            handler(&context, query)
                .map(|value| HttpResponse::Ok().json(value))
                .map_err(From::from)
        })
        .into_future()
};
```

Which introduces a low-level bytes collecting and looks much more complicated overall.

It's better to keep things simple when it's possible, isn't it?

And as an additional argument: what's the point of having public type (`dev::Payload`) without possibility to construct it?

fafhrd91:
would this work?

```rust
let index = move |request: HttpRequest, query: Json<Q>| -> FutureResponse {
        handler(&context, query)
            .map(|value| HttpResponse::Ok().json(value))
            .map_err(From::from)
}
```

popzxc:
Unfortunately no, because we collect different types of endpoint handlers as a closures, so having a non-generic signature is a requirement. Otherwise we won't be able to collect a bunch of handlers into a  `Vec<Box<dyn Fn(InputType) -> OutputType + ...>>`. In our case, `InputType` is `(HttpRequest, Payload)`.

fafhrd91:
ok, i see.

let's change to `Payload(pub ..)` and add `fn into_inner(self) -> Payload` method same as for query

fafhrd91:

fafhrd91 (approved):

fafhrd91:
thanks

PR #1114, opened 2019-10-01
Expose ContentDisposition in actix-multipart to fix broken doc link

### LINKED ISSUE #1109: Make actix_http::...::ContentDisposition public
In actix-multipart there is a method to get the content-disposition header of the request, `actix_multipart::Field.content_disposition`.

However, this returns a private type (I didn't think this was allowed).

The full path is `actix_http::header::common::content_disopsition::ContentDisposition`

Motivation: To be able to identify, and parse, regular form fields differently from files.

zacps:
The method `Field.content_disposition` returns `ContentDisposition` which is private, but publically exported by `actix-web`.

This PR re-exports `ContentDisposition` and its fields so that they are properly documented in `actix-multipart`.

Fixes #1109

### DIFF
--- a/actix-http/src/header/mod.rs
+++ b/actix-http/src/header/mod.rs
@@ -16,7 +16,6 @@ use crate::httpmessage::HttpMessage;
 mod common;
 pub(crate) mod map;
 mod shared;
-#[doc(hidden)]
 pub use self::common::*;
 #[doc(hidden)]
 pub use self::shared::*;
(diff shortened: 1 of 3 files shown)


zacps:
This could also be solved by exporting `ContentDisposition` in `actix-http`.

JohnTitor:
The doc for `actix-multipart` points to `actix-http`, so this should be fixed on the `actix-http`side, I think.

zacps:
Yeah I agree. It turns out they are actually exported in `actix-http` but have the `doc(hidden)` attribute. Removing that fixes the broken links.

There's also `actix-http::header::shared` which has `doc(hidden)`. I left that because it doesn't affect the specific problem I encountered, but it might be worth evaluating if there's any reason to hide these docs.

fafhrd91 (approved):

PR #1115, opened 2019-10-02
[actix-files] Allow user defined guards for NamedFile (actix#1113)

### LINKED ISSUE #1113: NamedFile can only be returned from a GET request
I've got a route that returns a NamedFile, and it works fine with GET requests.  But now I need to update some database state before I return the file, so I need a POST request.  But when I change my route to POST, I get a 405 error from actix_web with a body "This resource only supports GET and HEAD."  
I've traced the execution through my handler function up until I return the opened NamedFile to actix, so I know it's actually getting into my code and we're successfully opening the file handle.
Is it a requirement to only use NamedFile with GET?

naim94a:
I though this could be a good first issue, please tell me if you would have taken another approach implementing this. 
Should fix #1113.

### DIFF
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -15,7 +15,7 @@ use actix_http::body::SizedStream;
 use actix_web::http::header::{
     self, ContentDisposition, DispositionParam, DispositionType,
 };
-use actix_web::http::{ContentEncoding, Method, StatusCode};
+use actix_web::http::{ContentEncoding, StatusCode};
 use actix_web::middleware::BodyEncoding;
 use actix_web::{Error, HttpMessage, HttpRequest, HttpResponse, Responder};
 
@@ -322,16 +322,6 @@ impl Responder for NamedFile {
                 counter: 0,
             };
             return Ok(resp.streaming(reader));
-        }
-
-        match *req.method() {
-            Method::HEAD | Method::GET => (),
-            _ => {
-                return Ok(HttpResponse::MethodNotAllowed()
-                    .header(header::CONTENT_TYPE, "text/plain")
-                    .header(header::ALLOW, "GET, HEAD")
-                    .body("This resource only supports GET and HEAD."));
-            }
         }
 
         let etag = if self.flags.contains(Flags::ETAG) {
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -6,6 +6,8 @@
 
 * Bump up `percent-encoding` crate version to 2.1
 
+* Allow user defined request guards for `Files` #1113
+
 ## [0.1.4] - 2019-07-20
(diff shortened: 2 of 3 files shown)


actix-files/src/named.rs
```diff
@@ -37,7 +38,6 @@ impl Default for Flags {
 }
 
 /// A file with an associated name.
-#[derive(Debug)]
```
JohnTitor: I don't think it's a good idea to remove the Debug implementation.

actix-files/src/named.rs
```diff
@@ -216,6 +218,14 @@ impl NamedFile {
         self
     }
 
+    ///Specifies custom guards.
+    /// 
+    ///Default guards allow GET, HEAD.
```
JohnTitor:
```suggestion
    /// Default guards allow GET, HEAD. 
```

actix-files/src/named.rs
```diff
@@ -216,6 +218,14 @@ impl NamedFile {
         self
     }
 
+    ///Specifies custom guards.
```
JohnTitor:
```suggestion
    /// Specifies custom guards.
```

actix-files/src/named.rs
```diff
@@ -324,14 +334,20 @@ impl Responder for NamedFile {
...
-            Method::HEAD | Method::GET => (),
-            _ => {
-                return Ok(HttpResponse::MethodNotAllowed()
-                    .header(header::CONTENT_TYPE, "text/plain")
-                    .header(header::ALLOW, "GET, HEAD")
-                    .body("This resource only supports GET and HEAD."));
+        let are_guards_satisfied = if let Some(ref guard) = self.guard {
+            (**guard).check(req.head())
+        }
+        else {
```
JohnTitor: These lines(L339 and L340) can be one line.

JohnTitor (commented):
I'm not sure the current implementation is good, I'd like to hear other's opinions. Also, we should add the test to make sure that works well.

naim94a:
> I'm not sure the current implementation is good, I'd like to hear other's opinions. Also, we should add the test to make sure that works well.

The other approach I thought about was to make NamedFile generic: 
```rust
struct NamedFile<G: Guard>{ ... }
```

fafhrd91:
i think method check should be moved to `Files` type. `NamedFile` should not assume or imply any guard checks.

naim94a:
@fafhrd91 Do you think its a good idea to remove the existing checks from `NamedFile` (checks if the method is GET/HEAD)? Or should the existing checks be bypassed when using `Files`?

fafhrd91:
I think we should move guard like checks to Files. but HEAD handling should stay.

fafhrd91 (approved):

PR #1116, opened 2019-10-03
Update docs of guard.rs

koenhoeijmakers:


JohnTitor (approved):

src/guard.rs
```diff
@@ -29,11 +29,11 @@
...
 
-/// Trait defines resource guards. Guards are used for routes selection.
+/// Trait defines resource guards. Guards are used for route selection.
 ///
-/// Guard can not modify request object. But it is possible to
-/// to store extra attributes on request by using `Extensions` container,
-/// Extensions container available via `RequestHead::extensions()` method.
+//! Guards can not modify the request object. But it is possible
+//! to store extra attributes on a request by using the `Extensions` container.
+//! Extensions containers are available via the `RequestHead::extensions()` method.
```
JohnTitor: Ah, wait, we shouldn't use doc comment for modules here. Should be `///`.

fafhrd91: https://travis-ci.org/actix/actix-web/jobs/593196858#L1103

koenhoeijmakers: You're right, totally a copy paste problem haha :)

JohnTitor (approved):

PR #1119, opened 2019-10-06
 Let ResponseError render w/ 'text/plain; charset=utf-8' header (#1118)

### LINKED ISSUE #1118: ResposeError should always be rendered with charset utf-8 content-type
# Abstract

With actix-web-1.0.8, `error::InternalError` always response with content-type text/plain, but with no charset specification, which makes some unicode characters (i.e. Japanese kanas and chinese characters and emojis) unreadable in user-agent.

# Use case

```rust
fn test() -> Result<String, actix_web::Error> {
    Err(actix_web::error::ErrorForbidden("😋test"))
}
```

# Client Environment

* User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:70.0) Gecko/20100101 Firefox/70.0
* ${LANG} = zh_CN.UTF-8

# Test Result

When access this test case `😋test` , only got `馃構test`, which is actually unreadable. After I use burp suite to intercept the response and change the content-type to `text/plain; charset=utf-8` , the issue fixed.

# Fix suggestion

PR #1119

SuperHacker-liuan:
Trait ResponseError originally render Error messages with header
`text/plain` , which causes browsers (i.e. Firefox 70.0) with
Non-English locale unable to render UTF-8 responses with non-English
characters correctly. i.e. emoji.

This fix solved this problem by specifying the charset of `text/plain`
as utf-8, which is the default charset in rust.

Before actix-web consider to support other charsets, this hotfix is
enough.

Test case:

```rust
fn test() -> Result<String, actix_web::Error> {
    Err(actix_web::error::ErrorForbidden("😋test"))
}
```

Fixes #1118

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -6,6 +6,10 @@
 
 * Add support for serde_json::Value to be passed as argument to ResponseBuilder.body()
 
+### Fixed
+
+* To be compatible with non-English error responses, `ResponseError` rendered with `text/plain; charset=utf-8` header #1118
+
 
 ## [0.2.10] - 2019-09-11
 
--- a/actix-http/src/error.rs
+++ b/actix-http/src/error.rs
@@ -75,7 +75,7 @@ pub trait ResponseError: fmt::Debug + fmt::Display {
         let _ = write!(Writer(&mut buf), "{}", self);
         resp.headers_mut().insert(
             header::CONTENT_TYPE,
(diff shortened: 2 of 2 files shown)


fafhrd91 (approved):

JohnTitor (approved):

fafhrd91 (approved):

fafhrd91:
Thanks!

Issue #1060, opened 2019-08-28
awc::ClientRequest: serialize to URI


Ploppz:
It seems like functionality is already there? Because when you have `Query<T>` in some handler, `actix-web` automatically deserializes the incoming URI to the type `T`.

To name an example, in `reqwest` it's possible to serialize to URI: https://docs.rs/reqwest/0.9.20/reqwest/async/struct.RequestBuilder.html#method.query

I suggest to have a similar `query()` on [ClientRequest](https://docs.rs/awc/0.2.4/awc/struct.ClientRequest.html). What do you think?

fafhrd91:
Sounds good

fafhrd91:
fixed in master

PR #1130, opened 2019-10-11
awc: Add support for setting query from Serialize type for client request

Jonathas-Conceicao:
This solves #1060 but I have one minor issue resolve, I'm leaving a followup comment on the code to discuss.

### DIFF
--- a/awc/src/request.rs
+++ b/awc/src/request.rs
@@ -382,6 +382,27 @@ impl ClientRequest {
         }
     }
 
+    /// Sets the query part of the request
+    pub fn query<T: Serialize>(
+        mut self,
+        query: &T,
+    ) -> Result<Self, serde_urlencoded::ser::Error> {
+        let mut parts = self.head.uri.clone().into_parts();
+
+        if let Some(path_and_query) = parts.path_and_query {
+            let query = serde_urlencoded::to_string(query)?;
+            let path = path_and_query.path();
+            parts.path_and_query = format!("{}?{}", path, query).parse().ok();
+
+            match Uri::from_parts(parts) {
+                Ok(uri) => self.head.uri = uri,
+                Err(e) => self.err = Some(e.into()),
+            }
+        }
+
+        Ok(self)
+    }
+
     /// Freeze request builder and construct `FrozenClientRequest`,
     /// which could be used for sending same request multiple times.
     pub fn freeze(self) -> Result<FrozenClientRequest, FreezeRequestError> {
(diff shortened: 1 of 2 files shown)


awc/src/request.rs
```diff
@@ -382,6 +382,25 @@ impl ClientRequest {
         }
     }
 
+    /// Sets the query part of the request
+    pub fn query<T: Serialize>(mut self, query: &T) -> Self {
+        let mut parts = self.head.uri.clone().into_parts();
+
+        if let Some(path_and_query) = parts.path_and_query {
+            if let Ok(query) = serde_urlencoded::to_string(query) {
```
Jonathas-Conceicao: As the implementation is right now, it would silently fail if the serde_urlencoded fails to serialize the `query` argument, I believe that the `self.err` of the `ClientRequest` should be set, but didn't manage to find a way to create a `http::Error` from serde's or from `actix_http::Error`. Any suggestions on how to handle this?

fafhrd91: I think you should be able to to do `actix_http::Error::from(e)` for serde errors. you may need to add new request's attribute for serde_urlencoded errors, then [prep_for_sending](https://github.com/actix/actix-web/blob/master/awc/src/request.rs#L492) should accommodate that error.

fafhrd91:
ping

awc/src/request.rs
Jonathas-Conceicao: I can generate a `actix_http::Error` from a serde_urlencoded without any problems, but I didn't have a way to store that on the `awc::request::ClientRequest`, so I added a `actix_error` to that is handled by the prep_for_sending as well. But to make the conversion from `actix_http::Error` to a `actix_http::client::FreezeRequestError` I had to add a enum item to the `actix_http::client::FreezeRequestError`.

With this approach a update to the actix_http would have to come first, I've been testing setting the path from actix_http to '../actix_http', but I didn't add this to the PR.

Maybe a better way to handle this would be change the err on `awc::request::ClientRequest` itself to be a `actix_http::Error` insted of having two errors fields, and since we have a From `actix_http::Error` to `http::Error`, doing this should lead to any problems other than having to change a feel types and adding a from here and there.

What do you think? Should I keep two 'err' fields on ClientRequest, or should I change the err type and have only one?

fafhrd91: I don't like idea with actix_http::Error in general, it is fine when it get returned back to framework (ie into actix-web), but i think it is not good to expose it to user. let's make `query()` method return `Result<Self, ..>`

Jonathas-Conceicao:
I've set the query return type  to be a `Result<Self, E>` as suggested. Let me know if there's anything else you think I could improve on this or if you stop something I've missed.

fafhrd91 (approved):

fafhrd91:
thanks!

PR #1133, opened 2019-10-13
Update `syn` & `quote` to 1.0

DanSnow:
This PR include:

- Update `syn` & `quote`
- Provide better error message when argument is invalid. For example:

image

BTW, Can `actix-web-codegen` become a optional feature? I'm not actually using it. But it still pull the older `syn` & `quote` as dependencies into my project.

### DIFF
--- a/actix-web-codegen/CHANGES.md
+++ b/actix-web-codegen/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## [UNRELEASE]
+
+* Bump up `syn` & `quote` to 1.0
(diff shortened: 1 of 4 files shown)


fafhrd91:
changing it to optional dependency is fine. but it should be enabled by default

fafhrd91 (approved):

fafhrd91:
please fix travis failures

JohnTitor (approved):

fafhrd91:
thanks!

PR #1135, opened 2019-10-14
WIP Update webpki and rustls

### LINKED ISSUE #1030: Upgrade to `rustls v0.16.0`
`rustls v0.16.0` has upgraded `ring` which is no longer compatible with `rustls v0.15.2`.

svenstaro:
Still a bit broken apparently. Need to figure out why.

Fixes #1030.


svenstaro:
Ok frankly I got no idea why the tests now fail. This is probably related to the actix-server changes. Any help @fafhrd91?

JohnTitor:
Because `ring` has breaking changes. I added a WIP commit to follow up it but there are still some tasks to pass the build.

- Rename `ring::hmac::Key` to `SigningKey` or something else
- We cannot use `ring::hkdf::expand` anymore so we should find another way (perhaps `ring::hkdf::Prk.expand` but the signature is changed)

fafhrd91:
what is status?

svenstaro:
> what is status?

Well, it's broken and needs fixing :D

fafhrd91:
master is on rustls-0.16

PR #1141, opened 2019-10-16
Add DataRaw type.

fakeshadow:
I want to introduce a foreign type from other crates to `App::data`. 
Although it's already a `Send + Sync`type but the `web::Data` will still wrap it with another layer of `Arc`. So I added this `DataRaw` type which share the same principle with `Data` but it explicitly require the type to be `Send + Sync + Clone` so that it doesn't have to be wrapped in `Arc`.


fafhrd91:
I don't like new type, I'd rather add generic api so this functionality could be added as third-part crate

fakeshadow:
> I don't like new type, I'd rather add generic api so this functionality could be added as third-part crate

Ok. That's understandable.

PR #1142, opened 2019-10-17
Drop hashbrown dep in favour of core's

mocsy:


fafhrd91 (changes requested):
You have to use different hash function

fafhrd91:
ping

mocsy:
> ping

I'll probably have some time to take a look this weekend.

mocsy:
It might not be that simple. Event after switching to ahash, which is the default (at compile time) since hashbrown 5.0 (the current in actix-http), I still get test failures like this:

`---- h1::encoder::tests::test_camel_case stdout ----
thread 'h1::encoder::tests::test_camel_case' panicked at 'assertion failed: `(left == right)`
  left: `b"\r\nContent-Length: 0\r\nConnection: close\r\nContent-Type: plain/text\r\nDate: date\r\n\r\n"`,
 right: `b"\r\nContent-Length: 0\r\nConnection: close\r\nDate: date\r\nContent-Type: plain/text\r\n\r\n"`', actix-http/src/h1/encoder.rs:551:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.`

The Content-Type precedes the Date, I think this is the very point of this test.
Even if it would work, it wouldn't make much sense anyway because all it would accomplish is to replace the dependency on the hashbrown crate to a dependency on the ahash crate, while making the code more convoluted.

I think it's best to keep hashbrown as is for now, maybe updating to 6 from 5 instead?

fafhrd91:
I am fine with keeping hashbrawn

martingallagher:
Is it possible the PR has uncovered an issue; the test case requires deterministic ordering of header keys/values, but AFAIK hashbrown has arbitrary ordering - maybe `indexmap` is a better option.

PR #1151, opened 2019-11-01
Fix escaping/encoding problems in Content-Disposition header

Gowee:
I just find that #461 which I PR last year have some problems due to my misunderstanding then that may result in compatibility problems.

That is, the *Content-Disposition* response header of `actix_files::NameFile` use `DispositionParam::Filename` regardless of whether there are non-ASCII characters in file names or not, which is discouraged by RFC6266:
> Some user agents inspect the value (which defaults to ISO-8859-1 for the quoted-string form) and switch to UTF-8 when it seems to be more likely to be the correct interpretation.
> As with the approaches above, this is not interoperable and, furthermore, risks misinterpreting the actual value.

> Avoid using non-ASCII characters in the filename parameter. Although most existing implementations will decode them as ISO-8859-1, some will apply heuristics to detect UTF-8, and thus might fail on certain names.

For the current implementation, It is known that both Edge and IE11 fail to decode non-ASCII file names (e.g. `Content-Disposition: attachment; filename="文件.doc"` is treated as `content-disposition: inline; filename="æä»¶.html"`).

To fix this problem, `DispositionParam::FilenameExt` is used instead.

To make it more clear for both users and further maintenance, I also add more comments in *content-dispostion.rs*. Besides, a trivial escaping problem is also fixed.

### DIFF
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -13,7 +13,7 @@ use mime_guess::from_path;
 
 use actix_http::body::SizedStream;
 use actix_web::http::header::{
-    self, ContentDisposition, DispositionParam, DispositionType,
+    self, Charset, ContentDisposition, DispositionParam, DispositionType, ExtendedValue,
 };
 use actix_web::http::{ContentEncoding, StatusCode};
 use actix_web::middleware::BodyEncoding;
@@ -95,7 +95,13 @@ impl NamedFile {
             };
             let cd = ContentDisposition {
                 disposition: disposition_type,
-                parameters: vec![DispositionParam::Filename(filename.into_owned())],
+                parameters: vec![
+                    DispositionParam::FilenameExt(ExtendedValue {
+                        charset: Charset::Ext(String::from("UTF-8")),
+                        language_tag: None,
+                        value: filename.as_bytes().to_vec(),
+                    }),
+                ],
             };
             (ct, cd)
         };
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -93,15 +93,18 @@ impl NamedFile {
                 mime::IMAGE | mime::TEXT | mime::VIDEO => DispositionType::Inline,
                 _ => DispositionType::Attachment,
             };
+            let mut parameters =
+                vec![DispositionParam::Filename(String::from(filename.as_ref()))];
+            if !filename.is_ascii() {
+                parameters.push(DispositionParam::FilenameExt(ExtendedValue {
+                    charset: Charset::Ext(String::from("UTF-8")),
+                    language_tag: None,
+                    value: filename.into_owned().into_bytes(),
+                }))
+            }
             let cd = ContentDisposition {
                 disposition: disposition_type,
-                parameters: vec![
-                    DispositionParam::FilenameExt(ExtendedValue {
-                        charset: Charset::Ext(String::from("UTF-8")),
-                        language_tag: None,
-                        value: filename.as_bytes().to_vec(),
-                    }),
-                ],
+                parameters: parameters,
             };
             (ct, cd)
         };
--- a/actix-http/src/header/common/content_disposition.rs
+++ b/actix-http/src/header/common/content_disposition.rs
@@ -76,6 +76,11 @@ pub enum DispositionParam {
     /// the form.
     Name(String),
     /// A plain file name.
+    ///
+    /// It is [not supposed](https://tools.ietf.org/html/rfc6266#appendix-D) to contain any
+    /// non-ASCII characters when used in a *Content-Disposition* HTTP response header, where
+    /// [`FilenameExt`](DispositionParam::FilenameExt) with charset UTF-8 may be used instead
+    /// in case there are Unicode charaters in file names.
     Filename(String),
     /// An extended file name. It must not exist for `ContentType::Formdata` according to
     /// [RFC7578 Section 4.2](https://tools.ietf.org/html/rfc7578#section-4.2).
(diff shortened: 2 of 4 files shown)


actix-files/src/named.rs
```diff
@@ -95,7 +95,13 @@ impl NamedFile {
             };
             let cd = ContentDisposition {
                 disposition: disposition_type,
-                parameters: vec![DispositionParam::Filename(filename.into_owned())],
+                parameters: vec![
+                    DispositionParam::FilenameExt(ExtendedValue {
```
Gowee: Still one problem: by replacing Filename(`filename`) with FilenameExt(`filename*`), it brings risks that some old user agent implementations cannot understand `filename*`.

RFC6266 suggests that:
> When a "filename*" parameter is sent, to also generate a "filename" parameter as a fallback for user agents that do not support the "filename*" form, if possible.  This can be done by substituting characters with US-ASCII sequences (e.g., Unicode character point U+00E4 (LATIN SMALL LETTER A WITH DIARESIS) by "ae").  Note that this may not be possible in some locales.

This is not implemented in this PR, because I am not sure how to "substituting characters" generally. Do you have any  ideas?

Gowee: Or just leave `filename` parameters intact? Say,
```rust
let cd = ContentDisposition {
    disposition: disposition_type,
    parameters: vec![
        DispositionParam::Filename(String::from(filename.as_ref())),
        DispositionParam::FilenameExt(ExtendedValue {
            charset: Charset::Ext(String::from("UTF-8")),
            language_tag: None,
            value: filename.into_owned().into_bytes(),
        }),
    ],
};
```
so that there are both `filename*` and `filename` which might contains invalid charaters.

leizzer: > This is not implemented in this PR, because I am not sure how to "substituting characters" generally. Do you have any ideas?

It seems that you have to replace/translate Unicode characters to equivalent in ASCII that and even so they say not every Unicode char has an equivalent in ASCII.
So I prefer your second proposal.

leizzer:
@Gowee Some tests are failing, could you look at them?

Gowee:
I forgot to update tests in *actix-files/src/lib.rs* accordingly. These new commits should pass CI.

Now both `filename` and `filename*` are present in `Content-Disposition` header for non-ASCII file names.

@leizzer Could you please review these?

actix-files/src/named.rs
Gowee: > It seems that you have to replace/translate Unicode characters to equivalent in ASCII that and even so they say not every Unicode char has an equivalent in ASCII.

You are right. As suggested by @gumblex, iconv has a great [TRANSLIT](https://spin.atomicobject.com/2011/07/13/some-useful-iconv-functionality/) function, which could "convert characters in the origin encoding to the closest possible matching character in the target encoding". But the problem is still "not every Unicode char has an equivalent in ASCII" and in which case those chars are barely replaced by Unicode Replacement Character. So I also prefer the second way.

Gowee: Update:
As suggested by @lambdaq, there is something like https://github.com/chowdhurya/rust-unidecode which provides decent transliteration from Unicode to ASCII. But its functionality is still limited. And I do not think it is worth to introduce a new dependency. Just write it down here for future reference.

fafhrd91:
@JohnTitor what do you think?

actix-files/src/named.rs
lambdaq: Hi, I am wondering what if I suggested anything at all. I dont write Rust for now but I do want to learn this language.

actix-http/src/header/common/content_disposition.rs
```diff
@@ -76,6 +76,11 @@ pub enum DispositionParam {
     /// the form.
     Name(String),
     /// A plain file name.
+    ///
+    /// It is [not supposed](https://tools.ietf.org/html/rfc6266#appendix-D) to contain any
+    /// non-ASCII characters when used in a *Content-Disposition* HTTP response header, where
+    /// [`FilenameExt`](DispositionParam::FilenameExt) with charset UTF-8 may be used instead
+    /// in case there are Unicode charaters in file names.
```
JohnTitor:
```suggestion
    /// in case there are Unicode characters in file names.
```

actix-http/src/header/common/content_disposition.rs
```diff
@@ -774,8 +835,18 @@ mod tests {
 
     #[test]
     fn test_from_raw_uncessary_percent_decode() {
+        // In fact, RFC7578 (multipart/form-data) Section 2 and 4.2 suggests that filename with
+        // non-ASCII characters MAY be percent-encoded.
+        // To the contrary, RFC6266 or other RFCs related to Content-Disposition response header
```
JohnTitor:
```suggestion
        // On the contrary, RFC6266 or other RFCs related to Content-Disposition response header
```

actix-http/src/header/common/content_disposition.rs
```diff
@@ -774,8 +835,18 @@ mod tests {
...
+        // do not mention such percent-encoding.
+        // So, it appears to be undecidable whether to percent-decode or not without
+        // knowning the usage scenario (multipart/form-data v.s. HTTP response header) and
```
JohnTitor:
```suggestion
        // knowing the usage scenario (multipart/form-data v.s. HTTP response header) and
```

JohnTitor (approved):
LGTM with some nits

JohnTitor:
I think having two parameters(`filename` and `filename*`) is reasonable. On many browsers, `filename` is ignored if `filename*` is enabled.

fafhrd91:
@Gowee could you add entry to changes.md

Gowee:
@JohnTitor @fafhrd91 Thanks for reviewing. Done.

fafhrd91:
thanks!

Issue #1149, opened 2019-10-25
actix-http-test: Connect(Timeout)


Ploppz:
I randomly (maybe one time per 20) get `Connect(Timeout)` error in my tests where I use `actix_http_test::TestServer`, and when I block on a request like this:

```
        let req = srv.get("/api/v1/load").send();
        let mut response = srv.block_on(req).unwrap();
```
It's that last line that gives the error. The endpoints themselves should not take a long time, they are quite light.

Ploppz:
I see in the code that the timeout is set to 500ms. But I timed the handler that timed out, and it took 35ms. So I wonder whether the actix-web (test) framework is slow (in debug mode), which I doubt. But another possibility is that when you have a lot of tests running in parallel sending many requests to different TestServers, they might all get slowed waiting for IO? Is that possible? The timeout should probably be increased in any case.

leizzer:
I noticed you submited changes that were merged.
Can we close this?

Ploppz:
Yes :)

PR #1153, opened 2019-11-04
Increase timeouts in test-server

Ploppz:
Due to https://github.com/actix/actix-web/issues/1149, I think the timeout should be increased. I am not sure about the exact value, but I think it should not be too strict. When I test my code on a CI server, it spuriously fails about 10% of the time. Sometimes it fails several times in a row within a few minutes. This makes me suspect that it does have something to do with the IO load on the server. In any case, such spurious timeouts should never happen. It's difficult to reproduce because it has to do with the hardware, and on my own desktop it very rarely happens (but I have seen it).

### DIFF
--- a/test-server/CHANGES.md
+++ b/test-server/CHANGES.md
@@ -5,6 +5,7 @@
 ### Changed
 
 * Update serde_urlencoded to "0.6.1"
+* Increase TestServerRuntime timeouts from 500ms to 3000ms
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


fafhrd91:
you should also add same change to https://github.com/actix/actix-net/tree/master/actix-testing
actix-web will switch to actix-testing in next version

Ploppz:
I don't see any timeout there

fafhrd91:
ok, thanks

fafhrd91 (approved):

PR #1160, opened 2019-11-10
Handle fragmented websocket messages

### LINKED ISSUE #7: Fragmented ws frames are not supported

maciejhirsz:
Closes #7. This PR implements handing of fragmented messages (using continuation frames) by adding a couple new `Frame` variants, and then concating them within `actix-web-actors`.

Some points for discussion:
+ I added some `error!` logs to help myself debug stuff, those could be changed to `debug!` or removed if there are any concerns.
+ The concatenation could be done within the `Codec` itself, although that requires attaching a buffer to the `Codec` which makes it non-`Copy`. I went with adding new frame variants instead as it matches the protocol better (continuation frames are separate frames after all).
+ There concatenation that happens within `actix-web-actors` is currently unconstrained. It could use the same `max_size` limit the `Codec` is enforcing (that's probably the easiest and least confusing way to do it, it basically changes the limit from "frame limit" to "message limit"), or be a setting of its own.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -558,11 +558,13 @@ where
                 }
                 Ok(None) => break,
                 Err(ParseError::Io(e)) => {
+                    error!("IO error when parsing!");
                     self.client_disconnected();
                     self.error = Some(DispatchError::Io(e));
                     break;
                 }
                 Err(e) => {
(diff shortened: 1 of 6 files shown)


maciejhirsz:
I just noticed that #530 has been open for a long time and addresses the same issue, though at a much older root. I've no preference on how this is really solved :P.

fafhrd91:
fixed in master

PR #1167, opened 2019-11-20
Remove implementation of Responder for (). Fixes #1108.

### LINKED ISSUE #1108: impl Responder for () makes common mistake silent
This is a tentative suggestion, since I'm not too familiar with the Actix API, but:

I think it may be a mistake to implement `Responder` for `()`, because that means that a stray semicolon at the end of a responder's code is a silent error. Even when caught in testing, it can take a while to notice the source of the problem.

Rust's "omit a semicolon to return the value" rule would be intolerable for programmers - when I explain it to people, they always look skeptical - except that the type checker always catches the mistake. By permitting `()` as a return type, Actix has removed that safety net.

Without that impl, people would have to return `HttpResponse::Ok()`, which doesn't seem like much of a burden. All the examples in the documentation do this already.

jimblandy:
Rationale:

- In Rust, one can omit a semicolon after a function's final expression to make
  its value the function's return value. It's common for people to include a
  semicolon after the last expression by mistake - common enough that the Rust
  compiler suggests removing the semicolon when there's a type mismatch between
  the function's signature and body. By implementing Responder for (), Actix makes
  this common mistake a silent error in handler functions.

- Functions returning an empty body should return HTTP status 204 ("No Content"),
  so the current Responder impl for (), which returns status 200 ("OK"), is not
  really what one wants anyway.

- It's not much of a burden to ask handlers to explicitly return
  `HttpResponse::Ok()` if that is what they want; all the examples in the
  documentation do this already.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+## [2.0.0-alpha.2] - 2019-xx-xx
+
+### Changed
+
+* Remove implementation of `Responder` for `()`. (#1167)
+
+
 ## [1.0.9] - 2019-11-14
 
 ### Added
--- a/src/app.rs
+++ b/src/app.rs
@@ -84,14 +84,15 @@ where
     ///
     /// ```rust
     /// use std::cell::Cell;
-    /// use actix_web::{web, App};
+    /// use actix_web::{web, App, HttpResponse, Responder};
     ///
     /// struct MyData {
     ///     counter: Cell<usize>,
     /// }
     ///
-    /// async fn index(data: web::Data<MyData>) {
+    /// async fn index(data: web::Data<MyData>) -> impl Responder {
     ///     data.counter.set(data.counter.get() + 1);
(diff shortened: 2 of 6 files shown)


jimblandy:
Note, this is a breaking change.

fafhrd91:
could you rebase, and please add entry to changes.md

jimblandy:
I didn't run the tests with `--workspace`. Fixing.

jimblandy:
Okay, seems like I need 525c22de.

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # Changes
 
+## [1.0.10] - 2019-xx-xx
```
fafhrd91: change it to 2.0.0-alpha.2

jimblandy: Okay.

fafhrd91 (approved):

fafhrd91:
thanks!

Issue #934, opened 2019-06-22
Optional middlewares in 1.0 because of wrap generics


aschampion:
The equivalent code in prior versions used to allow for easy optional enabling of middlewares, but now because of the generic return of `wrap` it's not clear how to get a coherent type for `app` when multiple middlewares may or may not be enabled. This can't be done with `App::configure`, either.

```rust
let mut server = HttpServer::new(
        || {
            let mut app = App::new()
                .service(
                    // etc
                );
            if cors {
                app = app.wrap(Cors::new().send_wildcard());
            }
            app
        })
```
```
error[E0308]: mismatched types
   --> src/main.rs:474:23
    |
474 |                 app = app.wrap(Cors::new().send_wildcard());
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `actix_web::app_service::AppEntry`, found opaque type
    |
    = note: expected type `actix_web::App<actix_web::app_service::AppEntry, _>`
               found type `actix_web::App<impl actix_service::NewService, _>`
```

fafhrd91:
good question. there is no solution at the moment. we need `Optional` service.

KeenS:
How about writing a middleware that conditionally enable another middleware like this?
https://github.com/KeenS/actix-web-condition-middleware

If this is acceptable, I'll submit a PR to this repository.

fafhrd91:
@KeenS looks good. pr is welcomed!

fafhrd91:
Condition middleware has been added to master

aschampion:
I'm still not sure how to wire this up:

```sh
    Checking h2n5 v0.1.3 (/Users/championa/code/rust/h2n5)
error[E0277]: the trait bound `actix_web::middleware::Condition<actix_cors::Cors>: actix_service::transform::IntoTransform<_, actix_web::app_service::AppRouting>` is not satisfied
   --> src/main.rs:469:18
    |
469 |                 .wrap(actix_web::middleware::Condition::new(cors, Cors::new().send_wildcard()))
    |                  ^^^^ the trait `actix_service::transform::IntoTransform<_, actix_web::app_service::AppRouting>` is not implemented for `actix_web::middleware::Condition<actix_cors::Cors>`

error: aborting due to previous error
```

[...]

PR #1177, opened 2019-11-25
Update Condition middleware to support IntoTransform middlewares

KeenS:
As it revealed in https://github.com/actix/actix-web/issues/934#issuecomment-546983746 , current implementation of `Condition` middleware lacks support of `IntoTransform` middlewares.
Thus I'm adding a new function `enable_if`, which can take `IntoTransform` middlewares, without breaking exisitng APIs.

As old `new` is less featureful, I recommend deprecate it (but leave the decision to the actix-web developers).

BTW, it seems `IntoTransform` has disappeared in current master. I'm not much aware of what has been changed, so submitting this P-R to the 1.0 branch. Is this correct?

### DIFF
--- a/src/middleware/condition.rs
+++ b/src/middleware/condition.rs
@@ -1,9 +1,11 @@
 //! `Middleware` for conditionally enables another middleware.
-use actix_service::{Service, Transform};
+use actix_service::{IntoTransform, Service, Transform};
 use futures::future::{ok, Either, FutureResult, Map};
 use futures::{Future, Poll};
+use std::marker::PhantomData;
 
 /// `Middleware` for conditionally enables another middleware.
+///
 /// The controled middleware must not change the `Service` interfaces.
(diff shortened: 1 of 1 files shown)


src/middleware/condition.rs
```diff
@@ -1,9 +1,11 @@
 //! `Middleware` for conditionally enables another middleware.
-use actix_service::{Service, Transform};
+use actix_service::{IntoTransform, Service, Transform};
 use futures::future::{ok, Either, FutureResult, Map};
 use futures::{Future, Poll};
+use std::marker::PhantomData;
 
 /// `Middleware` for conditionally enables another middleware.
+///
```
KeenS, line 8: new line for one line document.

fafhrd91:
seems demand for this change is very low. i am not sure about new release for 1.0

KeenS:
Hmm... Then, I close this P-R. I'll re-open this when someone demand this.

PR #1182, opened 2019-11-27
Remove `rustls` from `package.metadata.docs.rs`

JohnTitor:
I think this is needed to disable `rustls` feature completely and build docs


fafhrd91 (approved):

fabricedesre:
Out of curiosity, why is rustls support removed?

PR #1189, opened 2019-12-03
actix-multipart: Fix multipart boundary reading

alexlarsson:
If we're not ready to read the first line after the multipart field
(which should be a "\r\n" line) then return NotReady instead of Ready(None)
so that we will get called again to read that line.

Without this I was getting MultipartError::Boundary from read_boundary()
because it got the "\r\n" line instead of the boundary.

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## [0.1.5] - 2019-xx-xx
+
+* Multipart handling now handles NotReady during read of boundary #1189
+
 ## [0.1.4] - 2019-09-12
 
 * Multipart handling now parses requests which do not end in CRLF #1038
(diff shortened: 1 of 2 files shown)


alexlarsson:
Note, this applies to 1.0 which i'm using, but looking at master:
  https://github.com/actix/actix-web/blob/0015a204aa223440405eb7ba97a50eac72a851a4/actix-multipart/src/server.rs#L613
It seems to have the same issue.

fafhrd91:
could you add test

fafhrd91:
And please add entry to changes.md

alexlarsson:
Added test and rebased with changes entry.

fafhrd91:
thanks!

would you create pr for master as well?

alexlarsson:
https://github.com/actix/actix-web/pull/1205 is for master

PR #1192, opened 2019-12-06
Add rich project metadata

krircc:


JohnTitor:
It's a very minor point but it'd be nice all the badges are the same style (currently, two badges are flat unlike others).

JohnTitor (approved):
CI failure is unrelated to this PR.

PR #1196, opened 2019-12-07
Fix dependency features.

daxpedda:
The `awc` dependency was incorreclty activating all default features instead of using the feature selection from `actix-web`.


fafhrd91 (approved):

fafhrd91:
Thanks

PR #1204, opened 2019-12-09
Fix broken docs

cevn:
Fixed un escaped brackets in lib.rs, and reflowed links to ConnectionInfo in app, config, and server.rs


fafhrd91 (approved):

fafhrd91:
thanks!

PR #1205, opened 2019-12-09
actix-multipart: Fix multipart boundary reading

alexlarsson:
If we're not ready to read the first line after the multipart field
(which should be a "\r\n" line) then return Pending instead of Ready(None)
so that we will get called again to read that line.

Without this I was getting MultipartError::Boundary from read_boundary()
because it got the "\r\n" line instead of the boundary.

Also tweaks the test_stream test to test partial reads.

This is a forward port of #1189 from 1.0

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## [2.0.0-alpha.4] - 2019-12-xx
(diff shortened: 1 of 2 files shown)


fafhrd91 (approved):

fafhrd91:
thanks!

PR #1208, opened 2019-12-11
Log path if it is not a directory

emigr2k1:
The previous code would not print the path that was provided by the user.


fafhrd91 (approved):

fafhrd91:
thanks!

PR #1225, opened 2019-12-19
feat: add access to the session also from immutable references

tglman:
Hi,

As the title this add the ability to access to the session from not mutable request instances, the current implementation allow that, but if for API cleanness you prefer have mutable request feel free to not accept this PR


fafhrd91 (approved):

fafhrd91:
thanks!

PR #1227, opened 2019-12-20
set the actix-web client send the http2 packet without ssl.

lgm-28:
I have created the interface to send http2 packets without ssl by following the method of sending http1.1 packets. And now I can send HTTP2 packets without ssl with use of set_http2() api.Do you guys think this is a feasible method? If not, can you tell me any good way to send HTTP2 packets without the SSL protocol. Thank you.

### DIFF
--- a/actix-http/src/client/connector.rs
+++ b/actix-http/src/client/connector.rs
@@ -206,6 +206,38 @@ where
         self
     }
 
+    /// Finish configuration process and create connector service wich http2 set.
+    pub fn finish_for_h2(
+        self,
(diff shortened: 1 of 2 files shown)


fafhrd91:
does this code work? 

at the moment there is no way to use http2 over tcp connection.

lgm-28:
Yes,I have made a test and it can work.

fafhrd91:
I am not sure this can work. could you add test

fafhrd91:
Reviewed again, this pr does not do what title says

lgm-28:
OK， I will make a test and push again with new pull request if it works.

PR #1232, opened 2019-12-22
Enable GitHub Actions for Windows and macOS builds

JohnTitor:
Note that the current check suite is fragile due to frequent timeout.

`cargo check` and `cargo test` require OpenSSL so windows-gnu is removed.
Also, somehow i686-pc-windows-msvc is stuck on Actions frequently so it is disabled temporarily.

See the check suite: https://github.com/JohnTitor/actix-web/commit/a28d1d7778da317083a8f332ca8bd796d453d9d7/checks?check_suite_id=370125307

### DIFF
--- a/actix-files/src/lib.rs
+++ b/actix-files/src/lib.rs
@@ -1222,7 +1222,10 @@ mod tests {
         );
 
         let bytes = test::read_body(resp).await;
+        #[cfg(unix)]
         assert!(format!("{:?}", bytes).contains("/tests/test.png"));
+        #[cfg(windows)]
+        assert!(format!("{:?}", bytes).contains("/tests\\\\test.png"));
     }
 
     #[actix_rt::test]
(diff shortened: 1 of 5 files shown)


actix-files/src/lib.rs
```diff
@@ -1222,7 +1222,10 @@ mod tests {
         );
 
         let bytes = test::read_body(resp).await;
+        #[cfg(unix)]
         assert!(format!("{:?}", bytes).contains("/tests/test.png"));
+        #[cfg(windows)]
+        assert!(format!("{:?}", bytes).contains("/tests\\\\test.png"));
```
fafhrd91: I think, this is wrong. We should always return `/` even for windows

JohnTitor: Added a fix, could you review it?

fafhrd91 (approved):
Could also add entry to changes and make new actix-files release

JohnTitor:
Sure, should it is a patch release?

fafhrd91:
Yes, 0.2.1

JohnTitor:
Released actix-files v0.2.1

PR #1233, opened 2019-12-22
Implement Body for Cow<'static, str>

SOF3:


fafhrd91:
i am not sure what is the purpose of `Cow<'static, _>`

SOF3:
For example, some code might either return a dynamic string or a constant literal:

```rust
fn divide(a: i32, b: i32) -> Cow<'static, str> {
    if b == 0 {
        Cow::Borrowed("Error: Division by zero")
    } else {
        Cow::Owned(format!("{}", a / b))
    }
}

#[actix_web::get("/divide/{a}/{b}")]
fn handler((a, b): Path<(i32, i32)>) -> impl Responder {
    HttpResponse::Ok().body(divide(a, b))
}
```

fafhrd91:
I think this wrong use for Cow. It is more Either case

PR #1236, opened 2019-12-26
Fix link to example in readme.md

vv9k:


JohnTitor:
It seems tls example renamed to openssl, not rustls. And we have two examples(openssl/rustls) so it'd be nice to have both links on README.

vv9k:
I've added both of the links to the README.

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1244, opened 2019-12-30
Added 2.0.0 rustls feature name change

jbrookins13:
Added comment to MIGRATION.md to reflect the feature name change from `rust-tls` to `rustls`.


JohnTitor (approved):
No blocking: it'd be nice that we also mention `openssl` feature which was renamed from `ssl`.

fafhrd91:
Thanks!

PR #1258, opened 2020-01-06
Replace sha1 dependency with sha-1

linkmauve:
This other crate is being maintained, and it offers better performances when using the `asm` feature (especially [on AArch64](https://github.com/RustCrypto/hashes/pull/97)).

PR #1259, opened 2020-01-06
Fix actix-http examples

linkmauve:
Did anyone ever run them?!

This was preventing #1258 from passing.

### DIFF
--- a/actix-http/examples/echo.rs
+++ b/actix-http/examples/echo.rs
@@ -7,7 +7,8 @@ use futures::StreamExt;
 use http::header::HeaderValue;
 use log::info;
 
-fn main() -> io::Result<()> {
+#[actix_rt::main]
+async fn main() -> io::Result<()> {
     env::set_var("RUST_LOG", "echo=info");
     env_logger::init();
 
@@ -37,4 +38,5 @@ fn main() -> io::Result<()> {
                 .tcp()
         })?
         .run()
+        .await
 }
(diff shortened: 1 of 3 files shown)


actix-http/examples/echo.rs
```diff
@@ -36,5 +36,6 @@ fn main() -> io::Result<()> {
                 })
                 .tcp()
         })?
-        .run()
+        .run();
+    Ok(())
```
JohnTitor: I think we should use `.await` here, so the function looks like:

```rust
#[actix_rt::main]
...
        .run()
        .await
}
```

actix-http/examples/echo2.rs
```diff
@@ -27,5 +27,6 @@ fn main() -> io::Result<()> {
         .bind("echo", "127.0.0.1:8080", || {
             HttpService::build().finish(handle_request).tcp()
         })?
-        .run()
+        .run();
+    Ok(())
```
JohnTitor: Same here.

actix-http/examples/hello-world.rs
```diff
@@ -23,5 +23,6 @@ fn main() -> io::Result<()> {
                 })
                 .tcp()
         })?
-        .run()
+        .run();
+    Ok(())
```
JohnTitor: Same here.

JohnTitor (changes requested):
I prefer to use `.await` over returning `Ok(())`, as other examples do so.

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1260, opened 2020-01-07
Ignore some tests on macOS and Windows

JohnTitor:
Some tests cause timeout on GitHub Actions (it's Actions specific problem? I'm not sure). So ignore them for now.
And I noticed vcpkg installation takes long time (8~10min), I'm investigating how we cache openssl.

### DIFF
--- a/actix-http/tests/test_openssl.rs
+++ b/actix-http/tests/test_openssl.rs
@@ -115,6 +115,7 @@ async fn test_h2_body() -> io::Result<()> {
     Ok(())
 }
 
+#[cfg(not(any(target_os = "macos", windows)))]
 #[actix_rt::test]
 async fn test_h2_content_length() {
     let srv = test_server(move || {
--- a/actix-http/tests/test_rustls.rs
+++ b/actix-http/tests/test_rustls.rs
@@ -121,6 +121,7 @@ async fn test_h2_body1() -> io::Result<()> {
     Ok(())
 }
 
+#[cfg(not(any(target_os = "macos", windows)))]
 #[actix_rt::test]
 async fn test_h2_content_length() {
     let srv = test_server(move || {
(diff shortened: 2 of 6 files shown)


tests/test_server.rs
```diff
@@ -741,6 +741,7 @@ async fn test_brotli_encoding_large_openssl() {
 }
 
 #[cfg(all(feature = "rustls", feature = "openssl"))]
+#[cfg(not(any(target_os = "macos", windows)))]
```
fafhrd91, line 744: this fails on osx only

awc/tests/test_client.rs
```diff
@@ -46,6 +46,7 @@ const STR: &str = "Hello World Hello World Hello World Hello World Hello World \
                    Hello World Hello World Hello World Hello World Hello World";
 
 #[actix_rt::test]
+#[cfg(not(windows))]
```
fafhrd91, line 49: does it fail on windows?

actix-http/tests/test_rustls.rs
```diff
@@ -121,6 +121,7 @@ async fn test_h2_body1() -> io::Result<()> {
     Ok(())
 }
 
+#[cfg(not(any(target_os = "macos", windows)))]
```
fafhrd91, line 124: does it fail?

actix-http/tests/test_openssl.rs
```diff
@@ -115,6 +115,7 @@ async fn test_h2_body() -> io::Result<()> {
     Ok(())
 }
 
+#[cfg(not(any(target_os = "macos", windows)))]
```
fafhrd91, line 118: i do development on osx, so disabling osx tests breaks my process

JohnTitor:
Their tests actually are failed during my debugging. But I noticed this approach isn't good (I didn't know you use macOS, sorry!) and we need more investigation. I'm closing this for now.

PR #1265, opened 2020-01-09
Update doc comment for `HttpRequest::app_data`

kardeiz:
This PR fixes the doc comment for `HttpRequest::app_data`, which currently references the non-existent `App::extension` method.

It also calls out a big change between actix-web 1.x and 2.x: if you use `App::data` to store data in 2.x, you need to use `req.app_data::<Data<T>>()` to extract it (in 1.x you would just use `req.app_data::<T>()`). This can lead to the data object being silently ignored.


src/request.rs
```diff
@@ -206,8 +206,14 @@ impl HttpRequest {
         &self.0.config
     }
 
-    /// Get an application data stored with `App::extension()` method during
-    /// application configuration.
+    /// Get an application data object stored with `App::data` or `App::app_data`
```
JohnTitor:
```suggestion
    /// Get an application data object stored with `App::data` or `App::app_data`
```

JohnTitor (approved):
Looks good to me with one minor nit

JohnTitor:
Thanks!

PR #1274, opened 2020-01-12
improve App::data and App::app_data docs

### LINKED ISSUE #1272: struggles with shared mutable state in 2.0
I'm trying to update my code from Actix 1.0 to 2.0, and am struggling to convert `register_data()` to `app_data()`. I've not figured out how to resolve the error, `App data is not configured, to configure use App::data()`.

Here's a link to the 1.0 code which was working -- setting up the data:
https://github.com/jeremyandrews/kakaia/blob/master/src/main.rs#L37

And then using the data in a thread:
https://github.com/jeremyandrews/kakaia/blob/master/src/speech.rs#L128

And this works as intended:
```
$ curl -v --request POST --data @test/test.base64 http://127.0.0.1:8088/convert/audio/text
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:8088...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8088 (#0)
> POST /convert/audio/text HTTP/1.1
> Host: 127.0.0.1:8088
> User-Agent: curl/7.65.3
> Accept: */*
> Content-Length: 45632
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 100 Continue
* We are completely uploaded and fine
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 5
< content-type: text/plain; charset=utf-8
< date: Sun, 12 Jan 2020 08:50:13 GMT
< 
test
* Connection #0 to host 127.0.0.1 left intact
```

However, having updated to 2.0 I can't seem to get this working -- here's where I set up the data:
https://github.com/jeremyandrews/kakaia/blob/actixweb2/src/main.rs#L38

And here's where I'm using the data in a thread:
https://github.com/jeremyandrews/kakaia/blob/actixweb2/src/speech.rs#L128

When I push data to the API, I get the following error:
```
$ curl -v --request POST --data @test/test.base64 http://127.0.0.1:8088/convert/audio/text
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:8088...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8088 (#0)
> POST /convert/audio/text HTTP/1.1
> Host: 127.0.0.1:8088
> User-Agent: curl/7.65.3
> Accept: */*
> Content-Length: 45632
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 100 Continue
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< content-length: 56
< content-type: text/plain; charset=utf-8
< date: Sun, 12 Jan 2020 08:44:36 GMT
* HTTP error before end of send, stop sending
< 
* Closing connection 0
App data is not configured, to configure use App::data()
```

I'm not seeing what I'm doing wrong -- any help would be greatly appreciated!

adwhit:
Closes #1272

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -78,9 +78,8 @@ where
     /// an application instance. Http server constructs an application
     /// instance for each thread, thus application data must be constructed
     /// multiple times. If you want to share data between different
-    /// threads, a shared object should be used, e.g. `Arc`. Internally `Data` type
-    /// uses `Arc` so data could be created outside of app factory and clones could
-    /// be stored via `App::app_data()` method.
+    /// threads, it is recommended instead to wrap your state in a `web::Data<T>`
+    /// and attach to the `App` with the `App::app_data` method.
     ///
     /// ```rust
     /// use std::cell::Cell;
@@ -137,13 +136,42 @@ where
         self
     }
 
-    /// Set application level arbitrary data item.
+    /// Set application-level arbitrary data item.
     ///
     /// Application data stored with `App::app_data()` method is available
     /// via `HttpRequest::app_data()` method at runtime.
     ///
-    /// This method could be used for storing `Data<T>` as well, in that case
-    /// data could be accessed by using `Data<T>` extractor.
+    /// If this method is used to store a `Data<T>`, that
+    /// data could be accessed by using the `Data<T>` extractor.
+    ///
+    /// # Examples
+    ///
+    /// This is an example to have shared global state
+    /// that would otherwise go inside an `Arc`.
+    ///
+    /// ```rust
+    /// use std::sync::Mutex;
+    /// use actix_web::{web, App, HttpServer, HttpResponse, Responder};
+    ///
+    /// struct MyData {
+    ///     counter: Mutex<usize>,
+    /// }
+    ///
+    /// async fn index(data: web::Data<MyData>) -> impl Responder {
+    ///     *data.counter.lock().unwrap() += 1;
+    ///     HttpResponse::Ok()
+    /// }
+    ///
+    /// let global_data = web::Data::new(MyData{ counter: Mutex::new(0) });
+    ///
+    /// let server = HttpServer::new(move || {
+    ///     App::new()
+    ///         .app_data(global_data.clone())
+    ///         .service(
+    ///             web::resource("/index.html").route(
+    ///                 web::get().to(index)))
+    /// });
+    /// ```
     pub fn app_data<U: 'static>(mut self, ext: U) -> Self {
         self.extensions.insert(ext);
         self


src/app.rs
```diff
@@ -77,10 +77,9 @@ where
     /// **Note**: http server accepts an application factory rather than
     /// an application instance. Http server constructs an application
     /// instance for each thread, thus application data must be constructed
-    /// multiple times. If you want to share data between different
-    /// threads, a shared object should be used, e.g. `Arc`. Internally `Data` type
-    /// uses `Arc` so data could be created outside of app factory and clones could
-    /// be stored via `App::app_data()` method.
+    /// multiple times. If you wish to share data between different
```
JohnTitor:
```suggestion
    /// multiple times. If you want to share data between different
```

src/app.rs
```diff
@@ -137,13 +136,40 @@ where
...
+    /// let global_data = web::Data::new(MyData{ counter: Mutex::new(0) });
+    ///
+    /// let server = HttpServer::new(move || {
+    ///     App::new()
+    ///         .app_data(global_data.clone())
+    ///         .service(
+    ///             web::resource("/index.html").route(
+    ///                 web::get().to(index)))
+    /// });
+    /// ```
```
JohnTitor: I think this should be in `Example` section.

src/app.rs
```diff
@@ -137,13 +136,40 @@ where
...
     ///
     /// Application data stored with `App::app_data()` method is available
     /// via `HttpRequest::app_data()` method at runtime.
     ///
-    /// This method could be used for storing `Data<T>` as well, in that case
-    /// data could be accessed by using `Data<T>` extractor.
+    /// If this method is used to store a `Data<T>`, that
+    /// data could be accessed by using the `Data<T>` extractor.
+    ///
+    /// This is the recommended method if you have globally-shared state
```
JohnTitor:
```suggestion
    /// This is an example to have shared global state
```

src/app.rs
```diff
@@ -137,13 +136,42 @@ where
...
     ///
-    /// This method could be used for storing `Data<T>` as well, in that case
-    /// data could be accessed by using `Data<T>` extractor.
+    /// If this method is used to store a `Data<T>`, that
+    /// data could be accessed by using the `Data<T>` extractor.
+    ///
+    /// # Examples
+    ///
+    /// This is an example to have shared global state
+    /// that would otherwise go inside an `Arc`.
```
robjtede, line 150: But it does end up in an Arc inside of Data. Feels like this example is the same as the example on `App::data`. What about an example of something that either wouldn't work or is more efficient using `.data`.

robjtede:
Changes have been and are being made in this area. Closing for now, these discussions are useful though so some of these clarifications may be used in future.

PR #1275, opened 2020-01-14
Simplified `Handle` trait

cetra3:
As discussed in the chat, there is no easy migration path for handlers from `0.7`.  The only way you can respond to requests is free floating functions.  You should be able to impl a trait on structs to easily encapsulate data and use this accordingly.

This PR, while it probably won't be accepted initially, is my take on what I would like to see to simplify this process.

It defines a much simpler trait than the existing `Service` trait which requires 2 trait impls and a number of associated types.

The trait is as follows (Could possibly be generic on `Error`):

```rust
pub trait Handle {
    type Future: Future<Output= Result<ServiceResponse, Error>>;

    fn call(&mut self, req: ServiceRequest) -> Self::Future;
}
```

Then a struct (as we can't use specialisation yet) which requires that your struct must be clone so that service factory can be implemented.  There might be other ways to accomplish this. 

Example of usage:

```rust
#[derive(Clone)]
struct TestService;

impl Handle for TestService {
    type Future = Ready<Result<ServiceResponse, Error>>;

    fn call(&mut self, req: ServiceRequest) -> Self::Future {
        ok(req.into_response(HttpResponse::Ok().body("Test")))
    }
}
```

```rust
    std::env::set_var("RUST_LOG", "actix_web=debug");
    env_logger::init();

    HttpServer::new(|| {
        App::new()
            .service(web::service("/").handler(TestService)))
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1,3 +1,5 @@
+use std::future::Future;
+use std::task::{Context, Poll};
 use std::cell::{Ref, RefMut};
 use std::rc::Rc;
 use std::{fmt, net};
@@ -9,7 +11,9 @@ use actix_http::{
     ResponseHead,
 };
 use actix_router::{IntoPattern, Path, Resource, ResourceDef, Url};
-use actix_service::{IntoServiceFactory, ServiceFactory};
+use actix_service::{IntoServiceFactory, ServiceFactory, Service};
+
+use futures::future::{ok, Ready};
 
 use crate::config::{AppConfig, AppService};
 use crate::data::Data;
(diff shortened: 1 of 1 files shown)


fafhrd91:
Your `Handle` trait is just specific version of `Service` trait. But it misses two features
Async initialization, which is `ServiceFactory` does. and backpressure support, `Service` implements backpressure via poll_ready() method.

i dont see how `Handle` trait can replace `Service` trait.

cetra3:
I'm not looking to replace the `Service` trait at all.  The `Handle` trait is just simplified down for cases when you don't need backpressure or async initialization (like in a standard function/extractor sense).

I've been investigating upgrading one of my projects from `0.7` to `2.0` and found that it's going to be hard without an analog to the existing `Handler` trait & there is no real direction in the `MIGRATIONS.MD` file.  I know there was a discussion that `Service` would be replacing the handler trait but that is much more complex than `Handler` was.

Another possibility is to look at using tower's service trait, but possibly behind a feature flag or as an external crate.

fafhrd91:
`FnMut(ServiceRequest) -> Future<ServiceRespinse>` is a service too

cetra3:
Yep, but the challenge with that is it's a free floating function.  If you could replace `FnMut` with a custom struct, then that would be a perfect solution

fafhrd91:
you can replace FnMut with any struct that implements Service trait.

cetra3:
Yes you can, but not without having to define a `ServiceFactory`, otherwise there is no way to wire it up.

I.e, without this change this is the *bare minimum* to have a struct somewhere in your web app instead of a function.

Lots of boilerplate:

```rust
impl Service for TestService {
    type Request = ServiceRequest;
    type Response = ServiceResponse;
    type Error = Error;
    type Future = Ready<Result<Self::Response, Self::Error>>;

    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        Poll::Ready(Ok(()))
    }

    fn call(&mut self, req: ServiceRequest) -> Self::Future {
        ok(req.into_response(HttpResponse::Ok().body("Test")))
    }
}

impl ServiceFactory for TestService {
    type Request = ServiceRequest;
    type Response = ServiceResponse;
    type Error = Error;
    type Future = Ready<Result<TestService, ()>>;
    type Config = ();
    type InitError = ();
    type Service = TestService;

    fn new_service(&self, _config: ()) -> Self::Future {
        ok(self.clone())
    }
}
```
If there is a simpler way than that currently, I could not find it in documentation or examples or can be lead there by the compiler.

Compare that with the handle trait:

```rust
impl Handle for TestService {
    type Future = Ready<Result<ServiceResponse, Error>>;

    fn call(&mut self, req: ServiceRequest) -> Self::Future {
        ok(req.into_response(HttpResponse::Ok().body("Test")))
    }
}
```

27 Lines of code vs. 7 and a lot more readable.

fafhrd91:
You can use fn_factory

PR #1278, opened 2020-01-15
Fixed documentation typo for actix-files

adamkewley:
Found when I was optimizing an application for memory usage. Setting the documented env var (`ACTIX_CPU_POOL`) essentially did nothing. I found the fixed env var in `actix-net/actix_threadpool`:

https://github.com/actix/actix-net/blob/master/actix-threadpool/src/lib.rs#L14

Which seemed to fix my problem.


fafhrd91:
thanks!

fafhrd91 (approved):

Issue #1286, opened 2020-01-19
impl downcast_ref for MessageBody


leptonyu:
`ResponseError` has a function `downcast_ref` which convert `dyn ResponseError` to concrete type of ResponseError. It is very useful.

`Box<dyn MessageBody>` is used in definition of `web_http::body::Body`. it's better to provide function `downcast_ref` for `MessageBody` which can be used for dynamic modify ReponseBody in middleware.

```rust
pub trait MessageBody {
    fn size(&self) -> BodySize;

    fn poll_next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<Bytes, Error>>>;

  //Code added
    #[doc(hidden)]
    fn __private_get_type_id__(&self) -> TypeId
    where
        Self: 'static,
    {
        TypeId::of::<Self>()
    }
}

impl dyn MessageBody + 'static {
    pub fn downcast_ref<T: MessageBody + 'static>(&self) -> Option<&T> {
        if self.__private_get_type_id__() == TypeId::of::<T>() {
            unsafe { Some(&*(self as *const dyn MessageBody as *const T)) }
        } else {
            None
        }
    }
}
```

PR #1287, opened 2020-01-19
impl downcast_ref for MessageBody

leptonyu:
#1286


cetra3 (approved):
This pattern is already used within [error](https://github.com/actix/actix-web/blob/1d12ba9d5ff691178c3fe61f3f662ca7d98080d2/actix-http/src/error.rs#L97), it is a cut & paste of the code basically so I don't see an issue with it

zzy (approved):

Dowwie:
This feels regressive.  Should anyone approve adding unsafe to this code base, even if prior work exists?  I'm not suggesting this work be rejected but rather the original unsafe and this one both refactor away the need for unsafe.

@actix/contributors Stop all new unsafe, regardless of precedent

mitsuhiko:
I think this generally makes sense. I would propose hiding this behind a macro so we can reuse the code safely in more than one place. There is no way to implement this without unsafe and neither std::error/any/failure/error or anything else does anything different here.

Dowwie:
@leptonyu if this PR is to move forward, it requires wrapping the relevant unsafe in a macro and using that macro where applicable

leptonyu:
@Dowwie ok, I need sometime to do it.

leptonyu:
@robjtede I've updated the code.

robjtede:
LGTM.

Ping @cetra3 @zzy

cetra3:
LGTM

zzy (approved):

JohnTitor:
Sorry for the delay! My bandwidth is quite tight now :/
I'm fine with this, any concerns? @actix/contributors also cc @Aaron1011 

@leptonyu btw, could you please stop requesting review frequently? We'll review PRs with our free time :)

robjtede (approved):

Aaron1011 (approved):

leptonyu:
@JohnTitor ok

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1297, opened 2020-01-21
Extensions module improvement and tests.

zero-systems:
Firstly `contains` method of extensions module used `self.map.get(...).is_some()` and I changed it to `self.map.contains_key`.

Secondly I added some tests for different types and operations.

Thirdly I see than there is no need in box casting in some functions and removed casts.

### DIFF
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -119,11 +119,11 @@ fn test_integers() {
     assert!(map.get::<i32>().is_some());
     assert!(map.get::<i64>().is_some());
     assert!(map.get::<i128>().is_some());
-    assert!(map.get::<i8>().is_some());
-    assert!(map.get::<i16>().is_some());
-    assert!(map.get::<i32>().is_some());
-    assert!(map.get::<i64>().is_some());
-    assert!(map.get::<i128>().is_some());
+    assert!(map.get::<u8>().is_some());
+    assert!(map.get::<u16>().is_some());
+    assert!(map.get::<u32>().is_some());
+    assert!(map.get::<u64>().is_some());
+    assert!(map.get::<u128>().is_some());
 }
 
 #[test]
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -71,6 +71,92 @@ impl fmt::Debug for Extensions {
 }
 
 #[test]
+fn test_remove() {
+    let mut map = Extensions::new();
+
+    map.insert::<i8>(123);
+    assert!(map.get::<i8>().is_some());
+
+    map.remove::<i8>();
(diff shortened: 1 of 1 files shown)


actix-http/src/extensions.rs
```diff
@@ -70,6 +70,92 @@ impl fmt::Debug for Extensions {
...
+    assert!(map.get::<i8>().is_some());
+    assert!(map.get::<i16>().is_some());
+    assert!(map.get::<i32>().is_some());
+    assert!(map.get::<i64>().is_some());
+    assert!(map.get::<i128>().is_some());
+    assert!(map.get::<i8>().is_some());
+    assert!(map.get::<i16>().is_some());
+    assert!(map.get::<i32>().is_some());
+    assert!(map.get::<i64>().is_some());
+    assert!(map.get::<i128>().is_some());
```
JohnTitor: They should be uints, I think.

actix-http/src/extensions.rs
```diff
@@ -70,6 +70,92 @@ impl fmt::Debug for Extensions {
...
+        pub guns: usize,
+    }
+
+    let mut map = Extensions::new();
+
+    map.insert(Magi(Madoka { god: false }));
+    map.insert(Magi(Homura { attempts: 0 }));
+    map.insert(Magi(Mami { guns: 999 }));
+
+    assert_eq!(false, map.get::<Magi<Madoka>>().unwrap().0.god);
```
JohnTitor: Yeah, the current assertion isn't wrong but I'd prefer using `assert!` to `assert_eq!` in this case.

```suggestion
    assert!(!map.get::<Magi<Madoka>>().unwrap().0.god);
```

actix-http/src/extensions.rs
```diff
@@ -70,6 +70,92 @@ impl fmt::Debug for Extensions {
...
+        pub god: bool,
+    }
+
+    struct Homura {
+        pub attempts: usize,
+    }
+
+    struct Mami {
+        pub guns: usize,
+    }
```
JohnTitor, line 143: btw, I like this naming.

actix-http/src/extensions.rs
```diff
@@ -70,6 +70,92 @@ impl fmt::Debug for Extensions {
...
 
+#[test]
+fn test_remove() {
+    let mut map = Extensions::new();
+
+    map.insert::<i8>(123);
+    assert!(map.get::<i8>().is_some());
+
+    map.remove::<i8>();
+    assert!(!map.get::<i8>().is_some());
```
JohnTitor: Could simplify

```suggestion
    assert!(map.get::<i8>().is_none());
```

JohnTitor:
Thanks!

PR #1298, opened 2020-01-21
Add benchmark for full stack request lifecycle

dunnock:
This PR is adding criterion and one benchmark of full stack client-server communication. 

Note: client server communication benchmarks are quite volatile and results might vary due to connection lags. To minimize impact of that benches::server::bench_async_burst issues requests in batches. Each batch is reusing same client connection (keep alive).

Other kinds of benchmarks which are not async will be quite easy to implement.


Cargo.toml
```diff
@@ -116,3 +117,7 @@ actix-session = { path = "actix-session" }
 actix-files = { path = "actix-files" }
 actix-multipart = { path = "actix-multipart" }
 awc = { path = "awc" }
+
+[[bench]]
+name = "server"
+harness = false
```
alekseysidorov: Newline

benches/server.rs
```diff
@@ -0,0 +1,66 @@
...
+                   Hello World Hello World Hello World Hello World Hello World \
+                   Hello World Hello World Hello World Hello World Hello World \
+                   Hello World Hello World Hello World Hello World Hello World \
+                   Hello World Hello World Hello World Hello World Hello World \
+                   Hello World Hello World Hello World Hello World Hello World";
+
+
+// benchmark sending all requests at the same time
+fn bench_async_burst(c: &mut Criterion) {
+    let srv = test::start(|| {
```
alekseysidorov, line 30: `srv` actually starts in another thread. So you may wait until it starts successfully.

alekseysidorov (commented):
This benchmark is highly dependent on what number of workers `actix-web` uses under the hood. I think that you may add additional benchmarks for the fixed number of workers: 1, 2, 4, 8, 16.

benches/server.rs
dunnock: Good point. I actually thought that it is waiting here:
https://github.com/actix/actix-web/blob/master/src/test.rs#L718
If not what is the best way to wait for that, cannot think of anything else but issuing web requests until they are successful but that seems like overkill..

dunnock:
I think it is now ready for review. 

Latest commit added benchmarks for CloneableService conversion to RefCell to validate solution for https://github.com/actix/actix-web/issues/1295, though few things appeared quite complicated:
1. There is service benchmark generic defined in `bench/service.rs`, which I wanted to reuse in `bench/cloneable.rs`. That appeared problematic as I cannot address actix_web::bench. From the other side moving generic under src folder would require adding `criterion` to dependencies (now it's in dependencies-dev).
2. From benchmark I needed to access CloneableService from actix_http crate, hence one way or another I had to copy wither generic or CloneableService. I have placed CloneableService copy under bench/experiments, as well as modified version.

Probably we should move benchmarks to another subcrate, preserving the experiments practice?

In general, not exactly relevant to this PR, though actix_web::{dev, test} IMHO might be moved to another crate too and made available only via feature flag.

JohnTitor:
Overall looks good to me, 28d8106cf92ffa87bbef26bfd57e88629050c00b can be dropped once #1303 is merged?

dunnock:
@JohnTitor indeed, except changes in benches/service.rs. I will make respective changes in #1303 if this lands on master first

dunnock:
@JohnTitor I have dropped cloneable experiment files and believe this is ready for review now

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1299, opened 2020-01-21
Remove codecoverage for tests and examples

godofdream:
Codecov checks Codecoverage for bench and tests folders.
By excluding these files from codecoverage we show more correct codecoverage statistics.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,5 @@
+ignore: # ignore codecoverage on following paths
+  - "**/tests"
+  - "test-server"
+  - "**/benches"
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,6 @@
+codecov:
```
JohnTitor: This line seems unnecessary.

godofdream:
I removed the "codecov:" line.

JohnTitor (approved):

JohnTitor:
Thanks!

Issue #1304, opened 2020-01-22
Error in data factory causes panic


runfalk:
I am using actix-web 2.0.0.

The following example has a confusing panic message:

```rust
use anyhow::Result;
use actix_web::{web, App, HttpServer, Responder};
async fn data() -> Result<u32, ()> {
    Err(())
}

async fn greet(_: web::Data<u32>) -> impl Responder {
    "Hello world"
}

#[actix_rt::main]
async fn main() -> Result<()> {
    Ok(HttpServer::new(move || {
        App::new().data_factory(data).route("/", web::get().to(greet))
    })
    .bind("127.0.0.1:5000")?
    .run()
    .await?)
}
```

If a data_factory returns a [Result](https://doc.rust-lang.org/std/result/enum.Result.html) and the result is error it triggers a panic within Actix. See stack trace below (three panics from a single request). Since the error message isn't triggered in my code it is hard to realize that the data factory is the culprit. Would it be possible to display this in a more helpful way?

```
thread 'actix-rt:worker:4' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/73528e339aae0f17a15ffa49a8ac608f50c6cf14/src/libcore/slice/mod.rs:2802:14
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
[...]

PR #1309, opened 2020-01-22
Incorrect handling of errors in data factory

dunnock:
This originates from #1304. 

When data_factory produces Error server is interrupting connection which is incorrect by web standard. It should return `StatusCode::INTERNAL_SERVER_ERROR` instead.

- [X] Add test case for #1304.
- [ ] Fix the issue

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -278,4 +278,31 @@ mod tests {
 
         assert_eq!(num.load(Ordering::SeqCst), 0);
     }
+    
+    #[actix_rt::test]
+    async fn test_error_data_service() {
+        let mut srv =
+            init_service(App::new().data_factory(|| async { Err::<u32,_>(()) }).service(
+                web::resource("/").to(|_: web::Data<usize>| HttpResponse::Ok()),
+            ))
+            .await;
+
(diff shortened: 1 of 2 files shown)


dunnock:
On fixing this issue - I am not sure it is feasible. After investigating actix-service connection lifecycle, connections decoupled of workers and in this case when workers cannot start connection will be just dropped. In this case if any Service failing to start due to error all Workers will fail to start.

What about merging this PR since it increases code coverage anyways and upgrading that #1304 to enhancement (not sure about priority).

One possible way to resolve is any service failing in AppInitResult, still let it start as ServiceUnavailable, e.g. here 
https://github.com/actix/actix-web/blob/master/src/app_service.rs#L192
spawn something like AppUnavailableService which will return 503 on all requests. Though that might be quite complicated change as one need to ensure that AppUnavailableService is permanenetly polled and if all services started upgrades to AppInitService as expected...

fakeshadow:
This simple replacement of all services has no runtime overhead.
https://github.com/actix/actix-web/pull/1927

robjtede:
data factory is being deprecated

PR #1327, opened 2020-01-28
Don't use actions/cache in Windows CI

JohnTitor:
Recent Windows builders are failed due to no disk space: https://github.com/actix/actix-web/actions
It should be an issue of cache so we don't cache on Windows builders for now. Probably there is some space to cache OpenSSL, it can be follow-up work.


leizzer (approved):
LGTM

tyranron (approved):

PR #1337, opened 2020-02-03
Dont REUSEADDR and upgrade to socket2

### LINKED ISSUE #1335: Port already in use does not cause error on Windows + WSL
## Expected Behavior
When you start up the HttpServer the bind("127.0.0.1:65000")? throws an error when the port referenced is under use.

## Current Behavior
You can start up as many instances of the same application as you want (tested up to 3) and none of the later started throw error on the fact that the port is already bound. In fact, the first instance catches the requests, the other ones does not receive it. When you terminate the first instance, the second will catch the requests. And if that is terminated then the third. And when the third is terminated, then the request fails as expected.

1. Use the barebone application code, what can be found in the examples
2. Run multiple instances

## Context
This can cause a mayor issue if you bind your port on something you don't know is already been used, for example by a running SQL server.

## Your Environment
Windows 10, WSL - Ubuntu

* Rust Version: rustc 1.41.0 (5e1a79984 2020-01-27)
* Actix Web Version: 2.0, 1.0.8

Rudi3:
Attempt to fix https://github.com/actix/actix-web/issues/1335 and upgrade to the still actively developed `socket2` crate while preserving the [`backlog` option](https://github.com/actix/actix-web/blob/6a0cd2dced8b085865e966e2d40baf1b8ad9d640/src/server.rs#L575).

If `socket2` is a bad idea, `builder.reuse_address(false)?;` would also work to fix this.
When removing `net2` I also modified the tests, specifically the utility function `unused_addr` to use `net::TcpListener`.


Rudi3:
I'm not sure about `create_tcp_listener`. Is it ultimately supposed to be in 

https://github.com/actix/actix-net/blob/6908b589435bac24fb69249ed0b9ebabc4f53b05/actix-server/src/builder.rs#L490

or in

https://github.com/actix/actix-web/blob/6a0cd2dced8b085865e966e2d40baf1b8ad9d640/src/server.rs#L565 or both? It doesn't seem to be used in/from `builder.rs` atm.

cetra3:
I don't think it's worth changing to `socket2` just to turn off REUSEADDR here.  I'm assuming that this could be simplified down, i.e, simply making `reuse_addr` an option to turn on/off?

Rudi3:
@cetra3 sure. Upon first glance it seemed like `net2` was not actively developed. I just thought that `socket2` could be more future-proof, but it's not required at all to fix https://github.com/actix/actix-web/issues/1335. Can do either.

Beyond that: do you think not using `REUSEADDR` could break anything? I didn't notice anything, but I'm also not sure...

cetra3:
@Rudi3 I can see `REUSEADDR` in use if you have a "backup" process or you are upgrading your process instance & don't want to drop connections when you kill off your old process.

I would leave it on by default, otherwise it is classed as a breaking change, but provide an option to turn it off.

robjtede:
> I would leave it on by default, otherwise it is classed as a breaking change

That said, we're looking toward a v3 right now so breaking changes are up for dicussion.

Rudi3:
@cetra3 so then you could specify it in `server.bind()`? But where would you put it? It seems like you can specify multiple `SocketAddrs`. Perhaps something [like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=89a236dbbd43fd378649a00002e0e40a)? It would preserve the current api. And it could allow multiple addresses, same config and different addresses, different config.

Should the creation of the listener/socket happen in the `actix_net::ServerBuilder` or in `actix_web::HttpServer`?

cdbattags (approved):

Rudi3:
@cdbattags while this PR should lead to consistant behavior, it would currently take away the potential functionality of reusing the port on windows. As well as binding to different addrs, same port in general.

Do you think that there should be an option to select that on bind? In other words: were the reuse mechanics intentional or is that currently not supported  anyways? If it's not supported rn, it might be acceptable to add that later on. But if it is, I wouldn't want to break that with my PR.

shirshak55:
hmm if we run multiple actix process for scaling etc. isn't it going to be problematic?

Rudi3:
@shirshak55 if my theory is right, you currently cannot run multiple instances on the same port on linux (where is would make sense with loadbalancing, etc). Also see https://github.com/actix/actix-web/issues/1335.
Perhaps using "socket options" where you can choose the flags that are used to bind the socket could improve functionality. But it's os-dependent behaviour, so you'd probably have to adjust for that out of actix...
I have a crude proposal for a possible way to add socket options [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=89a236dbbd43fd378649a00002e0e40a).

robjtede:
@Rudi3 are you still having issues with this? We migrated to socket2 separately so this PR should be rebased if still being considered.

PR #1343, opened 2020-02-06
Add dependencies to docs example

### LINKED ISSUE #1342: Needs `actix_rt` dependency to start
The crate `actix_rt` is already a dependency and thus should be available for use.

```
use actix_web::{get, web, App, HttpServer, Responder};

#[get("/{id}/{name}/index.html")]
async fn index(info: web::Path<(u32, String)>) -> impl Responder {
    format!("Hello {}! id:{}", info.1, info.0)
}

#[actix_rt::main]
    HttpServer::new(|| App::new().service(index))
        .bind("127.0.0.1:4444")?
        .run()
        .await
}
```

When I try to execute the simplest example from the docs I get an error.

```
error[E0433]: failed to resolve: use of undeclared type or module `actix_rt`
  --> src/main.rs:19:3
   |
19 | #[actix_rt::main]
   |   ^^^^^^^^ use of undeclared type or module `actix_rt`

error[E0433]: failed to resolve: use of undeclared type or module `env_logger`
  --> src/main.rs:22:5
   |
22 |     env_logger::init();
   |     ^^^^^^^^^^ use of undeclared type or module `env_logger`

error[E0277]: `main` has invalid return type `impl std::future::Future`
  --> src/main.rs:20:20
   |
20 | async fn main() -> std::io::Result<()> {
   |                    ^^^^^^^^^^^^^^^^^^^ `main` can only return types that implement `std::process::Termination`
   |
   = help: consider using `()`, or a `Result`

error: aborting due to 3 previous errors
```

The error disappears by adding `actix_rt` dependency in the `Cargo.toml`. If this is the expected behaviour then the docs needs to be updated.

elliotekj:
Hi,

This PR adds a dependencies code block to the example in the docs to highlight the need for `actix_rt` to be included as a dependency for the example to run.

closes #1342


src/lib.rs
```diff
@@ -8,6 +8,12 @@
 //! for Rust.
 //!
 //! ```rust,no_run
```
elliotekj: Setting this to `toml` didn't highlight the block but setting it to `rust` did, so I left it like this. Happy to change it.

elliotekj:
Huh, the tests failed. I didn't even thing to run `cargo test`, `cargo build` & `cargo doc` didn't raise anything. I guess it's because the block's marked as \`\`\`rust but doesn't contain valid Rust. Fixing.

JohnTitor:
I feel it'd be nice we could tell the user that actix-web requires actix-rt as a runtime, not just list dependencies.

elliotekj:
Sorry for the delay in following this up. Is my update the sort of thing you had in mind @JohnTitor?

JohnTitor (approved):
Sorry! I completely forgot about this.

PR #1344, opened 2020-02-07
Replace unsafe content length helper

robjtede:
Tackling the low-hanging fruit. Attempting to replace unsafe blocks in the actix-http h1 helpers.rs file with safe equivalents that are as performant as possible (though I don't expect a small decrease in the performance of these functions to materially impact an overall request's performance).

### DIFF
--- a/actix-http/Cargo.toml
+++ b/actix-http/Cargo.toml
@@ -93,8 +93,13 @@ actix-server = "1.0.0"
 actix-connect = { version = "1.0.0", features=["openssl"] }
 actix-http-test = { version = "1.0.0", features=["openssl"] }
 actix-tls = { version = "1.0.0", features=["openssl"] }
+criterion = "0.3"
 futures = "0.3.1"
 env_logger = "0.6"
 serde_derive = "1.0"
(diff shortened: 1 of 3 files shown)


cdbattags (dismissed):

robjtede:
@cdbattags not done yet, my dude :)

cdbattags:
> @cdbattags not done yet, my dude :)

Yezzir, sorry I didn't know we're gating PRs around approvals now. My culture for approvals usually stems as just "I'm good with your current approach" instead of a "this immutable chunk of changes is good to go" if that makes sense.

Obviously saw the draft as well.

robjtede:
Ready for review.

Here's the violin plot from the benchmark run. Currently the fast path covers responses <10Kb. We could easily add another case for <100Kb if desired.

safe

JohnTitor:
Looks good to me!

> We could easily add another case for <100Kb if desired.

Yeah, let's add it.

robjtede:
There are now fast paths for responses up to 1MB.

2020-02-24_20 54 56-8de68e0a

JohnTitor (approved):

PR #1346, opened 2020-02-07
Backport to 1.0: allow explicit SameSite=None cookies (#1282)

### LINKED ISSUE #1035: Support SameSite value "None" cookie attribute

theikkila:
fixes #1035
(cherry picked from commit a3287948d19bbdc9e7cf9957403961eeb2d8b94d)

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,3 +1,57 @@
+## Unreleased
+
+* Setting a cookie's SameSite property, explicitly, to `SameSite::None` will now
+  result in `SameSite=None` being sent with the response Set-Cookie header.
+  To create a cookie without a SameSite attribute, remove any calls setting same_site.
+
+## 2.0.0
+
+* `HttpServer::start()` renamed to `HttpServer::run()`. It also possible to
+  `.await` on `run` method result, in that case it awaits server exit.
+
+* `App::register_data()` renamed to `App::app_data()` and accepts any type `T: 'static`.
+  Stored data is available via `HttpRequest::app_data()` method at runtime.
+
+* Extractor configuration must be registered with `App::app_data()` instead of `App::data()`
+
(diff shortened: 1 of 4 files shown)


robjtede:
Thanks for the backport. Not sure what the best approach is for the changelogs. @actix/contributors

cdbattags (approved):

JohnTitor:
Uhm, I wouldn't backport this as discussed on former PR.

JohnTitor:
As doc says, `None` here means "No `SameSite` attribute" and we should expect so. One avoidance is to add "The _None_ `SameSite` attribute" and leave `None` variant. But naming is a problem and may cause confusion (yeah, current naming also causes it though :/).

robjtede:
This is cherry picked from my fix. The issue before was that None and Some(SameSite::None) are treated the same which is wrong.

JohnTitor:
Yeah, I'd just say when you want to fix it on 2.0 or 1.0, you should come up with another way that doesn't break current behavior.

robjtede:
Hmm, I stil consider this a bug and not a breaking change per se.
To do what you're asking there'd have to be either a feature flag or an extra field on Cookie to enable the behavior.

JohnTitor:
Uhmm, okay I'll re-visit here once we make a new alpha release for actix-http.

JohnTitor:
Okay, I was thinking about this and would like to accept. But we're going ahead to 3.0 so we should backport to 2.0.

robjtede:
In light of recent issues, I don't think we should backport this without also backporting the cookie crate migration... @actix/core-contributors ?

PR #1349, opened 2020-02-10
Update Session doc to refer to UserSession instead of RequestSession

dotdash:


robjtede:
Thanks for the doc update. We're migrating actix-session over to https://github.com/actix/actix-extras. Can you move this PR over to there please.

dotdash:
https://github.com/actix/actix-extras/pull/3

PR #1354, opened 2020-02-13
Disable coverage for PRs

JohnTitor:
It's meaningless _mostly_ since GHA disables secrets when PR comes from forks.


messense:
FYI, Codecov released a new version of the action which added support for tokenless uploads.

https://github.com/codecov/codecov-action/issues/29#issuecomment-595062189

JohnTitor:
@messense Wow, that's great! I'll re-setup codecov actions for PRs, thanks for the notice.

PR #1358, opened 2020-02-18
Adapt workspace crates to MessageBody::poll with Pin

dunnock:
After https://github.com/actix/actix-web/pull/1332 there are incompatibility issues in actix-web logger left as well as some tests need to be reenabled.

This PR is resolving incompatibility issues so that whole workspace would compile and run tests with local actix-http version. PR is still placed against msg-body branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ codegen-units = 1
 
 [patch.crates-io]
 actix-web = { path = "." }
-#actix-http = { path = "actix-http" }
+actix-http = { path = "actix-http" }
 actix-http-test = { path = "test-server" }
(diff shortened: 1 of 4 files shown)


robjtede (approved):

JohnTitor (approved):

JohnTitor:
Nice work!

JohnTitor:
@dunnock Hey, we're close to new alpha-release for actix-http, are there remaining things you come up with?

dunnock:
@JohnTitor I have reviewed changes once again and was only able to find one more discrepancy, created separate PR https://github.com/actix/actix-web/pull/1363 just in case those crates are still part of this project. Otherwise it all seems ok.

PR #1363, opened 2020-02-20
Reenable actix-session and actix-identity in msg-body branch

dunnock:
It was disabled previously in https://github.com/actix/actix-web/pull/1332 as some checks were broken. Now after recent changes in https://github.com/actix/actix-web/pull/1358 reenabling does not break checks and tests anymore.

This change is only affecting msg-body branch


robjtede:
wouldnt bother, they're being moved to actix-extras anyway

dunnock:
cool, makes sense

PR #1366, opened 2020-02-21
Upgrade `time` to 0.2.6

kevinpoitra:


JohnTitor:
Generally, updating patch versions are unnecessary. Do you have a strong opinion to use `0.2.6`?

kevinpoitra:
I updated actix-web mainly to keep it in sync with the `time` version used within `actix-extras`, wherein my PR will make those libraries use 0.2.6. Outside of that, the [`time` 0.2.6 release](https://github.com/time-rs/time/releases/tag/v0.2.6) mentions some inconsistencies in that were introduced 0.2.5 - none of which I believe actix-web is affected by, but seeing as there isn't a version of actix-web released that includes `time` 0.2, I figure updating its patch version to avoid being on a version with those inconsistencies would be best. It's not absolutely necessary though, and my changes within my actix-extras PR will still compile even if `time` is left on 0.2.5.

JohnTitor:
We use caret requirements for `time` so Cargo can skip yanked patch versions. I think it's no problem here.

kevinpoitra:
You're right - I just refreshed myself on how version formatting works within Cargo, and I forgot that the caret is added automatically if no other requirement operators are present, making this PR unneeded!

PR #1379, opened 2020-02-24
Add actix-http support for actix error messages

dunnock:
As part of actix:0.10 release actix will not have "http" feature flag https://github.com/actix/actix/issues/321 which allowed to break circular dependency. Though actix does not support handling of actor's error messages via ResponseError anymore. 

This PR introduces "actors" feature flags which would re-add support for handling of MailboxError and ResolverError (with "resolver" flag).

### DIFF
--- a/actix-http/Cargo.toml
+++ b/actix-http/Cargo.toml
@@ -15,7 +15,7 @@ license = "MIT/Apache-2.0"
 edition = "2018"
 
 [package.metadata.docs.rs]
-features = ["openssl", "rustls", "failure", "compress", "secure-cookies","actix"]
+features = ["openssl", "rustls", "failure", "compress", "secure-cookies","actors"]
 
 [lib]
 name = "actix_http"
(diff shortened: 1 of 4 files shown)


actix-http/Cargo.toml
```diff
@@ -15,7 +15,7 @@ license = "MIT/Apache-2.0"
 edition = "2018"
 
 [package.metadata.docs.rs]
-features = ["openssl", "rustls", "failure", "compress", "secure-cookies"]
+features = ["openssl", "rustls", "failure", "compress", "secure-cookies","actix"]
```
robjtede: feature name is actors or actix ?

robjtede (changes requested):

actix-http/Cargo.toml
JohnTitor: Yeah, should be `actors`.

```suggestion
features = ["openssl", "rustls", "failure", "compress", "secure-cookies","actors"]
```

dunnock: Oh, right, sorry, it was late night fix. I have pushed with few more documentation changes.

robjtede (approved):

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1381, opened 2020-02-25
Remove limitation that prohibits segments from starting with '.'

ctm:
The path "/.well-known", can be handy to use with Let's Encrypt, due to the requirements of https://tools.ietf.org/html/rfc8555 which inherits that name from https://tools.ietf.org/html/rfc5785

This commit removes the restriction that prevents actix-files::Files from serving files from paths that contain a segment that starts with a leading '.'

Without this commit I was unsuccessful getting [acme-lib](https://github.com/algesten/acme-lib) to work, but with it (and a [commit to acme-lib](https://github.com/algesten/acme-lib/pull/12)), I was successful.

### DIFF
--- a/actix-files/src/lib.rs
+++ b/actix-files/src/lib.rs
@@ -605,8 +605,6 @@ impl PathBufWrp {
         for segment in path.split('/') {
             if segment == ".." {
                 buf.pop();
-            } else if segment.starts_with('.') {
-                return Err(UriSegmentError::BadStart('.'));
             } else if segment.starts_with('*') {
                 return Err(UriSegmentError::BadStart('*'));
             } else if segment.ends_with(':') {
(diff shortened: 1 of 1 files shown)


robjtede:
This is a good idea. I think there's precedent, however, that serving "hidden files" should be opt-in.

- https://github.com/koajs/static#options
- https://github.com/expressjs/serve-static#dotfiles
- https://api.rocket.rs/v0.4/rocket_contrib/serve/struct.Options.html#associatedconstant.DotFiles

robjtede (changes requested):
This is a good idea. I think there's precedent, however, that serving "hidden files" should be opt-in.

- https://github.com/koajs/static#options
- https://github.com/expressjs/serve-static#dotfiles
- https://api.rocket.rs/v0.4/rocket_contrib/serve/struct.Options.html#associatedconstant.DotFiles

ctm:
Makes sense. I'll get on that today.

ctm:
I was wrong, in that there's no need for any change in order to use acme-lib, as long as the `.well-known` portion is in the `path` argument to `Files::new()`. As such, I no longer have a need to support the optional returning of hidden files.

FWIW, I've already started such functionality, and I like contributing to projects use, so I'm happy to finish my attempt and update this PR. However, as @robjtede pointed out, different web servers have different levels of options and by choosing one to implement when I don't need the functionality _at all_ might lead us down the wrong path.

So, I'm perfectly happy with closing this PR with no further work, or finishing my PR (it was when I was testing it that I realized my mistake). Just let me know.

cetra3:
I'd say to close this PR if not needed.  Can always revisit it later on.

PR #1391, opened 2020-03-01
Release awc v2.0.0-alpha.1

JohnTitor:


Dowwie:
I suggest delaying this release until the source of the ssl slowness is identified..

robjtede:
As it's just an alpha and the TLS fix won't require breaking changes, I'd say it's okay to get some testing underway. We should definitely patch it before stable release though.

Dowwie:
> As it's just an alpha and the TLS fix won't require breaking changes, I'd say it's okay to get some testing underway. We should definitely patch it before stable release though.

h2 support is "broken" in the awc

I was unsatisfied with the result of the awc analysis because the default behavior was "broken".  One does not need to manually add ssl Connectors in order to use https because [awc has default settings](https://github.com/actix/actix-web/blob/master/actix-http/src/client/connector.rs#L74) that will be used when either openssl or rustls feature flags are set.  Investigating this further with @dunnock , we found that h2 processing is the root cause of the performance issues.

Fixing this won't be a breaking change but releasing it as-is isn't exactly achieving a milestone either..

JohnTitor:
a9a475d changes current `test_server` fn behavior. I'll release it as v2-alpha and I believe it should be fine but I'd like to hear any opinions (even if it's just an approval).

awc/CHANGES.md
```diff
@@ -1,9 +1,10 @@
 # Changes
 
-## [NEXT]
+## [2.0.0-alpha.1] - 2020-03-08
```
JohnTitor:
```suggestion
## [2.0.0-alpha.1] - 2020-03-11
```

Issue #1393, opened 2020-03-03
h2 support in the awc is causing serious performance issues


Dowwie:
If you were to add the following alpn protocol configuration to [this awc example](https://github.com/actix/examples/blob/master/awc_https/src/main.rs#L6), you would essentially have [the same configuration](https://github.com/actix/actix-web/blob/master/actix-http/src/client/connector.rs#L74) as what is used by default within the awc when `openssl` feature flag is used:

```rust
let mut builder = SslConnector::builder(SslMethod::tls()).unwrap();
builder.set_alpn_protos(b"\x02h2\x08http/1.1").unwrap();
```

Running the example with the protocol declaration results in a major performance problem for https requests.  The source of this issue is the use of h2.  If you remove h2 from `set_alpn_protos`, the issue is resolved.

I don't know whether h2 works correctly for http-server but it clearly has issues with the awc.  It would be useful to know whether the http-server handles h2 better.

dunnock:
Attached is a flamegraph for awc example
[flamegraph-awc.svg.zip](https://github.com/actix/actix-web/files/4281322/flamegraph-awc.svg.zip)

Dowwie:
@actix/contributors Please note this issue.  This is a test of whether awc can be maintained going forward.  This also may be affecting the http server as well.

@fafhrd91 fyi

dunnock:
It seems reqwest had a similar issue and provide some details in the issue https://github.com/hyperium/hyper/issues/1813, especially from seanmonstar:
> And curiously, with hyper's new default HTTP2 window, ~875MB/s. My only guess as to why that's faster than maxing the window is that it pushes back pressure to the executor instead of the h2::Connection prioritizer juggling chunks.

Suggest we try to increase defaults as they did, will look into it
https://github.com/hyperium/hyper/blob/master/src/proto/h2/client.rs#L24-L28

dunnock:
BTW, actix-http::h2::Service is using default H2 parameters too, not sure if they are bad (default window size 64k sounds right for REST API, but bad for blobs)
https://github.com/actix/actix-web/blob/master/actix-http/src/h2/service.rs#L311

Dowwie:
Looks like Dunnock fixed the issue.  Great job, Dunnock!  We're fortunate that the tokio community figured this problem out.

PR #1394, opened 2020-03-03
Http2 client configuration to improve performance

dunnock:
This PR should be fixing https://github.com/actix/actix-web/issues/1393. 

Since implementation is using hardcoded const need your opinion (pls see question below). Implementation is pretty much based on [hyper's resolving similar problem](https://github.com/hyperium/hyper/blob/master/src/proto/h2/client.rs#L24-L28), though it sets initial buffer for quite large 5MB which gets some pressure off a task scheduler but requires more memory.

UPD: this PR also adding possibility to configure awc and Connector:
- initial_window_size
- initial_connection_window_size
- min_http_version

I have pointed awc_https example to this branch to show that it really helps:
https://github.com/dunnock/examples/tree/h2-speed-1393
Also it can be seen from benchmarks 
https://github.com/dunnock/awc_test/tree/benches

Speed with enabled alpn protocol:
```
	let mut builder = SslConnector::builder(SslMethod::tls()).unwrap();
	builder.set_alpn_protos(b"\x02h2\x08http/1.1").unwrap(); 
```
```
Benchmarking awc_test_openssl: 
Throughput 5994288B/s
Iters 10 total 101747060 elapsed 16974ms
awc_test_openssl        time:   [1.5834 s 1.6329 s 1.6794 s]                              
```
```
Benchmarking awc_test_default
Throughput 6327160B/s
Iters 10 total 101747060 elapsed 16081ms
awc_test_default        time:   [1.5967 s 1.7657 s 2.0360 s]                              
```
compared to reqwest
```
Benchmarking reqwest:
Throughput 6333067B/s
Iters 10 total 101747060 elapsed 16066ms
reqwest                 time:   [1.6207 s 1.7328 s 1.8409 s]                     
```

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -12,7 +12,11 @@
 
 * MessageBody is not implemented for &'static [u8] anymore.
 
-* Change defaul initial window size and connection window size for HTTP2 to 5MB and 2MB respectively. This 7 times improves download speed for awc when downloading large objects.
+* Change default initial window size and connection window size for HTTP2 to 2MB and 1MB respectively to improve download speed for awc when downloading large objects.
+
+* client::Connector accepts initial_window_size and initial_connection_window_size HTTP2 configuration
+
+* client::Connector allowing to set max_http_version to limit HTTP version to be used
 
 ### Fixed
 
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -12,6 +12,8 @@
 
 * MessageBody is not implemented for &'static [u8] anymore.
 
+* Change defaul initial window size and connection window size for HTTP2 to 5MB and 2MB respectively. This 7 times improves download speed for awc when downloading large objects.
+
 ### Fixed
 
 * Allow `SameSite=None` cookies to be sent in a response.
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -1,5 +1,15 @@
 # Changes
 
+## [NEXT] - 2020-02-27
+
+### Changed
+
+* Change default initial window size and connection window size for HTTP2 to 2MB and 1MB respectively to improve download speed for awc when downloading large objects.
+
+* client::Connector accepts initial_window_size and initial_connection_window_size HTTP2 configuration
+
+* client::Connector allowing to set max_http_version to limit HTTP version to be used
+
 ## [2.0.0-alpha.1] - 2020-02-27
 
 ### Changed
@@ -11,12 +21,6 @@
 * Breaking change: trait MessageBody requires Unpin and accepting Pin<&mut Self> instead of &mut self in the poll_next().
 
 * MessageBody is not implemented for &'static [u8] anymore.
-
-* Change default initial window size and connection window size for HTTP2 to 2MB and 1MB respectively to improve download speed for awc when downloading large objects.
-
-* client::Connector accepts initial_window_size and initial_connection_window_size HTTP2 configuration
-
-* client::Connector allowing to set max_http_version to limit HTTP version to be used
 
 ### Fixed
 
--- a/actix-http/src/client/config.rs
+++ b/actix-http/src/client/config.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 // These values are taken from hyper/src/proto/h2/client.rs
 const DEFAULT_H2_CONN_WINDOW: u32 = 1024 * 1024 * 2; // 2mb
-const DEFAULT_H2_STREAM_WINDOW: u32 = 1024 * 1024 * 1; // 1mb
+const DEFAULT_H2_STREAM_WINDOW: u32 = 1024 * 1024; // 1mb
 
 /// Connector configuration
 #[derive(Clone)]
--- a/actix-http/src/client/config.rs
+++ b/actix-http/src/client/config.rs
@@ -1,0 +1,39 @@
+use std::time::Duration;
+
+// These values are taken from hyper/src/proto/h2/client.rs
+const DEFAULT_H2_CONN_WINDOW: u32 = 1024 * 1024 * 2; // 2mb
+const DEFAULT_H2_STREAM_WINDOW: u32 = 1024 * 1024 * 1; // 1mb
+
+/// Connector configuration
+#[derive(Clone)]
+pub(crate) struct ConnectorConfig {
+    pub(crate) timeout: Duration,
+    pub(crate) conn_lifetime: Duration,
+    pub(crate) conn_keep_alive: Duration,
+    pub(crate) disconnect_timeout: Option<Duration>,
+    pub(crate) limit: usize,
+    pub(crate) conn_window_size: u32,
+    pub(crate) stream_window_size: u32,
+}
+
+impl Default for ConnectorConfig {
+    fn default() -> Self {
+        Self {
+            timeout: Duration::from_secs(1),
+            conn_lifetime: Duration::from_secs(75),
+            conn_keep_alive: Duration::from_secs(15),
+            disconnect_timeout: Some(Duration::from_millis(3000)),
+            limit: 100,
+            conn_window_size: DEFAULT_H2_CONN_WINDOW,
+            stream_window_size: DEFAULT_H2_STREAM_WINDOW,
+        }
+    }
+}
+
+impl ConnectorConfig {
+    pub(crate) fn no_disconnect_timeout(&self) -> Self {
+        let mut res = self.clone();
+        res.disconnect_timeout = None;
+        res
+    }
+}
--- a/actix-http/src/client/connector.rs
+++ b/actix-http/src/client/connector.rs
@@ -50,13 +50,6 @@ type SslConnector = ();
 pub struct Connector<T, U> {
     connector: T,
     config: ConnectorConfig,
-    /*    timeout: Duration,
-    conn_lifetime: Duration,
-    conn_keep_alive: Duration,
-    disconnect_timeout: Duration,
-    limit: usize,
-    conn_window_size: u32,
-    stream_window_size: u32,*/
     #[allow(dead_code)]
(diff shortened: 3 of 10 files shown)


Dowwie:
Great job!  This seems to have fixed the 10MB download issue.

JohnTitor:
Looks good! Could you also update changelog?

dunnock:
> Looks good! Could you also update changelog?

Done

dunnock:
On my question I also looked into implementation of [adaptive window](https://github.com/hyperium/hyper/blob/master/src/proto/h2/bdp.rs) and it seems an overkill: for every poll on connection it uses Mutex to retrieve calculation. Maybe that's the reason why it's disabled by default in hyper client.

I am thinking of adding 3 parameters to the ClientBuilder as part of another PR:
- h2_initial_window_size
- h2_initial_connection_window_size
- http1_only -- one off requests would benefit of using HTTP1

what do you think?

robjtede:
More options are always beneficial.

With HTTP3 on the horizon, I'd argue to drop the `h2_` prefixes and use a more forwards compatible option for `http1_only`; something like `http_version` to declare the max "supported" protocol.

But yes, we should discuss in a seperate issue/PR.

Dowwie:
@JohnTitor thoughts on releasing this fix in the near future?

Dowwie:
useful feedback from gitter channel:

```
Grant Jennings @gajbooks 20:54:
Actually, it is not an issue, but a misconfiguration in the usage. By default, the window size is set to 65k both on the connection level and on the stream level in h2. This results in a packet being sent every time 65k of data is received, greatly increasing latency.

Grant Jennings @gajbooks 21:03
For sub-65k payloads this would not be an issue, but it heavily caps throughput for larger files. H2 does not have built-in flow control adjustment.

Grant Jennings @gajbooks 21:09
It is also not specified in the spec, and likely will never be implemented in h2 as a result.
To pick a random number, 1MB seems to give very good performance, even over .5MB. However, anything over 1MB doesn't improve performance. (Not any sort of sane statistical test, just anecdotal evidence.)
```

dunnock:
Based on this feedback I will try to test if performance same with 1MB window.
That would be easy to test when I add configurable parameters, so probably will extend this PR with configuration, should be ready soon...

dunnock:
I have tested with variable buffer size: 500kb, 1MB and 2MB. 2MB gives little benefit over 1MB (10% for big downloads), 500kb 25% slower than 1MB. Based on this I have set defaults to 1MB stream buffer and 2MB connection buffer.
```
awc_test_default(1MB)        time:   [1.9620 s 2.0073 s 2.0645 s]
awc_test_window_2mb     time:   [1.7536 s 1.7912 s 1.8353 s] 
awc_test_window_500kb   time:   [2.3543 s 2.5123 s 2.6838 s] 
reqwest                 time:   [1.8177 s 1.8597 s 1.8981 s]   
```

This is now ready for final review with parameters configuration to awc::ClientBuilder and actix-http::client::Connector.

actix-http/src/client/connector.rs
```diff
@@ -71,42 +68,52 @@ impl Connector<(), ()> {
...
             connector: default_connector(),
-            timeout: Duration::from_secs(1),
+            config: ConnectorConfig::default(),
+            /*            timeout: Duration::from_secs(1),
             conn_lifetime: Duration::from_secs(75),
             conn_keep_alive: Duration::from_secs(15),
             disconnect_timeout: Duration::from_millis(3000),
             limit: 100,
+            conn_window_size: DEFAULT_H2_CONN_WINDOW,
+            stream_window_size: DEFAULT_H2_STREAM_WINDOW,*/
```
JohnTitor: Should be removed as well.

actix-http/CHANGES.md
```diff
@@ -12,6 +12,12 @@
 
 * MessageBody is not implemented for &'static [u8] anymore.
 
+* Change default initial window size and connection window size for HTTP2 to 2MB and 1MB respectively to improve download speed for awc when downloading large objects.
+
+* client::Connector accepts initial_window_size and initial_connection_window_size HTTP2 configuration
+
+* client::Connector allowing to set max_http_version to limit HTTP version to be used
+
```
JohnTitor: You should create new section.

actix-http/src/client/connector.rs
```diff
@@ -71,42 +68,52 @@ impl Connector<(), ()> {
...
+            let mut config = ClientConfig::new();
+            config.set_protocols(&protocols);
+            config
+                .root_store
+                .add_server_trust_anchors(&actix_tls::rustls::TLS_SERVER_ROOTS);
+            SslConnector::Rustls(Arc::new(config))
+        }
+        #[cfg(not(any(feature = "openssl", feature = "rustls")))]
+        {}
+    }
```
JohnTitor: Uhm, it's readability-less... Could you just separate and set cfgs?

actix-http/src/client/config.rs
```diff
@@ -0,0 +1,39 @@
+use std::time::Duration;
+
+// These values are taken from hyper/src/proto/h2/client.rs
+const DEFAULT_H2_CONN_WINDOW: u32 = 1024 * 1024 * 2; // 2mb
+const DEFAULT_H2_STREAM_WINDOW: u32 = 1024 * 1024 * 1; // 1mb
```
JohnTitor:
```suggestion
const DEFAULT_H2_STREAM_WINDOW: u32 = 1024 * 1024; // 1mb
```

awc/src/builder.rs
```diff
@@ -140,7 +169,27 @@ impl ClientBuilder {
 
     /// Finish build process and create `Client` instance.
     pub fn finish(self) -> Client {
-        Client(Rc::new(self.config))
+        let connector = if self.connector.is_some() {
+            self.connector.unwrap()
```
JohnTitor: Why isn't it `if let`?

JohnTitor (changes requested):
Overall looks good! Left some minor points.

actix-http/src/client/connector.rs
dunnock: Yes, removed in https://github.com/actix/actix-web/pull/1394/commits/cf7022521665384db3f18ab41a9b93d5c6af70c8

JohnTitor: No, cf7022521665384db3f18ab41a9b93d5c6af70c8 removed only struct members, this is in `fn new()` :)

actix-http/CHANGES.md
dunnock: Fixed with https://github.com/actix/actix-web/pull/1394/commits/1c2f7d4e23a1cdfad181985c7196bb36dba9fd67

actix-http/src/client/connector.rs
dunnock: Good point, fixed with https://github.com/actix/actix-web/pull/1394/commits/1c2f7d4e23a1cdfad181985c7196bb36dba9fd67, thanks!

actix-http/src/client/config.rs
dunnock: Fixed with https://github.com/actix/actix-web/pull/1394/commits/1c2f7d4e23a1cdfad181985c7196bb36dba9fd67

awc/src/builder.rs
dunnock: Indeed, fixed that with https://github.com/actix/actix-web/pull/1394/commits/3a8b4d70221b7089b356777e73bf77ffb5617492, thank you!

actix-http/CHANGES.md
dunnock: Resolved conflict with https://github.com/actix/actix-web/pull/1394/commits/a19cae8472feb7172dde673194ebf430b2df74b3

actix-http/src/client/connector.rs
dunnock: Ohh, right, I have completely overlooked this one, thanks! Fixed with https://github.com/actix/actix-web/pull/1394/commits/64b3cbbe737105ed6e12560614f3068efc8fe226

dunnock:
This poining to the common build problem https://users.rust-lang.org/t/failure-derive-compilation-error/39062

JohnTitor:
Let's pin quote version in this PR, could you?

dunnock:
Ok it seems that helped!

actix-http/src/client/connector.rs
```diff
@@ -163,12 +163,41 @@ where
         self
     }
 
+    /// Maximum supported http major version
+    /// When supplied 1 both HTTP/1.0 and HTTP/1.1 will be allowed
+    pub fn max_http_version(mut self, val: u8) -> Self {
```
robjtede: I think this should take a `http::Version`.

dunnock: @robjtede indeed that's better. what about providing minimum cap at http/1.1 not growing the list below?
```
            http::Version::HTTP_09 => vec![b"http/0.9".to_vec()],
            http::Version::HTTP_10 => vec![b"http/1.0".to_vec()],
            http::Version::HTTP_11 => vec![b"http/1.1".to_vec()],
            http::Version::HTTP_2 => vec![b"h2".to_vec(), b"http/1.1".to_vec()],
```

robjtede: yeah sure keeping 1.1 as minimum sounds good

dunnock: Should be fixed now

actix-web-codegen/Cargo.toml
```diff
@@ -12,7 +12,7 @@ workspace = ".."
 proc-macro = true
 
 [dependencies]
-quote = "^1"
+quote = "=1.0.2"
```
JohnTitor: The failure has been resolved, could you revert it?

dunnock: Reverted it back

JohnTitor (approved):
Great, thanks!

PR #1398, opened 2020-03-07
Add convenience functions for testing

torepettersen:
I want to add a few convenience functions that will make it a bit easier to write tests.

One for deserializing the response body in case you need the ServiceResponse for something else than just read the body.

And the second for making it a bit easier to send the request.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -204,6 +204,54 @@ where
     bytes.freeze()
 }
 
+/// Helper function that returns a deserialized response body of a ServiceResponse.
+///
+/// ```rust
+/// use actix_web::{App, test, web, HttpResponse, http::header};
+/// use serde::{Serialize, Deserialize};
+///
+/// #[derive(Serialize, Deserialize)]
+/// pub struct Person {
+///     id: String,
+///     name: String,
+/// }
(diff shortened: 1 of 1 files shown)


src/test.rs
```diff
@@ -204,6 +204,54 @@ where
...
+///         .set_payload(payload)
+///         .send_request(&mut app)
+///         .await;
+///
+///     assert!(resp.status().is_success());
+///
+///     let result: Person = test::read_json_body(resp).await;
+/// }
+/// ```
+pub async fn read_json_body<T, B>(res: ServiceResponse<B>) -> T
```
JohnTitor: The naming rule should be the same as `read_response_json`.

```suggestion
pub async fn read_body_json<T, B>(res: ServiceResponse<B>) -> T
```

JohnTitor (commented):
Thanks for the PR! At a glance, looks good to me. Also it'd be great if you could add tests for them.
And base branch should be master when you add new API, then backport to 2.0 or else.

torepettersen:
Thanks for your quick reply. Then I will add some tests, fix the naming and make a new PR against master.

PR #1401, opened 2020-03-07
Add convenience functions for testing

torepettersen:
Adding tests and fixing remarks from https://github.com/actix/actix-web/pull/1398


JohnTitor:
Looks good! Could you also update changelog?

torepettersen:
Yeah, I can.

JohnTitor:
master should fix the build, could you rebase?

src/test.rs
```diff
@@ -1068,6 +1143,28 @@ mod tests {
         assert_eq!(&result.id, "12345");
     }
 
+    #[actix_rt::test]
+    async fn test_body_json() {
```
pfrenssen, line 1147: Since the function is called `read_body_json()` it is probably a good idea to call this test `test_read_body_json()`.

JohnTitor: It's the same as `read_response_json` and `test_response_json`, we can follow-up later.

JohnTitor (approved):
Sorry for the delay, thanks!

PR #1410, opened 2020-03-11
Release actix-web 3.0.0-alpha.1

JohnTitor:


Jonathas-Conceicao:
@JohnTitor, should this rally be a major release? Did we have any breaking changes for this?

JohnTitor:
@Jonathas-Conceicao It depends on some alpha versions, we may add additional breaking changes here/there in the future. Also, we have 2.0 branch so we can backport if needed.

Jonathas-Conceicao:
I see, thanks for the explanation.

PR #1412, opened 2020-03-12
Convert from `derive_more` to `thiserror`

otavio:
The thiserror has the advantage of implementing std::error::Error and it
integrates better with the Rust ecosystem.

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,6 @@ actix-http = "2.0.0-alpha.2"
 awc = { version = "2.0.0-alpha.1", default-features = false }
 
 bytes = "0.5.3"
-derive_more = "0.99.2"
 encoding_rs = "0.8"
 futures = "0.3.1"
 fxhash = "0.2.1"
@@ -88,6 +87,7 @@ serde = { version = "1.0", features=["derive"] }
 serde_json = "1.0"
 serde_urlencoded = "0.6.1"
 time = { version = "0.2.7", default-features = false, features = ["std"] }
+thiserror = "1.0.11"
 url = "2.1"
 open-ssl = { version="0.10", package = "openssl", optional = true }
 rust-tls = { version = "0.17.0", package = "rustls", optional = true }
--- a/actix-files/Cargo.toml
+++ b/actix-files/Cargo.toml
@@ -24,12 +24,12 @@ actix-service = "1.0.1"
 bitflags = "1"
 bytes = "0.5.3"
 futures = "0.3.1"
-derive_more = "0.99.2"
 log = "0.4"
 mime = "0.3"
 mime_guess = "2.0.1"
 percent-encoding = "2.1"
 v_htmlescape = "0.4"
+thiserror = "1.0.11"
 
 [dev-dependencies]
 actix-rt = "1.0.0"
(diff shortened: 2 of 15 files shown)


otavio:
This does the change for the `http` crate but I can continue converting more crates if desired. Please give me the project view of the project on this.

JohnTitor:
At a glance, I'm fine with that since we shouldn't lose anything and can reduce dependencies size. But let's hear other thoughts from contributors.

JohnTitor:
Also it'd be great if you could resolve conflict with rebase.

otavio:
> Also it'd be great if you could resolve conflict with rebase.

Sure, just did it. Thanks for looking at it @JohnTitor

otavio:
The _CI Windows_ failure seems to be not related.

Dowwie:
Seems like a net gain for the ecosystem.  Actix error compatibility could use improving.  I think that this needs regression testing and feedback, just to be sure.

@mitsuhiko  as a resident expert about error handling, what do you think?

bitemyapp:
`thiserror` has seemed pretty unobtrusive by design from my use of it. I use `anyhow` and `thiserror` in my projects. I'd be in favor but I'll defer to someone with more experience and expertise, such as @mitsuhiko.

tyranron:
Error deriving in `derive_more` is about to be landed, so I'm unsure whether this change is meaningful. Some tips in favor of `derive_more`:
- Supports more `std` traits. `derive(From)` is used quite oftenly in my practice. Stuff [like this](https://github.com/actix/actix-web/blob/1fa02b5f1c331f26f188627effd0b8af8a675893/actix-http/src/header/shared/httpdate.rs#L35-L39) could be replaced with a single attribute making the codebase cleaner.
- Clever [about generics](https://github.com/JelteF/derive_more/blob/master/tests/display.rs#L210-L212).
- Won't impose problems in integration with `anyhow` as the latest integrates well with any `Error`s.

As for dependencies: both `thiserror` and `derive_more` use only `proc-macro2`, `syn` and `quote`, so there is no actual dependencies reduce.

otavio:
> Error deriving in `derive_more` is about to be landed, so I'm unsure whether this change is meaningful. 

I am fine either way. Considering `derive_more` adds it, it makes sense to keep using it as it can reduce even more core.

If everyone agrees, once `derive_more` is released, I can rework this PR.

robjtede:
> As for dependencies: both thiserror and derive_more use only proc-macro2, syn and quote, so there is no actual dependencies reduce.

I think this is what kills this PR for me.

If we can wait a small while and add the better error deriving with sources/backtraces using derive_more, I'm in favor of that.

JohnTitor:
> As for dependencies: both thiserror and derive_more use only proc-macro2, syn and quote, so there is no actual dependencies reduce.

Ohhh, somehow I was confused, perhaps saw other crate page :P

otavio:
We will be waiting for `derive_more` to finish adding `Error` support, so this PR should be closed.

In meanwhile I made #1422 to allow the `Error` use for more types.

Dowwie:
derive_more 0.99.5 just released, featuring std::Error support

thanks @tyranron  !

PR #1418, opened 2020-03-15
📝 Improve the code example for JsonConfig

Stigjb:
I initially submitted a new example in https://github.com/actix/examples/pull/268, and was recommended to add the example to the docstring in this repository.

This expands the JSON error handler to set a response code and return an error message in the body. It also 

I made a couple of changes compared to https://github.com/actix/examples/pull/268. I removed the 422 response, as I discovered that Serde sometimes gives an `is_data` type error even for malformed JSON. 422 is not supposed to be used if the request is syntactically incorrect. For example, `{"a": 5` will give an `is_data` error if `"a"` is expected to be a string, and not mention the missing closing curly bracket.

I could add the `curl` examples from the README in https://github.com/actix/examples/pull/268, but I'm not quite sure where they fit in best here. 

## Questions

- Should this example be in addition to, rather than replace, the current example?

### DIFF
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -223,8 +223,6 @@ where
 ///     format!("Welcome {}!", info.username)
 /// }
 ///
-/// /// Create custom error response
-///
 /// /// Return either a 400 or 415, and include the error message from serde
 /// /// in the response body
 /// fn json_error_handler(err: error::JsonPayloadError, _req: &HttpRequest) -> error::Error {
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -208,8 +208,10 @@ where
 
 /// Json extractor configuration
 ///
+/// # Examples
+///
 /// ```rust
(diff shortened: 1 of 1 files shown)


src/types/json.rs
```diff
@@ -221,6 +223,21 @@ where
 ///     format!("Welcome {}!", info.username)
 /// }
 ///
+/// /// Create custom error response
```
JohnTitor: This doc comment doesn't explain any items, I think it can be removed since later comment explains details.

JohnTitor (commented):
One concern is that this example is a bit long to be displayed on docs.rs/rustdoc but it's certainly useful. So I'm fine to accept it.

src/types/json.rs
Stigjb: Removed in 592d465

JohnTitor (approved):
Cool, thanks!

PR #1433, opened 2020-03-28
Change NormalizePath to append trailing slash

stevemk14ebr:
For paths like /blah/ requests like /////blah now work. Previously this would 404 as the trailing slash was missing and NormalizePath only replaces multiples and doesn't add the trailing slash when missing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ### Changed
 
 * Implement `std::error::Error` for our custom errors [#1422]
+* NormalizePath middleware now appends trailing / so that routes of form /example/ respond to /example resquests.
 
 [#1422]: https://github.com/actix/actix-web/pull/1422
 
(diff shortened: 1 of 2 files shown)


robjtede:
Can you add tests so we never see regressions.

robjtede (commented):
LGTM, thanks for the contribution

otavio (dismissed):

robjtede:
Can you update changelog also?

stevemk14ebr:
done, merge it?

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 
 * Implement `std::error::Error` for our custom errors [#1422]
+* NormalizePath middleware now appends trailing / so that routes of form /example/ respond to /example resquests.
```
JohnTitor:
```suggestion
* NormalizePath middleware now appends trailing / so that routes of form /example/ respond to /example requests.
```

JohnTitor (approved):
Looks good, thanks!

PR #1449, opened 2020-04-12
actix-http: Remove `failure` support

JohnTitor:
The `failure` crate will be deprecated, see [current maintainer's post on IRLO](https://internals.rust-lang.org/t/failure-crate-maintenance/12087?u=johntitor) and [withoutboats' post](https://boats.gitlab.io/blog/post/failure-to-fehler/).
It's no rush but we should land this sooner or later.

Someone may want to support another ecosystem but I think it's out of scope here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,15 +42,13 @@ members = [
 ]
 
 [features]
-default = ["compress", "failure"]
+default = ["compress"]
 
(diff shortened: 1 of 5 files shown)


robjtede (approved):
no complaints, we can try out derive_more Error support in another PR

PR #1452, opened 2020-04-14
set data container on default service calls

### LINKED ISSUE #1450: scoped default_service cannot access data from same scope
## Expected Behavior
default_service handler should be able to access data declared in the same scope 
## Current Behavior
Runtime error "App data is not configured, to configure use App::data()" if using handler parameter or error if using request.app_data

## Possible Solution
Using service instead of default_service works, but of course doesn't allow you to handle all requests within the scope as one might need to. 

A reproduction repo can be found here:  https://github.com/lilith/actix_scoped_state
Visiting /a works (service), but /b (default_service) fails to get data. 
```rust
use std::io;
use std::sync::atomic::{AtomicUsize, Ordering};
use actix_web::{middleware, web, App, HttpRequest, HttpResponse, HttpServer};

#[derive(Debug)]
struct Counter(AtomicUsize);

async fn scope_a(counter_a: web::Data<Counter>, _req: HttpRequest) -> HttpResponse {
    let body = format!(
        "counter a (increment by visiting /a): {}",
        counter_a.0.fetch_add(1, Ordering::SeqCst),
    );
    HttpResponse::Ok().body(body)
}

async fn scope_b(counter_b: web::Data<Counter>, _req: HttpRequest) -> HttpResponse {
    let body = format!(
        "counter b (increment by visiting /b): {}",
        counter_b.0.fetch_add(1, Ordering::SeqCst),
    );
    HttpResponse::Ok().body(body)
}

#[actix_rt::main]
    std::env::set_var("RUST_LOG", "actix_web=info");
    env_logger::init();

    HttpServer::new(move || {
        let counter_a =Counter(AtomicUsize::new(0usize));
        let counter_b = Counter(AtomicUsize::new(0usize));

        App::new()
            .service(
                web::scope("/a")
                    // This works, because we use .service
                    .data(counter_a)
                    .service(web::resource("").to(scope_a)))
            .service(
                web::scope("/b")
                    // This does not, because default_service loses its data
                    .data(counter_b)
                    .default_service(web::to(scope_b)))

            // enable logger
    })
        .bind("127.0.0.1:8080")?
        .run()
        .await
}
```

## Context
I have a modular server that needs to route requests to handlers based on their prefix. default_service works, but cannot access scoped data that I need for each module 

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.42.0 (b8cedc004 2020-03-09)
* Actix Web Version: 2.0.0

robjtede:
closes #1450

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,10 +4,12 @@
 
 ### Changed
 
+* `{Resource,Scope}::default_service(f)` handlers now support app data extraction. [#1452]
 * Implement `std::error::Error` for our custom errors [#1422]
 * NormalizePath middleware now appends trailing / so that routes of form /example/ respond to /example requests.
 
 [#1422]: https://github.com/actix/actix-web/pull/1422
+[#1452]: https://github.com/actix/actix-web/pull/1452
 
 ## [3.0.0-alpha.1] - 2020-03-11
 
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -542,6 +542,9 @@ impl Service for ResourceService {
             }
         }
         if let Some(ref mut default) = self.default {
+            if let Some(ref data) = self.data {
+                req.set_data_container(data.clone());
+            }
             Either::Right(default.call(req))
         } else {
             let req = req.into_parts().0;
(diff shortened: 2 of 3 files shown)


JohnTitor (approved):
Looks good!

robjtede:
how come codecov hasnt run on this?

JohnTitor:
> how come codecov hasnt run on this?

Hmm, it's weird, uploading was successful but the bot didn't comment here: https://github.com/actix/actix-web/pull/1452/checks?check_run_id=584354405

Perhaps we should tweak codecov config, let me check it.

JohnTitor:
Oh, they showed up late to the party.

PR #1459, opened 2020-04-20
Hide continuation frames from application layer.

wayeast:
Greeting `actix_web` maintainers.  Please find attached a suggestion floated by way of a pull request.  I try to explain my reasoning in the commit message that follows, and would be interested to read your reactions...

This commit asserts that continuation frames are an implementation
detail of the websocket layer, and should remain hidden from the
application layer.  That is:
  - a codec should write only frames to the wire, and read only frames from the wire
  - when sending messages, the websocket implementation should break large
    text and binary messages into continuation frames -- the application should
    not have to be aware of this.
  - when receiving messages, the websocket implementation should reconstitute
    continuation frames into their original messages -- the application should
    not have to handle this.
  - the application should only have to send and receive complete websocket messages

Here, the reconstitution of continuation frames into their original messages is
done by the `Stream` implementation of `actix_web_actors::ws::WsStream`, by adding
a continuation frame buffer and not issuing a `Poll::Ready(Some(Ok))` result until
a complete message has been buffered.  A test in `actix_web_actors::tests::test_ws.rs`
checks this.

The breaking of large message payloads into sequential continuation frames is
done by the addition of an `actix_http::ws::frame_iters` module, which introduces
two structs `ContinuationBins` and `ContinuationTexts`.  These are iterators over
either single `Frame::Binary` or `Frame::Text` frames, if the payloads are small,
or over sequences of `Frame::Continuation`'s with appropriate Items `FirstBinary/FirstText`,
`Continue`, and `Last`.  New tests in `actix_http::ws::frame_iters` verify this
functionality.

### DIFF
--- a/actix-http/src/ws/codec.rs
+++ b/actix-http/src/ws/codec.rs
@@ -12,8 +12,6 @@ pub enum Message {
     Text(String),
     /// Binary message
     Binary(Bytes),
-    /// Continuation
-    Continuation(Item),
     /// Ping message
     Ping(Bytes),
     /// Pong message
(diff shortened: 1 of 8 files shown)


fafhrd91:
so if I open 1000 ws clients and send 1mb of data without final frame, actix-web with this patch would consume 1tb of ram? nice

PR #1463, opened 2020-04-21
fix spelling errors in doc comments

robjtede:
Run over docs with a code spell checker.

Left in the relevant `.vscode` file with allow list but can remove if anyone has strong opinions.

Future fixes:
`ConnectError::Unresolverd` => `ConnectError::Unresolved` (breaking)

### DIFF
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -6,7 +6,7 @@
 //! It is possible to add guards to *scopes*, *resources*
 //! and *routes*. Actix provide several guards by default, like various
 //! http methods, header, etc. To become a guard, type must implement `Guard`
-//! trait. Simple functions coulds guards as well.
+//! trait. Simple functions could be guards as well.
 //!
 //! Guards can not modify the request object. But it is possible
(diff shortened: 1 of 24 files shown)


src/guard.rs
```diff
@@ -100,7 +100,7 @@ pub fn Any<F: Guard + 'static>(guard: F) -> AnyGuard {
     AnyGuard(vec![Box::new(guard)])
 }
 
-/// Matches if any of supplied guards matche.
+/// Matches if any of supplied guards match.
```
JohnTitor: I think it's clearer:

```suggestion
/// Matches any of supplied guards.
```

JohnTitor (commented):
Overall looks good! I'm reluctant to introduce `.vscode/settings.json` here, do we really need this?

robjtede:
> do we really need this?

Need? No. I understand the reluctance to introduce editor specific configs.

JohnTitor (approved):
Thanks!

PR #1465, opened 2020-04-21
Small ws codec fix

wayeast:
Minor fix.


JohnTitor:
I believe it was introduced accidentally and is reasonable to fix, but it will change the behavior. Could you also update the changelog?

wayeast:
@JohnTitor added a line to the changelog

JohnTitor (approved):
Thanks!

PR #1476, opened 2020-04-28
clarify resource/scope app data overriding

robjtede:
The current docs refer to "overriding" app data when using .app_data on Resource or Scope but it is not clear that this means overriding the entire parent app data container.

I would like to address this limitation but it seems tricky and will take some time to work out; I thought a docs update was called for, regardless.

Also centralizes the DataFactoryFn type alias and moves the tests for Extensions inside a conditional `mod test`.


JohnTitor (approved):
Lgtm!

PR #1483, opened 2020-05-01
conditional test compilation [range, charset]

robjtede:
Also contains some small explanatory docs and removals of deprecated syntax.


JohnTitor (approved):
Looks good!

PR #1484, opened 2020-05-02
refactor h1 status line helper to remove unsafe usage

robjtede:
Added benchmark of old vs new vs naive approaches. On average the new method loses about 20 nanoseconds but is quite a lot better than a naive method. Like with the content length refactor, you would not expect any noticable difference.

<details>
<summary>Violin plots from benchmarks:</summary>

violin09

violin10

violin11

</details>


JohnTitor (approved):
Looks great!

PR #1485, opened 2020-05-02
Fix audit issue logging by default peer address

pando85:
By default log format include remote address that is taken from headers.
This is very easy to replace making log untrusted.

Changing default log format value `%a` to peer address we are getting
this trusted data always. Also, remote address option is maintianed and
relegated to `%{r}a` value.

Related  kanidm/kanidm#191.

### DIFF
--- a/src/middleware/logger.rs
+++ b/src/middleware/logger.rs
@@ -80,8 +80,8 @@ use crate::HttpResponse;
 ///
 /// `%{FOO}e`  os.environ['FOO']
 ///
-/// ### **\* security warning**
-///  It is calculated using
+/// # Security
+///  **\*** It is calculated using
 ///  [`ConnectionInfo::realip_remote_addr()`](../dev/struct.ConnectionInfo.html#method.realip_remote_addr)
 ///
 ///  If you use this value ensure that all requests come from trusted hosts, since it is trivial
--- a/src/middleware/logger.rs
+++ b/src/middleware/logger.rs
@@ -55,7 +55,7 @@ use crate::HttpResponse;
 ///
 /// `%%`  The percent sign
 ///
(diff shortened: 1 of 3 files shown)


robjtede:
So if I have this right, you take issue with the fact that the current %a replacement uses the derived "Client IP" which could be taken from X-Forwarded-For headers which are obviously easy to replace if your proxy is misconfigured (or you don't have a proxy and the header is being sent manually).

There's a couple bits of terminology I want to nail down, document, and be strict to use in the right places.
- Client Address
- Peer Address
- Remote Address

Could you elaborate on the above so there is no confusion in reviewing? Might be worth adding doc comments to the ConnectionInfo fields to describe when and what they are populated with.

pando85:
@robjtede thanks for your comment!
I tried to be more clear about IP meanings choosing Nginx style names.

src/middleware/logger.rs
```diff
@@ -72,12 +72,21 @@ use crate::HttpResponse;
...
 ///
+/// `%{r}a`  Real IP remote address **\***
+///
 /// `%{FOO}i`  request.headers['FOO']
 ///
 /// `%{FOO}o`  response.headers['FOO']
 ///
 /// `%{FOO}e`  os.environ['FOO']
 ///
+/// ### **\* security warning**
```
JohnTitor:
```suggestion
/// # Security
```
We usually use this header.

JohnTitor (commented):
I think this is a reasonable change, could you also update the changelog?

pando85:
@JohnTitor both done.

Must I rebase all changes in one?

pando85:
Any update?

JohnTitor (approved):
I missed this, sorry. Thanks!

PR #1486, opened 2020-05-03
allow parent data containers to be accessed from child scopes

robjtede:
`ServiceRequest::set_data_container` now adds to an array of `Rc<Extensions>`s and lookups reverse iterate to find the closest app data of that type.

Adds the `tinyvec` dependency to keep a small set of references to data containers on the stack, up to 4 data containers, which should cover most use cases optimally.

### DIFF
--- a/src/app_service.rs
+++ b/src/app_service.rs
@@ -245,7 +245,7 @@ where
             inner.path.reset();
             inner.head = head;
             inner.payload = payload;
-            inner.app_data = self.data.clone();
+            inner.app_data.push(self.data.clone());
             req
         } else {
             HttpRequest::new(
--- a/src/service.rs
+++ b/src/service.rs
@@ -217,11 +217,13 @@ impl ServiceRequest {
     /// Get an application data stored with `App::data()` method during
     /// application configuration.
     pub fn app_data<T: 'static>(&self) -> Option<Data<T>> {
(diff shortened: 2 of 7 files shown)


src/service.rs
```diff
@@ -230,9 +232,12 @@ impl ServiceRequest {
     }
 
     #[doc(hidden)]
-    /// Set new app data container
+    /// Add app data container to request's resolution set.
     pub fn set_data_container(&mut self, extensions: Rc<Extensions>) {
```
robjtede: Since this is additive, should it now be called `add_data_container`?

src/app_service.rs
```diff
@@ -245,7 +245,7 @@ where
             inner.path.reset();
             inner.head = head;
             inner.payload = payload;
-            inner.app_data = self.data.clone();
+            inner.app_data.push(self.data.clone());
```
robjtede, line 248: I'm not 100% clear on the purpose and lifecycle of the request pool so not sure if this should be `.app_data.push(...)` or `.app_data = tiny_vec![...]`.

JohnTitor: I think the current is fine.

src/service.rs
JohnTitor: +1 for renaming.

JohnTitor (approved):
Seems reasonable to me. So, I think it's fine to try this in our next alpha release.

PR #1491, opened 2020-05-06
Remove content_length from ResponseBuilder

### LINKED ISSUE #1439: Cannot force Content-Length header
## Expected Behavior
I'm able to provide my value to Content-Length header.

## Current Behavior
When I use `HttpResponseBuilder` (e.g. by `HttpResponse::Ok()`), my custom-provided content-length header value is ignored/overridden by real length of the response body.  
This is very unfortunate when implementing HEAD endpoint where the headers should be equal to GET but with an empty body.

## Steps to Reproduce
```rust
#[head("/update")]
async fn get_head() -> impl Responder {
    HttpResponse::Ok()
        .content_length(20)
        .content_type("application/octet-stream")
        .finish()
}
```
I also tried:
```rust
HttpResponse::Ok()
  .content_length(20)
  .content_type("application/octet-stream")
  .body(Body::Empty);
```
```rust
let mut response = HttpResponse::Ok()
    .content_length(20)
    .content_type("application/octet-stream")
    .body(Body::Empty); // also with finish()

response
    .headers_mut()
    .insert(header::CONTENT_LENGTH, HeaderValue::from(20));

response
```
Nothing has helped.

## Context
I'm trying to implement HEAD endpoint. According to [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13):
```
The Content-Length entity-header field indicates ... the size of the entity-body that
would have been sent had the request been a GET.
```
This is currently impossible.

## Your Environment

* Rust Version: 1.42.0 stable
* Actix Web Version: 2.0.0

omid:
Remove content_length from ResponseBuilder, since it'll be overwritten by the response body
FIXES #1439


robjtede:
Worth a mention in CHANGELOG and MIGRATION.

MIGRATION.md
```diff
@@ -5,6 +5,8 @@
   To create a cookie without a SameSite attribute, remove any calls setting same_site.
 * actix-http support for Actors messages was moved to actix-http crate and is enabled 
   with feature `actors`
+* content_length function is removed from actix-files.
```
JohnTitor:
```suggestion
* content_length function is removed from actix-http.
```

JohnTitor (approved):
So, I think this is a reasonable removal, let's try it out in alpha releases.

MIGRATION.md
JohnTitor: --^ r=me once this is fixed.

omid:
@JohnTitor thanks, solved

PR #1496, opened 2020-05-08
Actix-files will always send SizedStream

### LINKED ISSUE #1384: actix-web/actix-files doesn't put a Content-Length on range responses
It seems that if you do a range request to a [`Files`](https://docs.rs/actix-files/0.2.1/actix_files/struct.Files.html)-backed resource, you will get a `Content-Range` header back in the 206, but not a `Content-Length` back:

```
  HTTP/1.1 206 Partial Content
  transfer-encoding: chunked
  last-modified: Wed, 26 Feb 2020 18:01:01 GMT
  accept-ranges: bytes
  cache-control: max-age=100
  content-type: application/octet-stream
  content-range: bytes 0-100/3238690816
  etag: "8e6d2f:c10a8000:5e56b25d:207070e2"
  content-disposition: attachment; filename="data.db"
  date: Wed, 26 Feb 2020 23:40:44 GMT
```

While [RFC 7223: Hypertext Transfer Protocol (HTTP/1.1): Range Requests](https://tools.ietf.org/html/rfc7233) doesn't seem to explicitly forbid omitting a `Content-Length` header, all of the examples in it do include the `Content-Length` in 206 responses, e.g.:

```
     HTTP/1.1 206 Partial Content
     Date: Wed, 15 Nov 1995 06:25:24 GMT
     Last-Modified: Wed, 15 Nov 1995 04:58:08 GMT
     Content-Range: bytes 21010-47021/47022
     Content-Length: 26012
     Content-Type: image/gif

     ... 26012 bytes of partial image data ...
```
(From [section 4.1](https://tools.ietf.org/html/rfc7233#section-4.1))

In addition, a reverse proxy/CDN I'm using in front of my app also requires the Content-Length header, which is how I noticed this issue.

I *think* this is a regression in actix-web 2.

omid:
Fixes #1384

PS: I'm new to the project, please review carefully :-)


robjtede (approved):
I see the issue now. Thanks for the investigation and fix.

Lesiuk:
We should create some basic test for this issue.

shirshak55:
yep test is needed so if other breaks we will know?

omid:
It's a common area between actix-files and actix-http.

Where should this test be? Should it be inside [here](https://github.com/actix/actix-web/blob/master/tests/test_server.rs) or [here](https://github.com/actix/actix-web/blob/master/actix-files/src/lib.rs) or somewhere else?

PS: I'm almost new to the source code of this project.

robjtede:
IMO in actix-files

omid:
Looks like with the current setup it's impossible to test it in _actix-files_!

The service call is happening [here](https://github.com/actix/actix-web/blob/master/src/test.rs#L142) but in HTTP1 implementation it's [here](https://github.com/actix/actix-web/blob/master/actix-http/src/h1/dispatcher.rs#L508). As far as I can see, in general, test flow is not similar to normal flow.

The only solution _can_ be to write the tests in the "test_server". Are you fine with this?

robjtede:
I guess? Isn't the test to check the response headers from a request with specified range?

omid:
Yes, but the content_length header will be added not in actix-files, but in actix-http.

robjtede:
I’m forgetting there’s file access to consider. In test_server is fine.

omid:
I cannot reproduce the CI errors locally!
Can somebody help me to understand them plz?

robjtede:
Yeah dw about the CI errors, it's because we merged a rust 1.40 feature in recently and the CI hasnt had the minimum rust version test bumped yet.

robjtede:
I take it back, this test is more appropriate to be part of actix-files given there was a similar test already there with commented out checks. It works fine running `cargo test --package actix-files` from root of actix-web when using `.index_file("tests/test.binary")` (seems that --package changes the working directory or something).

omid:
> I take it back, this test is more appropriate to be part of actix-files given there was a similar test already there with commented out checks. It works fine running `cargo test --package actix-files` from root of actix-web when using `.index_file("tests/test.binary")` (seems that --package changes the working directory or something).

@robjtede  I couldn't make it work with the current setup. As I said in [this comment](https://github.com/actix/actix-web/pull/1496#issuecomment-627522251), as far as I can see, service call doesn't happen via actix-http. Content-length header applied in actix-http and there is NO content-length header inside actix-files.

robjtede:
It does pass through actix-http. The evidence needed is that this test passes with new code and fails with old code. That's all we need to check.

omid:
I don't understand what you mean.
Please check this file first: https://github.com/actix/actix-web/blob/master/actix-http/src/h1/dispatcher.rs#L508

As you can see, it's calling the service at this line... but when you run tests inside actix-files, the call stack is like the following image, as you can see, it doesn't go through actix-http:
Screenshot from 2020-05-12 23-55-10

In the other hand, if I put the tests inside "test_server", the call stack will be like the following image:
Screenshot from 2020-05-12 23-59-14

Please pay attention to the existence of "actix-http" in the second image.

robjtede:
Ok I don’t understand why it matters. This patch modifies actix-files and provides a test that the header exists in a range call.

Tests are black boxes and independent of the underlying implementation. The test passes with new code and fails with old code. Isn’t that what we want?

omid:
Because actix-files will not set the Content-Length (CL) header, so the only test I can write is to check that CL doesn't exist.

Actix-http is responsible to add CL header based on some situations. For example, if a response is a Stream, actix-http will NOT ass CL, but if it is a SizedStream, then it'll add CL.

To be able to test CL in requests to actix-files, I have to test them in test_server AFAICS.

robjtede:
I stand corrected. Turns out that `test::init_service` is the cause (since it only allows testing of "services"). We can still write the tests in actix-files however, using the `test::start` method, similar to the tests you have in this PR (since it uses the full stack and an HTTP client to make requests to the given service). I maintain they should live there and not in test_server.

I opened a draft PR with my version of the tests here: https://github.com/actix/actix-web/pull/1505. Take a look, I think we have both leared something from this. After changes, we should merge _this_ PR since you've put effort and research into it :)

omid:
So cool, thanks, was not aware of `test::start` so much!
I'll use your tests inside the code, just replace some of "get" reqeusts to "head".

actix-files/src/lib.rs
```diff
@@ -952,137 +952,72 @@ mod tests {
...
-
-        let bytes = test::read_body(response).await;
+        // Check file contents
+        let bytes = response.body().await.unwrap();
         let data = Bytes::from(fs::read("tests/test.binary").unwrap());
         assert_eq!(bytes, data);
     }
 
-    #[actix_rt::test]
-    async fn test_head_content_length_headers() {
```
robjtede, line 1063: I think leave the head tests in their own method. I also think theres a separate issue with HEAD requests hence the previous TODO in this method. (It should be addressed in a seperate PR)

robjtede (approved):
Edit: wrong button

robjtede (changes requested):
- Leave HEAD test in own method
- Add CHANGELOG entry

actix-files/src/lib.rs
omid: NP, I'll put two types of tests, one for GET another simple one for HEAD.
The HEAD problem/todo should be solved with this PR. But for sure, we shouldn't use init_service also in this case.

robjtede (approved):

JohnTitor (approved):
Looks good! Thanks both, for investigating and fixing.

PR #1507, opened 2020-05-13
replace mem::replace with mem::take where possible

robjtede:
thanks to bump to 1.40


therealprof (approved):
Nice cleanup!

JohnTitor (approved):
Looks good, nice catch!

PR #1509, opened 2020-05-14
Provide impl<T> From<Arc<T>> for Data<T>

### LINKED ISSUE #1300: impl From<Arc<T>> for web::Data<T>
actix-web is just part of my application. The data I am trying to store comes from the upper level, which passes the value around as an `Arc<T>`. I could only store this value as `web::Data<Arc<T>>`, but this would become double indirection.

svenallers:
Resolves #1300


robjtede (approved):
Thanks 👍

robjtede:
Can you add a CHANGELOG entry?

robjtede (approved):

PR #1519, opened 2020-05-19
bench: Reduce sample count to remove warning

JohnTitor:


robjtede:
It's possible to pass command line args to achieve this which I think would be a better approach requiring no code changes.

For the two packages, -web and -http, with benches need to add the following to the Cargo.toml:

```toml
[lib]
bench = false
```

and we'd need to use the equivalent of this to run the tests

```sh
cargo bench -- --sample-size=15
cargo bench --package actix-http -- --sample-size=15
```

source: https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options

edit: thinking about it, maybe not running the -http benches on CI is okay

JohnTitor:
I modified code since I've seen this error on my local and CI:
```
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 99.8s or reduce sample count to 10.
```

Have you ever seen this?

> thinking about it, maybe not running the -http benches on CI is okay

IIUC, we don't run actix-http benchmarks since we run `cargo bench` without `--workspace` flag on CI. Given from some CI errors, we shouldn't run the async_service_direct, i.e., service benchmark. I'll propose another PR to account that.

robjtede:
true, still seems a shame to hardcode the sample count in the benchmark which menas that locally you'd need to change it back to get the more statically accurate result.

I think it would be enough for CI just to run:
```
cargo bench --bench server -- --sample-size=15
```

JohnTitor:
Hm, fair enough. We may be able to increase target time by default instead but anyway it isn't a critical problem and the way you suggested is enough here, updated.

robjtede (approved):
looks good

PR #1520, opened 2020-05-19
log message as error instead of debug

octave99:
from_request implementation for Json should log the message as error instead of debug during error


fafhrd91:
This change allows malicious behavior. User can flood server with error log

PR #1530, opened 2020-05-21
Use `itoa` in the content-length helper

NickKolpinskiy:
In #1344 an unsafe implementation was replaced with a safe one, however `actix-http` already depends on `itoa` indirectly (see dependencies of `http`, `serde_json` or `serde_urlencoded`), it makes sense to make use of that crate instead of writing (and then maintaining) your own implementation.

It's also only slightly slower than the original function on inputs larger than 1M:

violin

Benchmarked on a 1.3 Ghz dual-core i5 from 2013.


robjtede:
From the looks of things, the original implementation is identical to the implementation used in `itoa`, that being based on unsafe blocks with calls to `mem::copy_nonoverlapping` and `mem::transmute`. Not passing any judgement for now but interesting to note.

NickKolpinskiy:
I think I should've used [`itoa::Buffer`](https://docs.rs/itoa/0.4.5/itoa/struct.Buffer.html) directly instead of io::Write, to avoid [an unnecessary copy](https://docs.rs/itoa/0.4.5/src/itoa/lib.rs.html#72-73).

> the original implementation is identical to the implementation used in `itoa`

Which, in turn, is pretty much what [the std implementation](https://doc.rust-lang.org/1.4.0/src/core/fmt/num.rs.html#246-314) is, minus the formatting code.

NickKolpinskiy:
Removed the intermediate buffer:

violin_no_intermediate_buffer

robjtede:
> pretty much what the std implementation is

Good point, and I'm much happier relying on this crate without really understanding its implementation given it has tests and used by a bunch of other crates.

cetra3 (approved):
I think it's a good change considering that crates already depend on `itoa`.

JohnTitor (approved):
Awesome! Since `itoa` is a well-known crate and perf looks good, I'd accept this change.

PR #1532, opened 2020-05-22
Add links to Gitter on the issue template

JohnTitor:
I don't know if there's a big difference between these channels, but it's better to have them both listed.

r? @robjtede since you're active on Gitter than me.


robjtede (approved):
Yeah, really good to have these listed alongside the issue template prompt.

We don't mind people posting in the "wrong" channel (since the `actix` chat is linked from the actix-web readme) so both is good.

PR #1537, opened 2020-05-25
framed: Bump up to 0.4.0-alpha.1

JohnTitor:


fafhrd91:
I think this package could be deleted. I don’t think if anybody uses it

JohnTitor:
> I think this package could be deleted. I don’t think if anybody uses it

Indeed we don't have "Dependent crates" for this and the download count is low. Okay, I'd make a release 0.3.1 with a deprecation note.

PR #1539, opened 2020-05-25
framed: Bump up to 0.3.1 with deprecation notice

JohnTitor:


JohnTitor:
cc @fafhrd91

fafhrd91:
I added this crate because we used it internally, but now we use different crates. I don't think there are anyone who uses it

JohnTitor:
Hm, got it, tweaked the wording.

PR #1548, opened 2020-05-31
Fix NormalizePath trailing slash behavior

### LINKED ISSUE #1536: NormalizePath inconsistent trailing slash
The `NormalizePath` middleware is inconsistent in the `3.0.0-alpha.3` version. Using a route `/user/{id}/` and the `NormalizePath` middleware, the trailing slash is only added when the request path contains duplicated slashes somewhere else.

Minimal example to reproduce:

```
#[actix_rt::main]
    HttpServer::new(move || {
        App::new()
            .wrap(middleware::NormalizePath)
            .route("/user/{id}/", web::get().to(| | HttpResponse::Ok().body("hello world")))
        })
    .bind("127.0.0.1:8088")?
    .run()
    .await
}
```

## Expected Behavior
I expect the trailing slash to be added regardless of the amount of duplicate slashes in the rest of the path.
| Request Path  | Expected Response |
| ------------- | ------------- |
| /user/1 | 200  |
| /user//1  | 200  |

## Current Behavior

The trailing slash is added only if there are duplicate slashes somewhere in the path.
| Request Path  | Actual Response |
| ------------- | ------------- |
| /user/1 | 404 |
| /user//1  | 200  |
## Possible Solution

The current implementation of NormalizePath [looks like this:](https://github.com/actix/actix-web/blob/b9e268e95fb433aa69f3ac08cbbebd820afe0e05/src/middleware/normalize.rs#L78)
```
...
        // always add trailing slash, might be an extra one
        let path = head.uri.path().to_string() + "/";
        let original_len = path.len();

        // normalize multiple /'s to one /
        let path = self.merge_slash.replace_all(&path, "/");

        if original_len != path.len() {
...
```
The path is only updated if `original_len != path.len()`. But `original_len` is the length with the trailing slash added. If adding the trailing slash is the only change, then the URI will not be updated.

A possible solution could be to calculate the `original_len` before adding the extra slash, like this:
```
...
        let path = head.uri.path().to_string();
        let original_len = path.len();

        // always add trailing slash, might be an extra one
        let path = path + "/";

        // normalize multiple /'s to one /
        let path = self.merge_slash.replace_all(&path, "/");

        if original_len != path.len() {
...
```

1. Run the example code
2. Make a GET request to `127.0.0.1:8088/user/1`

## Context

I wish to use the NormalizePath middleware to append a trailing slash to the route regardless of the rest of the path.

## Your Environment
* Rust Version (I.e, output of `rustc -V`): `1.43.1 (8d69840ab 2020-05-04)`
* Actix Web Version: `3.0.0-alpha.3`

Fabianstelmach:
Fixes #1536

NormalizePath would append a trailing slash if and only if it would
merge multiple slashes into one. For example, the following path would
not get a trailing slash appended:

```"/sample/path"```

On the other hand, this path would get a trailing slash appended:

```"/sample//path"```

After this commit, the trailing slash is appended for both the paths.

Also, the documentation for NormalizePath is updated to inform the user of the trailing slash behavior.

### DIFF
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -16,6 +16,7 @@ use crate::Error;
 /// Performs following:
 ///
 /// - Merges multiple slashes into one.
+/// - Appends a trailing slash if one is not present.
 ///
 /// ```rust
 /// use actix_web::{web, http, middleware, App, HttpResponse};
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -144,6 +144,10 @@ mod tests {
         let req3 = TestRequest::with_uri("//v1//////something").to_request();
         let res3 = call_service(&mut app, req3).await;
         assert!(res3.status().is_success());
+
+        let req4 = TestRequest::with_uri("/v1//something").to_request();
+        let res4 = call_service(&mut app, req4).await;
+        assert!(res4.status().is_success());
     }
 
     #[actix_rt::test]
@@ -169,6 +173,10 @@ mod tests {
         let req3 = TestRequest::with_uri("//v1///something").to_srv_request();
         let res3 = normalize.call(req3).await.unwrap();
         assert!(res3.status().is_success());
+
+        let req4 = TestRequest::with_uri("/v1//something").to_srv_request();
+        let res4 = normalize.call(req4).await.unwrap();
+        assert!(res4.status().is_success());
     }
 
     #[actix_rt::test]
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -75,14 +75,26 @@ where
     fn call(&mut self, mut req: ServiceRequest) -> Self::Future {
         let head = req.head_mut();
 
+        let original_path = head.uri.path();
+
         // always add trailing slash, might be an extra one
-        let path = head.uri.path().to_string() + "/";
-        let original_len = path.len();
+        let path = original_path.to_string() + "/";
 
         // normalize multiple /'s to one /
         let path = self.merge_slash.replace_all(&path, "/");
 
-        if original_len != path.len() {
+        // Check whether the path has been changed
+        //
+        // This check was previously implemented as string length comparison
+        //
+        // That approach fails when a trailing slash is added,
+        // and a duplicate slash is removed,
+        // since the length of the strings remains the same
+        // 
+        // For example, the path "/v1//s" will be normalized to "/v1/s/"
+        // Both of the paths have the same length, 
+        // so the change can not be deduced from the length comparison
+        if path != original_path {
             let mut parts = head.uri.clone().into_parts();
             let pq = parts.path_and_query.as_ref().unwrap();
 
@@ -178,11 +190,11 @@ mod tests {
     }
 
     #[actix_rt::test]
-    async fn should_normalize_nothing_notrail() {
+    async fn should_normalize_notrail() {
         const URI: &str = "/v1/something";
 
         let srv = |req: ServiceRequest| {
-            assert_eq!(URI, req.path());
+            assert_eq!(URI.to_string() + "/", req.path());
             ok(req.into_response(HttpResponse::Ok().finish()))
         };
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 * Fix actix_http::h1::dispatcher so it returns when HW_BUFFER_SIZE is reached. Should reduce peak memory consumption during large uploads. [#1550]
 
+### Fixed
+
+* `NormalizePath` improved consistency when path needs slashes added _and_ removed.
+
 ## [3.0.0-alpha.3] - 2020-05-21
 
 ### Added

### CODE REVIEW

**Comment by stevemk14ebr** [comment]
There's no need for the let append_trailing_slash = !path.ends_with("/"); or the branching. Just move the  let original_len = path.len(); check before the append like originally discussed in the issue.

**Comment by Fabianstelmach** [comment]
Comparing original length and the new path length is not enough in some cases. As an example, take the following path of length 5: ```/s//p``` This path would get normalized to: ```/s/p/```

But this path also has length of 5. Therefore, comparing the length of those paths is not enough to determine whether the path has changed. I do not see how this can be solved using just path length comparison, but maybe I'm missing something?

In my eyes, there are two possible approaches for this:

1) Check whether a trailing slash has been appended OR whether any slashes have been removed with the `replace_all` method.

2) Compare the old path and new path using string comparison directly.

In this PR, the first method has been implemented. If preferred, I could implement the second method instead. 

Please let me know whether either of those solutions is correct.

**Comment by stevemk14ebr** [comment]
Ah I understand, with the case you just mentioned this seems fine.

**Comment by robjtede** [comment]
I'd be in favor of documenting the whys and edge cases covered in code comments.

**Comment by cetra3** [comment]
My understanding is the intent is to see whether the path has had any changes to it by doing a length check, which fails in certain scenarios.  I think `replace_all` will return `Cow` so it should be enough to check if it's borrowed (I.e, no changes). I.e,

```rust
match self.merge_slash.replace_all(&path, "/")
  Cow::Borrowed(_var) => self.service.call(req),
  Cow::Owned(new_path) => {
  ....// we have a new path here
} 
```

You should also add tests for the example you have provided

Fabianstelmach:
> 
> 
> I'd be in favor of documenting the whys and edge cases covered in code comments.

Agreed, this is now covered in the code.



> 
> 
> My understanding is the intent is to see whether the path has had any changes to it by doing a length check, which fails in certain scenarios. I think `replace_all` will return `Cow` so it should be enough to check if it's borrowed (I.e, no changes). I.e,
> 
> ```rust
> match self.merge_slash.replace_all(&path, "/")
>   Cow::Borrowed(_var) => self.service.call(req),
>   Cow::Owned(new_path) => {
>   ....// we have a new path here
> } 
> ```

This works well for determining whether any duplicate slash has been dropped, but it does not help determine whether a trailing slash has been appended, as the latter does not use regex.

> You should also add tests for the example you have provided

Agreed, and implemented.

I have changed the implementation to use string comparison instead. The code is much simpler now. Please let me know if this solution is appropriate.

cetra3:
I still can't see a test for the scenario you're trying to test against, although the implementation doesn't matter now, but I would still like to see something to guard against it if the implementation changes.

src/middleware/normalize.rs
```diff
@@ -178,11 +199,11 @@ mod tests {
     }
 
     #[actix_rt::test]
-    async fn should_normalize_nothing_notrail() {
+    async fn should_normalize_notrail() {
         const URI: &str = "/v1/something";
 
         let srv = |req: ServiceRequest| {
-            assert_eq!(URI, req.path());
+            assert_eq!(URI.to_string() + "/", req.path());
```
Fabianstelmach, line 206: Testing the case mentioned in the issue #1536, a simple path gets trailing slash appended.

src/middleware/normalize.rs
```diff
@@ -156,6 +173,10 @@ mod tests {
         let req3 = TestRequest::with_uri("//v1///something").to_srv_request();
         let res3 = normalize.call(req3).await.unwrap();
         assert!(res3.status().is_success());
+
+        let req4 = TestRequest::with_uri("/v1//something").to_srv_request();
+        let res4 = normalize.call(req4).await.unwrap();
+        assert!(res4.status().is_success());
```
Fabianstelmach, line 179: This test case would not work with the previous implementation using just string length. The path `/v1//something` will get normalized to `/v1/something/` and so its length won't change.

src/middleware/normalize.rs
```diff
@@ -131,6 +144,10 @@ mod tests {
         let req3 = TestRequest::with_uri("//v1//////something").to_request();
         let res3 = call_service(&mut app, req3).await;
         assert!(res3.status().is_success());
+
+        let req4 = TestRequest::with_uri("/v1//something").to_request();
+        let res4 = call_service(&mut app, req4).await;
+        assert!(res4.status().is_success());
```
Fabianstelmach, line 150: This test case would not work with the previous implementation using just string length. The path `/v1//something` will get normalized to `/v1/something/` and so its length won't change.

Fabianstelmach (commented):
I have updated an old test case and added a new one, based on the issue #1548 and the discussion here. I highlighted those tests in the review comments. 

Do those tests not suffice? If so, what kind of tests would you suggest to add?

cetra3 (approved):
Approved.

robjtede (approved):

robjtede:
Just a changelog entry to add now.

PR #1550, opened 2020-06-01
Enforce HW_BUFFER_SIZE inside h1::dispatcher

### LINKED ISSUE #1551: Http1 request bodies allocate extremely large buffers
I filled a PR #1550 here but it looks like there is a bug in the dispatcher code for http1 that can cause large request bodies to grow the buffer beyond its intended capacity, `HW_BUFFER_SIZE`.

## Expected Behavior
`actix_web::web::Payload` when consumed as a Stream of bytes should have some fixed upper bound of memory usage.

## Current Behavior
Large requests, even when using the Stream api for `actix_web::web::Payload` seem consume cause the server's memory footprint to grow considerably and memory possibly seems to be live leaked. 

## Possible Solution
See #1550 this seems to fix the problem and the root cause seems fairly obvious, assuming the behavior of this bug is indeed a bug.

Don't have a minimal example yet but I can build one. I can build a minimal one and post here with instruments allocation traces if needed.

1. Create actix view that consumes a stream of bytes
2. Repeatedly send large 0.5gb requests from memory via reqwest to it
3. note the large memory footprint

## Context
I think the network speed/TCP window size and http client library probably play a big factor here since a large number of bytes need to be available to be read at once for this bug to trigger. It originally was seen in GKE/google cloud but I have produced with a slightly more minimal repro on my laptop using the loopback interface.

Without this fix streaming in large uploads to our web service we are building will not be possible/would have to bypass actix.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.41.0 (5e1a79984 2020-01-27)
* Actix Web Version: 2.0 (synced to the 2.0 branch currently)
* OS: MacOs 10.15.4 and linux on GKE/gcloud

robo-corg:
This fixes what I presume to be a bug in the `actix-http` http1 protocol body reading where a large amount of bytes being available to be read from the underlying TCP socket can cause `read_buf` to be resized far beyond `HW_BUFFER_SIZE`.

The end result before this patch was `read_buf` more or less becoming the size of the entire payload--defeating any streaming the user might be trying to in their view and causing a very large memory foot print. In my test environment a 512mb file could easily cause the server process to balloon out to 10gb fairly quickly. After this fix it stays at a nice constant.

From what I can tell this was probably caused by a misunderstanding of how `BufMut` is implemented for a `BytesMut`. I only discovered it while debugging that the `BufMut` impl for `BytesMut` effectively does not have an upper bound in side because it can just resize the underlying BytesMut. See https://docs.rs/bytes/0.5.4/src/bytes/bytes_mut.rs.html#965 This is why I have added a `MaxBuf`.

Not super happy with the MaxBuf wrapper but it seemed like continuing to use `BufMut` +`AsyncRead::poll_read_buf` is the right call--otherwise we would have to introduce a bunch of unsafe buffering code ourselves that probably would end up being just as messy.

closes #1551

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -861,7 +861,14 @@ where
     T: AsyncRead + Unpin,
 {
     let mut read_some = false;
+
     loop {
+        // If buf is full return but do not disconnect since
+        // there is more reading to be done
+        if buf.len() >= HW_BUFFER_SIZE {
+            return Ok(Some(false));
+        }
+
         let remaining = buf.capacity() - buf.len();
         if remaining < LW_BUFFER_SIZE {
             buf.reserve(HW_BUFFER_SIZE - remaining);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## [Unreleased]
+
+### Changed
+
+* Fix actix_http::h1::dispatcher so it returns when HW_BUFFER_SIZE is reached. Should reduce peak memory consumption during large uploads. [#1550]
+
 ## [3.0.0-alpha.3] - 2020-05-21
 
 ### Added


robjtede:
How come this is based on 2.0 and not master?

robo-corg:
No reason other than that is what I needed the fix for. I will rebase!

robjtede:
Ok we can consider backporting the fix when it's in master.

fafhrd91:
I hope you guys understand that when you read from socket you actually read from kernel memory.

Dowwie:
> I hope you guys understand that when you read from socket you actually read from kernel memory.

with this given, what was missed?

fafhrd91:
this patch fixes side-effect of the bug

Dowwie:
Thanks, Nik!  @fafhrd91

robjtede:
Sounds like you're saying there's a deeper issue here Nikolay? What might that be? Would save us some time if its known to you.

fafhrd91:
I don’t know what is the problem, I don’t use actix anymore. It is community task to find it.

Dowwie:
@JohnTitor fyi

Dowwie:
@robo-corg thanks for fixing this important issue

Dowwie:
@actix/contributors Note this serious bug.  Also, the current fix adds unsafe, so would like to hear from others about concerns.  Large uploads are showing this behavior.   This issue requires escalating an emergency release once all are comfortable with the solution.

robjtede:
I am investigating the root cause of this bug.

vmalloc:
I'm not sure we're talking about the same thing, but if this is about backpressure when streaming payload to other destinations (e.g. streaming user payload to awc connections) I also think this is a serious issue. I tried implementing S3 uploads from an Actix backend and it ended up exposing my service to DoS when S3 was performing poorly.

Dowwie:
> I'm not sure we're talking about the same thing, but if this is about backpressure when streaming payload to other destinations (e.g. streaming user payload to awc connections) I also think this is a serious issue. I tried implementing S3 uploads from an Actix backend and it ended up exposing my service to DoS when S3 was performing poorly.

@vmalloc could you replicate your scenario using the patch provided here and report?   you may have to simulate a struggling s3 api

robo-corg:
It is possible a slow consumer might exacerbate this. For my own repro I just ran for_each on the stream of bytes. The problem that I saw was that `read_available` would would call `read` with the buffer as a `BufMut`. The `BufMut` for `BytesMut` can be resized arbitrarily which is presumably what the `AsyncRead` was doing. @fafhrd91 is probably right that this shouldn't be that big if we are reading whatever is available without blocking and so there is probably a bug at a lower level.

I was seeing this on loopback on a mac, and inside k8s on GKE. I might be able to build a minimal repro today if that is helpful. If we have minimal repro that will probably make debugging the underlying issue a lot easier.

robo-corg:
Tested this under macos with instruments: https://github.com/robo-corg/actix-repro I might have time later today to investigate its behavior in linux. Running it in docker with top didn't seem to show it using very much memory.

On my machine it peaks at around 2gb and has multiple 1gb allocations coming from dispatch:

Screenshot 2020-06-08 13 35 33

After the patch with instruments peaks around 30mb no allocation by actix larger than 128kb (besides threads):
Screenshot 2020-06-08 13 45 50

actix-http/src/h1/dispatcher.rs
```diff
@@ -867,13 +890,21 @@ where
...
     let mut read_some = false;
+
     loop {
         let remaining = buf.capacity() - buf.len();
+
+        // If buf is full return but do not disconnect since
+        // there is more reading to be done
+        if buf.len() >= HW_BUFFER_SIZE {
+            return Ok(Some(false));
+        }
```
robjtede, line 870: In testing I've found that this check is all that's needed to keep memory usage in check; without the MaxBuf struct.

robjtede:
The problem is the loop in `read_available` which will allocate as much memory as it needs as long as the socket keeps polling as ready and providing data, not passing the buffer to the next step of the pipeline until it `read` returns a `Ok(Some(false))`. Hard to say exactly what client conditions are necessary to trigger this but i've been able to reproduce very consitently using httpie but not curl. 

```
dd if=/dev/urandom of=1.dat bs=1g count=1
curl -iv 127.0.0.1:8080 --data-binary @- < 1.dat
http POST 127.0.0.1:8080 < 1.dat
```

~Worth noting that this patch has significant performance regression on the test case of posting very large bodies; full call times are about 8x on my machine.~ Might have been the copious amounts of extra logging I had enabled. Gathering some more concrete performance metrics now.

cetra3:
I agree with @robjtede that if we don't need to introduce a new `MaxBuf` (and the *unsafe* that goes along with it) we shouldn't.

But maybe 32k is not appropriate a value, and should be bumped up to something a bit larger?  Or even more of a refactor of the code here.

Is there also a way to easily test this behaviour that we can ensure there are no regressions in RAM usage?

robjtede:
Some (admitedly non-exhaustive) performance numbers, appears to not be too harmful.


time (seconds) for 1GB upload on macos localhost | original code | patched (32k buffer) | patched (128k buffer)
-- | -- | -- | --
httpie fresh | 1.951 | 1.951 | 4.087
httpie 2 | 1.981 | 3.994 | 1.575
httpie 3 | 1.995 | 1.959 | 1.785
  |   |   |  
curl fresh | 2.129 | 2.407 | 1.788
curl 2 | 1.717 | 1.646 | 2.185
curl 3 | 1.900 | 1.698 | 1.638


I would be okay to go ahead with the simplified patch (with a small tweak) but also agree that a larger ongoing refactor of the h1 dispatcher is needed. It may also relate to some keep alive issues that have been reported.

Dowwie:
@robo-corg seems like all but the code that @robjtede identified could be cleared from the PR

robo-corg:
I can update the patch to remove the MaxBuf. Probably run it through instruments and again and to verify.

robo-corg:
You are right, seems to work exactly the same with MaxBuf gone so I removed it from this PR. It is a little disconcerting that the AsyncRead + BufMut contract allows it to enlarge the buffer but I guess that is fine. Same instruments output with no allocations going beyond 128k.

Do you want HW_BUFFER_SIZE modified? I could imagine that might be a nice thing to have as something configurable but maybe that should wait.

robjtede (approved):

robjtede:
Think we'll leave the buffer size alone for now, it didn't appear to have much of any effect in the large file tests.

robjtede:
Hey @robo-corg, can you add in a changelog entry and I'll get this merged.

robo-corg:
I added an entry to `CHANGES.md`. I bump the version too if you want.

robjtede (approved):

robjtede (approved):

adelbertc:
What is the protocol for getting this backported to actix-web 2.x (or actix-http 1.x)?

robjtede:
@adelbertc we'll look at backporting after the current buffer of changes have been flushed to beta releases.

robjtede:
@adelbertc just to follow up, we've decided not to backport any changes (for a multitude of reasons) but upgrading to v3 should be easy enough. There's a migration guide in the root of the repo that should guide you with anything that needs changing.

PR #1553, opened 2020-06-05
Remove uses of pin_project::project attribute

taiki-e:
pin-project will deprecate the project attribute due to some unfixable
limitations.

Refs: https://github.com/taiki-e/pin-project/issues/225

*Although the deprecation of the project attribute has not been released yet, this PR is submitted in advance to avoid the CI from being broken by deprecated warnings.*


JohnTitor (approved):
Thanks for the heads up!

PR #1558, opened 2020-06-09
migrate cookie handling to cookie crate

### LINKED ISSUE #1556: Copying the cookie crate without attribution violates its licenses
Hello!

It appears that the entirety of the [`cookie`](https://github.com/SergioBenitez/cookie-rs) crate, with the exception of its license(s), has been [copied into the `actix-http` crate](https://github.com/actix/actix-web/tree/4f9a1ac3b761dd2b8aa31e086ad84d8455e7893b/actix-http/src/cookie) . This is in violation of the two license options (MIT and Apache 2.0) at your disposal, both of which require attribution. Furthermore, it looks like [upstream](https://github.com/SergioBenitez/cookie-rs/commit/50ef66357921673d8292aeecd344d2645ba65fe4) [commits](https://github.com/SergioBenitez/cookie-rs/commit/451928eada489322e4c49f9c38dd2f8604e5ec50) are simply being [copied](https://github.com/actix/actix-web/commit/63864ecf9e3e6b8dbf883d3030267a525fb50a18#diff-dc185b5b55e47631d97ea058d6e405b5) [here](https://github.com/actix/actix-web/commit/d5ceae20740837bddef85053c3354452b437877a#diff-dc185b5b55e47631d97ea058d6e405b5), again without attribution. And, while this latter point doesn't further violate any licenses, I would consider it poor practice and poor open-source citizenship.

Please either attribute the `cookie` code as per either the MIT license or the Apache 2.0 license or remove the code from your repository. Better yet: why not use `cookie` as a dependency? You'd benefit from upstream improvements automatically, and any improvements you've made here could potentially benefit the community at large. In true open-source fashion: it's a win-win. Please do consider this latter option.

Thank you!

robjtede:
Bonus: removes dependency on `ring`.

closes #1556


cetra3 (approved):
Looks good to me

awc/src/ws.rs
```diff
@@ -246,16 +243,18 @@ impl WebsocketsRequest {
...
         if let Some(ref mut jar) = self.cookies {
-            let mut cookie = String::new();
-            for c in jar.delta() {
-                let name = percent_encode(c.name().as_bytes(), USERINFO);
-                let value = percent_encode(c.value().as_bytes(), USERINFO);
-                let _ = write!(&mut cookie, "; {}={}", name, value);
+            let cookie: String = jar
+                .delta()
+                // ensure only name=value is written to cookie header
+                .map(|c| Cookie::new(c.name(), c.value()).encoded().to_string())
```
therealprof, line 249: I don't understand why a new cookie needs to created here with the same information as the existing one. Can't we just render the cookie directly?

robjtede: Because there's no distinction between a plain cookie and one with attributes the cookie header would get printed with the attributes. You can see that the previous code manually set up printing for name=value as well.

therealprof: Sounds odd to me. I thought the whole point of the cookies crate was to make that seamless. https://docs.rs/cookie/0.14.1/src/cookie/lib.rs.html#1065 looks like it would do the right thing to my eyes but maybe I'm missing something.

robjtede: The key is the call to `self.0.fmt_parameters(f)`. So if for whatever reason you set a .domain on a cookie you pass to the client the cookie header will end up with `name=value; domain=example.com` which when parsed is actually 2 cookies.

That's to say, the cookie crate makes it seamless to generate set-cookie headers as a server but a little more nuanced to generate cookie headers for clients.

robjtede: @SergioBenitez is my understanding correct?

therealprof (commented):
Does this actually remove the cookie code? I don't see any relevant deletions...

robjtede:
> Does this actually remove the cookie code? I don't see any relevant deletions...

🤦

awc/src/ws.rs
SergioBenitez: That's correct. If the cookie has parameters, they will be rendered. To ease this, I would be open to implementing an `unset_all()` method on `Cookie` that clears all parameters besides `name` and `value`, which are required, of course.

robjtede: Alternatively, or additionally, zero cost Display wrappers (eg .plain and .plain_encoded) that don't print the parameters could be cheaper.

robjtede: @therealprof if you're happy with this explanation I'd like to go ahead and merge. There's also a PR open on the cookie crate for adding zero cost display wrappers which we can migrate to later.

therealprof: Yes, works for me, thanks.

therealprof (approved):

awc/src/ws.rs
SergioBenitez: @robjtede Note that you can write the same implementation outside of `cookie`. Even better, you can write a wrapper around all of `CookieJar` that also writes out the `;` so you needn't allocate at all. Pseudocode:

```rust
struct CookieJarDisplay(CookieJar);

impl Display for CookieJarDisplay {
    fn fmt(&self, f: &mut Formatter) -> Result {
        for (i, cookie) in self.0.delta().enumerate() {
            if i != 0 { write!(f, ';')?; }
            write!(f, "{}", Cookie::new(cookie.name(), cookie.value()).encoded())?;
        }

        Ok(())
    }
}
```

PR #1559, opened 2020-06-09
re-export actix_rt::main macro

robjtede:
Re-exports `actix_rt::main` as `actix_web::main` (and `actix_rt::test` as `actix_web::test_rt` since test is already a module). Also bumps minimum rt version to 1.1.1.

Would like to hear thoughts on this. Seems advantageous to remove the requirement for users to add `actix-rt` as a direct dependency. The question gets asked reasonably often of why two deps are needed for the hello world example.


cetra3:
I think that makes sense, but would need to update documentation for it.  Is there any reason why `actix_rt` would be used besides the macro?  I don't think I've used it directly in my code.

robjtede:
Using `actix_rt::spawn` is probably the second most common case for the crate. Then the test macro (I'm not completely sold on re-exporting it like this, if there are objections speak up 😄). If this idea is accepted would update docs, website and examples.

therealprof (approved):
Sounds like a good idea to me.

cetra3 (approved):
Approved

PR #1562, opened 2020-06-13
exclude default -web features on -actors

### LINKED ISSUE #1561: Option to compile without brotli
We use actix for a webserver that runs behind nginx. As such, we dont use any of the compression features of actix. It would be nice if we could exclude brotli from the compile, because it is quite slow (about 30s). Besides, we are trying to remove native dependencies to make cross compilation easier.

robjtede:
closes #1561

### DIFF
--- a/actix-web-actors/Cargo.toml
+++ b/actix-web-actors/Cargo.toml
@@ -17,7 +17,7 @@ path = "src/lib.rs"
 
 [dependencies]
 actix = "0.10.0-alpha.2"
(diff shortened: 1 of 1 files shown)


cetra3 (commented):
Seems like a sensible change, should we do this for other dependencies?

robjtede:
> Seems like a sensible change, should we do this for other dependencies?

@cetra3  Similar changes were made to packages in actix-extras.

PR #1564, opened 2020-06-14
Make current actix client demo work

esemeniuc:
I was unable to make the existing `client.rs` demo work as it was missing `Cargo.toml`. I have updated the demo to work out of the box.


robjtede:
What command wasn’t working for you?

esemeniuc:
I was getting `Response: Err(Connect(SslIsNotSupported))` as the openssl/rustls dependency was not explicit.

Another thing improvement is that my example does not require the `awc` or `actix_http` crates to be  explicitly added.

robjtede:
Would be better to keep the example in this repo as a single file and maybe guard the main method on the requred feature with an alternative main method showing the command needed to run it properly.

JohnTitor:
So the problem here is root Cargo.toml doesn't have an entry for the client example.
You should add this instead of adding another `Cargo.toml`:
```toml
[[example]]
name = "client"
required-features = ["rustls"]
```

robjtede:
closing in favor of #1625

PR #1566, opened 2020-06-16
add method to extract matched resource pattern

### LINKED ISSUE #1565: Ability to get raw match string from request
There is a middleware for prometheus: https://github.com/nlopes/actix-web-prom/issues/20

It currently uses `req.path()` in order to get path that should appear in metrics. The issue is that it includes values for dynamic parts of URLs, thus inflating number of labels and essentially makes such metrics useless.

This feature request is about having an ability for route `/path/{id}` to get literal string `/path/{id}` from `HttpRequest` somehow, such that middleware can use it if necessary.

The closest thing currently is `req.match_info()`, but it requires additional work to transform it back to original match string and even then doesn't seem to expose enough info in public API to do this.

robjtede:
A commonly requested feature for logging and metrics reasons. Not possible to do with existing methods.

Closes #1565

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 ### Added
 
 * Re-export `actix_rt::main` as `actix_web::main`.
+* `HttpRequest::match_pattern` and `ServiceRequest::match_pattern` for extracting the matched
+  resource pattern.
 
 ### Changed
 
--- a/src/request.rs
+++ b/src/request.rs
@@ -126,6 +126,17 @@ impl HttpRequest {
         &mut Rc::get_mut(&mut self.0).unwrap().path
     }
 
+    /// The resource definition pattern that matched the path. Useful for logging and metrics.
+    ///
+    /// For example, when a resource with pattern `/user/{id}/profile` is defined and a call is made
+    /// to `/user/123/profile` this function would return `Some("/user/{id}/profile")`.
+    ///
+    /// Returns a None when no resource is fully matched, including default services.
+    #[inline]
+    pub fn match_pattern(&self) -> Option<String> {
+        self.0.rmap.match_pattern(self.path())
+    }
+
     /// Request extensions
     #[inline]
     pub fn extensions(&self) -> Ref<'_, Extensions> {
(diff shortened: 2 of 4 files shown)


zizhengtai:
Thank you @robjtede. Is this also accessible through `ServiceRequest`s so that we can use it in middleware?

zizhengtai:
There is another thing that I don't know if is possible with the current approach in this PR. When creating resources, we can specify a "name":
```rust
web::resource("/foos/{id}").name("get_foo").route(...)
```
... so that later in a handler we can do:
```rust
let get_foo_url = req.url_for("get_foo", &[123]);
```
which is very useful for reliable reverse routing.

I'm wondering given a `ServiceRequest` or an `HttpRequest`, whether it's possible to get the name `"get_foo"`, in addition to the matched pattern `"/foos/{id}"`.

robjtede:
> I'm wondering given a ServiceRequest or an HttpRequest, whether it's possible to get the name "get_foo"

@ZizhengTai I can't really see the use of this since the resource names are often used to refer to other resources not the "this" resource. Maybe for logging reasons, too? Do shout if you have a legit case we can consider. Either way should be part of a separate PR.

robjtede:
Anyone from @actix/contributors want to sanity check this while Yuki is taking a break?

zizhengtai:
@robjtede Yeah definitely should be in a separate PR.

The use case is, in our current codebase, we have a route table that stores the names (instead of match patterns) of each route. This route table is used for things like reverse routing.

We also have an authentication middleware that has a list of whitelisted paths for things like sign-in and sign-out. Currently, a whitelist is described with a regex pattern. But this is not ideal, because if possible we don't want to scatter raw URL patterns everywhere. Ideally, if in the middleware we can get the "name" of the route, then the whitelist can be constructed simply with route names from our static route table.

robjtede:
Cool we’ll probably best to make an issue so we don’t forget. Should be pretty easy using a similar method to this.

cetra3:
Are there tests for multiple matches. I.e, `/user/{id}/{action}` ?

robjtede:
added in some multi variable resource tests for posterity @cetra3

cetra3 (approved):
LGTM!

nazar-pc:
Can this be backported to stable 2.x too? I'd really like to use this as soon as possible.

PR #1575, opened 2020-06-21
re-export rt in web and add main macro

robjtede:
I made a mistake in #1559. The new example file compiled but only because actix-rt is still an explicit dependency of actix-web. This PR re-exports the bundled actix-rt as `actix_web::rt`, thus allowing access to `actix_web::rt::System` and copies the main macro out of actix-macros, modifying it to use the new re-export.

Also migrated the main lib.rs module doc comment to use the actix_web::main macro as well as some reduced the scope of some clippy lints.


cetra3 (approved):
Yeah this change looks fine if it's all working now

PR #1577, opened 2020-06-22
add method to extract matched resource name

### LINKED ISSUE #1568: Ability to get route name from request
It'd be nice to be able to get the _name_ of the route given a request. See #1566 for context.

> The use case is, in our current codebase, we have a route table that stores the names (instead of match patterns) of each route. This route table is used for things like reverse routing.
>
> We also have an authentication middleware that has a list of whitelisted paths for things like sign-in and sign-out. Currently, a whitelist is described with a regex pattern. But this is not ideal, because if possible we don't want to scatter raw URL patterns everywhere. Ideally, if in the middleware we can get the "name" of the route, then the whitelist can be constructed simply with route names from our static route table.

Takashiidobe:
This PR adds a method to match a path to a route.
This PR is for issue #1568.
Fixes #1568

I've given it a shot @robjtede, looking for your feedback, guidance, and corrections.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@
 * Re-export `actix_rt::main` as `actix_web::main`.
 * `HttpRequest::match_pattern` and `ServiceRequest::match_pattern` for extracting the matched
   resource pattern.
-* `HttpRequest::match_name` for matching a path to a route.
+* `HttpRequest::match_name` for matching a path to a named route.
 
 ### Changed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@
 * Re-export `actix_rt::main` as `actix_web::main`.
 * `HttpRequest::match_pattern` and `ServiceRequest::match_pattern` for extracting the matched
   resource pattern.
-* `HttpRequest::match_name` for matching a path to a named route.
+* `HttpRequest::match_name` and `ServiceRequest::match_name` for extracting matched resource name.
 
 ### Changed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 * Re-export `actix_rt::main` as `actix_web::main`.
 * `HttpRequest::match_pattern` and `ServiceRequest::match_pattern` for extracting the matched
   resource pattern.
+* `HttpRequest::match_name` for matching a path to a route.
 
 ### Changed
 
--- a/src/request.rs
+++ b/src/request.rs
@@ -135,6 +135,7 @@ impl HttpRequest {
     #[inline]
     pub fn match_pattern(&self) -> Option<String> {
         self.0.rmap.match_pattern(self.path())
+    }
 
     /// Checks if a given path matches a route
     #[inline]
@@ -645,14 +646,14 @@ mod tests {
                         move |req: HttpRequest| {
                             assert_eq!(
                                 req.match_name(),
-                                Some("/user/{id}/profile".to_owned())
+                                Some(&ResourceDef::new("/profile"))
                             );
 
                             HttpResponse::Ok().finish()
                         },
                     )))
                     .default_service(web::to(move |req: HttpRequest| {
-                        assert!(req.match_pattern().is_none());
+                        assert!(req.match_name().is_none());
                         HttpResponse::Ok().finish()
                     })),
             ),
--- a/src/request.rs
+++ b/src/request.rs
@@ -480,9 +480,12 @@ mod tests {
 
         assert!(rmap.has_resource("/index.html"));
 
-        let req = TestRequest::default().rmap(rmap).to_http_request();
-        let route_name = req.0.rmap.match_name("/index.html");
-        assert_eq!(route_name.unwrap(), "index".to_owned());
+        let req = TestRequest::default()
+            .uri("/index.html")
+            .rmap(rmap)
+            .to_http_request();
+
+        assert_eq!(req.match_name(), Some("index"));
     }
 
     #[test]
--- a/src/request.rs
+++ b/src/request.rs
@@ -137,9 +137,11 @@ impl HttpRequest {
         self.0.rmap.match_pattern(self.path())
     }
 
-    /// Checks if a given path matches a route
+    /// The resource name that matched the path. Useful for logging and metrics.
+    ///
+    /// Returns a None when no resource is fully matched, including default services.
     #[inline]
-    pub fn match_name(&self) -> Option<String> {
+    pub fn match_name(&self) -> Option<&str> {
         self.0.rmap.match_name(self.path())
     }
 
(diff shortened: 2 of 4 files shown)


src/rmap.rs
```diff
@@ -95,6 +95,11 @@ impl ResourceMap {
         false
     }
 
+    pub fn match_path_to_route(&self, path: &str) -> Option<&ResourceDef> {
+        let path = if path.is_empty() { "/" } else { path };
+        self.named.get(path)
```
robjtede: Implementation looks good, nice handy hashmap there.

src/rmap.rs
```diff
@@ -95,6 +95,11 @@ impl ResourceMap {
         false
     }
 
+    pub fn match_path_to_route(&self, path: &str) -> Option<&ResourceDef> {
```
robjtede: I think `match_name` is concise and descriptive.

src/request.rs
```diff
@@ -126,6 +126,12 @@ impl HttpRequest {
         &mut Rc::get_mut(&mut self.0).unwrap().path
     }
 
+    /// Checks if a given path matches a route
+    #[inline]
+    pub fn match_path_to_route(&self) -> Option<&ResourceDef> {
+        self.resource_map().match_path_to_route(&self.path())
```
robjtede: Can access `self.0.rmap` directly.

src/request.rs
```diff
@@ -126,6 +126,12 @@ impl HttpRequest {
         &mut Rc::get_mut(&mut self.0).unwrap().path
     }
 
+    /// Checks if a given path matches a route
+    #[inline]
+    pub fn match_path_to_route(&self) -> Option<&ResourceDef> {
```
robjtede: Same as resource map, `match_name`.

src/request.rs
```diff
@@ -452,6 +458,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_match_path_to_route() {
```
robjtede: Include a test that match names from nested resource maps. See the match_pattern PR for example of how to set this up.

robjtede (changes requested):
Rebase on master and place near the previously added match_pattern functions.
A doc test would be good on HttpRequest addition.
Extended inline documentation needed on resource map addition.
Changelog entry needed.

Thanks for picking this up :)

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 * Re-export `actix_rt::main` as `actix_web::main`.
 * `HttpRequest::match_pattern` and `ServiceRequest::match_pattern` for extracting the matched
   resource pattern.
+* `HttpRequest::match_name` for matching a path to a route.
```
robjtede: specify "named route"

robjtede:
Do you have a specific reason for wanting to return the entire ResourceDef? Issue #1568 only talks about returning the name.

Edit: started looking at the tests too it seems to be different to the idea proposed in the issue. Can you clarify the intent of the PR.

src/request.rs
```diff
@@ -462,6 +468,27 @@ mod tests {
...
+        let mut rmap = ResourceMap::new(ResourceDef::new(""));
+        rmap.add(&mut rdef, None);
+
+        assert!(rmap.has_resource("/index.html"));
+
+        let req = TestRequest::with_uri("/test")
+            .header(header::HOST, "www.rust-lang.org")
+            .rmap(rmap)
+            .to_http_request();
+        let route_name = req.0.rmap.match_name("index");
```
robjtede: I think this should be `match_name("/index.html")` and the assert below should be that `route_name.unwrap` equals `"index"` ?
_Since the function is: path in => Option(name) out?_

src/request.rs
```diff
@@ -610,6 +637,38 @@ mod tests {
...
+    async fn test_path_routes() {
+        let mut srv = init_service(
+            App::new().service(
+                web::scope("/user/{id}")
+                    .service(web::resource("/profile").route(web::get().to(
+                        move |req: HttpRequest| {
+                            assert_eq!(
+                                req.0.rmap.match_name("/user/22/profile"),
+                                Some(ResourceDef::new("/profile").name().to_owned())
+                            );
```
robjtede: Again here, we need to set .name("x") on the resource and assert that req.match_name().unwrap() equals "x", right?

Takashiidobe:
Thanks for your patience; I wasn't able to get the last test block (test_path_routes) to work, but match_name should now take a path and return the resource name.

I can't quite figure that test out, with setting the name of the route and matching to said route.

robjtede:
@Takashiidobe Great work. I'll have a tinker with that other test case tomorrow then we should be able to get this merged.

robjtede:
I think we should strive that this test should pass: (Note the nested "user_post" resource)

```rust
    #[test]
    fn extract_matched_name() {
        let mut root = ResourceMap::new(ResourceDef::root_prefix(""));

        let mut rdef = ResourceDef::new("/info");
        *rdef.name_mut() = "root_info".to_owned();
        root.add(&mut rdef, None);
        println!("{:?}", &rdef);

        let mut user_map = ResourceMap::new(ResourceDef::root_prefix(""));
        let mut rdef = ResourceDef::new("/");
        user_map.add(&mut rdef, None);
        println!("{:?}", &rdef);

        let mut rdef = ResourceDef::new("/post/{post_id}");
        *rdef.name_mut() = "user_post".to_owned();
        user_map.add(&mut rdef, None);
        println!("{:?}", &rdef);

        root.add(
            &mut ResourceDef::root_prefix("/user/{id}"),
            Some(Rc::new(user_map)),
        );

        let root = Rc::new(root);
        root.finish(Rc::clone(&root));

        // sanity check resource map setup

        assert!(root.has_resource("/info"));
        assert!(!root.has_resource("/bar"));

        assert!(root.has_resource("/user/22"));
        assert!(root.has_resource("/user/22/"));
        assert!(root.has_resource("/user/22/post/55"));

        println!("{:?}", &root);

        // extract patterns from paths

		// non-existent resources
        assert!(root.match_name("/bar").is_none());
        assert!(root.match_name("/v44").is_none());

        // resources that exist but are not named
        assert!(root.match_name("/user/22").is_none());
        assert!(root.match_name("/user/22/").is_none());

        // named resources
        assert_eq!(root.match_name("/info"), Some("root_info"));
        assert_eq!(root.match_name("/user/22/post/55"), Some("user_post"));
    }
```

To do that, there will need to be some recursive search in match_name. Check out has_resource and match_pattern.

Other considerations:
- we can return `Option<&str>` and save an allocation
- Empty names should return None

Takashiidobe:
I gave it a shot, but I can't figure out how to recurse through the route and return the name of the route, unfortunately.  I can give the issue up to someone else or let you take it over the finish line.

robjtede:
Okay. Thanks for your effort!

robjtede (approved):

JohnTitor (approved):
Awesome!

Takashiidobe:
Thanks for your work on this issue Rob, your help was much appreciated.

PR #1578, opened 2020-06-22
move pull request template

robjtede:
Turns out this doesnt follow the same format as the issue templates, only one is allowed and the front-matter isnt interpreted.

Tested on my fork.


JohnTitor (approved):
Oh, I didn't know that, looks good!

Issue #1523, opened 2020-05-20
Memory leak with actix-web 2.0 and actix-rt 1.0.0 for echo websocket server


jeandeducla:
I took this actix-web example: [websocket echo server](https://github.com/actix/examples/tree/master/websocket), I just changed the client so it spawns `N` `ChatClient` that sends message every 1 seconds for `T` seconds (I pasted the client code below) and then close. 

I did the following test: I start the server, I spawn 200 sessions for 5min, wait some minutes, spawn another 200 sessions for 5min, wait... I expect the server memory to increase on the first 200 connections and then decrease to some threshold. On the next 200 connections I also expect the server memory to increase and then decrease to the same previous threshold. However that is not what I observe: the memory does not decrease once the connections close and it does not go back to the original threshold, it just keeps increasing. 

I ran this test using docker containers in kubernetes, and here is the container cpu and memory timeseries (the 200 sessions spawning correspond to the cpu bursts):
`cpu`
![image](https://user-images.githubusercontent.com/28356783/82465914-12524000-9ac0-11ea-8afd-89f3434a39c3.png)

`memory`
![image](https://user-images.githubusercontent.com/28356783/82465958-1da56b80-9ac0-11ea-9c91-84668d3490ad.png)

I observe the same thing when running this on my machine: the memory of the server process keeps increasing.

The versions used are the one from the example's `Cargo.toml` (I enforced actix-rt to `1.0.0` as suggested in this [issue](https://github.com/actix/actix-net/issues/129)), so:
```
[dependencies]
actix = "0.9.0"
actix-codec = "0.2.0"
actix-web = "2.0.0"
actix-web-actors = "2.0.0"
actix-files = "0.2.1"
actix-rt = "=1.0.0"
awc = "1.0.1"
env_logger = "0.7"
futures = "0.3.1"
bytes = "0.5.3"
```

## Expected Behavior
When the server handles connections, memory may increase and once the connections are all closed and the server is idle, memory should stay at some level that remains the same during the server lifetime (when the server is idle its memory should be around this threshold).

## Current Behavior
Server memory never decreases even when the server is idle.

## Steps to Reproduce
As I said I ran this in kubernetes, but you could see the same thing on your machine, I could not test on other platforms:
1. run the server and check the initial process memory
2. spawn 200 peers or more with the provided client
3. when the sessions close, check again the server process memory.
4. repeat

## Your Environment
[...]

Issue #1464, opened 2020-04-21
ConnectorPoolSupport leaks in each pool call


fafhrd91:
https://github.com/actix/actix-web/blob/751253f23e54b4eae6665b858430d9c1fc79b8d7/actix-http/src/client/pool.rs#L95

cdbattags:
Any more data/sample code to support this claim?

fafhrd91:
Did you actually check link I posted?

jtescher:
@cdbattags that future is spawned on every awc request, and you can see that it will never complete in https://github.com/actix/actix-web/blob/d5ceae20740837bddef85053c3354452b437877a/actix-http/src/client/pool.rs#L435-L484

jtescher:
Was this meant to be tied to the lifecycle of the pool rather than one per request?

robjtede:
Tenatively fixed with #1580.

Issue #1428, opened 2020-03-24
possible memory leak


shaitao:
I build a server with Actix-web 2.0, and the memory was increasing with the increasing of requests. I don't know where is the problem, I just used awc(with tls), web.  when I change all things to warp the problem is gone. I don't know how to find it.

robjtede:
@shaitao you'll need to provide us with a minimal example so we can help

Lesiuk:
My actix server is leaking memory when It's behind nginx (using proxy_pass). And everything is fine when connecting to directly. It's like few megabytes per 3000 requests.  I will work on minimal reproduction example this weekend.

pythoneer:
@Lesiuk is this related to your recent issue in https://github.com/actix/actix-web/issues/1446 ?

Lesiuk:
@pythoneer yes, It's this issue and during investigation I was debugging a little a didnt found any other major leak

robjtede:
wait, this issue was filed before the release of actix-rt 1.1 so can't be the same issue right?

nazar-pc:
For me with 2.0.0 even this simple example leaks memory:
```rust
HttpServer::new(|| App::new().route("/", web::get().to(HttpResponse::Ok)))
```

Tested like this:
```
ab -c 100 -n 100000 http://localhost:8081/
```

With above command process consumes 315M of memory, with second call 626M, with third 939M, you get the idea.

Behavior is the same with both debug and release build.

I think this is quite a critical issue, surprised not many people have noticed it.

Lesiuk:
Unfortunately I cannot reproduce it. Could you give us more info? Which operating system are you using? Actix-rt version?

nazar-pc:
Ubuntu 20.10 (development branch), actix-rt 1.1.0.
Updating to actix-rt 1.1.1 seems to fix the issue for me.

Lesiuk:
@nazar-pc thats the reason why actix-rt 1.1.0 was yanked https://github.com/actix/actix-net/issues/129

nazar-pc:
Yeah, but I had it in lock file, so it installed fine anyway.
Unless there is additional evidence I believe this issue can be closed.

jtescher:
This is also likely related to https://github.com/actix/actix-web/issues/1464

singggum3b:
I'm also played with basic actix http2 `3.0.0-alpha.3` and also notice some kind of memory leak
Simple h2 server reponse with some json literal:
```

[...]

PR #1580, opened 2020-06-24
Fix leaks with actix_http's client

ptescher:
Each request was creating a new ConnectorPoolSupport future which never
finished. That future also referenced the pool’s inner so even when the
entire pool dropped we leaked the internals. This fixes both.

## PR Type
What kind of change does this PR make?

Bug fix

## PR Checklist

## Overview
There was a leak in actin-http's client where each request would span a future that would never finish. That future also referenced the pool so the pool would never drop.

This PR fixes both cases. The future now only spawns once when the pool is created and when the pool drops the future finishes.

This should fix:
- https://github.com/actix/actix-web/issues/1523
- https://github.com/actix/actix-web/issues/1464
- https://github.com/actix/actix-web/issues/1428

### DIFF
--- a/awc/src/request.rs
+++ b/awc/src/request.rs
@@ -586,16 +586,16 @@ mod tests {
     use super::*;
     use crate::Client;
 
-    #[test]
-    fn test_debug() {
+    #[actix_rt::test]
+    async fn test_debug() {
         let request = Client::new().get("/").header("x-test", "111");
         let repr = format!("{:?}", request);
         assert!(repr.contains("ClientRequest"));
         assert!(repr.contains("x-test"));
     }
 
-    #[test]
-    fn test_basics() {
+    #[actix_rt::test]
+    async fn test_basics() {
         let mut req = Client::new()
             .put("/")
             .version(Version::HTTP_2)
@@ -621,8 +621,8 @@ mod tests {
         let _ = req.send_body("");
     }
 
-    #[test]
-    fn test_client_header() {
+    #[actix_rt::test]
+    async fn test_client_header() {
         let req = Client::build()
             .header(header::CONTENT_TYPE, "111")
             .finish()
@@ -639,8 +639,8 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_client_header_override() {
+    #[actix_rt::test]
+    async fn test_client_header_override() {
         let req = Client::build()
             .header(header::CONTENT_TYPE, "111")
             .finish()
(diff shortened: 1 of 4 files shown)


JohnTitor:
I don't look in detail yet, but it looks like this breaks some (awc) tests.

ptescher:
@JohnTitor hmm interesting, this shouldn't really affect anything. All the tests pass locally for me.

ptescher:
@JohnTitor ok actually something was out of sync. I see the failures now, will fix.

ptescher:
Ok fixed the tests. One change now is that an actix-rt system needs to be started before a client is created, not just before the first request. This seems fine to me, especially if this is part of a version bump.

JohnTitor (approved):
I think that makes sense, looks great to me!

robjtede:
think we should get one more eye over this before merging, pretty big deal if this fixes the root cause of those memory leak issues

awc/src/request.rs
```diff
@@ -586,16 +586,16 @@ mod tests {
     use super::*;
     use crate::Client;
 
-    #[test]
-    fn test_debug() {
+    #[actix_rt::test]
```
cetra3, line 589: I can't see clearly these tests include using multiple clients in parallel.  I am assuming this is where the `ConnectorPool` issues will show up, but I could be wrong.

cetra3 (approved):
Appears to be a simple change to ensure that the `ConnectorPoolSupport` will eventually resolve the future (if there are no references to it).

I would be interested in seeing a few parallel request tests as well or any test that covers the issues linked to this PR.

Dowwie:
> Ok fixed the tests. One change now is that an actix-rt system needs to be started before a client is created, not just before the first request. This seems fine to me, especially if this is part of a version bump.

@actix/contributors if anyone objects to this change, please share a plausible scenario where there is a conflict

awc/src/request.rs
ptescher: Because the client now creates a future when it is launched, instead of one per request, these tests now require actix_rt. But the tests themselves don't really use the connectorpool at all.

The benches hit it hard though, since they generate a single client and make multiple requests.

ptescher:
@cetra3 a proper regression test here would check with Tokio to make sure that the number of current futures being worked on is the same before and after a request. There is an open issue for Tokio to support this: https://github.com/tokio-rs/tokio/issues/2510.

robjtede:
> make sure that the number of current futures being worked on is the same before and after a request

That would be a *really* elegant way to detect leaks of this type.

cetra3:
I don't think we should hold up the PR for that change in tokio to land, but maybe another issue raised?

robjtede:
100%. I think get this merged in the next 24hrs if there are no objections from contributors.

PR #1584, opened 2020-06-27
Fix example on README.md

tivrfoa:
It was missing the actix-rt dependency. Also it's actix_rt::main, and not actix_web::main.

## PR Type
What kind of change does this PR make?

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / code style change (no functional or public API changes)
- [x] Other

## PR Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Dependencies:
 
 ```toml
 [dependencies]
+actix-rt = "1"
 actix-web = "2"
 ```
 
@@ -65,7 +66,7 @@ async fn index(info: web::Path<(u32, String)>) -> impl Responder {
     format!("Hello {}! id:{}", info.1, info.0)
 }
 
-#[actix_web::main]
+#[actix_rt::main]
 async fn main() -> std::io::Result<()> {
     HttpServer::new(|| App::new().service(index))
         .bind("127.0.0.1:8080")?


robjtede:
The readme is valid for the master branch where this code does work and will be the recommended way in version 3.

tobymurray:
(hopefully) constructive feedback, as a new Rust user I went to https://actix.rs/, copy/pasted the code there, went to the "Install" page and followed the current instructions - it does not work (as described above). Noting that version 3 is upcoming, it appears the documentation is not consistent in a way that is confusing for brand new users. The "install" page shows version 2, the main page shows code code that only works with version 3.

robjtede:
@tobymurray I believe you're having a different issue.

The website has received no updates for v3 yet. The main page shows a valid hello world example for v2. The common pitfall here is that the installation page does not mention `actix_rt` (the runtime crate that allows usage of the `actix_rt::main` macro) until reaching the next page (Getting Started).

We're making a change in v3 so that the recommended way to start the runtime is with a new `actix_web::main` macro which will not require an explicit import of `actix_rt` and therefore make onboarding easier for new users.

It's also true that the website often needs some love and we are happy to accept PRs to improve it :)

tobymurray:
Yep - you're absolutely right. I did a shamefully bad job at reading, and then again when looking at this PR. The README change here wouldn't make sense, as you mentioned, as it's consistent with the relevant branches.

> We're making a change in v3 so that the recommended way to start the runtime is with a new actix_web::main macro which will not require an explicit import of actix_rt and therefore make onboarding easier for new users.

Sounds like a good change, so those of us who don't read have a smoother time onboarding!

Coming from not-Rust, it feels unexpected to need to install two dependencies to make use of the example on the landing page. I'm sure there are good reasons for it that I'd come to understand. As analogs, Spring came out with their "Spring Boot starter", Rails makes some decisions on the user's behalf so it "just works", etc.  It sounds like you're already addressing that so probably not worth stressing too much over the v2 example.

If you're keen to change and there's an aesthetically pleasing way to do so, adding the:
```
[dependencies]
actix-web = "2.0"
actix-rt = "1.0"
```
to the landing page would have (I think) made it clear to me, and avoided having me end up here.

PR #1589, opened 2020-07-03
Updated README to reflect '3.0.0-alpha.3' in Example. Updated list of…

seenickcode:
## PR Type
README

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor / code style change (no functional or public API changes)
- [x] Other

## Overview
Updated README to be more friendly to newcomers of Rust. *Important note*: __many__ newcomers to a new language will go straight to a web framework to try things out. This is one of the leading web frameworks out there and if the Getting Started / Example isn't quick and easy (if it doesn't "just work") this will not only hurt adoption of the library but **Rust in general**. 

Therefore, I recommend that: 
1. Only one single link to the "API Documentation" is available. I'm not sure why there was a link to the docs on `actix.rs` as well as `docs.rs` so I removed one of them.

2. Make it __easy__ for developers to update their dependencies and get the example to "just work", without requiring newcomers to know what `actix-rs` is. 

Therefore I've changed:

- Updated README to reflect '3.0.0-alpha.3' in Example. 
- Updated list of Documentation links in the README as well.

I recommend keeping the README up to date w/ the latest version of dependencies to ensure better adoption rates.

Thanks for considering my PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,8 @@ Actix web is a simple, pragmatic and extremely fast web framework for Rust.
 
 ## Docs
 
-* [API documentation (master)](https://actix.rs/actix-web/actix_web)
-* [API documentation (docs.rs)](https://docs.rs/actix-web)
-* [User guide](https://actix.rs)
+* [Website](https://actix.rs)
+* [API documentation](https://docs.rs/actix-web)
 
 ## Example
 
@@ -52,7 +51,8 @@ Dependencies:
 
 ```toml
 [dependencies]
-actix-web = "2"
+actix-web = "3.0.0-alpha.3"
+actix-rt = "1.1.1"
 ```
 
 Code:
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Actix web is a simple, pragmatic and extremely fast web framework for Rust.
 ## Docs
 
 * [Website](https://actix.rs)
-* [API documentation](https://docs.rs/actix-web)
+* [API documentation (master)](https://actix.rs/actix-web/actix_web)
 
 ## Example
 


README.md
```diff
@@ -42,17 +42,17 @@ Actix web is a simple, pragmatic and extremely fast web framework for Rust.
 
 ## Docs
 
-* [API documentation (master)](https://actix.rs/actix-web/actix_web)
```
robjtede, line 45: This is the rendered version of master branch docs.

README.md
```diff
@@ -42,17 +42,17 @@ Actix web is a simple, pragmatic and extremely fast web framework for Rust.
...
-* [API documentation (docs.rs)](https://docs.rs/actix-web)
```
robjtede, line 46: This should arguably say "stable" and be higher than master docs link.

robjtede:
As I said in the issue, you will not need the explicit actix-rt dependency with v3 and I'd like the master branch readme to reflect this. A note making this clear above the example block redirecting newcomers to the stable docs is a good idea though.

README.md
seenickcode: Ok I've switched it.

seenickcode:
> As I said in the issue, you will not need the explicit actix-rt dependency with v3 and I'd like the master branch readme to reflect this. A note making this clear above the example block redirecting newcomers to the stable docs is a good idea though.

So that's what I was confused about. There's no `v3.0.0` yet, correct? Until that is available, it seems like you do you need this `actix-rt` dependency, no?

robjtede:
> Until that is available, it seems like you do you need this actix-rt dependency, no?

On v2 and previous v3 alphas, yes. But master branch is not either of those. For that reason, I won't accept a change that puts `actix-rt` back in the readme. The best course of action is just to make this limbo state clear so people don't try to use that example code in their v2 project.

seenickcode:
> > Until that is available, it seems like you do you need this actix-rt dependency, no?
> 
> On v2 and previous v3 alphas, yes. But master branch is not either of those. For that reason, I won't accept a change that puts `actix-rt` back in the readme. The best course of action is just to make this limbo state clear so people don't try to use that example code in their v2 project.

Ok so let me ask you this: if I'm a newcomer to this library, what version should I use? v3.0.0-alpha.3 or master?

The Example section is for newcomers to the project. The vast majority of users will not be using v2 or having any existing projects. So

robjtede:
> what version should I use? v3.0.0-alpha.3 or master?

You should use v2 as per: 
- https://docs.rs/actix-web/
- https://actix.rs/
- https://github.com/actix/examples

robjtede:
After some dicussion on Gitter, we'll make some changes to the docs soon.

PR #1594, opened 2020-07-10
Make web::Path a tuple struct with a public inner value

jplatte:
## PR Type

Feature

## PR Checklist

## Overview

Currently, it is impossible to destructure a `web::Path` to obtain the inner value in a pattern (for example an argument definiton in a function signature), while the same is supported for other extractors, such as `web::Json` and `web::Query`. This means it is common to have one of the first lines of a handler be

```rust
let x = x.into_inner();
```

for single-path-parameter routes, or

```rust
let (a, b, c) = path.into_inner();
```

unless borrowing from the `web::Path` is feasible, in which case `Deref` takes care of the job.

This change makes `web::Path` behave the same way as `web::Json` and `web::Query`, which is breaking change because derefs to tuples stop working. In practice, I would expect destructuring to be more concise over `.into_inner` in many cases because it can be done right in the function signature and often, every parameter is on its own line already. It also moves assignment of path parameter names closer to the route definition if using route attributes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 * Fix actix_http::h1::dispatcher so it returns when HW_BUFFER_SIZE is reached. Should reduce peak memory consumption during large uploads. [#1550]
 * Migrate cookie handling to `cookie` crate. Actix-web no longer requires `ring` dependency.
 * MSRV is now 1.41.1
+* `web::Path` now has a public representation: `web::Path(pub T)`
+  * This means it can now be destructured the same way as `web::Json`: `async fn some_route(web::Path(s): web::Path<String>)`
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -14,6 +14,8 @@
 * Fix actix_http::h1::dispatcher so it returns when HW_BUFFER_SIZE is reached. Should reduce peak memory consumption during large uploads. [#1550]
 * Migrate cookie handling to `cookie` crate. Actix-web no longer requires `ring` dependency.
 * MSRV is now 1.41.1
+* `web::Path` now has a public representation: `web::Path(pub T)`
+  * This means it can now be destructured the same way as `web::Json`: `async fn some_route(web::Path(s): web::Path<String>)`
```
JohnTitor: This should be moved.

JohnTitor (approved):
Sorry for the delay! I'd be in support of this, nice =)

Dowwie:
nice update

robjtede (approved):

PR #1600, opened 2020-07-13
prepare actix-web v3.0.0-beta.1 release

robjtede:
bumps actix-web and codegen versions for release

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,38 +1,32 @@
 # Changes
 
-## [Unreleased]
+## Unreleased - 2020-xx-xx
 
+
+## 3.0.0-beta.1 - 2020-07-13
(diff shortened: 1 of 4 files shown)


robjtede:
There may be other things we want to bump before this?

Silentdoer:
app_data should be rename?

robjtede:
Not in this PR. Let's keep discussions of specific items to their issues.

JohnTitor (approved):
Though it's a dev-dep, it still refers alpha - let's sync :)

https://github.com/actix/actix-web/blob/2065c4744afd670e3313c6944ea250e6f5a3b602/awc/Cargo.toml#L59

Otherwise looks good to me.

PR #1602, opened 2020-07-14
prepare awc v2.0.0-beta.1 release

robjtede:
following the dep graph backwards shows this should be done before actix-web


JohnTitor (approved):
Yep, we've actually done that in the past :+1:

PR #1610, opened 2020-07-19
also try extracting payload config as Data<T>

### LINKED ISSUE #1608: JsonConfig limit has no effect
actix-web version: `3.0.0-alpha.3`

My App starts with
```rust
    HttpServer::new(move || {
        App::new()
            .wrap(actix_web::middleware::Logger::new(
                "%a \"%r\" (size: %{Content-Length}i) Response: %s %b",
            ))
            .data(web::JsonConfig::default().limit(1024 * 1024 * 50))
```

This is the signature of the request handler I invoke:
```rust
pub async fn post(
    body: web::Json<NewAnalysisMethod>,
) -> Result<Json<AnalysisMethodId>, Error> {
```
This is what it logs when I send about 5MB of payload:
```
Jul 16 11:03:40.391 DEBG Failed to deserialize Json from payload. Request path: /api/v1/data/analysis_method
Jul 16 11:03:40.391 DEBG Error in response: Overflow
Jul 16 11:03:40.391 INFO 127.0.0.1:58804 "POST /api/v1/data/analysis_method HTTP/1.1" (size: 4915402) Response: 413 46
```
Where the format of the Logger middleware is set to `"%a \"%r\" (size: %{Content-Length}i) Response: %s %b"`.

robjtede:
## PR Type

Ergonomics Improvement

## PR Checklist

## Overview

When setting payload config it seems to make sense to check Data<T> too to allow for programmer error (for lack of better phrase) in using `App::data` instead of `App::app_data` to set config, especially since a mis-configuration currently does not result in an error message (because default fallback config is used).

## Follow-up Work
Apply same change to `Json` and `Form` extractors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,15 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+### Changed
+* `PayloadConfig` is now also considered in `Bytes` and `String` extractors when set
+  using `App::data`. [#1610]
+
 ### Fixed
 * Memory leak of app data in pooled requests. [#1609]
 
 [#1609]: https://github.com/actix/actix-web/pull/1609
+[#1610]: https://github.com/actix/actix-web/pull/1610
 
 
 ## 3.0.0-beta.1 - 2020-07-13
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Sounds good idea, lgtm.

PR #1614, opened 2020-07-20
fix soundness concern in h1 decoder

robjtede:
## PR Type

Unsoundness Fix

## PR Checklist

## Overview

From all the sources I can see, its *never* okay to do:
```rust
let arr:[T; n] = unsafe { MaybeUninit::uninit().assume_init() };
```
unless `T` is `MaybeUninit<T>` (see [the nomicon](https://doc.rust-lang.org/nomicon/unchecked-uninit.html)). 

Parsing HTTP headers in the h1 decoder currently uses code like this which Miri flags as unsound. [See this playground for a simplified example.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f5d9aa166aa48df8dca05fce2b6c3915)

This PR removes all the unsound memory initializations in h1 decoder and opts to use a pre-filled const array to be handed to `httparse`.

I believe this to be the fastest sound solution without requiring changes to `httparse`. (Aside: I think in theory, `httparse` could provide a parse function that fills in a MaybeUninit header array and allows caller to declare the array intitialized with the number of headers it parsed. You can imagine it's data flow with [the setup in this modifed playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=88c1376bbb6c9bb0bb085300578ab152). However, I opted for the safer approach for now)

Benchmark code is included. Using test cases from `httparse` we see the following regressions:

Short Response:
```
Original (Unsound) [short]        time:   [162.17 ns 166.87 ns 172.44 ns]
New (safe) [short]                time:   [207.06 ns 210.67 ns 214.79 ns]
```

Realistic Response:
```
Original (Unsound) [realistic]    time:   [516.38 ns 526.95 ns 538.79 ns]
New (safe) [realistic]            time:   [532.47 ns 543.05 ns 556.96 ns]
```


JohnTitor (approved):
Considering benchmarks, I think it's reasonable trade-off. Thanks for detail explanation and implementation, great work :)

Shnatsel:
This RFC will allow getting performance back safely: https://github.com/rust-lang/rfcs/pull/2930

PR #1618, opened 2020-07-21
match HttpRequest app_data behavior in ServiceRequest

### LINKED ISSUE #1613: `ServiceRequest::app_data` only allows extracting `Data<T>` types
## Expected Behavior
`ServiceRequest::app_data` should allow extracting any type registered with either `App::app_data` or `App::data` without having to resort to using into_parts to get access to the HttpRequest which does allow any T to be extracted.

## Current Behavior
As evidenced by the function, only checks for Data<T> in each data layer.

```rust
pub fn app_data<T: 'static>(&self) -> Option<Data<T>> {
    for container in (self.0).0.app_data.iter().rev() {
        if let Some(data) = container.get::<Data<T>>() {
           ...
        }
    }
    ...
}
```

## Possible Solution
Pass type param T to get directly and require authors to specifiy Data<T> as type parameter if that is their intention.

```rust
pub fn app_data<T: 'static>(&self) -> Option<Data<T>> {
    for container in (self.0).0.app_data.iter().rev() {
        if let Some(data) = container.get::<T>() {
            return Some(Data::clone(&data));
        }
    }

}
```

## Context
Brought up during a troubleshooting discussion on Discord and discovering the disparity between ServiceRequest and HttpRequest.

<details>
<summary>Brought to attention with this discussion on community Discord.</summary>

dragonn
it fails at getting app_data inside Middleware, I have no idea why since it works in other places without any problems
it this helps in any way `[file containing Option::unwrap panic]`

robjtede
are you registering with App::app_data or App::data

dragonn
```rust
let mut server = HttpServer::new(|| {
    App::new()
        .app_data(config::init())
```

config::init() is really simple:

```rust
use config;
use crate::models::config::Settings;

pub fn init() -> Settings {
    let mut settings = config::Config::default();

    settings.merge(config::File::with_name("Default")).unwrap();
    settings.merge(config::File::with_name("Settings")).unwrap();

    
    settings.try_into::<Settings>().unwrap()
}
```

it uses the config crate

robjtede
yeah ServiceRequest::app_data can only extract when set with App::data
its an oddity and I have some plans to address it at some point
if you use into_parts you can access the HttpRequest
where app_data allows access to non Data wrapped data

dragonn
Works now, big thank you :). Yeah this is a litte bit confusing

</details>

robjtede:
## PR Type

Fix

## PR Checklist

## Overview

Changes `ServiceRequest::app_data` to match `HttpRequest::app_data`'s return type since it doesn't make much sense to have different return types given you can just split a ServiceRequest and get at the HttpRequest anyway. 

closes #1613


fakeshadow (approved):
I like this change. It also fully enabled custom data types where it can be extracted both in handler and requests.

PR #1621, opened 2020-07-23
Re-export all error types from awc

### LINKED ISSUE #1474: Re-export awc::error::JsonPayloadError
I'm trying to decode a returned json response from a web server while handling the response appropriately.  However, the error from the `actix_web::client::ClientResponse::json` method cannot be handled easily because there is no way to refer to `awc::error::JsonPayloadError` without adding an additional dependency on `awc`, which seems to defeat the purpose of `actix_web` reexporting most of the `awc` interface in the first place.

One solution would be to re-export JsonPayloadError in `actix_web::client`.  Another is to just add a re-export for the whole `awc` crate to solve future issues of this type.

mwilliammyers:
## PR Type

Other

## PR Checklist

- [x] Tests for the changes have been added / updated. (_N/A_)
- [x] Documentation comments have been added / updated. (_N/A_)

## Overview

This makes using error handling while using the `awc::Client` *much easier*.

Closes #1474

<br>

Or I can update this PR to do one of the following (in order of ascending levels of change):

1. Import the rest of the missing `errors` (and the rest of the API as well?) like `FrozenClientRequest` _individually_.
1. Change the `error` imports to:
   ```rust
   pub use awc::error::*;
   ```
1. Change the whole `client` module to:
   
   1. Re-export everything while keeping the current structure:
        ```rust
        pub use awc::error::*;
        pub use awc::*;
        ```
    1. Re-export everything to keep the same structure as `awc` (__breaking change__):
        ```rust
        pub use awc::*;
        ```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
   destructuring. [#1594]
 * `ServiceRequest::app_data` allows retrieval of non-Data data without splitting into parts to
   access `HttpRequest` which already allows this. [#1618]
+* Re-export all error types from `awc`. [#1621]
 * MSRV is now 1.42.0.
 
 ### Fixed
@@ -17,6 +18,7 @@
 [#1609]: https://github.com/actix/actix-web/pull/1609
 [#1610]: https://github.com/actix/actix-web/pull/1610
 [#1618]: https://github.com/actix/actix-web/pull/1610
+[#1621]: https://github.com/actix/actix-web/pull/1621
 
 
 ## 3.0.0-beta.1 - 2020-07-13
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
I have no objections for this. Could you also update the changelog?

mwilliammyers:
Should I _not_ add any other imports (or do any of the other options listed above)?

robjtede:
I would be in favour of re exporting all the error types as you mentioned in 2.

JohnTitor:
We'd be happy to include this in the next beta release if you could update in a few days :)

mwilliammyers:
Done!

I also noticed the change log has a bad link:
```
[#1618]: https://github.com/actix/actix-web/pull/1610
```

I can fix that as well, unless you wanted to keep that separate?

robjtede:
> I can fix that as well, unless you wanted to keep that separate?

might as well since you spotted it 👍

robjtede (approved):

PR #1625, opened 2020-08-08
require rustls feature for client example

robjtede:
## PR Type
Example Fix

## Overview
Requires client example is run with `rustls` feature.

Error when run without feature is now:
```plain
error: target `client` in package `actix-web` requires the features: `rustls`
Consider enabling them by passing, e.g., `--features="rustls"`
```

instead of:
```plain
Error: Connect(SslIsNotSupported)
```

Superceeds #1564


JohnTitor (approved):
Thanks!

PR #1626, opened 2020-08-09
fix awc pool leak

### LINKED ISSUE #1622: Memory leak with awc::Client
It appears that creating multiple `awc::Client` instances results in a memory leak, even once the `Client` objects leave scope. The documentation rightfully points out that a `Client` should be reused rather than created on a per-request basis, but even so we should expect no memory leaks.

It does not appear that the recent PR (#1580) addresses this leak.

##  Expected Behavior

Memory associated with a `Client` is released after the client (and its associated requests) are complete.

## Current Behavior

Instantiating multiple clients in a loop results in excessive and unbounded memory usage.

Here is a simple `awc-mem-leak` application.

`Cargo.toml`:
```
[dependencies]
actix-rt = "1.1.1"
awc = { version = "2.0.0-beta.2", features = ["compress"] }
```

`main.rs`:
```
// Change this to target an example HTTP server.
const URL: &str = "http://localhost:8000";

#[actix_rt::main]
async fn main() {
    let mut i = 0;
    loop {
        i += 1;
        if let Err(_) = awc::Client::build().finish().get(URL).send().await {
            eprintln!("{}: error", i);
        } else {
            eprintln!("{}: ok", i);
        }
    }
}
```

**Usage**: Stand up a local HTTP server (e.g., `python3 -m http.server`), adjust `URL` above, and run: `cargo run --release`. Memory will increase unboundedly.

Running this client application with valgrind & massif makes the memory leak very clear.

1. Build the binary: `cargo build --release`
2. Profile it: `valgrind --tool=massif target/release/awc-mem-leak`
3. After a couple minutes, Ctrl+C the process. This will produce a `massif.out.PID` file.
4. Visualize: `massif-visualizer massif.out.PID`

**Heap usage**. After running for 5 minutes (about 31k iterations of the above loop), Massif shows:

image

The original Massif capture is [here](https://gist.github.com/ajsyp/12fcaaaabcab65cc7060e22e5d60eddd).

## Environment:
```
OS: ArchLinux (x86_64)
cargo 1.45.0 (744bd1fbb 2020-06-15
rustc 1.45.0 (5c1f21c3b 2020-07-13)
```

fakeshadow:
## PR Type
Bug Fix

## PR Checklist

## Overview
When `awc::Client` goes out of scope it doesn't wake up it's support future which could result in the future stuck in executor and not polled again causing memory leak.

This PR explicitly wake up the support future when `awc::Client` is dropped and resolve the future. 

It also goes back to use `Rc` instead of `Weak` for `Inner<T>` field of `ConnectorPoolSupport`. As `Weak` doesn't prevent the leak and it also have a little overhead with every poll of future.

Closes #1622


actix-http/src/client/pool.rs
```diff
@@ -66,7 +66,7 @@ where
         // start support future
         actix_rt::spawn(ConnectorPoolSupport {
             connector: connector_rc.clone(),
-            inner: Rc::downgrade(&inner_rc),
+            inner: inner_rc.clone(),
```
robjtede: would prefer explicit Rc::clone

actix-http/src/client/pool.rs
```diff
@@ -82,6 +82,13 @@ where
     }
 }
 
+impl<T, Io> Drop for ConnectionPool<T, Io> {
+    fn drop(&mut self) {
+        // We wake up the ConnectorPoolSupport when dropping so it can exit properly.
```
robjtede: drop the "We"

fakeshadow:
fixed

robjtede (approved):

robjtede:
Good work on this 👍

PR #1636, opened 2020-08-17
web: Bump up to 3.0.0-beta.2

JohnTitor:


robjtede:
shouldnt we also bump awc version?

therealprof:
@robjtede Yes, please. Would also be nice if we could get the rustls bump some time...

JohnTitor:
Well, it's not necessary, I think. Any notable changes there?

robjtede:
Isn't it just to get the -http versions the same?

JohnTitor:
We can already pull actix-http 2.0.0-beta.3, cannot we? I think it's enough here to avoid making release harder.

robjtede (approved):

JohnTitor:
@robjtede Thanks for tagging it, do you mind publish a release, too?

PR #1639, opened 2020-08-18
Configurable trailing slash behaviour for NormalizePath

### LINKED ISSUE #1612: customization of NormalizePath behavior
excerpt from #959 

> I think had hoped that NormalizePath might let you create a route like /hello and accept either /hello or /hello/ I suppose this can also be achieved via route matching (e.g. #[get(r#"/hello{tail:/?}"#)]) (as an aside: I also wonder if users would prefer to accept both distinct routes or have one version 301 redirect to the other?).
>
> When I tried to check v3 beta, seems like my usage of NormalizePath did not work for me? Not sure if I have something mixed up here, but wasn't able to access either /hello or /hello/.

So, middleware should take option `TrailingSlash::{Always, Trim}` and possibly an option to 301 redirect to the normalized path.

l-2-j:
## PR Type

Feature

## PR Checklist

## Overview

Currently NormalizePath always adds a trailing slash. This change would make it configurable to either follow that style, or to always trim the trailing slash instead.

Will be a breaking change to `NormalizePath` usage, since `middleware::NormalizePath` will need to be replaced by `middleware::NormalizePath::default()` or `::new(...)` when creating the middleware.

Closes #1612

### DIFF
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -10,20 +10,29 @@ use regex::Regex;
 use crate::service::{ServiceRequest, ServiceResponse};
 use crate::Error;
 
-#[derive(Default, Clone, Copy)]
+/// To be used when constructing `NormalizePath` to define it's behavior.
+#[derive(Clone, Copy)]
+pub enum TrailingSlash {
+    /// Always add a trailing slash to the end of the path.
+    Always,
+    /// Trim trailing slashes from the end of the path.
+    Trim,
+}
+
+#[derive(Clone, Copy)]
 /// `Middleware` to normalize request's URI in place
 ///
 /// Performs following:
 ///
 /// - Merges multiple slashes into one.
-/// - Appends a trailing slash if one is not present.
+/// - Appends a trailing slash if one is not present, or removes one if present, depending on the supplied `TrailingSlash`.
 ///
 /// ```rust
 /// use actix_web::{web, http, middleware, App, HttpResponse};
 ///
 /// # fn main() {
 /// let app = App::new()
-///     .wrap(middleware::NormalizePath)
+///     .wrap(middleware::NormalizePath::default())
 ///     .service(
 ///         web::resource("/test")
 ///             .route(web::get().to(|| HttpResponse::Ok()))
@@ -32,7 +41,20 @@ use crate::Error;
 /// # }
 /// ```
 
-pub struct NormalizePath;
+pub struct NormalizePath(TrailingSlash);
+
+impl NormalizePath {
+    /// Create new `NormalizePath` middleware with the specified trailing slash style.
+    pub fn new(trailing_slash_style: TrailingSlash) -> Self {
+        NormalizePath(trailing_slash_style)
+    }
+}
+
+impl Default for NormalizePath {
+    fn default() -> Self {
+        NormalizePath::new(TrailingSlash::Always)
+    }
+}
 
 impl<S, B> Transform<S> for NormalizePath
 where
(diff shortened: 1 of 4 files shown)


src/middleware/normalize.rs
```diff
@@ -50,13 +72,18 @@ where
...
+                TrailingSlash::Always => false,
+            },
         })
     }
 }
 
 pub struct NormalizePathNormalization<S> {
     service: S,
     merge_slash: Regex,
+    trim_last_slash: bool,
```
robjtede: Given how we're using it, makes sense to keep the enum here too. Makes for more expressive matches later and possible ease of extensibility.

src/middleware/normalize.rs
```diff
@@ -32,7 +41,20 @@ use crate::Error;
...
+    pub fn new(trailing_slash_style: TrailingSlash) -> Self {
+        NormalizePath(trailing_slash_style)
+    }
+}
+
+impl Default for NormalizePath {
+    fn default() -> Self {
+        NormalizePath::new(TrailingSlash::Always)
+    }
+}
```
robjtede: If we impl Default for TrailingSlash we can keep the derive on NormalizePath.

src/middleware/normalize.rs
```diff
@@ -10,20 +10,29 @@ use regex::Regex;
 use crate::service::{ServiceRequest, ServiceResponse};
 use crate::Error;
 
-#[derive(Default, Clone, Copy)]
+/// To be used when constructing `NormalizePath` to define it's behavior.
+#[derive(Clone, Copy)]
+pub enum TrailingSlash {
+    /// Always add a trailing slash to the end of the path.
```
robjtede, line 17: Make a note that all routes defined must end in trailing slash for them to to be accessible.

src/middleware/normalize.rs
```diff
@@ -150,14 +180,36 @@ mod tests {
         assert!(res4.status().is_success());
     }
 
+    #[actix_rt::test]
+    async fn trim_trailing_slashes() {
```
robjtede, line 183: Also include a test request without trailing slash.

src/middleware/normalize.rs
```diff
@@ -78,8 +105,11 @@ where
 
         let original_path = head.uri.path();
 
-        // always add trailing slash, might be an extra one
-        let path = original_path.to_string() + "/";
+        // Either adds a string to the end (duplicates will be removed anyways) or trims all slashes from the end
+        let path = match self.trim_last_slash {
+            false => original_path.to_string() + "/",
+            true => original_path.trim_end_matches(|c| c == '/').to_string(),
```
robjtede: I believe this is equivalent.

```suggestion
            true => original_path.trim_end_matches('/').to_string(),
```

src/middleware/normalize.rs
```diff
@@ -10,20 +10,29 @@ use regex::Regex;
 use crate::service::{ServiceRequest, ServiceResponse};
 use crate::Error;
 
-#[derive(Default, Clone, Copy)]
+/// To be used when constructing `NormalizePath` to define it's behavior.
+#[derive(Clone, Copy)]
+pub enum TrailingSlash {
```
robjtede, line 16: Can you stick a #[non_exhaustive] on this please. Unlikely folks will want to match on it but would let us do things in future without breaking changes.

robjtede (changes requested):
Looks really good so far. Just left some initial thoughts and obviously the checks are failing from the doc comment in condition.rs.

robjtede:
Thanks for those updates. Add a CHANGELOG.md entry and a note to MIGRATION.md. Also check the job failure to see what needs updating.

robjtede (approved):

robjtede:
Great work, thanks 👍

patrickelectric:
Hey, any plan to release it ? It's an awesome feature, I was able to have a `/path|path/` to the same get method, and now with 2.0.0 appears to be not possible at all sadly.

PR #1641, opened 2020-08-21
extract more config types from Data<T> as well

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
Follow up to #1610. Extracts Form and Json config objects as T and then tries Data\<T\> as well.

### DIFF
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -20,7 +20,7 @@ use crate::dev::Decompress;
 use crate::error::{Error, JsonPayloadError};
 use crate::extract::FromRequest;
 use crate::request::HttpRequest;
-use crate::responder::Responder;
+use crate::{responder::Responder, web};
 
 /// Json helper
 ///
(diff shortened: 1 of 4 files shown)


robjtede:
@actix/contributors could use your help to get v3 over the last couple hurdles :D

src/types/json.rs
```diff
@@ -283,15 +286,26 @@ impl JsonConfig {
...
+    /// back to the default payload config.
+    fn from_req(req: &HttpRequest) -> &Self {
+        req.app_data::<Self>()
+            .or_else(|| req.app_data::<web::Data<Self>>().map(|d| d.as_ref()))
+            .unwrap_or_else(|| &DEFAULT_CONFIG)
+    }
 }
 
+// Allow shared refs to default.
+static DEFAULT_CONFIG: JsonConfig = JsonConfig {
```
fakeshadow: would a const be better fit here?

fakeshadow (approved):
looks great to me and I like this approach than a single type of app state

src/types/json.rs
robjtede: nice spot, changed for Payload<T> too

PR #1644, opened 2020-08-23
PoC: multipart form derive

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
Currently, multipart forms are quite difficult to wrangle. Requiring reasonably detailed knowledge of the streaming interface to extract fields and no standard serde interface for deserializing, for example. This has good reason, it would be prone to abuse from large files filling memory, especially since multipart forms are used primarily for file uploads; often these files are binary and not necessarily suitable for "deserializing".

The goal of this PR is to introduce a new derive crate for multipart that allows simpler declaration of field names, value types, maximum field sizes, content type assertions, and anything else that might be useful for multipart extraction.

The design is subject to change, but at the moment it uses a custom FromRequest impl on the Form struct, a builder struct for the Form, and a trait that allows building values from chunks of Bytes.

Simple Example:

```rust
#[derive(Debug, Clone, Default, MultipartForm)]
struct Form {
    name: String,

    #[multipart(max_size = 1024)]
    file: BytesMut,
}

#[post("/")]
async fn handler(form: Form) -> &'static str {
    println!("{:?}", &form);
    "Hello world!\r\n"
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ members = [
   "actix-http",
   "actix-files",
   "actix-multipart",
+  "actix-multipart-derive",
   "actix-web-actors",
   "actix-web-codegen",
   "test-server",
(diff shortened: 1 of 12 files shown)


robjtede:
decided to release this not under actix org initially

PR #1654, opened 2020-09-07
Improve awc websocket docs

popzxc:
## PR Type

Docs

## PR Checklist

## Overview

When I first tried to use `awc`, I had some problems with understanding how to use it with WebSocket.

I thought an example will be very helpful for readers.

### DIFF
--- a/awc/src/lib.rs
+++ b/awc/src/lib.rs
@@ -193,7 +193,8 @@ impl Client {
         self.request(Method::OPTIONS, url)
     }
 
-    /// Construct WebSockets request.
+    /// Initialize a WebSockets connection.
+    /// This method returns a builder structure to establish a WebSocket connection.
     pub fn ws<U>(&self, url: U) -> ws::WebsocketsRequest
     where
         Uri: TryFrom<U>,
--- a/awc/src/lib.rs
+++ b/awc/src/lib.rs
@@ -193,8 +193,8 @@ impl Client {
         self.request(Method::OPTIONS, url)
     }
 
-    /// Initialize a WebSockets connection.
-    /// This method returns a builder structure to establish a WebSocket connection.
+    /// Initialize a WebSocket connection.
+    /// Returns a WebSocket connection builder.
     pub fn ws<U>(&self, url: U) -> ws::WebsocketsRequest
     where
         Uri: TryFrom<U>,
--- a/awc/src/ws.rs
+++ b/awc/src/ws.rs
@@ -1,7 +1,6 @@
 //! Websockets client
 //!
-//! This module contains type definitions required to use [`awc::Client`](../struct.Client.html) as a WebSocket client.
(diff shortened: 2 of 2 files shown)


popzxc:
Hm. I'm not sure why CI fails on windows, but passes on Linux / macOS. It doesn't seem that doctest has some platform-specific code.

Neopallium:
Looks like an issue getting a cert for TLS:

"the handshake failed: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl\\statem\\statem_clnt.c:1924:: unable to get local issuer certificate"

I don't think it is related to this PR.

Neopallium:
Would it work with "ws://echo.websocket.org" ws instead of wss?

awc/src/lib.rs
```diff
@@ -193,7 +193,8 @@ impl Client {
         self.request(Method::OPTIONS, url)
     }
 
-    /// Construct WebSockets request.
+    /// Initialize a WebSockets connection.
```
robjtede:
```suggestion
    /// Initialize a WebSocket connection.
```

awc/src/lib.rs
```diff
@@ -193,7 +193,8 @@ impl Client {
...
+    /// This method returns a builder structure to establish a WebSocket connection.
```
robjtede:
```suggestion
    /// Returns a WebSocket connection builder.
```

awc/src/ws.rs
```diff
@@ -1,4 +1,33 @@
 //! Websockets client
+//!
+//! This module contains type definitions required to use [`awc::Client`](../struct.Client.html) as a WebSocket client.
```
robjtede:
```suggestion
//! Type definitions required to use [`awc::Client`](../struct.Client.html) as a WebSocket client.
```

awc/src/ws.rs
```diff
@@ -1,4 +1,33 @@
...
+//! To use `awc::Client`
```
robjtede:
```suggestion
```

Neopallium (approved):
Looks good.

awc/src/ws.rs
```diff
@@ -1,4 +1,33 @@
...
+//!     connection
+//!         .send(ws::Message::Text("Echo".to_string()))
+//!         .await
+//!         .unwrap();
+//!     let response = connection.next().await.unwrap().unwrap();
+//!
+//!     assert_eq!(response, ws::Frame::Text("Echo".as_bytes().into()));
+//! }
+//! ```
+//!
```
robjtede:
```suggestion
```

robjtede (changes requested):
just some wording adjustments, otherwise LGTM

robjtede (approved):
thanks for the contribution 👍

PR #1656, opened 2020-09-08
awc: improve module documentation

popzxc:
## PR Type

Docs

## PR Checklist
Check your PR fulfills the following:

- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [ ] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview

Subjectively, one great part of `reqwest` is its documentation, which provides usage examples for most common users' needs right on the main page of `docs.rs`.

This PR extends the `awc main doc-comment with the similar list of examples, so that users can see how to use library without digging deep into the crate docs (or even the source code).

PR #1659, opened 2020-09-09
prepare release beta 4

robjtede:
i think this will be the last beta 🚀


Neopallium (approved):
looks good.

popzxc (approved):
Can it wait for #1656 to get merged, so that new version will have better docs already?

Issue #1237, opened 2019-12-26
JSON Deserialization errors are no longer outputted by default


Darkeye9:
Hi all and extensive congratulation on putting v2.0.0 out! Support for async/await and updated dependencies were all very welcomed by our team.

I wanted to raise a question, more than a bug. I noticed, that in the first project were are doing with the updated `actix_web 2.0.0`, JSON deserialization errors that used to be included in the HTTP 400 responses are now gone.

We did not change any of our code handling errors and neither it seems that we were changing actix-web native error handling, so I am supposing it has to do with the new release, as we were tracking new releases of actix-web, and only now I am missing those messages.

For reference, the messages I am talking about are the ones that occur when a JSON body does not conform to some input model an endpoint requires. They were along the lines of:

```
Json deserialize error: missing field `description` at line 3 column
```

I set the project logging configuration to TRACE, and I could not see any JSON validation error. So this seems like a developer-unfriendly failure here. By default, the HTTP error messages should come back, or a log line should be outputted, for developer sanity.

Had a good christmas and tell me what you think! :santa:

fafhrd91:
`std::env::set_var("RUST_LOG", "actix_web=trace");`

```
[2019-12-26T17:13:21Z DEBUG actix_web::types::json] Failed to deserialize Json from payload. Request path: /extractor
[2019-12-26T17:13:21Z DEBUG actix_web::middleware::logger] Error in response: Deserialize(Error("EOF while parsing a value", line: 1, column: 0))
[2019-12-26T17:13:21Z INFO  actix_web::middleware::logger] 127.0.0.1:52676 "POST /extractor HTTP/1.1" 400 0 "-" "curl/7.64.1" 0.000573
```

Darkeye9:
Ok, I can now see that message. I am amazed by that not being in the `error` level, or at least, something higher than `debug`.

Is the output in the HTTP Response silenced on purpose in this release? And can I enable it back?

Greetings and thank you for your time ;)

fafhrd91:
Nothing changed since 1.0 release

fafhrd91:
You can build your own extractor, it is easy

surfingtomchen:
Hi,

I just find the extractor way to solve this issue is a little bit complicated, because I need to write multiple extractors for each JSON structure. For example:

there is a user structure in user.rs:

``` rust
pub struct User {
    pub _id: Option<ObjectId>,
    pub name: String,
    pub password: String,
}
```

error.rs:

``` rust
[...]

PR #1661, opened 2020-09-10
Improve json, form and query extractor config docs

Neopallium:
Improve the `JsonConfig` docs to show example of both a global and per-type json extractor configuration.

### DIFF
--- a/src/types/json.rs
+++ b/src/types/json.rs
@@ -209,9 +209,7 @@ where
 
 /// Json extractor configuration
 ///
-/// # Examples
-///
-/// Global extractor configuration:
+/// # Example
 ///
 /// ```rust
 /// use actix_web::{error, web, App, FromRequest, HttpResponse};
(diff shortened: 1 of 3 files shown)


Neopallium:
See issue #1237

src/types/json.rs
```diff
@@ -211,6 +211,42 @@ where
...
+/// struct Info {
+///     username: String,
+/// }
+///
+/// /// deserialize `Info` from request's body, max payload size is 4kb
+/// async fn index(info: web::Json<Info>) -> String {
+///     format!("Welcome {}!", info.username)
+/// }
+///
+/// fn main() {
```
popzxc, line 228: `fn main()` is not required in doc-tests (unless it's `async`).

src/types/json.rs
```diff
@@ -242,7 +278,7 @@ where
 ///     let app = App::new().service(
 ///         web::resource("/index.html")
 ///             .app_data(
-///                 // change json extractor configuration
+///                 // Json extractor configuration for only the `Info` struct.
```
popzxc: IMHO it would read better like that:

```suggestion
///                 // JSON extractor configuration for the `Info` struct only.
```

src/types/json.rs
```diff
@@ -211,6 +211,42 @@ where
 ///
 /// # Examples
 ///
+/// Global extractor configuration:
```
popzxc: I think it would be right to add an empty line between text and code sample.

src/types/json.rs
Neopallium: These are example for the users.  Shouldn't it show `main()`?

popzxc: Well, the `main` function doesn't contain any logic aside of creating an `App`.
I believe it's a matter of taste, but I think that `main` function should not be shown unless it provides more context / meaning for the user.

Neopallium: The other doc examples still include the `main` function, so I will keep it for now.

robjtede (approved):
~Looks great, thanks~

Edit: see below

robjtede:
> Looks great...

I got slightly suspicious after approving about how the Json struct knew wether to use the "global" config or it's own "specific" config. I've touched some of the config code recently and had to investigate.

The associated config type for `Json<T>` is `JsonConfig` which is not generic over `T` which means that the extraction phase won't distinguish between config set with `JsonConfig...` and `Json<T>::configure...`. Therefore, using `Json<T>::configure...` is misleading and should be replaced with using `JsonConfig`.

Example program below, you'll see that 2 routes using different structs are affected by the same limit set by only 1 struct's configuration.

```rust
use actix_web::{
    error, post, web, App, FromRequest, HttpRequest, HttpResponse, HttpServer, Responder,
};
use serde::Deserialize;

#[derive(Deserialize)]
struct Info {
    name: String,
}

#[post("/1")]
async fn greet(name: web::Json<Info>) -> impl Responder {
    HttpResponse::Ok().body(format!("Hello {}!", name.name))
}

#[derive(Deserialize)]
struct Info2 {
    name: String,
}

#[post("/2")]
async fn greet2(name: web::Json<Info2>) -> impl Responder {
    HttpResponse::Ok().body(format!("Hello {}!", name.name))
}

fn json_error_handler(err: error::JsonPayloadError, _req: &HttpRequest) -> error::Error {
    use actix_web::error::JsonPayloadError;

    let detail = err.to_string();
    let resp = match &err {
        JsonPayloadError::ContentType => {
            HttpResponse::UnsupportedMediaType().body(detail)
        }
        JsonPayloadError::Deserialize(json_err) if json_err.is_data() => {
            HttpResponse::UnprocessableEntity().body(detail)
        }
        _ => HttpResponse::BadRequest().body(detail),
    };
    error::InternalError::from_response(err, resp).into()
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .app_data(web::Json::<Info>::configure(|cfg| {
                cfg.error_handler(json_error_handler).limit(20)
            }))
            .service(greet)
            .service(greet2)
    })
    .bind("127.0.0.1:8088")?
    .run()
    .await
}
```

Neopallium:
Should we just remove the example for "Per-type" configuration, since it seems that is broken?  Or do we want to try fixing it before the final 3.0 release? @robjtede

robjtede:
> Should we just remove the example for "Per-type" configuration

Yes. I think that's best.

> try fixing it before the final 3.0 release

I wouldn't say it's broken really, since it's possible to apply these configs per route which is effectively the same as a config per form type.

Neopallium:
How about the Form/Query extractor configuration examples?  They have the same problem.  I can convert them to the generic "global" format.

robjtede:
> How about the Form/Query extractor configuration examples

Yep, good idea.

robjtede (approved):

Issue #1560, opened 2020-06-10
actix_web::client doesn't work on windows


Pzixel:
## Steps to Reproduce (for bugs)
Code:

```rust
use actix_web::client::Client;

#[actix_rt::main]
async fn main() {
    let client = Client::default();

    // Create request builder and send request
    let response = client
        .get("https://www.rust-lang.org") // <--- notice the "s" in "https://..."
        .header("User-Agent", "Actix-web")
        .send()
        .await; // <- Send http request

    println!("Response: {:?}", response);
}
```

When toml is:

```toml
[dependencies]
actix-rt = "1.1.1"
actix-web = { version = "3.0.0-alpha.3", features=["rustls"] }
rustls = "0.17.0"
```

Error is

```
2020-05-15T13:42:33.392505900+03:00 DEBUG trust_dns_resolver::async_resolver::background - trust-dns resolver running
2020-05-15T13:42:33.398505100+03:00 DEBUG trust_dns_proto::xfer::dns_handle - querying: www.rust-lang.org A
2020-05-15T13:42:33.399506400+03:00 DEBUG trust_dns_resolver::name_server::name_server - reconnecting: NameServerConfig { socket_addr: V6([fec0:0:0:ffff::1]:53), protocol: Udp, tls_dns_name: None }
2020-05-15T13:42:33.399506400+03:00 DEBUG trust_dns_resolver::name_server::name_server - reconnecting: NameServerConfig { socket_addr: V6([fec0:0:0:ffff::2]:53), protocol: Udp, tls_dns_name: None }
2020-05-15T13:42:33.399506400+03:00 DEBUG trust_dns_resolver::name_server::connection_provider - connecting: Udp { socket_addr: V6([fec0:0:0:ffff::1]:53), timeout: 5s }
2020-05-15T13:42:33.400506700+03:00 DEBUG trust_dns_proto::xfer - enqueueing message: [Query { name: Name { is_fqdn: false, labels: [www, rust-lang, org] }, query_type: A, query_class: IN }]
2020-05-15T13:42:33.400506700+03:00 DEBUG trust_dns_resolver::name_server::connection_provider - connecting: Udp { socket_addr: V6([fec0:0:0:ffff::2]:53), timeout: 5s }
2020-05-15T13:42:33.400506700+03:00 DEBUG trust_dns_proto::xfer - enqueueing message: [Query { name: Name { is_fqdn: false, labels: [www, rust-lang, org] }, query_type: A, query_class: IN }]
[...]

PR #1662, opened 2020-09-10
ameliorate awc timeouts on windows

robjtede:
## PR Type
Bug Workaround

## PR Checklist

## Overview
Works around issues on Windows seemingly caused by slow DNS lookups by doubling the default timeout period when built for Windows.

See conversation in #1560 

cc: @Pzixel @popzxc

### DIFF
--- a/awc/CHANGES.md
+++ b/awc/CHANGES.md
@@ -1,7 +1,8 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
-
+### Fixed
+* Reduced chance of timeouts on Windows.
 
 ## 2.0.0-beta.4 - 2020-09-09
 ### Changed
--- a/awc/src/builder.rs
+++ b/awc/src/builder.rs
@@ -40,7 +40,14 @@ impl ClientBuilder {
             allow_redirects: true,
             max_redirects: 10,
             headers: HeaderMap::new(),
+
+            #[cfg(not(target_os = "windows"))]
             timeout: Some(Duration::from_secs(5)),
+
+            // HACK: Issue #1560 - ameliorate connect timeouts from DNS lookups
+            #[cfg(target_os = "windows")]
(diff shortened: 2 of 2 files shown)


popzxc (approved):
I guess, it is now possible to make `awc` doctests use `https`. However, I'm not sure if it's right thing to do, as the Windows tests will still be very slow and may randomly fail.

Neopallium (approved):

robjtede:
> However, I'm not sure if it's right thing to do...

I'm not totally sure either, but the other thing to do for now is nothing... :/

Pzixel:
I still believe 10 sec delay for any request is insane. It doesn't actually fix the problem as I see it, it only changes its kind

popzxc:
But as we already pointed before, the problem is not located within any of the actix crates.

Proper ways of resolving this issue (which are either replacing the dependency or fixing the issue in the dependency, and then waiting for it to release a new version) will take much longer.

IMHO, slowly working code is better that code that doesn't work. So I think that it's an acceptable temporary workaround.

By the way, any help with a proper solution will be really appreciated 🙃

Pzixel:
I aggree. But I don't think 10 sec NS lookup is anywhere near acceptable.

So I'd propose to either not deliver the crate on windows at all (with CFG and explanation) or no actions to "fix" it in this crate. It's indeed dependencies problem, so the proper solution is fixing/replacing dependency. I probably should create an issue in this dependency repository if you didn't already. 

Any fix is appreciated if it helps, but this one just changes one problem (not working at all) to another (working inapplicable slowly). 

But I'm not strong on this point so you may right with this fix

cetra3:
Looks like we're working around a bug in `trust-dns` so I'm not in support of this change.  I have had issues with that crate before, so I think a more appropriate fix is to raise it in there.

Rademenes16:
Yes

PR #1667, opened 2020-09-11
fix clippy async_yields_async lints

robjtede:
## PR Type
Clippy

## PR Checklist

## Overview
Addresses a (new?) clippy lint `async_yields_async`. It's complaint is that return types from `HttpResponseBuilder::{json, body}` are themselves futures and don't require the async block.


popzxc (approved):
LGTM

PR #1671, opened 2020-09-11
Don't hand-roll slice.align_to_mut(), use the stdlib function

Shnatsel:
## PR Type
Refactor

## PR Checklist

## Overview
Use the standard library function [`slice::align_to_mut()`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.align_to_mut) that's available since Rust 1.30.0 instead of hand-rolling a custom implementation of it. Event Debian Stable ships Rust 1.34.2 now, so the minimum supported Rust version should not be a blocker.

I'll add the changelog entry if this gets the go-ahead.

### DIFF
--- a/actix-http/src/ws/mask.rs
+++ b/actix-http/src/ws/mask.rs
@@ -66,43 +66,16 @@ fn xor_short(buf: ShortSlice<'_>, mask: u64) {
     }
 }
 
-/// # Safety
-/// Caller must ensure the buffer has the correct size and alignment.
-#[inline]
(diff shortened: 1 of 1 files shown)


Shnatsel:
Huh, I've accidentally constructed ShortSlice bypassing the constructor with the checks. Which is apparently possible.

Shnatsel:
Oh wait, there is _other_ unsafe code that depends on the invariant of short slices being less than 9 bytes in length, which the standard library function does not provide. Nevermind then, this would be a bigger refactoring job than I anticipated.

PR #1673, opened 2020-09-12
make TrailingSlash enum accessible

### LINKED ISSUE #1675: TrailingSlash isn't exported
#1639 added `TrailingSlash` to configure the behaviour of `NormalizePath` middleware. However, as far as I can tell the type isn't actually exported (`middleware::normalize` is a private module) so there's no way to use it.

Lesiuk:
## PR Type
Bug Fix

## PR Checklist

## Overview
Wanted to try middleware::NormalizePath::new(TrailingSlash::Trim) introduced in 3.0.0 but I noticed that TrailingSlash enum is inside private package and It's not currently possible.


robjtede:
Needs a cargo fmt.

Can we add a test to tests/test_server.rs something like:

```rust
#[actix_rt::test]
async fn test_normalize() {
    let srv = test::start_with(test::config().h1(), || {
        App::new()
            .wrap(Compress::new(NormalizePath::new(TrailingSlash::Trim)))
            .service(web::resource("/one").route(web::to(|| HttpResponse::Ok().finish())))
    });

    let mut response = srv
        .get("/one/")
        .send()
        .await
        .unwrap();
    assert!(response.status().is_success());
}
```

robjtede:
Just a changelog entry and we can get this released.

robjtede (approved):

PR #1674, opened 2020-09-12
Provide attribute macro for multiple HTTP methods

### LINKED ISSUE #1360: attributes macros for multiple methods
How does one use the route attributes macros (`#[http_method("route")]`) to bind multiple http methods to one function?

For example, I have this trivial endpoint:
```rust
/// Returns a UUID4.
#[get("/uuid")]
async fn uuid_v4() -> impl Responder {
    HttpResponse::Ok().json(Uuid {
        uuid: uuid::Uuid::new_v4(),
    })
}
```
I would like to have the same endpoint handle `HEAD` requests, **how do I do this?**
My initial approach was to just stack up the macros:
```rust
/// Returns a UUID4.
#[get("/uuid")]
#[head("/uuid")]
async fn uuid_v4() -> impl Responder {
    HttpResponse::Ok().json(Uuid {
        uuid: uuid::Uuid::new_v4(),
    })
}
```
But I do get a compilation error:
```
    |
249 | async fn uuid_v4() -> impl Responder {
    |          ^^^^^^^ the trait `actix_web::handler::Factory<_, _, _>` is not implemented for `<uuid_v4 as actix_web::service::HttpServiceFactory>::register::uuid_v4`
```

_**I have gone through the actix-web and actix-web-codegen docs and didn't find anything addressing this_

mattgathu:
## PR Type
Feature

## PR Checklist

## Overview
Provide attribute macro for multiple HTTP methods

Define a new `route` attribute macro that supports defining multiple
HTTP methods to be routed to (handled by) a single handler.

The attribute macro syntax looks like this
```rust
use actix_web::route;

#[route("/multi", method = "GET", method = "POST")]
async fn multi_methods() -> &'static str {
    "Hello world!\r\n"
}
```

This implementation extends the [`GuardType`][1] enum in actix-web-codegen to have a
new `GuardType::Multi` variant that denotes when multiple method guards
are used.

A new `method` attribute in the `route` attribute macro provides an  HTTP method to provide a guard for.
This attribute can be used multiple times for multiple HTTP methods.

The code matches the methods to the respective
`GuardType` and uses the [`AnyGuard`][2] struct to combine them together.

The generated code looks like this:
```rust
pub struct multi_methods;
impl actix_web::dev::HttpServiceFactory for multi_methods {
    fn register(self, __config: &mut actix_web::dev::AppService) {
        async fn multi_methods() -> &'static str {
            "Hello world!\r\n"
        }
        let __resource = actix_web::Resource::new("/multi")
            .name("multi_methods")
            .guard(actix_web::guard::Any(actix_web::guard::Get()).or(actix_web::guard::Post()))
            .to(multi_methods);
        actix_web::dev::HttpServiceFactory::register(__resource, __config)
    }
}
```

This Closes https://github.com/actix/actix-web/issues/1360

[1]: https://github.com/actix/actix-web/blob/master/actix-web-codegen/src/route.rs#L21
[2]: https://github.com/actix/actix-web/blob/master/src/guard.rs#L104s

### DIFF
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -145,6 +145,8 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
 ///
 /// Syntax: `#[route("path"[, attributes])]`
 ///
+/// Example: `#[route("/", method="GET", method="HEAD")]`
+///
 /// ## Attributes
 ///
 /// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -141,6 +141,20 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
     route::generate(args, input, route::GuardType::Patch)
 }
 
+/// Creates route handler with Multiple HTTP methods guards.
+///
+/// Syntax: `#[route("path"[, attributes])]`
+///
+/// ## Attributes
+/// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
+/// - `methods="HTTP_METHOD_1,HTTP_METHOD_2"` - Registers HTTP methods to provide guards for.
+/// - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
+/// - `wrap="Middleware"` - Registers a resource middleware.
+#[proc_macro_attribute]
+pub fn route(args: TokenStream, input: TokenStream) -> TokenStream {
+    route::generate(args, input, route::GuardType::Multi)
+}
+
 /// Marks async main function as the actix system entry-point.
 ///
 /// ## Usage
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -147,7 +147,7 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
 ///
 /// ## Attributes
 /// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
-/// - `methods="HTTP_METHOD_1,HTTP_METHOD_2"` - Registers HTTP methods to provide guards for.
+/// - `method="HTTP_METHOD"` - Registers HTTP method to provide guard for.
 /// - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
 /// - `wrap="Middleware"` - Registers a resource middleware.
 #[proc_macro_attribute]
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -141,7 +141,7 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
     route::generate(args, input, route::GuardType::Patch)
 }
 
-/// Creates route handler with Multiple HTTP methods guards.
+/// Creates resource handler, allowing multiple HTTP method guards.
 ///
 /// Syntax: `#[route("path"[, attributes])]`
 ///
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -146,6 +146,7 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
 /// Syntax: `#[route("path"[, attributes])]`
 ///
 /// ## Attributes
+///
 /// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
 /// - `method="HTTP_METHOD"` - Registers HTTP method to provide guard for.
 /// - `guard="function_name"` - Registers function as guard using `actix_web::guard::fn_guard`
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -1,6 +1,7 @@
 extern crate proc_macro;
 
 use std::collections::HashSet;
+use std::convert::TryFrom;
 
 use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
@@ -57,6 +58,28 @@ impl ToTokens for GuardType {
     }
 }
 
+impl TryFrom<&syn::LitStr> for GuardType {
+    type Error = syn::Error;
+
+    fn try_from(value: &syn::LitStr) -> Result<Self, Self::Error> {
+        match value.value().as_str() {
+            "CONNECT" => Ok(GuardType::Connect),
+            "DELETE" => Ok(GuardType::Delete),
+            "GET" => Ok(GuardType::Get),
+            "HEAD" => Ok(GuardType::Head),
+            "OPTIONS" => Ok(GuardType::Options),
+            "PATCH" => Ok(GuardType::Patch),
+            "POST" => Ok(GuardType::Post),
+            "PUT" => Ok(GuardType::Put),
+            "TRACE" => Ok(GuardType::Trace),
+            _ => Err(syn::Error::new_spanned(
+                value,
+                &format!("Unexpected HTTP Method: `{}`", value.value()),
+            )),
+        }
+    }
+}
+
 struct Args {
     path: syn::LitStr,
     guards: Vec<Ident>,
@@ -104,34 +127,12 @@ impl Args {
                         }
                     } else if nv.path.is_ident("method") {
                         if let syn::Lit::Str(ref lit) = nv.lit {
-                            let guard_type: Option<GuardType> =
-                                match lit.value().as_str() {
-                                    "CONNECT" => Some(GuardType::Connect),
-                                    "DELETE" => Some(GuardType::Delete),
-                                    "GET" => Some(GuardType::Get),
-                                    "HEAD" => Some(GuardType::Head),
-                                    "OPTIONS" => Some(GuardType::Options),
-                                    "PATCH" => Some(GuardType::Patch),
-                                    "POST" => Some(GuardType::Post),
-                                    "PUT" => Some(GuardType::Put),
-                                    "TRACE" => Some(GuardType::Trace),
-                                    _ => None,
-                                };
-                            if let Some(guard) = guard_type {
-                                if !methods.insert(guard) {
-                                    return Err(syn::Error::new_spanned(
-                                        &nv.lit,
-                                        &format!(
-                                            "HTTP Method defined more than once: `{}`",
-                                            lit.value()
-                                        ),
-                                    ));
-                                }
-                            } else {
+                            let guard = GuardType::try_from(lit)?;
+                            if !methods.insert(guard) {
                                 return Err(syn::Error::new_spanned(
                                     &nv.lit,
                                     &format!(
-                                        "Unexpected HTTP Method: `{}`",
+                                        "HTTP Method defined more than once: `{}`",
                                         lit.value()
                                     ),
                                 ));
@@ -262,7 +263,7 @@ impl ToTokens for Route {
         let resource_name = name.to_string();
         let methods = methods.iter();
 
-        let guard_gen = if guard == &GuardType::Multi {
+        let guard_gen = if *guard == GuardType::Multi {
             quote! {
                 .guard(actix_web::guard::AnyGuard::new(vec![#(Box::new(actix_web::guard::#methods())),*]))
             }
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -102,7 +102,7 @@ impl Args {
                         }
                     } else if nv.path.is_ident("method") {
                         if let syn::Lit::Str(ref lit) = nv.lit {
-                            match lit.value().to_uppercase().as_str() {
+                            match lit.value().as_str() {
                                 "CONNECT" => methods.push(GuardType::Connect),
                                 "DELETE" => methods.push(GuardType::Delete),
                                 "GET" => methods.push(GuardType::Get),
@@ -202,7 +202,7 @@ impl Route {
         if guard == GuardType::Multi && args.methods.is_empty() {
             return Err(syn::Error::new(
                 Span::call_site(),
-                "The #[route(..)] macro requires at least one `method` attribute!",
+                "The #[route(..)] macro requires at least one `method` attribute",
             ));
         }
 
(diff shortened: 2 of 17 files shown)


Silentdoer:
I think this api should be:
```rust
#[route("/multi", method = "GET", method = "POST")]
async fn multi_methods() -> &'static str {
    "Hello world!\r\n"
}
```

otherwise it's not consistent with:
```rust
#[get("/index2", wrap = "middleware::SayHi", wrap = "middleware::SayHi2")]
```

Silentdoer:
or may be:
```rust
#[route("/multi", methods = ["GET","POST"])]
async fn multi_methods() -> &'static str {
    "Hello world!\r\n"
}
```

mattgathu:
Thanks for the feedback @Silentdoer 

My initial approach was `#[route("/multi", methods = ["GET","POST"])]` but unfortunately only literal arguments are allowed in this position, and that ruled out the `[...]` syntax.

I can reimplement the macro to have the `#[route("/multi", method = "GET", method = "POST")]` syntax. If the core contributor(s) can affirm that this is the preferred syntax. maybe @robjtede or @JohnTitor

mattgathu:
I've updated the syntax to be `#[route("/multi", method = "GET", method = "POST")]`

actix-web-codegen/src/route.rs
```diff
@@ -96,10 +100,38 @@ impl Args {
                                 "Attribute wrap expects type",
                             ));
                         }
+                    } else if nv.path.is_ident("method") {
+                        if let syn::Lit::Str(ref lit) = nv.lit {
+                            match lit.value().to_uppercase().as_str() {
```
popzxc: I don't think that `to_uppercase()` is required. HTTP methods are commonly written in upper case and supporting values such as "cOnNeCt" or "Get" may be confusing.

actix-web-codegen/src/route.rs
```diff
@@ -57,13 +59,15 @@ struct Args {
...
     wrappers: Vec<syn::Type>,
+    methods: Vec<GuardType>,
 }
 
 impl Args {
     fn new(args: AttributeArgs) -> syn::Result<Self> {
         let mut path = None;
         let mut guards = Vec::new();
         let mut wrappers = Vec::new();
+        let mut methods = Vec::new();
```
popzxc: What will happen if user will provide multiple methods with the same name? E.g. `method = "POST", method = "POST"`.

actix-web-codegen/src/route.rs
```diff
@@ -166,6 +199,13 @@ impl Route {
 
         let args = Args::new(args)?;
 
+        if guard == GuardType::Multi && args.methods.is_empty() {
+            return Err(syn::Error::new(
+                Span::call_site(),
+                "The #[route(..)] macro requires at least one `method` attribute!",
```
popzxc: Let's keep ourselves calm 🙃 

```suggestion
                "The #[route(..)] macro requires at least one `method` attribute",
```

actix-web-codegen/src/route.rs
```diff
@@ -201,25 +241,47 @@ impl ToTokens for Route {
...
-                    #ast
-                    let __resource = actix_web::Resource::new(#path)
-                        .name(#resource_name)
-                        .guard(actix_web::guard::#guard())
-                        #(.guard(actix_web::guard::fn_guard(#guards)))*
-                        #(.wrap(#wrappers))*
-                        .#resource_type(#name);
-
-                    actix_web::dev::HttpServiceFactory::register(__resource, __config)
+        let stream = if guard != &GuardType::Multi {
```
popzxc: `if` branches differ in one line only. I suggest generating only that line inside `if`, and then use it as a part of the shared `quote`, like:

```rust
let guard = if guard != &GuardType::Multi {
    // ...
} else {
    // ...
};

let stream = quote! { ... };
```

actix-web-codegen/src/route.rs
mattgathu: we would register an extra redundant guard. This can be avoided by making the methods vec a set or returning a compile time error. _not sure which is the better approach_

popzxc: I would prefer using a set AND returning an error if method is met more than once. Luckily, `HashSet::insert` returns a bool indicating whether this method was already present inside.

actix-web-codegen/src/lib.rs
```diff
@@ -141,6 +141,20 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
     route::generate(args, input, route::GuardType::Patch)
 }
 
+/// Creates route handler with Multiple HTTP methods guards.
+///
+/// Syntax: `#[route("path"[, attributes])]`
+///
+/// ## Attributes
+/// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
```
popzxc:
```suggestion
/// ## Attributes
///
/// - `"path"` - Raw literal string with path for which to register handler. Mandatory.
```

actix-web-codegen/src/route.rs
```diff
@@ -201,10 +255,23 @@ impl ToTokens for Route {
...
                     guards,
                     wrappers,
+                    methods,
                 },
             resource_type,
         } = self;
         let resource_name = name.to_string();
+        let methods = methods.iter();
+
+        let guard_gen = if guard == &GuardType::Multi {
```
popzxc: For a one-byte types it's better to derive `Copy` and compare them by value, rather than using references.

actix-web-codegen/src/route.rs
```diff
@@ -96,10 +102,50 @@ impl Args {
                                 "Attribute wrap expects type",
                             ));
                         }
+                    } else if nv.path.is_ident("method") {
+                        if let syn::Lit::Str(ref lit) = nv.lit {
```
popzxc, line 129: Would it make sense to provide an implementation of `TryFrom<&syn::Lit> for GuardType`?

With that, the code here will be simplified to just:

```rust
let guard = GuardType::try_from(&nv.lit)?;
if !methods.insert(guard) { ... };
```

mattgathu: I managed this:
```rust
                    } else if nv.path.is_ident("method") {
                        if let syn::Lit::Str(ref lit) = nv.lit {
                            let guard = GuardType::try_from(lit)?;
                            if !methods.insert(guard) {
                                return Err(syn::Error::new_spanned(
                                    &nv.lit,
                                    &format!(
                                        "HTTP Method defined more than once: `{}`",
                                        lit.value()
                                    ),
                                ));
                            }
                        } else {
                            return Err(syn::Error::new_spanned(
                                nv.lit,
                                "Attribute method expects literal string!",
                            ));
                        }
                    } else {
``` 
So doing the conversation from `&syn::LitStr` because using `&syn::Lit` would mean having the `"Attribute method expects literal string!"` error inside the `GuardType` conversion which seems wrong.

actix-web-codegen/src/route.rs
mattgathu: I guess I can do an deref `let guard_gen = if *guard == GuardType::Multi`?

popzxc: Isn't it possible to destruct a `Copy` field by value from a reference? If not, then yep, deref seems right.

actix-web-codegen/src/route.rs
popzxc: Makes sense.

robjtede:
Hey @mattgathu, it's great to see you tackling this!

I'd like to land #1677 first so we can test some failure cases in CI and make sure the compile error outputs make sense.

actix-web-codegen/src/route.rs
mattgathu: I still need to a deref even when I derive` Copy` and `Clone` for `GuardType`. I'm guessing the compiler doesn't treat enums as primitives types such as i32 or u8 ¯\_(ツ)_/¯

popzxc (approved):
Looks really nice. Thanks!

robjtede:
@mattgathu #1677 is merged so it'd be great to get some trybuild tests added as part of this. Try to get in as many weird edge case failures as you can think of.

src/guard.rs
```diff
@@ -104,6 +104,10 @@ pub fn Any<F: Guard + 'static>(guard: F) -> AnyGuard {
 pub struct AnyGuard(Vec<Box<dyn Guard>>);
 
 impl AnyGuard {
+    /// Create AnyGuard from a vector of Guards
+    pub fn new(guards: Vec<Box<dyn Guard>>) -> Self {
+        AnyGuard(guards)
+    }
```
robjtede: Seems to me this change is unnecessary, it should be possible to use `Any(first_guard).or(second_guard)...`.

mattgathu:
@robjtede I got an issue with tests passing on **rustc 1.42.0** but failing on the **current stable**. 
Here is the diff:
```
EXPECTED:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error: The #[route(..)] macro requires at least one `method` attribute
##[error] --> $DIR/route-missing-method-fail.rs:3:1
  |
3 | #[route("/")]
  | ^^^^^^^^^^^^^

error[E0425]: cannot find value `index` in this scope
##[error]  --> $DIR/route-missing-method-fail.rs:10:49
   |
10 |     let srv = test::start(|| App::new().service(index));
   |                                                 ^^^^^ not found in this scope
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
ACTUAL OUTPUT:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error: The #[route(..)] macro requires at least one `method` attribute
##[error] --> $DIR/route-missing-method-fail.rs:3:1
  |
3 | #[route("/")]
  | ^^^^^^^^^^^^^
  |
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find value `index` in this scope
##[error]  --> $DIR/route-missing-method-fail.rs:10:49
   |
10 |     let srv = test::start(|| App::new().service(index));
   |                                                 ^^^^^ not found in this scope
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
note: If the actual output is the correct output you can bless it by rerunning
      your test with the environment variable TRYBUILD=overwrite

```
note the **= note: this error originates in an attribute macro...** line.

Do you have any suggestions on how to handle this?

src/guard.rs
mattgathu: I think it is necessary since there is no way (_at least for me_) to repeatedly call `Any(first_guard).or(second_guard)` when doing code generation using `quote!` in `actix-web-codegen`

robjtede:
> Do you have any suggestions on how to handle this?

Hmm... I'll have a look.

In the mean time, I feel that the macro should be called "resource" as it is more consistent with the manual routing methods. A resource can define multiple methods on a path where a route is a single method.

actix-web-codegen/src/lib.rs
```diff
@@ -141,6 +141,21 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
     route::generate(args, input, route::GuardType::Patch)
 }
 
+/// Creates route handler with Multiple HTTP methods guards.
```
robjtede:
```suggestion
/// Creates resource handler, allowing multiple HTTP method guards.
```

actix-web-codegen/src/lib.rs
```diff
@@ -141,6 +141,21 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
...
+///
+/// Syntax: `#[route("path"[, attributes])]`
```
robjtede, line 146: Part of this doc block should clearly show that this macro is designed to be used with multiple methods.

mattgathu:
> In the mean time, I feel that the macro should be called "resource" as it is more consistent with the manual routing methods. A resource can define multiple methods on a path where a route is a single method.

I would argue that `route` is more fitting as it expresses the intention: to route one or more http methods to the same handler.  It is also, I would argue, more consistent with the HTTP verbs `get`, `put` etc as it is also a *verb*.

`resource` is a noun and doesn't capture/express the user's intent. 

_I'm viewing `route` as a verb and not its noun meaning_

robjtede:
Alright, fair argument. Would like to get more opinons from @actix/contributors then.

When you read it as a verb it makes sense even though you'd have to sort of read it as `#[route_to("/abc")]`. I think there's value in keeping the nomenclature consitent with existing routing methods. But, again, more input here will be good.

actix-web-codegen/src/lib.rs
mattgathu: I've added an `Example` line, similar to the `Syntax` line.

robjtede:
I'm coming around to #[route(...)] for brevity sake.

In other news, I think I've found a solution to the compile error difference while letting us test on stable too. Great find with that rustversion macro; using that you can do this:

```rust
#[test]
fn compile_macros() {
    let t = trybuild::TestCases::new();

    t.pass("tests/trybuild/simple.rs");
    t.compile_fail("tests/trybuild/simple-fail.rs");

    t.pass("tests/trybuild/route-ok.rs");
    t.compile_fail("tests/trybuild/route-duplicate-method-fail.rs");
    t.compile_fail("tests/trybuild/route-unexpected-method-fail.rs");
    missing_method_fail_hack(&t);
}

#[rustversion::stable(1.42)]
fn missing_method_fail_hack(t: &trybuild::TestCases) {
    t.compile_fail("tests/trybuild/route-missing-method-fail-msrv.rs");
}

#[rustversion::not(stable(1.42))]
fn missing_method_fail_hack(t: &trybuild::TestCases) {
    t.compile_fail("tests/trybuild/route-missing-method-fail.rs");
}
```

Then have two versions of the test file...:
- route-missing-method-fail.rs
- route-missing-method-fail-msrv.rs

(Even cleaner, you can symlink the -msrv version to the original.)

...and two versions of the compiler output:
- route-missing-method-fail.stderr
- route-missing-method-fail-msrv.stderr

fakeshadow:
I'm not against this poll but would like to share my two cents.

I feel attributes should follow the rust std style. For example: 
#[route(method = "")]
#[route(any(method = "", method = "" ))]
#[route(not(method = ""))]

mattgathu:
@robjtede regarding the trybuild tests, the Actix-web's CI matrix has 1.42, current stable and nightly tests with the compilers never guaranteed to produce similar outputs. There will be a probability that each time nightly or the current stable is updated the tests will break. Pinning the tests to 1.42 avoids this likely breakage.

Having multiple tests for different compilers is great but might not be maintainable IMHO.

robjtede:
I think the maintenance cost is small enough using this approach, testing for MSRV and latest stable is enough; we should not run any trybuild tests on nightly.

Worst case we will need to update the stderr files every 6 weeks. But I think theres value in knowing when compiler output changes for our macro errors, anyway.

mattgathu:
@robjtede I haven't found a way to differentiate between the 1.42 stable and other stable versions using the rustversion macros :-( 

You have `#[rustversion::not(stable(1.42))]` in your example, but it doesn't eliminate nightly runs.

**Edit:** I ended up building a dummy shim and stacking the macros 🤞

robjtede:
Great! I'm happy to go ahead with that test set up.

Check out this commit (02e16c6538ae31351e7618a714d42180046e4e90) for how to do the multi guard without requiring changes to actix-web. The generated code looks like this for the example in that commit:

```rust
let __resource = actix_web::Resource::new("/")
    .name("index")
    .guard(
        actix_web::guard::Any(actix_web::guard::Get())
            .or(actix_web::guard::Trace())
            .or(actix_web::guard::Post())
            .or(actix_web::guard::Head()),
    )
    .to(index);
```

robjtede (approved):
Thanks so much for working on this 👍 great job!

robjtede:
@fakeshadow RE your comment on attribute style: I don't think there's a use case for not(...) here and I think it's safe to assume the syntax is any(...) without having to explicitly write that out. This is also inkeeping with current wrap="..." style.

Speak now if you still feel strongly on it.

fakeshadow:
I'm fine with the current attributes. You guys did a good job.

robjtede:
[And awaaayyyyy we go...](https://media.giphy.com/media/WOYKaXG2xJsBO/giphy.gif)

PR #1689, opened 2020-09-17
Update `tinyvec` to 1.0

Lokathor:
Hi all, `tinyvec` lead here. Just bumping your `tinyvec` usage.

There's *likely* to be no break in practice between 0.3 and 1.0 (IIRC it was a small change in one macro match arm), but we'll let CI find out!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ time = { version = "0.2.7", default-features = false, features = ["std"] }
 url = "2.1"
 open-ssl = { package = "openssl", version = "0.10", optional = true }
 rust-tls = { package = "rustls", version = "0.18.0", optional = true }
-tinyvec = { version = "0.3", features = ["alloc"] }
+tinyvec = { version = "1", features = ["alloc"] }
 
 [dev-dependencies]
 actix = "0.10.0"


popzxc (approved):
Thanks! It's always cool to see that amount of 1.0+ projects in Rust ecosystem is increasing. ❤️

robjtede:
Hey @Lokathor, thanks for submitting this 👍 

Hopefully a helpful note from a happy consumer of tinyvec: I had considered getting v1 upgrade in for our v3 release but was put off slightly by the lack of changelog. Keeping track of this stuff, especially for a v0.x => v1 release, would give consumers of the library much greater confidence about bumping versions.

Anyway, you mentioned the changes are minimal and ci is happy so we'll get this merged :)

Lokathor:
Ahhh, yeah that is my fault a bit. I usually don't keep a changelog with stuff that is pre-1.0. I do try to do so when doing updates to a 1.0 crate though (eg: `bytemuck` has a changelog I'm pretty careful with).

This is our last break until const generics becomes stable and we can change around how the ArrayVec type works to be all const generics based.

robjtede (approved):

PR #1692, opened 2020-09-18
remove generic type B for app::App

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

This generic type is unnecessary as type `B` is constraint by `T` so we can leave it out of `App`.
Not big improvement as the major block for easy binding/returning `App` is the type `T` but with this PR if no middleware and/or middleware function is warpped it would be pretty easy to bind the `App`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,15 @@
 
 ## Unreleased - 2020-xx-xx
 ### Changed
-* remove generic type `B` for `app::App` and it's not bound to `actix_http::body::MessageBody` trait anymore. [#1692]
+* Remove generic type `B` for `app::App`, `app_service::{AppInit, AppInitResult, AppInitService}`.
+  Change `middleware::{Compressed, Logger}` to return `ServiceResponse<Body>` directly. [#1692]
+* Add `TrailingSlash::MergeOnly` behaviour to `NormalizePath`, which allow `NormalizePath`
+  to keep the trailing slash's existance as it is. [#1695]
+* Fix `ResourceMap` recursive references when printing/debugging. [#1708]
 
 [#1692]: https://github.com/actix/actix-web/pull/1692
+[#1695]: https://github.com/actix/actix-web/pull/1695
+[#1708]: https://github.com/actix/actix-web/pull/1708
 
 
 ## 3.0.2 - 2020-09-15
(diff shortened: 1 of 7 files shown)


popzxc (approved):

JohnTitor (approved):
Looks reasonable to me!

robjtede:
Is this breaking?

JohnTitor:
> 
> 
> Is this breaking?

This removes the private field of `App` and loosens bounds so these changes should be semver-minor, right?

fakeshadow:
This could be a break change if people managed to return an `App` from a function call somehow. They would have to remove the second type signature of `App`.

Neopallium:
How about leaving `B` on `App`, but not using it internally?  Schedule it for removal in 3.1.x?

robjtede:
Does this provide any tangible improvements? We'd need a very good reason to break semver despite the low number of projects that would be affected.

fakeshadow:
The type signautre of returning an `App` would change from
```
App<impl ServiceFactory<..B>, B>
```
to
```
App<impl ServiceFactory<..B>>
```

And it can be simplified to `App<AppEntry>` if no `wrap` and `warp_fn` is used on it.
So no it's not a big improvement.

fakeshadow:
I just found out all the middle types of `Appxx` can be simplied a litte as well. So this pr needs some more work.

fakeshadow:
`App` and it's middle types now is completely not bound to `B`. The break comes from with two parts: 
1. middlewares (Logger, Compress) used to return a wrapper type as `ServiceResponse<Wrapper<B>>`. This is now changed to return `ServiceRequest<B>` directly. The wrapper types are transformed to it in place with:
```rust
Response::Other(Body::Message(Box<dyn MessageBody + Unpin>))
```

2. return an `App` from function can be simplied to `App` or
```rust
App<impl ServiceFactory<
        Config = (),
        Request = ServiceRequest,
        Response = ServiceResponse,
        Error = Error,
        InitError = (),
    >>
```

fakeshadow:
I think this PR can be closed for now.

I didn't think through before making the changes. Wrapping the stream type returned in a Boxed trait object could be hurting performance noticeably when there are multiple middleware used that would alter the output payload stream.

I would give it another try where the B type is not boxed.

PR #1695, opened 2020-09-18
Add TrailingSlash::MergeOnly behavior

### LINKED ISSUE #1694: NormalizePath in actix-web-3 is not compatible with that in actix-web-2
##  Expected Behavior
In actix-web-2.0.0, `NormalizePath` will convert the  paths as following:

1. `//` => `/`
2. `//v1//` => `/v1/`
3. `//v1//something` => `/v1/something`

### When using `TrailingSlash::Trim`

1. `//` => `/`
2. `//v1//` => `/v1`
3. `//v1//something` => `/v1/something`

### When using `TrailingSlash::Always`

1. `//` => `/`
2. `//v1//` => `/v1/`
3. `//v1//something` => `/v1/something/`

## Possible Solution

Add an option called `MergeOnly` in `TrailingSlash`, which keeps the trailing slash's existance as it is.
This option should make `NormalizePath` compatible w/ actix-web-2.

```rust
#[actix_rt::test]
async fn keep_trailing_slash_unchange() {
    let mut app = init_service(
        App::new()
            .wrap(NormalizePath(TrailingSlash::MergeOnly))
            .service(web::resource("/").to(HttpResponse::Ok))
            .service(web::resource("/v1/something").to(HttpResponse::Ok))
            .service(web::resource("/v1/").to(HttpResponse::Ok)),
    )
    .await;

    let tests = vec![
        ("/", true), // root paths should still work
        ("/?query=test", true),
        ("///", true),
        ("/v1/something////", false),
        ("/v1/something/", false),
        ("//v1//something", true),
        ("/v1/", true),
        ("/v1", false),
        ("/v1////", true),
        ("//v1//", true),
        ("///v1", false),
    ];

    for (path, success) in tests {
        let req = TestRequest::with_uri(path).to_request();
        let res = call_service(&mut app, req).await;
        assert_eq!(res.status().is_success(), success);
    }
}
```

## Context

I wanna migrate from actix-web-2.0 to 3.0, but failed to access some of my resource.

## Your Environment

* Rust Version: rustc 1.46.0 (04488afe3 2020-08-24)
* Actix Web Version: 3.0.2

SuperHacker-liuan:
Behavior of `NormalizePath` in actix-web 2.0 would keep the trailing slash's existance as it is,
which is different from that in 3.0.

This patch add a new option called `MergeOnly` in `TrailingSlash`, which only merge the trailing
slashes into one if there exist. This option makes `NormalizePath` able to be compatible with
actix-web 2.0 and makes it eaiser for users who want to migrate from 2.0 to 3.0.

This will close #1694.

## PR Type

Bug Fix

## PR Checklist

## Overview

New option in `TrailingSlash` will makes it easy for developers who want to migrate
from 1.0 / 2.0 to 3.0, because the behavior of this option is compatible with legacy versions.

No breaking changes.

close #1694.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+### Changed
+* Add `TrailingSlash::MergeOnly` behaviour to `NormalizePath`, which allow `NormalizePath`
+  to keep the trailing slash's existance as it is. fix [#1694] [#1695]
 
 
 ## 3.0.2 - 2020-09-15
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## Unreleased - 2020-xx-xx
 ### Changed
 * Add `TrailingSlash::MergeOnly` behaviour to `NormalizePath`, which allow `NormalizePath`
-  to keep the trailing slash's existance as it is. fix [#1694] [#1695]
+  to keep the trailing slash's existance as it is. [#1695]
 
 
 ## 3.0.2 - 2020-09-15
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -17,8 +17,9 @@ pub enum TrailingSlash {
     /// Always add a trailing slash to the end of the path.
     /// This will require all routes to end in a trailing slash for them to be accessible.
     Always,
-    /// Neither add nor trim slash at end of path. Merge multiple slashes only
-    /// This is compatible with actix-web 2.0
+    /// Only merge any present multiple trailing slashes.
+    ///
+    /// Note: This option provides the best compatibility with the v2 version of this middlware.
     MergeOnly,
     /// Trim trailing slashes from the end of the path.
     Trim,
@@ -36,7 +37,8 @@ impl Default for TrailingSlash {
 /// Performs following:
 ///
 /// - Merges multiple slashes into one.
-/// - Appends a trailing slash if one is not present, or removes one if present, or keep trailing slash's existance as it was, depending on the supplied `TrailingSlash`.
+/// - Appends a trailing slash if one is not present, removes one if present, or keeps trailing
+///   slashes as-is, depending on the supplied `TrailingSlash` variant.
 ///
 /// ```rust
 /// use actix_web::{web, http, middleware, App, HttpResponse};
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -17,6 +17,9 @@ pub enum TrailingSlash {
     /// Always add a trailing slash to the end of the path.
     /// This will require all routes to end in a trailing slash for them to be accessible.
     Always,
+    /// Neither add nor trim slash at end of path. Merge multiple slashes only
+    /// This is compatible with actix-web 2.0
+    MergeOnly,
     /// Trim trailing slashes from the end of the path.
     Trim,
 }
@@ -33,7 +36,7 @@ impl Default for TrailingSlash {
 /// Performs following:
 ///
 /// - Merges multiple slashes into one.
-/// - Appends a trailing slash if one is not present, or removes one if present, depending on the supplied `TrailingSlash`.
+/// - Appends a trailing slash if one is not present, or removes one if present, or keep trailing slash's existance as it was, depending on the supplied `TrailingSlash`.
 ///
 /// ```rust
 /// use actix_web::{web, http, middleware, App, HttpResponse};
@@ -108,6 +111,7 @@ where
         // Either adds a string to the end (duplicates will be removed anyways) or trims all slashes from the end
         let path = match self.trailing_slash_behavior {
             TrailingSlash::Always => original_path.to_string() + "/",
+            TrailingSlash::MergeOnly => original_path.to_string(),
             TrailingSlash::Trim => original_path.trim_end_matches('/').to_string(),
         };
 
@@ -235,6 +239,38 @@ mod tests {
         let req4 = TestRequest::with_uri("//v1//something").to_request();
         let res4 = call_service(&mut app, req4).await;
         assert!(res4.status().is_success());
+    }
+
+    #[actix_rt::test]
+    async fn keep_trailing_slash_unchange() {
+        let mut app = init_service(
+            App::new()
+                .wrap(NormalizePath(TrailingSlash::MergeOnly))
+                .service(web::resource("/").to(HttpResponse::Ok))
+                .service(web::resource("/v1/something").to(HttpResponse::Ok))
+                .service(web::resource("/v1/").to(HttpResponse::Ok)),
+        )
+        .await;
+
+        let tests = vec![
+            ("/", true), // root paths should still work
+            ("/?query=test", true),
+            ("///", true),
+            ("/v1/something////", false),
+            ("/v1/something/", false),
+            ("//v1//something", true),
+            ("/v1/", true),
+            ("/v1", false),
+            ("/v1////", true),
+            ("//v1//", true),
+            ("///v1", false),
+        ];
+
+        for (path, success) in tests {
+            let req = TestRequest::with_uri(path).to_request();
+            let res = call_service(&mut app, req).await;
+            assert_eq!(res.status().is_success(), success);
+        }
     }
 
     #[actix_rt::test]
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -48,7 +48,7 @@
 
 * `middleware::NormalizePath` can now also be configured to trim trailing slashes instead of always keeping one.
   It will need `middleware::normalize::TrailingSlash` when being constructed with `NormalizePath::new(...)`,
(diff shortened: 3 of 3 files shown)


Neopallium (approved):
Looks good.

robjtede:
I think it should be discussed whether this should be considered a permanent option or just for migration from v2 (and then removed with the next breaking changes).

I personally can't see the use case for wanting the distinction and consider it a foot gun to allow it in this middleware. In my view, either all your routes are defined with trailing slashes or they aren't.

SuperHacker-liuan:
Browsers are affected by the trailing slash policies when they are trying to find resource via relative paths.
When browsers trying to request a resource via relative path, TrailingSlash::Always & TrailingSlash::Trim will cause a resource missing in some situation.

## Example 1: 

I've got a site provides `/app/` which will response a page `A`, and `/app/style.css` which will response a style sheet file being used for `/app/`. In page A, there will be a tag link to the css as following.
```html
<link href="style.css" type="text/css" rel="stylesheet"/>
```

No matter how the route I defined, users' access to `/app` & `/app/` will get difference. because browsers which trying to request `/app` will request `/style.css` which is in fact not exist in my routes.

To fix the problem above, I'd have to write the routes which response `/app` to `HttpResponse::SeeOther().header(header::LOCATION, "app/").finish()` to solve this problem.

## Example 2:

Consider I want `/app` and `/app/` both redirect to `/app/index`.

When I use `TrailingSlash::Trim`, they will both use the `/app` route, which will reponse a HTTP 303 to `app/index`. Browsers which are trying to request `/app/` will be redirected to `/app/app/index` which is in fact not exist.

When I use `TrailingSlash::Always`, they will both use the `/app/` route, which will response a HTTP 303 to `index`. Browsers which are trying to request `/app` will be redirected to `index` which may be another page.

Only when I use `TrailingSlash::MergeOnly`, and redirect `/app` & `/app/` to different locations, users' browsers are able to perform the correct action.

robjtede:
It seems most of your problems could be solved, client-side, by not relying so heavily on relative paths. That being said, the case of redirection and relative path resolving in browsers is compelling to try solving anyway.

From the perspective of an actix app serving markup, an alternate way to solve this would be an option within this middleware to redirect to the normalized path rather than transparently normalizing and serving that resource. I think this idea has been mentioned before. Prior art for this approach includes Apache and Nginx rewrite rule flags.

SuperHacker-liuan:
> It seems most of your problems could be solved, client-side, by not relying so heavily on relative paths. That being said, the case of redirection and relative path resolving in browsers is compelling to try solving anyway.
> 
> From the perspective of an actix app serving markup, an alternate way to solve this would be an option within this middleware to redirect to the normalized path rather than transparently normalizing and serving that resource. I think this idea has been mentioned before. Prior art for this approach includes Apache and Nginx rewrite rule flags.

Well, I'm afraid that w/ this approach, I'll not able to deploy actix-web sites directly to 80/tcp or 443/tcp, but have to set a reverse proxy for it.

robjtede:
> Well, I'm afraid that w/ this approach, I'll not able to deploy actix-web sites directly to 80/tcp or 443/tcp, but have to set a reverse proxy for it.

I want to try to fully understand your needs here so we can come up with the best solution.

It occurs to me that this is a solved problem on the reverse proxy/static page servers like apache and nginx where they have chosen to use the rewrite rules system that can optionally redirect to the normalized page. As far as I can tell, this solves the same problem and is worth considering as an alternative.

You've even suggested this will solve the problem:

> To fix the problem above, I'd have to write the routes which response /app to HttpResponse::SeeOther().header(header::LOCATION, "app/").finish() to solve this problem.

...so providing a built-in way to redirect to normalized paths would work and be more familiar to folks coming from apache/nginx.

Example 1: using a relative link here is dangerous anyway, using `href="/app/style.css"` would be better and serves `/app` and `/app/` the same file.

Example 2: same idea, you want to redirect to `/app/index`, so use that as the redirect path instead of trying to use relative urls.

There's lots of things to consider when adding a feature like this so I'd like to be convinced this is solving the right problem.

SuperHacker-liuan:
Well, what I'm facing is a little troublesome.

# 0x0. Working background

I'm working in an company w/ tens of branches and a huge intranet. **The company I'm working for is not a company w/ many professioned server operators, hence many employees are not able to write any config files. In fact, it is a company in traditional industries.** As a part of them but not a full-time developer, I often develop some small HTTP services to reduce my pressure. Sometimes, employees in other branches would tell me that they want to deploy my service in theirs branches, bad sadly, they are neither developers nor professioned server operators. In past years, I'd like to provide my HTTP services executables to them but they are unable to prepare the executing environment themselves. (i.e. NodeJS, docker & etc.) What the worse is I'm also wasn't able to access their server becasue the cyber departments in those branches would never allow theirs firewall policy request, in which will aceept the connection from me.
**Therefore I'll have to design my service whose releases will be deployed as easy as possible** .

We can consider most of them are only able copy my executable to their server's `/home/user` dir with `Filezilla` and then execute `./app --port=12345 --db=10.4.5.6:3306 --daemon` to start it, and finally request their cyber department to set the load balance at their F5 Big-IP load balancer's graphical interface.

In order to make the deployment as easy as possible, I have use the `rust-embed` to store the statics folder into the release executable, and implement a trait which will make actix-web able to load static files from `struct StaticEmbed` directly and automatically implement a Responder for it.

# 0x2. Why I used to face the problems like example 1 & 2 

In some branches which are financial challenged or far away from big metropolics, they would not able to hire a professional server operator to maintaince their servers. Services which are not provided by parent company have to be deployed by employees who are not familiar with servers. Those branches would provide a hardware load balancer like F5 Big-IP, but cannot provide independent hostnames, because DNS server in our intranet is controlled by parent company, and to applicate a new IP is also a painful work for them. Hence they often use the same hostname but different access folders for different service. Of course I know that's legacy way and not the best practise, but I cannot affect their choice.

In my experience, I found that use the relative path is the best way to avoid the troubles for them, because there were no need to config the root path when deploying the app, and compatible with any URL prefix. E.g. In branch A, they may map http://foo.A.intranet/ to `/foo/`. In branch B, they may map http://inner.B.intranet/foo to `/foo`. And in branch C, they may map https://service.C.intranet/inner/foo to `/foo`. **If I use absolute path, I'll have to face problem that if the deployers want to perform a config-free deployment or an one-stop deployment, only the deployment in branch B is rendered well.**

I believe that not only me but also other friends may face the similar problem in traditional industries, especially in some developing countries. So I hope this PR will be accept, otherwise I'll have to write a independent external crate for similar requirements😭

CHANGES.md
```diff
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+### Changed
+* Add `TrailingSlash::MergeOnly` behaviour to `NormalizePath`, which allow `NormalizePath`
+  to keep the trailing slash's existance as it is. fix [#1694] [#1695]
```
robjtede:
```suggestion
  to keep the trailing slash's existance as it is. [#1695]
  
[#1695]: https://github.com/actix/actix-web/pull/1695
```

src/middleware/normalize.rs
```diff
@@ -17,6 +17,9 @@ pub enum TrailingSlash {
     /// Always add a trailing slash to the end of the path.
     /// This will require all routes to end in a trailing slash for them to be accessible.
     Always,
+    /// Neither add nor trim slash at end of path. Merge multiple slashes only
+    /// This is compatible with actix-web 2.0
```
robjtede:
```suggestion
    /// Only merge any present multiple trailing slashes.
    ///
    /// Note: This option provides the best compatibility with the v2 version of this middlware.
```

src/middleware/normalize.rs
```diff
@@ -33,7 +36,7 @@ impl Default for TrailingSlash {
 /// Performs following:
 ///
 /// - Merges multiple slashes into one.
-/// - Appends a trailing slash if one is not present, or removes one if present, depending on the supplied `TrailingSlash`.
+/// - Appends a trailing slash if one is not present, or removes one if present, or keep trailing slash's existance as it was, depending on the supplied `TrailingSlash`.
```
robjtede:
```suggestion
/// - Appends a trailing slash if one is not present, removes one if present, or keeps trailing
///   slashes as-is, depending on the supplied `TrailingSlash` variant.
```

robjtede (changes requested):

robjtede:
Thank you for the detailed explanation; it is a compelling reason to add it.

src/middleware/normalize.rs
SuperHacker-liuan: Thank you for your suggestion😊

CHANGES.md
SuperHacker-liuan: Thank you for your suggestion😊

src/middleware/normalize.rs
SuperHacker-liuan: Thank you for your suggestion😊

SuperHacker-liuan:
Thank you for your suggestion  @robjtede . I have appended your suggestions to the new commit.

robjtede (approved):

PR #1702, opened 2020-09-20
prepare codegen 0.4.0 release

robjtede:
## PR Type
Release

## PR Checklist

## Overview
Releasing as 0.4.0 as actix-web does not automatically use it, given a short public trial period we can bump the dependency in actix-web to use this.

### DIFF
--- a/actix-web-codegen/CHANGES.md
+++ b/actix-web-codegen/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+
+
+## 0.4.0 - 2020-09-20
 * Added compile success and failure testing. [#1677]
 * Add `route` macro for supporting multiple HTTP methods guards.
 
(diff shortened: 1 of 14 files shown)


robjtede:
interesting... trybuild failures because it doesn't respect local patches or something, will check before progressing

robjtede:
Problem with trybuild was because tests used `actix_web::*` re-exports; have changed that and tweaked some of the stderr files accordingly.

Revamped the main page documentation for folks who land on the codegen docs.

Tweaked some of the attribute docs to encourage use of actix_web re-exports while keeping doc tests using self versions. Weird case here is `actix_web_codegen::main` which can't simply be refered to as `#[main]` because compiler says no; a clarification is given.

robjtede:
Alright, tests are passing 🎉 ready for a review @actix/contributors

Neopallium (approved):
Looks good.

PR #1704, opened 2020-09-21
Fix Multipart consuming payload before header checks

### LINKED ISSUE #1513: multipart always consumes body
## Expected Behavior
Multipart only consumes payload in extractor after header checks.

## Current Behavior
Payload is always consumed preventing optional extraction.

## Possible Solution
Perform header checks in FromRequest to prevent always calling `payload.take()`

## Context
Precedent set with `web::Json` extraction. It receives a `&mut Payload` and does not modify it until header content type checks passed.

mattgathu:
## PR Type

Bug Fix

## PR Checklist

## Overview

Enable headers checks before payload consumption by introducing two new `Multipart` constructor methods:
- **from_boundary:** build Multipart from boundary and stream
- **from_error:** build Multipart for MultipartError

We then refactor the `from_request` extractor to perform a header checks using the existing `Multipart::boundary` method and then call the appropriate constructor based on the result of the checks.

Closes #1513

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -1,6 +1,7 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+* Fix multipart consuming payload before header checks #1513
 
 
 ## 3.0.0 - 2020-09-11
--- a/actix-multipart/src/extractor.rs
+++ b/actix-multipart/src/extractor.rs
@@ -36,6 +36,9 @@ impl FromRequest for Multipart {
 
     #[inline]
     fn from_request(req: &HttpRequest, payload: &mut Payload) -> Self::Future {
-        ok(Multipart::new(req.headers(), payload.take()))
+        ok(match Multipart::boundary(req.headers()) {
+            Ok(boundary) => Multipart::from_boundary(boundary, payload.take()),
+            Err(err) => Multipart::from_error(err),
+        })
     }
 }
--- a/actix-multipart/src/server.rs
+++ b/actix-multipart/src/server.rs
@@ -828,6 +828,8 @@ mod tests {
     use actix_http::h1::Payload;
     use actix_utils::mpsc;
     use actix_web::http::header::{DispositionParam, DispositionType};
+    use actix_web::test::TestRequest;
+    use actix_web::FromRequest;
     use bytes::Bytes;
     use futures_util::future::lazy;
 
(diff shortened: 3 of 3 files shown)


Neopallium (commented):
Can you add a test?

robjtede:
As mentioned biefly in the original issue, I'd like to see a test that uses the `Option<T>` and/or `Result<T, E>` extractors to catch failures.

mattgathu:
> As mentioned biefly in the original issue, I'd like to see a test that uses the `Option<T>` and/or `Result<T, E>` extractors to catch failures.

Hi, I don't understand exactly what that means, and its implications. I went over the issue again and still can't connect the dots.

I have however added a test that tests that the bug is fixed. Took me a while to nail it but I think it suffices. _Is this what you meant?_

robjtede:
We should be able to use Multipart in a handler as a fallible extractor like this:

```rust
#[get("/")]
async fn handler(form: Result<Multipart, Error>) -> impl Responder {
	...
}
```

It's worth adding a test to make sure this is (and remains) possible.

mattgathu:
> We should be able to use Multipart in a handler as a fallible extractor like this:
> 
> ```rust
> #[get("/")]
> async fn handler(form: Result<Multipart, Error>) -> impl Responder {
> 	...
> }
> ```
> 
> It's worth adding a test to make sure this is (and remains) possible.

My understanding of actix-web is very limited and I'm really struggling to understand how the changes I made relate to multipart use in handlers. 

_Is the idea to test the extraction api i.e. the `from_request` method?_

I went over the [tests](https://github.com/actix/actix-web/blob/master/src/types/json.rs#L499-L556) for `web::Json` which is mentioned in the issue, is this the type of test in mind? Otherwise I would be grateful is you pointed me to an existing extraction test that I can learn from, thanks.

robjtede:
> My understanding of actix-web is very limited

Well in that case, first I'd like to say you've done a great job with this PR so far.

> Is the idea to test the extraction api i.e. the from_request method?

Yes. If you like I can probably add commits to this branch to address the things I'm talking about and let you look it over before merging?

mattgathu:
Yes, that would be great!

robjtede (approved):
In making the tweaks I was mentioning, it's occured to me they are not in scope for this PR. Thanks for picking this up. Good work 👍

PR #1705, opened 2020-09-21
Unify route macros

arniu:
## PR Type

Refactor

## PR Checklist

### DIFF
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -58,7 +58,7 @@ use proc_macro::TokenStream;
 /// - `wrap = "Middleware"` - Registers a resource middleware.
 #[proc_macro_attribute]
 pub fn get(args: TokenStream, input: TokenStream) -> TokenStream {
-    route::with_method(Some(route::Method::Get), args, input)
+    route::with_method(Some(route::MethodType::Get), args, input)
 }
 
 /// Creates route handler with `POST` method guard.
@@ -68,7 +68,7 @@ pub fn get(args: TokenStream, input: TokenStream) -> TokenStream {
 /// Attributes are the same as in [get](attr.get.html)
 #[proc_macro_attribute]
 pub fn post(args: TokenStream, input: TokenStream) -> TokenStream {
-    route::with_method(Some(route::Method::Post), args, input)
+    route::with_method(Some(route::MethodType::Post), args, input)
 }
 
 /// Creates route handler with `PUT` method guard.
@@ -78,7 +78,7 @@ pub fn post(args: TokenStream, input: TokenStream) -> TokenStream {
 /// Attributes are the same as in [get](attr.get.html)
 #[proc_macro_attribute]
 pub fn put(args: TokenStream, input: TokenStream) -> TokenStream {
-    route::with_method(Some(route::Method::Put), args, input)
+    route::with_method(Some(route::MethodType::Put), args, input)
 }
 
 /// Creates route handler with `DELETE` method guard.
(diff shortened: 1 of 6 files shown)


Neopallium (approved):
looks good.

Neopallium (approved):

actix-web-codegen/src/lib.rs
```diff
@@ -17,10 +17,12 @@
 //! - [options](attr.options.html)
 //! - [trace](attr.trace.html)
 //! - [patch](attr.patch.html)
+//! - [route](attr.route.html)
 //!
 //! ### Attributes:
 //!
-//! - `"path"` - Raw literal string with path for which to register handle. Mandatory.
+//! - `"path"` - *Required*, raw literal string with path for which to register handle
+//! - `method="HTTP_METHOD"` - Secondary HTTP method accepted, uppercased string. "GET", "POST" for example
```
robjtede: please make it clear this only applies to the route macro

robjtede: also: uppercased => uppercase or upper-cased

actix-web-codegen/src/lib.rs
```diff
@@ -155,7 +63,52 @@ pub fn patch(args: TokenStream, input: TokenStream) -> TokenStream {
...
+    ) => {
+        $(
+            doc_comment! {
+                concat!("Creates route handler with `", stringify!($upper), "` method guard.")
+                concat!("")
+                concat!("Syntax: `#[", stringify!($method), "(\"path\" [, attributes])]`")
+                concat!("")
+                concat!("## Attributes:")
+                concat!("")
+                concat!("- `\"path\"` - *required* Raw literal string with path for which to register handler")
```
robjtede: i don't think this required marker is necessary, it's specified in the syntax part

actix-web-codegen/src/lib.rs
```diff
@@ -17,10 +17,12 @@
 //! - [options](attr.options.html)
 //! - [trace](attr.trace.html)
 //! - [patch](attr.patch.html)
+//! - [route](attr.route.html)
 //!
 //! ### Attributes:
```
robjtede: it's not clear this section should be here any more

actix-web-codegen/src/route.rs
```diff
@@ -20,79 +20,81 @@ impl ToTokens for ResourceType {
...
-    Put,
-    Delete,
-    Head,
-    Connect,
-    Options,
-    Trace,
-    Patch,
-    Multi,
-}
+macro_rules! method_type {
```
robjtede, line 23: this is a great idea 👍

robjtede (commented):
Pretty nice clean up overall. Only major point at the moment is wether the attribute section in the main page docs should exist; possibly should remove it and move the notes section into each individual macro.

actix-web-codegen/src/lib.rs
arniu: Shall I take them all down, or just remove `route`?

robjtede: yeah i think remove it and move the notes section

actix-web-codegen/src/lib.rs
arniu: Now, other macros can take `method` too.

For example, `put("/form", method="POST")` will be accepted.

Do you think it is a feature, or a bug?

robjtede: no, that's not desired, multiple methods should only be allowed on route macro

robjtede:
I'd like to make sure the main doc page has some more guidance than this but I'll submit it as the next PR.

robjtede (approved):

PR #1708, opened 2020-09-25
Fix cyclic references in ResourceMap

### LINKED ISSUE #1582: ResourceMap debug printing is infinitely recursive for any meaningful routing structure
## Expected Behavior

Debug printing a resource map with nested resource maps should not stack overflow.

## Current Behavior

Debug printing a resource map with nested resource maps recurses infinitely and overflows the stack.

## Possible Solution

Provide an explicit Debug impl that accounts for the cyclic structure.

## Steps to Reproduce
```rust
println!("{:?}", req.resource_map());

> thread 'xxx' has overflowed its stack
```

## Context

Trying to determine the finalized structure of a resource map while testing a new feature.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): 1.40.0
* Actix Web Version: master (3.0.0-alpha.3)

mattgathu:
## PR Type

Bug Fix

## PR Checklist

## Overview

What
--
Refactor `ResourceMap` struct to remove hard references to parent ResourceMap.

How
--
Assuming the child resource maps are always nested within a parent, then
we can always assume the parent will be there and use a weak reference to
refer back to it.

> A Weak pointer is useful for keeping a temporary reference to the allocation managed by Rc without preventing its inner value from being dropped. It is also used to prevent circular references between Rc pointers, since mutual owning references would never allow either Rc to be dropped. For example, a tree could have strong Rc pointers from parent nodes to children, and Weak pointers from children back to their parents.

[_from the stdlib docs_](https://doc.rust-lang.org/nightly/std/rc/struct.Weak.html)

Closes https://github.com/actix/actix-web/issues/1582

### DIFF
--- a/src/rmap.rs
+++ b/src/rmap.rs
@@ -1,5 +1,5 @@
 use std::cell::RefCell;
-use std::rc::Rc;
+use std::rc::{Rc, Weak};
 
 use actix_router::ResourceDef;
 use fxhash::FxHashMap;
@@ -11,7 +11,7 @@ use crate::request::HttpRequest;
 #[derive(Clone, Debug)]
 pub struct ResourceMap {
     root: ResourceDef,
-    parent: RefCell<Option<Rc<ResourceMap>>>,
+    parent: RefCell<Weak<ResourceMap>>,
     named: FxHashMap<String, ResourceDef>,
     patterns: Vec<(ResourceDef, Option<Rc<ResourceMap>>)>,
 }
@@ -20,7 +20,7 @@ impl ResourceMap {
     pub fn new(root: ResourceDef) -> Self {
         ResourceMap {
             root,
-            parent: RefCell::new(None),
+            parent: RefCell::new(Weak::new()),
             named: FxHashMap::default(),
             patterns: Vec::new(),
         }
@@ -38,7 +38,7 @@ impl ResourceMap {
     pub(crate) fn finish(&self, current: Rc<ResourceMap>) {
         for (_, nested) in &self.patterns {
             if let Some(ref nested) = nested {
-                *nested.parent.borrow_mut() = Some(current.clone());
+                *nested.parent.borrow_mut() = Rc::downgrade(&current);
                 nested.finish(nested.clone());
             }
         }
@@ -210,7 +210,7 @@ impl ResourceMap {
         U: Iterator<Item = I>,
         I: AsRef<str>,
     {
-        if let Some(ref parent) = *self.parent.borrow() {
+        if let Some(ref parent) = self.parent.borrow().upgrade() {
             parent.fill_root(path, elements)?;
         }
         if self.root.resource_path(path, elements) {
@@ -230,7 +230,7 @@ impl ResourceMap {
         U: Iterator<Item = I>,
         I: AsRef<str>,
     {
-        if let Some(ref parent) = *self.parent.borrow() {
+        if let Some(ref parent) = self.parent.borrow().upgrade() {
             if let Some(pattern) = parent.named.get(name) {
                 self.fill_root(path, elements)?;
                 if pattern.resource_path(path, elements) {
--- a/src/rmap.rs
+++ b/src/rmap.rs
@@ -394,7 +394,7 @@ mod tests {
         root.finish(Rc::clone(&root));
 
         let output = format!("{:?}", root);
-        assert!(output.starts_with("ResourceMap"));
-        assert!(output.ends_with("}"));
+        assert!(output.starts_with("ResourceMap {"));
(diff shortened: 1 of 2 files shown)


mattgathu:
@robjtede I started working on a fix for #1582 .

It's still a draft PR but the bug is essentially fixed now. I managed to have the following program compile and have debug working.

```rust
use actix_web::{get, App, web,HttpResponse, HttpRequest,  HttpServer};

#[get("/rmap")]
async fn rmap(req:HttpRequest) -> String {
    format!("{:#?}\r\n", req.resource_map())
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(rmap)
            .service(
                web::scope("/x/")
                .service(web::resource("/path1").to(|| async { HttpResponse::Ok() }))
                .service(web::resource("/path2").route(web::get().to(|| HttpResponse::Ok())))
                .service(web::resource("/path3").route(web::head().to(|| HttpResponse::MethodNotAllowed())))
            )
        })
    .bind("127.0.0.1:8080")?
    .workers(1)
    .run()
    .await
}
```

robjtede:
Nice! Can you paste here what it looks like for that config?

mattgathu:
This is the output:
```
> http localhost:8080/rmap
HTTP/1.1 200 OK
content-length: 3907
content-type: text/plain; charset=utf-8
date: Fri, 25 Sep 2020 14:23:34 GMT

ResourceMap {
    root: ResourceDef {
        id: 0,
        tp: Static(
            "",
        ),
        name: "",
        pattern: "",
        elements: [
            Str(
                "",
            ),
        ],
    },
    parent: RefCell {
        value: (Weak),
    },
    named: {
        "rmap": ResourceDef {
            id: 0,
            tp: Static(
                "/rmap",
            ),
            name: "rmap",
            pattern: "/rmap",
            elements: [
                Str(
                    "/rmap",
                ),
            ],
        },
    },
    patterns: [
        (
            ResourceDef {
                id: 0,
                tp: Static(
                    "/rmap",
                ),
                name: "rmap",
                pattern: "/rmap",
                elements: [
                    Str(
                        "/rmap",
                    ),
                ],
            },
            None,
        ),
        (
            ResourceDef {
                id: 1,
                tp: Prefix(
                    "/x/",
                ),
                name: "",
                pattern: "/x/",
                elements: [
                    Str(
                        "/x/",
                    ),
                ],
            },
            Some(
                ResourceMap {
                    root: ResourceDef {
                        id: 0,
                        tp: Prefix(
                            "/x/",
                        ),
                        name: "",
                        pattern: "/x/",
                        elements: [
                            Str(
                                "/x/",
                            ),
                        ],
                    },
                    parent: RefCell {
                        value: (Weak),
                    },
                    named: {},
                    patterns: [
                        (
                            ResourceDef {
                                id: 0,
                                tp: Static(
                                    "/path1",
                                ),
                                name: "",
                                pattern: "/path1",
                                elements: [
                                    Str(
                                        "/path1",
                                    ),
                                ],
                            },
                            None,
                        ),
                        (
                            ResourceDef {
                                id: 1,
                                tp: Static(
                                    "/path2",
                                ),
                                name: "",
                                pattern: "/path2",
                                elements: [
                                    Str(
                                        "/path2",
                                    ),
                                ],
                            },
                            None,
                        ),
                        (
                            ResourceDef {
                                id: 2,
                                tp: Static(
                                    "/path3",
                                ),
                                name: "",
                                pattern: "/path3",
                                elements: [
                                    Str(
                                        "/path3",
                                    ),
                                ],
                            },
                            None,
                        ),
                    ],
                },
            ),
        ),
    ],
}
```

src/rmap.rs
```diff
@@ -367,4 +367,32 @@ mod tests {
...
+        root.add(&mut ResourceDef::new("/v{version:[[:digit:]]{1}}"), None);
+        root.add(
+            &mut ResourceDef::root_prefix("/user/{id}"),
+            Some(Rc::new(user_map)),
+        );
+
+        let root = Rc::new(root);
+        root.finish(Rc::clone(&root));
+
+        println!("{:?}", root);
```
robjtede: This would print a huge block in --nocapture scenarios (like our CI), can we use format! and make assertions about the last (and first) characters?

mattgathu: good catch! 👍

robjtede: almost, clippy wants a char pattern for the } 😅

robjtede (approved):
elegant solution to the problem 👍

mattgathu:
@robjtede I don't know what to do about updating docs. The changes are internal to `ResourceMap` and not to consumer facing 🙂

PR #1712, opened 2020-09-26
Remove Sized Bound for web::Data

### LINKED ISSUE #1710: Avoid Double Arc for Arc<dyn> with T: ?Sized
## Expected Behavior
It should be possible to create web::Data from an **Arc\<dyn Xyz>**. 

## Current Behavior
```rust
    // trait T{}, struct A{}, impl T for A{}
    // This works as of now
    let test: Arc<Box<dyn T>> = Arc::new(Box::new(A {}));
    let data = web::Data::from(test);
    // This should also work (see above with Sized)
    let test2: Arc<dyn T> = Arc::new(A {});
    let data2 = web::Data::from(test2);
```
=> the size for values of type `dyn T` cannot be known at compilation time 

## Possible Solution
Add Data<T: ?Sized>
Extensions.insert uses TypeId::of::\<T\>, where in this case T is Data<T: ?Sized>. 
To extract we can use extensions.get::<Data\<dyn T>>()

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7f31f526ea15a6c03ae768a900d822d0

Compare <T: ?Sized> (new version) with \<T\> (current version)

## Context
I want to dynamically pass a Trait object to actix. This allows me to swap e.g. Service implementations with mock objects for unit tests.  

## Your Environment
* Rust Version (I.e, output of `rustc -V`): rustc 1.46.0 (04488afe3 2020-08-24)
* Actix Web Version: 3.0

Question is: Would it break anything else when adding ?Sized. Are there places relying on a known size? Quick glance shows that even the Arc is additionally Boxed in the Extensions HashMap.

PeterUlb:
## PR Type
Improvement

## PR Checklist
- [X] A changelog entry has been made in CHANGES.md

## Overview
See #1710. It's now possible to use Arc<dyn Trait> as a source for constructing web::Data. Arc itself already supports it, so the Data wrapper should too. Otherwise there's yet another wrapper required.

Closes #1710

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 * Add `TrailingSlash::MergeOnly` behaviour to `NormalizePath`, which allow `NormalizePath`
   to keep the trailing slash's existance as it is. [#1695]
 * Fix `ResourceMap` recursive references when printing/debugging. [#1708]
+* Remove bound `std::marker::Sized` from `web::Data` to support storing `Arc<dyn Trait>` via `web::Data::from` [#1710]
 
 [#1708]: https://github.com/actix/actix-web/pull/1708
 
--- a/src/data.rs
+++ b/src/data.rs
@@ -66,7 +66,7 @@ pub(crate) type FnDataFactory =
 /// }
 /// ```
 #[derive(Debug)]
-pub struct Data<T>(Arc<T>);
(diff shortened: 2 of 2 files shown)


robjtede (approved):
Makes sense given this is a glorified Arc 👍

PR #1714, opened 2020-09-27
add files preference for utf8 text responses

### LINKED ISSUE #1683: actix-files having trouble with Chinese characters in file content (not filename)
## Expected Behavior

Chinese characters (and other unicode characters) should be displayed correctly.

## Current Behavior

They are not.

## Possible Solution

I noticed that there was a pull request related to this (#1151), that PR did fix the issue with filename, but it seems that there are still issues with file contents.

See: https://github.com/lucifer1004/actix-file-utf-8

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.47.0-nightly (30f0a0768 2020-08-18)
* Actix Web Version: 3.0.1
* Actix Files Version: 0.3.0

robjtede:
## PR Type
Fix/Feature

## PR Checklist

## Overview
Adds new method on NamedFile & Files service builder to prefer UTF-8 responses. For now this only applies to MIME types with UTF-8 variants according to the [`mime` crate main doc page](https://docs.rs/mime/0.3.16/mime/).

The default for the new option is `false` but in future it should be `true`.

Fixes #1683


robjtede:
@lucifer1004 thoughts on this approach?

lucifer1004:
> @lucifer1004 thoughts on this approach?

What if we want UTF-8 support for other mime types?

robjtede:
The way I understand it is that other mime types already demand a specific encoding. Do you have any others in mind?

I think that if we come across issues with any more types we can sadd them to this transformer.

PR #1715, opened 2020-09-28
improve clarity in header setting method names

robjtede:
## PR Type
Refactor

## PR Checklist

## Overview
- `ResponseBuilder`
  - Deprecate `header`. Rename to `append_header` for immediate clarity on append behavior.
  - Deprecate `set_header`. Rename to `insert_header` for naming parity with `HeaderMap` method.
  - Show `set` in docs. Kinda special method; not sure what to do about this yet. Seems useful and would be a shame to deprecate/remove it. However, `set` is kinda meh as a method name but obivously renaming wouldn't be able to use any of the above existing or renamed names.

### DIFF
--- a/actix-http/src/header/common/content_type.rs
+++ b/actix-http/src/header/common/content_type.rs
@@ -1,5 +1,6 @@
-use crate::header::CONTENT_TYPE;
 use mime::Mime;
+
+use crate::header::CONTENT_TYPE;
 
 header! {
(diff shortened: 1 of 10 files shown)


Neopallium (commented):
two usages of deprecated `header` method in tests of file `actix-http/src/response.rs`

robjtede:
have some better ideas

PR #1719, opened 2020-09-30
Content Negotiation

### LINKED ISSUE #1718: Add Content-Negotiation to actix-files
For the Semantic Web/Linked Data community, content negotiation is an important feature.

We could implement it similar to [how it's described here](https://pieterheyvaert.com/blog/2019/02/25/nginx-conneg/), where requesting `http://localhost:8080/test` with `Accept: text/turtle` then makes the server look for `test.ttl` on disk.

This could be enabled or disabled in the actix-files service.

alexkreidler:
## PR Type
Feature

## PR Checklist

## Overview
Enables content negotiation in the actix-files `FilesService`. Uses built-in Actix `Accept` header parsing, sorts by [`q` parameter](https://restfulapi.net/q-parameter-in-http-accept-header/), uses `mime_guess` to determine file extensions from the content-type. Tries all file exts for a given type, then tries the next content type.

## TODOs:

- [ ] make it optional, add a flag and method to enable in the `ServiceFactory`
- [ ] add limits to the number of content types to try and the number of extensions for each content type to prevent DOS scenarios
- [ ] think about/change error handling

Closes #1718

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ members = [
   "actix-multipart",
   "actix-web-actors",
   "actix-web-codegen",
+  "actix-example-server",
   "test-server",
 ]
 
--- a/actix-example-server/Cargo.toml
+++ b/actix-example-server/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "actix-example-server"
+version = "0.1.0"
+edition = "2018"
+
+[[bin]]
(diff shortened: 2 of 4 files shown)


robjtede:
Thanks for the submission, but I don't think this feature is strong enough to integrate into the official files lib. It seems to me you could build a service on top of actix-files to achieve this functionality.

PR #1721, opened 2020-10-02
Update return type of Extensions::insert to match documented behavior

jplatte:
This is a breaking change, but Extensions::insert was strictly less useful with the previous behavior.

## PR Type

Not sure whether to consider this a bug fix or feature :shrug: 

## PR Checklist

- [ ] ~~Documentation comments have been added / updated.~~ N/A

## Overview

Before, `Extensions::insert` did not return anything, contradicting its documentation. Now it does return the previously present value, if any. This is a breaking change.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Unreleased - 2020-xx-xx
 
+### Changed
+
+* Change the return type of `Extensions::insert` from `()` to `Option<T>`
 
 ## 2.0.0 - 2020-09-11
 * No significant changes from `2.0.0-beta.4`.
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -3,8 +3,7 @@
 ## Unreleased - 2020-xx-xx
 
 ### Changed
-
-* Change the return type of `Extensions::insert` from `()` to `Option<T>`
+* Change the return type of `Extensions::insert` from `()` to `Option<T>` [#1721]
 
 ## 2.0.0 - 2020-09-11
 * No significant changes from `2.0.0-beta.4`.
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -24,8 +24,10 @@ impl Extensions {
     ///
     /// If a extension of this type already existed, it will
     /// be returned.
-    pub fn insert<T: 'static>(&mut self, val: T) {
-        self.map.insert(TypeId::of::<T>(), Box::new(val));
+    pub fn insert<T: 'static>(&mut self, val: T) -> Option<T> {
+        self.map
+            .insert(TypeId::of::<T>(), Box::new(val))
+            .and_then(|boxed| boxed.downcast().ok().map(|boxed| *boxed))
     }
 
     /// Check if container contains entry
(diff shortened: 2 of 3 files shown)


actix-http/CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased - 2020-xx-xx
 
+### Changed
+
+* Change the return type of `Extensions::insert` from `()` to `Option<T>`
```
robjtede:
```suggestion
* Change the return type of `Extensions::insert` from `()` to `Option<T>` [#1721]
```

robjtede (changes requested):
AFAICT, this isnt breaking. Would appreciate a test though.

jplatte:
This is definitely a breaking change. I even had to update another module in actix-http because it was using `extensions.insert(foo)` as the last expression in the body of a function with a return type of `()`.

robjtede:
Ahh yeah of course, that's a shame then. It's conceivable that folks have similar methods to http where a return type of of () is required.

actix-http/src/extensions.rs
```diff
@@ -75,6 +75,19 @@ impl fmt::Debug for Extensions {
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_insert() {
+        let mut map = Extensions::new();
+
+        let prev = map.insert("foo");
+        assert_eq!(prev, None);
```
robjtede, line 83:
```suggestion
        assert!(prev.is_none());
```

jplatte: I prefer `assert_eq!` for `Option` checks (where applicable) because it tells you the value if the option being tested is unexpectedly `Some`.

jplatte:
Windows CI failure looks unrelated.

robjtede:
~@actix/core-contributors What do you think about a crate feature to enable breaking / unstable / next-gen changes like this one (documenting that there are no semver guarantees on it).~

robjtede:
Alright, I think the best approach here is to provide, simply, a new method that has the desired behavior, ammending docs of insert, possibly deprecating it.

Options:
- `update`: some synonym for "insert"; not obvious from name which the behavioural difference is
- `replace`: sort of same as above but has a more standard semantic meaning in Rust world
- `insert2`: sounds hacky but more or less makes clear the intention is to replace the current behavior
- `insert_and_return`: longer but makes behavior obvious

jplatte:
> ~@actix/core-contributors What do you think about a crate feature to enable breaking / unstable / next-gen changes like this one (documenting that there are no semver guarantees on it).~

Why not sth. like this? I really don't like having the right behavior in a new method because `insert` is the name every other associative container in Rust uses for this purpose.

robjtede:
> insert is the name every other associative container in Rust uses for this purpose

Yep, I agree. We should have done it right from the beginning. Certainly, introducing a breaking change is not an option for this. A feature flag seems attractive but it would be very easy to break third party crates that "rely" on the current signature (as we have done internally). This is what deprecations are for, a transition period.

robjtede:
@jplatte Will you amend the PR? Could get it into v3.2 if merged in the next couple of days?

jplatte:
I don't actually care that much about this change (I don't think I need the old value in any of my code). Just noticed the discrepancy between the current behavior and what other `insert` methods do (and the docs for this one say). So I think I'll just close this and you can select whichever solution you prefer and implement that yourself or open it as an easy issue for people looking to contribute.

PR #1723, opened 2020-10-03
Implement Logger middleware regex exclude pattern

### LINKED ISSUE #1682: Logger middleware exclude pattern
So that a regex can be used for exclusion, for path prefixes and such.

```rust
impl Logger {
	fn exclude_pattern(mut self, pattern: impl Pattern) -> Self
}
```

mattgathu:
## PR Type

Feature

## PR Checklist

## Overview
Support use of regex for logger exclusion rules

part of #1682

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
   to retain any trailing slashes. [#1695]
 * Remove bound `std::marker::Sized` from `web::Data` to support storing `Arc<dyn Trait>`
   via `web::Data::from` [#1710]
+* Implement Logger middleware regex exclude pattern [#1723]
 
 ### Fixed
 * `ResourceMap` debug printing is no longer infinitely recursive. [#1708]
@@ -16,6 +17,7 @@
 [#1695]: https://github.com/actix/actix-web/pull/1695
 [#1708]: https://github.com/actix/actix-web/pull/1708
 [#1710]: https://github.com/actix/actix-web/pull/1710
+[#1723]: https://github.com/actix/actix-web/pull/1723
 
 
 ## 3.0.2 - 2020-09-15
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,9 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+* Implement Logger middleware regex exclude pattern [#1723]
 
+[#1723]: https://github.com/actix/actix-web/pull/1723
 
 ## 3.1.0 - 2020-09-29
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -9,13 +9,15 @@
   to retain any trailing slashes. [#1695]
 * Remove bound `std::marker::Sized` from `web::Data` to support storing `Arc<dyn Trait>`
   via `web::Data::from` [#1710]
+* Implement Logger middleware regex exclude pattern [#1723]
```
robjtede: move this and link line to unreleased section

src/middleware/logger.rs
```diff
@@ -123,6 +138,7 @@ impl Default for Logger {
         Logger(Rc::new(Inner {
             format: Format::default(),
             exclude: HashSet::new(),
+            exclude_regex: RegexSet::new(Vec::<&str>::new()).unwrap(),
```
robjtede:
```suggestion
            exclude_regex: RegexSet::empty(),
```

src/middleware/logger.rs
```diff
@@ -100,6 +101,7 @@ impl Logger {
         Logger(Rc::new(Inner {
             format: Format::new(format),
             exclude: HashSet::new(),
+            exclude_regex: RegexSet::new(Vec::<&str>::new()).unwrap(),
```
robjtede: I got `RegexSet::empty` added to `regex` v1.4.

```suggestion
            exclude_regex: RegexSet::empty(),
```

robjtede: also will need the cargo toml regex version bumping

src/middleware/logger.rs
```diff
@@ -111,6 +113,19 @@ impl Logger {
...
         self
     }
+
+    /// Ignore and do not log access info for paths that match regex
+    pub fn exclude_regex<T: Into<String>>(
+        mut self,
+        path: T,
+    ) -> Result<Self, regex::Error> {
+        let inner = Rc::get_mut(&mut self.0).unwrap();
+        let mut patterns: Vec<String> = inner.exclude_regex.patterns().into();
```
robjtede:
```suggestion
        let mut patterns = inner.exclude_regex.patterns().to_vec();
```

robjtede (changes requested):
Looks alright so far. I think that middleware constructors should panic on invalid input so we can either just panic or require that a regex is passed in that we extract the known good pattern from.

mattgathu:
> Looks alright so far. I think that middleware constructors should panic on invalid input so we can either just panic or require that a regex is passed in that we extract the known good pattern from.

I went with panic, to avoid the user having to rely on the regex crate and also avoid incompatibility if the user's regex crate version significantly differs from the one used by Actix.

robjtede (approved):
I agree that not forwarding (leaking) the regex error type is a compelling reason to panic instead. Thanks for tackling this.

PR #1728, opened 2020-10-09
Fix clippy::rc_buffer

jplatte:
## PR Type

Refactor

## PR Checklist

- [ ] ~~Tests for the changes have been added / updated.~~ N/A
- [ ] ~~Documentation comments have been added / updated.~~ N/A
- [ ] ~~A changelog entry has been made for the appropriate packages.~~ N/A

## Overview

Fixed clippy lint rc_buffer in all places where it made sense.

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::rc_buffer)] // inner value is mutated before being shared (`Rc::get_mut`)
+
 use std::future::Future;
 use std::pin::Pin;
 use std::rc::Rc;
--- a/src/app.rs
+++ b/src/app.rs
@@ -459,8 +459,8 @@ where
 {
     fn into_factory(self) -> AppInit<T, B> {
         AppInit {
-            data: Rc::new(self.data),
-            data_factories: Rc::new(self.data_factories),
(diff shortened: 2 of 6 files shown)


src/route.rs
```diff
@@ -41,6 +41,7 @@ type BoxedRouteNewService<Req, Res> = Box<
 /// If handler is not explicitly set, default *404 Not Found* handler is used.
 pub struct Route {
     service: BoxedRouteNewService<ServiceRequest, ServiceResponse>,
+    #[allow(clippy::rc_buffer)]
     guards: Rc<Vec<Box<dyn Guard>>>,
```
JohnTitor: Could we add a comment to describe why this is allowed?

jplatte: So for one of these `guards` in this file, there is a `Rc::get_mut` usage, which basically invalidates this lint. The other ones are copied from / to that one. Not sure how to put that into a succinct commit. Maybe just move it to the top of the file as

```rust
#![allow(clippy::rc_buffer)] // because of Rc::get_mut
```
?

JohnTitor: Yeah, it would be helpful!

JohnTitor (approved):
Thanks!

PR #1730, opened 2020-10-10
check with cargo-audit in CI

arniu:
## PR Type

Other

## PR Checklist

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,18 @@
+name: Security audit
+
+on:
+  push:
+    paths: 
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -12,7 +12,7 @@ jobs:
   security_audit:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions-rs/audit-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/audit.yml
```diff
@@ -0,0 +1,18 @@
+name: Security audit
+
+on:
+  push:
+    paths: 
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
```
JohnTitor, line 7: This doesn't make sense as we don't add Cargo.lock in this repo, right?

.github/workflows/audit.yml
```diff
@@ -0,0 +1,18 @@
...
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
JohnTitor:
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/audit.yml
```diff
@@ -0,0 +1,18 @@
...
+      - uses: actions-rs/audit-check@v1
```
JohnTitor, line 16: Shouldn't we generate lockfile before running check?

JohnTitor (commented):
It seems this workflow is just copied/pasted from actions-rs/audit-check's README, do you have any context?
I know, cargo-audit is a great tool and it's _ideal_ to use here but I'm not sure if we have enough bandwidth to deal with issues from cargo-audit.

robjtede:
It's worth noting that a deps.rs badge on the readme (+ dependabot PRs maybe) would be plenty enough to catch insecure deps. I'm not convinced about adding this into the CI.

.github/workflows/audit.yml
arniu: Yes, we didn't. But just in case.

.github/workflows/audit.yml
arniu: My bad! I'll fix it.

.github/workflows/audit.yml
arniu: `actions-rs/audit-check` will generate lockfile before checking.

arniu:
> It's worth noting that a deps.rs badge on the readme (+ dependabot PRs maybe) would be plenty enough to catch insecure deps. I'm not convinced about adding this into the CI.

Dependabot is fine, except too annoying.

JohnTitor:
So, no one from the team acks this change, I'm in favor of just closing this.
@actix/contributors If anyone is interested in introducing `cargo-audit`, feel free to leave a comment :) I'll keep this open in a few days then close it.

robjtede:
No joy. FWIW, I'm intending to add a deps.rs badge to the README soon when we've implemented the "latest published crate status" feature.

PR #1737, opened 2020-10-16
use idenity encoding on client if no compression features are enabled

### LINKED ISSUE #1731: Disable compression headers when disabling compression feature
## Expected Behavior
I expect awc to set the Accept-Encoding header to `identity` when the compression feature in actix-web is disabled.

## Current Behavior
The Accept-Encoding header is set to `br`. Any HTTP response from a webserver that makes use of the header will send brotli-encoded bodies, which awc can not decode because the feature is disabled, resulting in a "useless" body for the user.

## Possible Solution
See expected behavior. I am unfamiliar with the source, but https://github.com/actix/actix-web/blob/master/awc/src/request.rs#L24 and below seems like the culprit.

This code snippet does it just fine. The web server that caused me to run into the issue requires authentication, so I cannot post that here.
```rust
#[actix_web::main]
async fn main() {
    let client = actix_web::client::Client::new();
    let response = client
        .get("https://httpbin.org/anything")
        .send()
        .await
        .unwrap()
        .body()
        .await
        .unwrap();
    // {
    //  "args": {},
    //  "data": "",
    //  "files": {},
    //  "form": {},
    //  "headers": {
    //    "Accept-Encoding": "br",
    //    "Content-Length": "0",
    //    "Host": "httpbin.org",
    //    "X-Amzn-Trace-Id": "Root=1-5f833c44-343c4ab909998d701ad5df2d"
    //  },
    //  "json": null,
    //  "method": "GET",
    //  "origin": "not-gonna-leak-my-ip",
    //   "url": "https://httpbin.org/anything"
    // }
    println!("{}", String::from_utf8(response.to_vec()).unwrap());
}
```

I was able to fix it manually with 
```rust
                    .header(
                        header::ACCEPT_ENCODING,
                        HeaderValue::from_static("identity"),
                    )
```
but this should not be the default with these feature flags.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.48.0-nightly (7402a3944 2020-09-13)
* Actix Web Version: 3.1.0, `default-features = false` to disable compression

j-brn:
## PR Type
Bugfix

## Overview
Fixes https://github.com/actix/actix-web/issues/1731 by using `Accept-Encoding: identity` when no compression feature is enabled.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,14 +1,9 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
-
 * Implement Logger middleware regex exclude pattern [#1723]
+
 [#1723]: https://github.com/actix/actix-web/pull/1723
-
-### Fixed
-
-* awc now uses `Accept-Encoding: identity` instead of `Accept-Encoding: br` when no compression feature is enabled
-[#1737](https://github.com/actix/actix-web/pull/1737)
 
 ## 3.1.0 - 2020-09-29
 ### Changed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased - 2020-xx-xx
 
+### Fixed
+
+* awc now uses `Accept-Encoding: identity` instead of `Accept-Encoding: br` when no compression feature is enabled
+[#1737](https://github.com/actix/actix-web/pull/1737)
 
 ## 3.1.0 - 2020-09-29
 ### Changed
--- a/awc/src/request.rs
+++ b/awc/src/request.rs
@@ -23,8 +23,17 @@ use crate::ClientConfig;
 
 #[cfg(any(feature = "flate2-zlib", feature = "flate2-rust"))]
 const HTTPS_ENCODING: &str = "br, gzip, deflate";
-#[cfg(not(any(feature = "flate2-zlib", feature = "flate2-rust")))]
+#[cfg(all(
+    not(any(feature = "flate2-zlib", feature = "flate2-rust")),
+    feature = "compress"
+))]
 const HTTPS_ENCODING: &str = "br";
+#[cfg(not(any(
+    feature = "flate2-zlib",
+    feature = "flate2-rust",
+    feature = "compress"
+)))]
+const HTTPS_ENCODING: &str = "identity";
 
 /// An HTTP Client request builder
 ///
--- a/awc/src/request.rs
+++ b/awc/src/request.rs
@@ -21,19 +21,15 @@ use crate::frozen::FrozenClientRequest;
 use crate::sender::{PrepForSendingError, RequestSender, SendClientRequest};
 use crate::ClientConfig;
 
-#[cfg(any(feature = "flate2-zlib", feature = "flate2-rust"))]
-const HTTPS_ENCODING: &str = "br, gzip, deflate";
-#[cfg(all(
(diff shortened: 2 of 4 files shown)


awc/src/request.rs
```diff
@@ -23,8 +23,17 @@ use crate::ClientConfig;
...
+    not(any(feature = "flate2-zlib", feature = "flate2-rust")),
+    feature = "compress"
+))]
 const HTTPS_ENCODING: &str = "br";
+#[cfg(not(any(
+    feature = "flate2-zlib",
+    feature = "flate2-rust",
+    feature = "compress"
+)))]
+const HTTPS_ENCODING: &str = "identity";
```
robjtede: would it be clearer to use `cfg-if` ?

j-brn: Do you mean https://docs.rs/cfg-if/1.0.0/cfg_if ? I can't find anything like `cfg-it` it under https://doc.rust-lang.org/reference/conditional-compilation.html

j-brn: That looks nice indeed and its already in the dependency tree. I'll change it.

robjtede: should avoid any situations where a combination of feature flags causes compile errors

j-brn: Done. The solution i originally had i mind doesn't work with the macro unfortunately, but it allows to reduce the complexity of the cfg attributes, so its still a win i guess.
```rust
const HTTPS_ENCODING: &str = cfg_if::cfg_if! {
    if #[cfg(any(feature = "flate2-zlib", feature = "flate2-rust"))] {
         "br, gzip, deflate";
    } else if #[cfg(feature = "compress")] {
        "br";
    } else {
        "identity";
    }
};
```

JohnTitor (approved):
Looks good, could you add an entry to the changelog as this changes the current behavior as a bug fix?

j-brn:
> 
> 
> Looks good, could you add an entry to the changelog as this changes the current behavior as a bug fix?

Added in c2aeea512d4770370f80423d4217bc5074f8b09c

CHANGES.md
```diff
@@ -1,10 +1,15 @@
...
-* Implement Logger middleware regex exclude pattern [#1723]
 
+* Implement Logger middleware regex exclude pattern [#1723]
 [#1723]: https://github.com/actix/actix-web/pull/1723
 
+### Fixed
+
+* awc now uses `Accept-Encoding: identity` instead of `Accept-Encoding: br` when no compression feature is enabled
+[#1737](https://github.com/actix/actix-web/pull/1737)
+
```
robjtede: this should be in `awc/CHANGEs.md`

CHANGES.md
```diff
@@ -1,10 +1,15 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
-* Implement Logger middleware regex exclude pattern [#1723]
 
+* Implement Logger middleware regex exclude pattern [#1723]
```
robjtede: undo the diff noise

CHANGES.md
j-brn: -> 06841e111c58a30537455fb2e0ba8e72df0ffcfe

CHANGES.md
j-brn: -> 06841e111c58a30537455fb2e0ba8e72df0ffcfe

robjtede (approved):

PR #1741, opened 2020-10-19
always construct shortslice using debug checked new constructor

### LINKED ISSUE #1672: ShortSlice is sometimes constructed bypassing checks
## Expected Behavior
The `ShortSlice` type is designed to contain a slice of 8 bytes or less, and guarantee that invariant so that unsafe code can rely on it. The constructor is unsafe to ensure that the invariant is upheld:

https://github.com/actix/actix-web/blob/64a2c13cdfed3616bf6f3ac2c39a5e1b40bfada4/actix-http/src/ws/mask.rs#L6-L20

## Current Behavior
It is currently possible to create a ShortSlice without any `unsafe` and bypassing the debug assertion.
In fact, this line does exactly that:
https://github.com/actix/actix-web/blob/64a2c13cdfed3616bf6f3ac2c39a5e1b40bfada4/actix-http/src/ws/mask.rs#L99

Notice the `ShortSlice(tail)` that doesn't go through `::new()`

## Possible Solution
The `ShortSlice` struct shoud not be directly constructable, e.g. by having a private zero-sized field and coercing into the underlying slice via Deref trait.

Despite allowing violation of an unsafe invariant, this does not have security implications in practice given the current surrounding code.

robjtede:
## PR Type
Fix

## PR Checklist

## Overview
Tiny safety fix. Not invocable outside the module.

closes #1672

### DIFF
--- a/actix-http/src/ws/mask.rs
+++ b/actix-http/src/ws/mask.rs
@@ -4,7 +4,9 @@ use std::ptr::copy_nonoverlapping;
 use std::slice;
 
 // Holds a slice guaranteed to be shorter than 8 bytes
-struct ShortSlice<'a>(&'a mut [u8]);
+struct ShortSlice<'a> {
+    inner: &'a mut [u8],
+}
 
 impl<'a> ShortSlice<'a> {
     /// # Safety
@@ -12,10 +14,11 @@ impl<'a> ShortSlice<'a> {
     unsafe fn new(slice: &'a mut [u8]) -> Self {
         // Sanity check for debug builds
         debug_assert!(slice.len() < 8);
-        ShortSlice(slice)
+        ShortSlice { inner: slice }
     }
+
     fn len(&self) -> usize {
-        self.0.len()
+        self.inner.len()
     }
 }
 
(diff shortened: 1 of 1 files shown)


popzxc (approved):

actix-http/src/ws/mask.rs
```diff
@@ -4,16 +4,17 @@ use std::ptr::copy_nonoverlapping;
 use std::slice;
 
 // Holds a slice guaranteed to be shorter than 8 bytes
-struct ShortSlice<'a>(&'a mut [u8]);
+struct ShortSlice<'a>(&'a mut [u8], ());
```
jplatte: This is just so one can't accidentally construct `ShortSlice(foo)`, right? Haven't seen this accomplished using an extra unit field before, why not

```suggestion
struct ShortSlice<'a> {
    inner: &'a mut [u8],
}
```

robjtede: It's kind of a meaningless change either way. The "fix" was making sure to use `::new` everywhere. Just changing the constructor in any way made sure the compiler flagged any additional offences. I like this structure better so will swap.

jplatte (approved):

Issue #1571, opened 2020-06-20
awc 2 does not follow redirects by default, contradicting documentation


Shnatsel:
## Expected Behavior
Documentation describes that HTTP redirects are followed by default.

## Current Behavior
Running the example code from docs.rs page for `awc` outputs HTTP code 301 instead of following the redirect and getting to 200.

## Steps to Reproduce (for bugs)
1. Run the code from the top of the page on https://docs.rs/awc/2.0.0-alpha.2/awc/
2. Observe 301 code in the output instead of 200

## Your Environment
* Rust Version (I.e, output of `rustc -V`): rustc 1.44.1 (c7087fe00 2020-06-17)
* awc: 2.0.0

robjtede:
contradicts docs here: https://docs.rs/awc/2.0.0/awc/struct.ClientBuilder.html#method.disable_redirects

robjtede:
I havent the foggiest idea where the builder methods are from for redirects because they do nothing at all and there is no hint of a redirect system anywhere in the client code. Whatsmore, it not possible to support redirections with the existing constructs without breaking changes. Will look into how a larger set of changes would impact the crate.

Needless to say, this is no longer a "good first issue" unless the task is to remove the redirect claims from the docs and deprecate the builder methods.

thiagoarrais:
I have been trying to add redirection with no success (I'm somewhat of a Rust newbie, btw). I'll pause that to work on removing the references to redirect from docs and submit a PR soon.

I'm still willing to add support for redirects, though. I just will need some guidance. Would anyone help me here? I may need some mentoring regarding the borrow checker, futures and all.

robjtede:
@thiagoarrais Yes, we'd much prefer to actually have the feature. I'd be happy to mentor this feature. I'm considering this issue the tracking issue for the feature.

Feel free to put up a draft PR if you'd like, so we can see what's happening as you make progress. If you have questions I'm active in the [Gitter chat](https://gitter.im/actix/actix-web).

robjtede:
Some notes based research I've done so far:
- [This MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections) is a great reference for the semantics of each redirect type (there are a few and some of them are kinda funky).
- For security reasons, we should detect cross-origin redirects and omit sensitive headers.
- If we can, request bodies should be re-used but it's okay to document a known limitation that streamed request bodies are not resent in redirects.

spikecodes:
[...]

PR #1745, opened 2020-10-19
Deprecate functions for following redirects

thiagoarrais:
## PR Type

Documentation

## PR Checklist

## Overview

I'm planning on adding support for following redirects. For the time being, here's a PR that adds deprecation notices to the
builder functions that reference redirects since they are effectively no-ops for the time being.

Related to #1571

### DIFF
--- a/awc/src/builder.rs
+++ b/awc/src/builder.rs
@@ -17,8 +17,6 @@ use crate::{Client, ClientConfig};
 /// builder-like pattern.
 pub struct ClientBuilder {
     default_headers: bool,
-    allow_redirects: bool,
-    max_redirects: usize,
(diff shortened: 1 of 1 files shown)


robjtede:
Do you think it's worth doing this? I assume you'll be adding them back straight away if you're intending to tackle adding in redirect support?

thiagoarrais:
I've made this one just in case I couldn't finish the real redirect one. This way HEAD at least is consistent.

Feel free to close this if you wish.

robjtede:
I think so, for now. If we don't see redirects coming soonish then we'll deprecate them. My hope is that now it has attention it can get coded up pretty quick.

PR #1743, opened 2020-10-19
Print unconfigured `Data<T>` type when attempting extraction

cquintana-verbio:
## PR Type

Other

## PR Checklist
Check your PR fulfills the following:

- [x] A changelog entry has been made for the appropriate packages.
- [X] Format code with the latest stable rustfmt

## Overview

This PR aims to help developers when facing the 

```
Failed to construct App-level Data extractor. Request path: "/your-endpoint"
```

Now, it will print something like the following:

```
Failed to construct App-level Data extractor. Request path: "/your-endpoint" (type = THE_TYPE_WILL_APPEAR_HERE)
```

PR #1729, opened 2020-10-09
Add web::ReqData extractor

jplatte:
## PR Type

Feature

## PR Checklist

- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview

Adds a `ReqData` extractor for request data / request extensions, analogous to `web::Data`.

PR #1748, opened 2020-10-20
Add `web::ReqData<T>` extractor

### LINKED ISSUE #1742: request data extractor similar to web::Data
work started in #1729

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
Superceeds #1729 but includes original commits by @jplatte.

Adds a `ReqData` extractor for request data / request extensions, analogous to `web::Data`.

Changes from original:
- remove `DerefMut` impl
  - i see this as a footgun, devs may infer that data is mutated in place (in the extensions map)
- make inner field private
  - matches Data
  - can always be make pub but not the other way without breaking change
- prints a failed extraction's type name (inspired by #1743)
- provide an into_inner
- improve docs
  - adding an example
  - clarify that data is cloned
- couple words changed on app data extractor to distingush from req data
  - plus formatting tweaks

Future work:
- support unsized
- mention possibility of `Rc<RefCell<T>>` in docs for in-place mutation (theres a test demonstrating this)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,17 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
-* Implement Logger middleware regex exclude pattern [#1723]
-* Print unconfigured `Data<T>` type when attempting extraction. [#1743]
+### Added
+* Implement `exclude_regex` for Logger middleware. [#1723]
+* Add request-local data extractor `web::ReqData`. [#1729]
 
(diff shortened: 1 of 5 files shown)


jplatte:
> can always be make pub but not the other way without breaking change

This isn't true due to `Deref` behavior, as we have seen with making the `Path` extractors inner value public. That said, I'm fine with this.

PR #1749, opened 2020-10-21
register fns for custom request-derived logging units

jdeepee:
## PR Type
Feature

## PR Checklist

## Overview
Adds possibility to register a closure to be executed in request middleware logging. Closure is implemented to work in the request scope and thus takes `ServiceRequest` as input.

### Our use case:
We wanted to log the UserID associated with all incoming requests to our API. UserID's in our case are stored inside JWT tokens. This means we cannot log the header directly but need to make some operations on it to get the internal encoded UserId. By allow the injection of this closure we can do the following:

```rust
Logger::new().custom_request_replace("jwt", |req: &ServiceRequest| {
      let user_token = req.headers().get("authorization");
      decode_jwt(user_token)
    }
)
```


robjtede:
For this to be considered, I need to know how this is different from just using a middleware function (`App::wrap_fn`) apart from it being on the same line as the Logger output.

jdeepee:
Ah, I didn't think about using `App::wrap_fn`. Is there someway that inside the `wrap_fn` closure I can call the logger? If there is not then the fact the logs are output in two different lines is a game breaker for our intended implementation.

If I have a middleware logging incoming requests and a `wrap_fn()` logging our user_id's there is no way to guarantee that other requests do not arrive in between middleware logging and `wrap_fn()` logging. Ive validated this problem with the following simple hello world app and a bash script that sends requests in parallel.

Actix app
```rust
#[actix_web::main]
    std::env::set_var("RUST_LOG", "actix_web=info");
    env_logger::init();

    HttpServer::new(|| {
        App::new()
            // enable logger
            .wrap(middleware::Logger::new(r#"%a "%r" %s %b "%{Referer}i" "%{User-Agent}i" %T "#).register_request_closure(|req: &ServiceRequest| {
                req.headers().get("Test").unwrap().to_str().unwrap().to_owned()
            }))
            .wrap_fn(|req, srv| {
                let header = req.headers().get("Test").unwrap().to_str().unwrap().to_owned();
                let fut = srv.call(req);
                async move {
                    let res = fut.await?;
                    println!("{:#?}", header);
                    Ok(res)
                }
            })
            .service(web::resource("/").to(|| async { "Hello world!" }))
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

Bash script
```sh
#!/bin/bash

curl -H 'Test: test1' localhost:8080 && echo "done1" &
curl -H 'Test: test3' localhost:8080 && echo "done2" &
curl -H 'Test: test2' localhost:8080 && echo "done3" &

wait
```

Actix-web Output
```
"test2"
"test3"
"test1"
[2020-10-22T09:57:53Z INFO  actix_web::middleware::logger] 127.0.0.1:51172 "GET / HTTP/1.1" 200 12 "-" "curl/7.68.0" 0.000057 test2
[2020-10-22T09:57:53Z INFO  actix_web::middleware::logger] 127.0.0.1:51170 "GET / HTTP/1.1" 200 12 "-" "curl/7.68.0" 0.000062 test3
[2020-10-22T09:57:53Z INFO  actix_web::middleware::logger] 127.0.0.1:51174 "GET / HTTP/1.1" 200 12 "-" "curl/7.68.0" 0.000052 test1
```

Here you can see the headers being printed in a way that causes a disjoint between the middleware logger and the `wrap_fn()`.
I could of course implement a custom logger in the `wrap_fn()` closure but would prefer to use an existing solution.

robjtede:
Alright, there is a reasonable use case for this to exist as you've demonstrated with log lines being disjoint. I think there is a more flexible way to achieve this: being able to define custom replacements to use in the format string.

We can currently use `%{FOO}i` to log the request header FOO. I think a map of string keys to closures (using a similar (&ServiceRequest) -> String) would be better as you could then add these parameters anywhere in the output.

Proposal:

Allow the format syntax `%{jwt}xi` (x => custom, i => incoming/request) to call the registered closure. (we should leave open the possibility for `%{FOO}xo` in the future).

```rust
Logger::custom_request_replace(&mut self, label: &str, f: Fn(&ServiceRequest) -> String) -> Self
```

(Open to better method name.)

jdeepee:
Okay above proposal has been implemented. Thanks for the outline; this is working quite nicely. 

You may notice that I have removed `Debug` from `FormatText`. There is a few reasons for this; some of which may not be valid, some verification/assistance here would be greatly appreciated. 

Firstly when trying to add the `CustomLog(String, Option<fn(req: &ServiceRequest) -> String>)` variant I was getting a rather strange error:

```
error: implementation of `std::fmt::Debug` is not general enough
   --> src/middleware/logger.rs:424:23
    |
424 |       CustomLog(String, Option<fn(req: &ServiceRequest) -> String>),
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `std::fmt::Debug` is not general enough
    | 
   ::: /home/josh/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:552:1
    |
552 | / pub trait Debug {
553 | |     /// Formats the value using the given formatter.
554 | |     ///
555 | |     /// # Examples
...   |
583 | |     fn fmt(&self, f: &mut Formatter<'_>) -> Result;
584 | | }
    | |_- trait `std::fmt::Debug` defined here
    |
    = note: `std::fmt::Debug` would have to be implemented for the type `for<'r> fn(&'r ServiceRequest) -> std::string::String`
    = note: ...but `std::fmt::Debug` is actually implemented for the type `fn(&'0 ServiceRequest) -> std::string::String`, for some specific lifetime `'0`
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error

error: could not compile `actix-web`
```

I tried messing around with lifetimes in order to create a more "specific" lifetime but had no luck in getting this to work. I also stumbled across this [issue](https://github.com/rust-lang/rust/issues/70263) whilst researching this problem. This made my wonder if the problem is caused by my shallow knowledge of lifetimes or if its a rust issue. 

In any case I noticed that `Debug` is being used anywhere so for now have removed this to test other functionality. Please let me know if `Debug` should be added again.

src/middleware/logger.rs
```diff
@@ -400,6 +421,7 @@ pub enum FormatText {
     RequestHeader(HeaderName),
     ResponseHeader(HeaderName),
     EnvironHeader(String),
+    CustomLog(String, Option<fn(req: &ServiceRequest) -> String>),
```
robjtede: since this is technically pub, we should indicate "request" like `CustomRequest`.

robjtede: So for the function, a common way around the Debug thing is a struct wrapping a function with a manual Debug impl.

I recommend using something like [this struct in actix-cors](https://docs.rs/actix-web/3.1.0/actix_web/struct.Resource.html#method.name):
```rust
struct CustomRequestFn {
    pub(crate) boxed_fn: Rc<dyn Fn(&HeaderValue, &RequestHead) -> bool>,}
```

Also using `Fn` over `fn` allows closures to capture from the environment (eg. a lazy instantiated regex). It's worth adding a test for this too.

src/middleware/logger.rs
```diff
@@ -123,6 +123,25 @@ impl Logger {
         inner.exclude_regex = regex_set;
         self
     }
+
+    /// Register a closure to be run on the request output of the logger.
+    /// Label passed into this function will be used to match output of given closure to label found in log format line with format %{label}xi.
+    pub fn custom_request_replace(
+        mut self,
+        label: &'static str,
+        closure: fn(req: &ServiceRequest) -> String,
```
robjtede:
```suggestion
        closure: impl Fn(req: &ServiceRequest) -> String,
```

src/middleware/logger.rs
```diff
@@ -123,6 +123,25 @@ impl Logger {
...
+    ) -> Self {
+        let inner = Rc::get_mut(&mut self.0).unwrap();
+
+        for tf in inner.format.0.iter_mut() {
+            if let FormatText::CustomLog(inner_label, _inner_closure) = tf {
+                if inner_label == label {
+                    *tf = FormatText::CustomLog(label.to_string(), Some(closure))
```
robjtede: since this is in an iter_mut could we not simply assign to `_inner_closure`?

robjtede (changes requested):
I'm liking how this is turning out. Good work. LMK if you have time to fix up these tweaks or not :)

src/middleware/logger.rs
jdeepee: Awesome thanks for the heads up on this

jdeepee:
Ive added most of the changes you outlined. Thanks for the assistance with that; was nice to go a little deeper into the world of closures.

Right now in `custom_request_replace()`  I am cloning the `Fn` into `Rc` as to satisfy the `'static` lifetime? There might be a way to do this without clone, but I don't know it...

Also didn't add any test for capturing env for the `Fn`. If you could take care of these pieces that would be much appreciated :)

robjtede:
np.. nice update. actix-cors managed to avoid require the Clone bound on the closure because it's being put in a Rc wrapper anyway (see https://github.com/actix/actix-extras/blob/master/actix-cors/src/builder.rs#L185). I'd need to have a mess with the code but it's forsure possible.

still would like that enum variant to elude to it referencing Requests, since it's public this will be locked as a name for at least the next major version

robjtede (approved):
I've tweaked and cleaned up a few things. This is looking ready.

robjtede:
Quick Q for @actix/contributors? Strictly this is breaking change to anyone who is exhaustively matching over the Logger format types. I would consider it harmless since it has always been hidden from docs.

Ergo, I would propose our policy should be server exemptions for doc(hidden) items. Can I get a :+1: or :-1: on that?

PR #1750, opened 2020-10-23
Re-export bytes::Buf and bytes::BufMut as well

therealprof:
## PR Type

Other

## PR Checklist

## Overview

actix-web is re-exporting `bytes::Bytes` and `bytes::BytesMut`, however it does not re-export the traits `bytes::Buf` and `bytes::BufMut` which are required to actually do something with the types in applications which means that users of those types have to import the traits from `bytes` directly which (as it currently the case) can cause version clashes. By re-exporting the traits as well we can eliminate this stumbling block and simplify the use.

Signed-off-by: Daniel Egger <daniel.egger@axiros.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,10 +7,12 @@
 
 ### Changed
 * Print non-configured `Data<T>` type when attempting extraction. [#1743]
+* Re-export bytes::Buf{Mut} in web module. [#1750]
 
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
Yeah, makes sense for me, could you add an entry to the changelog?

robjtede (approved):
Nice, this is more important now that bytes 0.6 is out.

PR #1752, opened 2020-10-23
Add redirect support

### LINKED ISSUE #1571: awc 2 does not follow redirects by default, contradicting documentation

thiagoarrais:
## PR Type

Feature

## PR Checklist

TODO:

- [x] allow configuration (enable/disable and max redirects)
- [x] trigger body resend per status code
- [x] trigger method switching per status code
- [ ] document that won't resend streamed bodies
- [ ] should also work with send_request_extra
- [ ] sanitize headers on cross-origin redirect
- [x] refactor unwraps out

## Overview

This makes AWC follow redirects by default and allows configuration to disable it.

This may be a *breaking change* because, even if the documentation says redirects are followed by default, they weren't followed in practice.

Closes #1571

### DIFF
--- a/awc/CHANGES.md
+++ b/awc/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Unreleased - 2020-xx-xx
 
+### Fixed
+* Follows redirects and respects `disable_redirects` and `max_redirects` options.
+
 
 ## 2.0.3 - 2020-11-29
 ### Fixed
--- a/awc/Cargo.toml
+++ b/awc/Cargo.toml
@@ -56,6 +56,7 @@ serde_json = "1.0"
 serde_urlencoded = "0.7"
 open-ssl = { version = "0.10", package = "openssl", optional = true }
 rust-tls = { version = "0.18.0", package = "rustls", optional = true, features = ["dangerous_configuration"] }
+if_chain = "1.0.1"
 
 [dev-dependencies]
 actix-connect = { version = "2.0.0", features = ["openssl"] }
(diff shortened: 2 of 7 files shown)


fakeshadow:
Why double layer of Rc<RefCell<>> and an additional dep?

None of these seems necessary and only slow down compile time and runtime.

btw redirect should be handled on lower level where redirect future can be pinned on stack or in poll methods. A recursive boxed future is just wasteful.

I would prefer a redirect with proper actix-web abstract and dispatch static future. For example an additional service layer around connector services handle retry and/or make `SendClientRequest` accept a static dispatched future as generic type. So servicel::call's typed future or a stack async block can handle redirect logic. This would limit every redirect future cost to one heap allocation instead of two.

Edit: After some look into the redirect should be done in `actix_http:client::h1proto` and `h2proto`. You can reuse the stream in it and write stacked loop async with redirect logic. Write it at higher level would result in additional overhead with constantly popping and pushing operation on connection pool(not including previous stated overhead) because sadly the existing code base depending on passing the ownership of stream to inner logic till it's owned by a framed reader/writer

robjtede:
we added redirect support as a middleware in #2013

there may be things missing or wrong still if you're interested in continuing with redirects

Issue #1727, opened 2020-10-09
Support MTLS


howard0su:
Support MTLS, the request handler is able to validate the certificate from the client.

If someone can point out the code to start with, I can try to do this task as well.

robjtede:
If possible, can you post some details/links about the rationale for this feature, some details just to spec this out and maybe some snippets of what the code should look like, ideally, when *using* the feature. Will help anyone looking into this about what is required.

howard0su:
## Definition
When TLS server accept the client connection, it is optional to request the client present its own certificate. This is a feature supported in all TLS library. It can be enabled by the configuration of the TLS server.

## Use Cases
This is common used feature when you are building a API server and the API server is using client certificate to authenticate itself to the server. It is common in the microservice architecture.

## API
This is not proposal of API but rather a way to give the reader on how to use MTLS.

When setup TLS listener, the code need to specific one root certificate or a list of certificate so that the server will only accept a client certificate is trusted by those root certificate. Or the code can configure the server can accept the connection with/without the certificate as well.

In the handler, the server code can get the client certificate's information, including but not limiting to the subject, altname, thumbprint, if it is trusted by a predefined root certificates. Based on the information, the handler is able to return the different responses.

robjtede:
Great, thanks for that.

If you want to have a go at implementing this, a good place to start is in the actix-server + actix-tls crates. It may be possible already to use this feature manually at that level. Check out https://github.com/actix/actix-net/blob/master/actix-tls/examples/basic.rs and perhaps see if it can be modified to achieve this result. If not, modifying one of those libraries will probably be the first step.

howard0su:
thank you for your suggestion. I will look into this.

robjtede:
Also potentially relevant for higher level work: https://github.com/actix/actix-web/pull/1482

robjtede:
@howard0su I've made progress on exposing this in #1754. It can provide access to `rustls::Certificate` or `openssl::x509::X509` and should be enough to support this use case.

robjtede:
actix-web v3.2 is released which exposes `on_connect`.

Example of extracting client certificate: https://github.com/actix/examples/tree/HEAD/rustls-client-cert

PR #1754, opened 2020-10-25
expose on_connect v2

robjtede:
## PR Type
Feature

## PR Checklist
  - [x] Client certs can be accessed.

## Overview
Builds on #1482 (which will superceed if this is merged), which attempted to use the existing on_connect methods in -http to add data into the request-local extensions container. This was along the right lines, but I believe the approach in this PR is better for a couple of reasons:
- doesn't add a type parameter to `HttpServer`
- multiple types can be added to request extensions

I believe this will unblock #1727.

As you created the original PR @MikailBag, I'd appreciate your thoughts on this modifed approach.


MikailBag (approved):
This is much better than my PR: ugly hack 1 (storing extention type in Server signature) successfully dropped!

Just as a reminder, there is also ugly hack 2: passing &dyn Any as handler input. I still don't have ideas how to make it more typed.

robjtede:
I don't think there's a way to remove that unless we do one on connect method for each type like on_connect_openssl etc.

robjtede:
heres the extract client certificate example, i'm going to move it to the examples repo and revert the in-repo example to the simpler one as before:

<details>
<summary>Example Code</summary>

```rust
use std::{any::Any, env, fs::File, io::BufReader};

use actix_tls::rustls::{ServerConfig, TlsStream};
use actix_web::{
    dev::Extensions, rt::net::TcpStream, web, App, HttpResponse, HttpServer, Responder,
};
use log::info;
use rust_tls::{
    internal::pemfile::{certs, pkcs8_private_keys},
    AllowAnyAnonymousOrAuthenticatedClient, Certificate, RootCertStore, Session,
};

const CA_CERT: &str = "examples/certs/rootCA.pem";
const SERVER_CERT: &str = "examples/certs/server-cert.pem";
const SERVER_KEY: &str = "examples/certs/server-key.pem";

#[derive(Debug, Clone)]
struct ConnectionInfo(String);

async fn route_whoami(
    conn_info: web::ReqData<ConnectionInfo>,
    client_cert: Option<web::ReqData<Certificate>>,
) -> impl Responder {
    if let Some(cert) = client_cert {
        HttpResponse::Ok().body(format!("{:?}\n\n{:?}", &conn_info, &cert))
    } else {
        HttpResponse::Unauthorized().body("No client certificate provided.")
    }
}

fn get_client_cert(connection: &dyn Any, data: &mut Extensions) {
    if let Some(tls_socket) = connection.downcast_ref::<TlsStream<TcpStream>>() {
        info!("TLS on_connect");

        let (socket, tls_session) = tls_socket.get_ref();

        let msg = format!(
            "local_addr={:?}; peer_addr={:?}",
            socket.local_addr(),
            socket.peer_addr()
        );

        data.insert(ConnectionInfo(msg));

        if let Some(mut certs) = tls_session.get_peer_certificates() {
            info!("client certificate found");

            // insert a `rustls::Certificate` into request data
            data.insert(certs.pop().unwrap());
        }
    } else if let Some(socket) = connection.downcast_ref::<TcpStream>() {
        info!("plaintext on_connect");

        let msg = format!(
            "local_addr={:?}; peer_addr={:?}",
            socket.local_addr(),
            socket.peer_addr()
        );

        data.insert(ConnectionInfo(msg));
    } else {
        unreachable!("socket should be TLS or plaintext");
    }
}

#[actix_web::main]
    if env::var("RUST_LOG").is_err() {
        env::set_var("RUST_LOG", "info");
    }

    env_logger::init();

    let ca_cert = &mut BufReader::new(File::open(CA_CERT)?);

    let mut cert_store = RootCertStore::empty();
    cert_store
        .add_pem_file(ca_cert)
        .expect("root CA not added to store");
    let client_auth = AllowAnyAnonymousOrAuthenticatedClient::new(cert_store);

    let mut config = ServerConfig::new(client_auth);

    let cert_file = &mut BufReader::new(File::open(SERVER_CERT)?);
    let key_file = &mut BufReader::new(File::open(SERVER_KEY)?);

    let cert_chain = certs(cert_file).unwrap();
    let mut keys = pkcs8_private_keys(key_file).unwrap();
    config.set_single_cert(cert_chain, keys.remove(0)).unwrap();

    HttpServer::new(|| App::new().route("/", web::get().to(route_whoami)))
        .on_connect(get_client_cert)
        .bind(("localhost", 8080))?
        .bind_rustls(("localhost", 8443), config)?
        .workers(1)
        .run()
        .await
}

```
</details>

Test with cURL or HTTPie:

```sh
curl https://127.0.0.1:8443/ --cacert examples/certs/rootCA.pem --cert examples/certs/client-cert.pem --key examples/certs/client-key.pem
# or
http https://127.0.0.1:8443/ --verify=false --cert=examples/certs/client-cert.pem --cert-key=examples/certs/client-key.pem
```

fakeshadow:
The Any type here should be either TcpStream or UnixStream and their wrapper types. They both have actix_server::FromStream trait and technically we can add additional trait bound to HttpServer for it. The problem is we would kick the problem down the line to actix-http where we don't have matched generic trait method for something we want from these streams.(I could be wrong as I don't actually get it compiled as the trait origins in actix-net repo)

I believe it can be worked out later.

fakeshadow (commented):
lgtm.

PR #1758, opened 2020-10-26
Adding app_data to ServiceConfig

### LINKED ISSUE #1757: add `app_data` equivalent to `ServiceConfig`

augustocdias:
## PR Type

Adding `app_data` for `ServiceConfig` struct.

## PR Checklist

## Overview

Currently it is not possible to register `app_data` through the `App::configure` builder.
This will essentially merge the `app_data` from the `ServiceConfig` with the `App`'s.

### DIFF
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -62,9 +62,9 @@ impl Extensions {
         self.map.clear();
     }
 
-    /// Extends it with the content for another `Extensions`
-    pub fn extend(&mut self, extensions: Extensions) {
-        self.map.extend(extensions.map);
+    /// Extends self with the items from another `Extensions`.
+    pub fn extend(&mut self, other: Extensions) {
+        self.map.extend(other.map);
     }
 }
 
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -61,6 +61,11 @@ impl Extensions {
     pub fn clear(&mut self) {
         self.map.clear();
     }
+
+    /// Extends it with the content for another `Extensions`
+    pub fn extend(&mut self, extensions: Extensions) {
+        self.map.extend(extensions.map);
+    }
 }
 
 impl fmt::Debug for Extensions {
--- a/actix-http/src/extensions.rs
+++ b/actix-http/src/extensions.rs
@@ -66,6 +66,11 @@ impl Extensions {
     pub fn extend(&mut self, other: Extensions) {
         self.map.extend(other.map);
     }
+
+    /// Returns `true` if no extension is registered
+    pub fn is_empty(&self) -> bool {
+        self.map.is_empty()
+    }
 }
 
 impl fmt::Debug for Extensions {
@@ -93,6 +98,8 @@ mod tests {
     fn test_clear() {
         let mut map = Extensions::new();
 
+        assert!(map.is_empty());
+
         map.insert::<i8>(8);
         map.insert::<i16>(16);
         map.insert::<i32>(32);
@@ -100,12 +107,14 @@ mod tests {
         assert!(map.contains::<i8>());
         assert!(map.contains::<i16>());
         assert!(map.contains::<i32>());
+        assert!(!map.is_empty());
 
         map.clear();
 
         assert!(!map.contains::<i8>());
         assert!(!map.contains::<i16>());
         assert!(!map.contains::<i32>());
+        assert!(map.is_empty());
 
         map.insert::<i8>(10);
         assert_eq!(*map.get::<i8>().unwrap(), 10);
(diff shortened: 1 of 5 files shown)


actix-http/src/extensions.rs
```diff
@@ -61,6 +61,11 @@ impl Extensions {
     pub fn clear(&mut self) {
         self.map.clear();
     }
+
+    /// Extends it with the content for another `Extensions`
+    pub fn extend(&mut self, extensions: Extensions) {
+        self.map.extend(extensions.map);
+    }
```
robjtede:
```suggestion
    /// Extends self with the items from another `Extensions`.
    pub fn extend(&mut self, other: Extensions) {
        self.map.extend(other.map);
    }
```

src/config.rs
```diff
@@ -254,13 +267,16 @@ mod tests {
     async fn test_data() {
         let cfg = |cfg: &mut ServiceConfig| {
             cfg.data(10usize);
+            cfg.app_data(10usize);
```
robjtede: can you make this a different number/type as a guard against it doing the same as `Data<T>`.

src/config.rs
```diff
@@ -198,6 +200,17 @@ impl ServiceConfig {
         self
     }
 
+    /// Set application level arbitrary data item.
+    ///
+    /// Application data stored with `App::app_data()` method is available
+    /// via `HttpRequest::app_data()` method at runtime.
+    ///
+    /// Note: this method is ignored for `Scope::configure` method
```
robjtede:
```suggestion
    /// Set arbitrary data item.
    ///
    /// This is same as `App::data()` method.
```

robjtede (changes requested):
I think that this should include `Scope::configure`.

actix-http/src/extensions.rs
```diff
@@ -61,6 +61,11 @@ impl Extensions {
     pub fn clear(&mut self) {
         self.map.clear();
     }
+
+    /// Extends it with the content for another `Extensions`
+    pub fn extend(&mut self, extensions: Extensions) {
```
popzxc: Wouldn't implementing [`std::iter::Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) make more sense?

Also, unit-tests for this functionality would be appreciated.

augustocdias: I agree. Though I'm not being able to make it compile... As I wanted to extend with another object of the same type and not with the internal structure...

```rust
impl Extend<Extensions> for Extensions {

    fn extend<T: IntoIterator<Item = Extensions>>(&mut self, iter: T) {
        self.map.extend(iter);
    }
}

error[E0271]: type mismatch resolving `<T as std::iter::IntoIterator>::Item == (std::any::TypeId, std::boxed::Box<(dyn std::any::Any + 'static)>)`
  --> actix-http/src/extensions.rs:75:18
   |
75 |         self.map.extend(iter);
   |                  ^^^^^^ expected struct `extensions::Extensions`, found tuple
   |
   = note: expected struct `extensions::Extensions`
               found tuple `(std::any::TypeId, std::boxed::Box<(dyn std::any::Any + 'static)>)`
```

popzxc: I believe it would require you to implement `IntoIterator` for `Extensions` as well.

actix-http/src/extensions.rs
```diff
@@ -66,6 +66,11 @@ impl Extensions {
     pub fn extend(&mut self, other: Extensions) {
         self.map.extend(other.map);
     }
+
+    /// Returns `true` if no extension is registered
+    pub fn is_empty(&self) -> bool {
```
robjtede, line 71: lets limit public API

```suggestion
    pub(crate) fn is_empty(&self) -> bool {
```

src/scope.rs
```diff
@@ -209,6 +209,12 @@ where
 
             self.data = Some(data);
         }
+
+        if !cfg.extensions.is_empty() {
+            let mut data = self.data.unwrap_or_else(Extensions::new);
+            data.extend(cfg.extensions);
+            self.data = Some(data);
+        }
```
robjtede: im not sure this is more efficient than simply: `self.data.extend(cfg.extensions)`

augustocdias: What if `self.data` is `None`?

actix-http/src/extensions.rs
augustocdias: @robjtede what is your opinion here? should we implement `IntoIterator` for `Extensions`?

robjtede: the current seems fine to me, it's not really useful on the developer side, only for actix-http => actix-web... i would make it pub(crate) if we could

src/scope.rs
robjtede: an if let with a mut ref then?

augustocdias: you mean something like this?

```rust
        if let Some(ref mut data) = self.data {
            data.extend(cfg.extensions);
        } else {
            self.data = Some(cfg.extensions);
        }
```

jplatte: If you want to be really concise:

```rust
self.data.get_or_insert_with(Extensions::new).extend(cfg.extensions);
```

actix-http/src/extensions.rs
popzxc: I'm not insisting either.

robjtede (approved):

PR #1760, opened 2020-10-26
deprecate builder if-x methods

robjtede:
## PR Type
Deprecation

## PR Checklist

## Overview
These methods serve little purpose where standard if-let constructions exists.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased - 2020-xx-xx
 * Upgrade `base64` to `0.13`.
 * Upgrade `pin-project` to `1.0`.
+* Deprecate `ResponseBuilder::{if_some, if_true}`.
 
 ## 2.0.0 - 2020-09-11
 * No significant changes from `2.0.0-beta.4`.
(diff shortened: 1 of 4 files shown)


popzxc (approved):
👍

jplatte (approved):
I guess the point was that you wouldn't need to bind the builder to a local variable with this in almost all cases. But I agree it's a bit weird and wouldn't personally use it because of that.

pka:
I'm struggling to find a replacement for the following `if_true` usage:
```rust
HttpResponse::Ok()
    .content_type("application/x-protobuf")
    .if_true(gzip, |r| {
        // data is already gzip compressed
        r.encoding(ContentEncoding::Identity)
            .header(header::CONTENT_ENCODING, "gzip");
    })
    .header(header::CACHE_CONTROL, format!("max-age={}", cache_max_age))
    .body(tile)
```

As soon as I split this into multiple statements I get `error[E0716]: temporary value dropped while borrowed` for the `ResponseBuilder` object, which isn't easy to solve.

robjtede:
~I admit this is a little ugly:~ Edit: improved.

```rust
let mut res = HttpResponse::Ok();
res.header(header::CACHE_CONTROL, format!("max-age={}", cache_max_age));
res.content_type("application/x-protobuf");

if gzip {
    // data is already gzip compressed
    res.encoding(ContentEncoding::Identity)
        .header(header::CONTENT_ENCODING, "gzip");
}

res.body(tile)
```

pka:
Thanks @robjtede, this is working!

PR #1767, opened 2020-10-30
Drop content length headers from 101 responses

Sebmaster:
## PR Type
Bug Fix

## PR Checklist

## Overview

After the refactor in 3901239128ac9d076cdaaeb46220117a8043f7ad 101 responses started to be forced to emit `transfer-encoding: chunked` headers. However, similarly to 100, 102 and 204, the HTTP standard says 101 MUST NOT contain either content-length or transfer-encoding.

This leads to incompatibilities in the ecosystem. While browsers just ignore the header, envoy is unable to process responses like this.

This change goes a bit further than just restoring the old behaviour and extends the skip_len flag to all of these status codes, meaning even if they were user provided, they'll be ignored in the future to better match the standard.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -1,6 +1,10 @@
 # Changes
 
 ## Unreleased - 2020-xx-xx
+### Fixed
+* Started dropping `transfer-encoding: chunked` and `Content-Length` for 1XX and 204 responses. [#1767]
+
+[#1767]: https://github.com/actix/actix-web/pull/1767
 
 
 ## 2.1.0 - 2020-10-30
--- a/actix-http/src/h1/encoder.rs
+++ b/actix-http/src/h1/encoder.rs
@@ -64,14 +64,17 @@ pub(crate) trait MessageType: Sized {
         // Content length
         if let Some(status) = self.status() {
             match status {
-                StatusCode::NO_CONTENT
-                | StatusCode::CONTINUE
-                | StatusCode::PROCESSING => length = BodySize::None,
-                StatusCode::SWITCHING_PROTOCOLS => {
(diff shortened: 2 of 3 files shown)


robjtede:
Cool. Can you drop a link to the section in the spec that states this for future reference.

Sebmaster:
Added to the commit, it's described here: https://tools.ietf.org/html/rfc7230#section-3.3.1

> A server MUST NOT send a Transfer-Encoding header field in any
   response with a status code of 1xx (Informational) or 204 (No
   Content).  A server MUST NOT send a Transfer-Encoding header field in
   any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of
   [RFC7231]).

Similar language can be found in the section after for `Content-Length`.

robjtede:
Somethings gone arwy with this branch. Needs a rebase. Can you also apply this patch?

```diff
             match status {
-                StatusCode::NO_CONTENT
-                | StatusCode::CONTINUE
+                StatusCode::CONTINUE
                 | StatusCode::SWITCHING_PROTOCOLS
-                | StatusCode::PROCESSING => {
+                | StatusCode::PROCESSING
+                | StatusCode::NO_CONTENT => {
+                    // skip content-length and transfer-encoding headers
+                    // See https://tools.ietf.org/html/rfc7230#section-3.3.1
+                    // and https://tools.ietf.org/html/rfc7230#section-3.3.2
                     skip_len = true;
                     length = BodySize::None
                 }
-                _ => (),
+                _ => {}
             }
```

Sebmaster:
Sure, done.

robjtede (approved):

JohnTitor (approved):
Looks reasonable to me!

PR #1786, opened 2020-11-13
Multipart client implementation

### LINKED ISSUE #195: Multipart writer
Add multipart writer support

rubdos:
## PR Type
Feature

## PR Checklist

## Overview

`awc` should be able to send multipart forms. Currently, a programmer needs to revert to additional, often unmaintained crates.  This adds a multipart body generator to `actix-multipart`, which currently contains just a parser for the Actix web server.

I am basing myself of https://github.com/cetra3/mpart-async/blob/master/src/client.rs and https://github.com/ferristseng/rust-multipart-rfc7578/pull/19

The former is by @cetra3, who wrote a few years ago in #195 that it was okay to base this on their code. The latter is my update to `multipart-rfc7578` to support the latest Actix, but it is an unresponsive PR.

I decided to make the `Form` type generic in the `dyn Stream` lifetime, as opposed to `mpart-async`, which is generic in the `Stream` implementation itself. Both have advantages and disadvantages, feel free to give me pointers to decide which approach is better for Actix.

Closes #195

### DIFF
--- a/actix-multipart/Cargo.toml
+++ b/actix-multipart/Cargo.toml
@@ -18,13 +18,16 @@ path = "src/lib.rs"
 [dependencies]
 actix-web = { version = "3.0.0", default-features = false }
 actix-service = "1.0.6"
+actix-http = "2.1.0"
 actix-utils = "2.0.0"
 bytes = "0.5.3"
 derive_more = "0.99.2"
 httparse = "1.3"
+futures = { version = "0.3.5", default-features = false }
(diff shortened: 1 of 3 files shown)


rubdos:
Actually, `mpart-async` seems to fill the void and *is* maintained. Sorry for the noise, I'm not going to take it on any more if `mpart-async` is that well maintained. I'll leave my branch open if anyone else would want to get started.

PR #1788, opened 2020-11-15
add either extractor

### LINKED ISSUE #1724: implement `FromRequest` for `Either<A, B>`
Make it easier to use fallback extractors.

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
Add extractor helper using the existing `Either<A, B>` type to handle fallbacks, like `Either<Json<Login>, Form<Login>>` for example.

closes #1724


src/types/either.rs
```diff
@@ -0,0 +1,254 @@
...
+///         )
+///     }
+/// }
+/// # fn is_a_variant() -> bool { true }
+/// # fn main() {}
+/// ```
+#[derive(Debug, PartialEq)]
+pub enum Either<A, B> {
+    /// First branch of the type
+    A(A),
```
jplatte, line 40: Why `A` over `Left` if the docs and test code use left / right terminology rather than A / B?

src/types/either.rs
```diff
@@ -0,0 +1,254 @@
...
+            EitherExtractError::B(err) => err.into(),
+        }
+    }
+}
+
+/// Provides a mechanism for trying two extractors, a primary and a fallback. Useful for
+/// "polymorphic payloads" where, for example, a form might be JSON or URL encoded.
+///
+/// It is important to note that this extractor, by necessity, buffers the entire request payload
+/// as part of it's implementation. Though, it does respect a `PayloadConfig`'s maximum size limit.
```
jplatte:
```suggestion
/// as part of its implementation. Though, it does respect a `PayloadConfig`'s maximum size limit.
```

src/types/either.rs
robjtede: It's because this isnt a new type (i just moved the either parts out of responder.rs), it's in my head to rename the variants in next breaking release.

jplatte (approved):

fakeshadow (approved):
lgtm

fakeshadow:
I like this extractor. The only thing I would like to improve on is avoid using another box.
Maybe something like this https://gist.github.com/fakeshadow/7f42a2b82ab46f7ae417a6233764dba7.
It's not necessary or make things better by any means though.

robjtede:
We'll leave it for now. A box is very little considering the rest of this change.

One thing I'd just like an explicit thought on is the choice to return the error for A in the case of both failing. Currently the B error is never made available. The alternatives are to only show B's error or to return both errors in a tuple.

PR #1797, opened 2020-11-24
refactor quality and use TryFrom instead of custom trait

robjtede:
## PR Type
Refactor

## PR Checklist

## Overview
Refactor Quality struct and associated methods. Now uses TryFrom for u16 and f32 conversions instead of the old custom sealed trait IntoQuality. Since it was sealed this change is fully compatible (the `q` function still panics on out-of-range numbers) but now Quality can be created by lib consumer using these standard traits as well.

### DIFF
--- a/actix-http/src/header/shared/quality_item.rs
+++ b/actix-http/src/header/shared/quality_item.rs
@@ -1,10 +1,17 @@
-use std::{cmp, fmt, str};
+use std::{
+    cmp,
+    convert::{TryFrom, TryInto},
+    fmt, str,
+};
 
-use self::internal::IntoQuality;
+use derive_more::{Display, Error};
+
+const MAX_QUALITY: u16 = 1000;
+const MAX_FLOAT_QUALITY: f32 = 1.0;
 
 /// Represents a quality used in quality values.
 ///
-/// Can be created with the `q` function.
+/// Can be created with the [`q`] function.
 ///
 /// # Implementation notes
 ///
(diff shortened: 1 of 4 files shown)


actix-http/src/header/shared/quality_item.rs
```diff
@@ -21,9 +28,51 @@ use self::internal::IntoQuality;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Quality(u16);
 
+impl Quality {
+    /// # Panics
+    /// Panics in debug mode when value is not in the range 0.0 <= n <= 1.0.
+    fn from_f32(value: f32) -> Self {
+        // Check that `value` is within range should be done before calling this method.
+        // Just in case, this debug_assert should catch if we were forgetful.
+        debug_assert!(
```
jplatte, line 37: Why not a regular assert?

actix-http/src/header/shared/quality_item.rs
```diff
@@ -21,9 +28,51 @@ use self::internal::IntoQuality;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Quality(u16);
 
+impl Quality {
+    /// # Panics
+    /// Panics in debug mode when value is not in the range 0.0 <= n <= 1.0.
+    fn from_f32(value: f32) -> Self {
```
jplatte, line 34: Why is there a panicking c'tor for f32, but not for u16?

jplatte:
I guess you could also provide an infallible `From<u8>` conversion, but not sure whether that would be useful.

actix-http/src/header/shared/quality_item.rs
robjtede: only because it's private and private ctor for u16 is just `Quality(n)`

actix-http/src/header/shared/quality_item.rs
robjtede: Only because it's private and checks for the range *should* be done before using this somewhere in the lib code (as it is in the header parsing code). So it's wasteful to check the range twice in release mode.

actix-http/src/header/shared/quality_item.rs
jplatte: Oh, I didn't notice the missing `pub`. Must have assumed it's public because it has docs :smile:

jplatte (approved):

robjtede:
> I guess you could also provide an infallible From<u8> conversion, but not sure whether that would be useful.

Hmm, yeah... an infallible u8 constructor wouldn't be able to use what is arguably the most useful value of 1000.

PR #1798, opened 2020-11-27
Fix match_pattern() returning None for scope with resource of empty path

jdeepee:
## PR Type
Bug Fix

## PR Checklist

## Overview

`match_pattern()` was incorrectly returning a None match if a `web::scope()` was proceed by a resource with path "". Not completely sure if setting up scopes/resources in this configuration is some kind of anti-pattern but for our uses it works very well.


JohnTitor (approved):
Looks good! Could you add an entry to the changelog as this will change the behavior as bugfix?

jdeepee:
@JohnTitor thanks :) Changelog now also updated.

robjtede:
Thanks for this. I'll get a release together later with the latest patches.

PR #1803, opened 2020-11-30
Upgrade socket2 dependency

faern:
## PR Type
Bug Fix / Other: Upgrades a dependency to a version not making invalid assumptions about the memory layout of standard library types (`std::net::SocketAddr`).

Helps unblock https://github.com/rust-lang/rust/pull/78802. See that PR for more details. This PR simply helps adoption of the fixed version(s) of `socket2`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ futures-util = { version = "0.3.5", default-features = false }
 fxhash = "0.2.1"
 log = "0.4"
 mime = "0.3"
-socket2 = "0.3"
+socket2 = "0.3.16"
 pin-project = "1.0.0"
 regex = "1.4"
 serde = { version = "1.0", features = ["derive"] }


JohnTitor:
Since we don't lock `socket2` patch version and cargo tries to pull the newest `0.3.x` version, this doesn't make much sense for me.

faern:
Cargo only tries to pull the latest version when you have no lockfile. Anyone with a lockfile will continue to pull the version they already have until they do `cargo update` or `cargo update -p socket2`. This change would force the upgrade. It does not hurt `actix-web` in any way. All it does is help the adoption of the patch relesea of `socket2` that stops making invalid assumptions about the standard library.

The faster we can make the community adopt these fixed versions the faster the linked to PR on the standard library can advance.

Other crates that accepted the same type of bump in order to help out with adoption rate: https://github.com/alexcrichton/curl-rust/pull/365, https://github.com/hyperium/hyper/pull/2349 and https://github.com/rust-lang/cargo/pull/8909

JohnTitor:
> Cargo only tries to pull the latest version when you have no lockfile. Anyone with a lockfile will continue to pull the version they already have until they do `cargo update` or `cargo update -p socket2`. This change would force the upgrade.

But then you should do `cargo update -p actix-web`. Hm but yeah, we usually don't accept this type of PR, in this case, "invalid assumptions" seems a reasonable reason to accept.

JohnTitor (approved):

faern:
Thank you!

robjtede:
Yep I agree with this merge, in this case.

PR #1804, opened 2020-11-30
remove one indirection for HttpServiceHandlerResponse

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

This indirect is not necessary.


actix-http/src/cloneable.rs
```diff
@@ -10,7 +10,7 @@ use actix_service::Service;
 /// # Panics
 /// CloneableService might panic with some creative use of thread local storage.
 /// See https://github.com/actix/actix-web/issues/1295 for example
-pub(crate) struct CloneableService<T: Service>(Rc<RefCell<T>>);
+pub struct CloneableService<T: Service>(Rc<RefCell<T>>);
```
robjtede, line 13: I'm of the opinion that it's easier to grok the scope of these when they're pub(crate) vs pub without a lib re-export.

fakeshadow: It has to be public right? As HttpServiceHandlerResponse is now a public enum

robjtede: aite, can we hide it from docs then. The DataFactory trait too, I don't want to assume semver responsibility for either of these things.

fakeshadow:
Closing this for now. Looking back this is totally not a meaningful change.

PR #1806, opened 2020-12-01
fix clippy warnings

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Adresses many clippy warnings, either fixing or marking allowed.

FWIW, I did a quick benchmark of the entity validation change and saw no statistically significant change.

### DIFF
--- a/awc/src/ws.rs
+++ b/awc/src/ws.rs
@@ -70,9 +70,14 @@ impl WebsocketsRequest {
         <Uri as TryFrom<U>>::Error: Into<HttpError>,
     {
         let mut err = None;
-        let mut head = RequestHead::default();
-        head.method = Method::GET;
-        head.version = Version::HTTP_11;
+
+        #[allow(clippy::field_reassign_with_default)]
+        let mut head = {
+            let mut head = RequestHead::default();
+            head.method = Method::GET;
+            head.version = Version::HTTP_11;
+            head
+        };
 
         match Uri::try_from(uri) {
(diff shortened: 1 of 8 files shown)


awc/src/ws.rs
```diff
@@ -70,9 +70,14 @@ impl WebsocketsRequest {
         <Uri as TryFrom<U>>::Error: Into<HttpError>,
     {
         let mut err = None;
-        let mut head = RequestHead::default();
-        head.method = Method::GET;
-        head.version = Version::HTTP_11;
+
+        #[allow(clippy::field_reassign_with_default)]
+        let mut head = {
```
robjtede, line 75: Necessary because `flags` field is private and doesn't work with clippy's suggestion here.

Arguably a bug in clippy.

examples/enum_path.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[get("/res/{name}")]
+async fn index(name: web::Path<MyThing>) -> String {
+    println!("REQ: {:?}", name);
+
+    "Hello".to_owned()
+}
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
```
robjtede: oops, reminder to self to remove this

awc/src/ws.rs
JohnTitor: Yeah, seems a known FP: https://github.com/rust-lang/rust-clippy/issues/6344

PR #1825, opened 2020-12-12
feat config : service_if

nebnes:
## PR Type

add flag to service to conditionally add it or not

## PR Checklist


robjtede:
I disagree that this is cleaner than a built-in if statement.

For instance, registering multiple services using the same condition now requires evaluating the condition multiple times.

PR #1828, opened 2020-12-15
Add Chinese document  in README.md

zzy:
## PR Type

Other

## Overview

I have translated the website and User Guide of actix-web into Chinese, building it depends on mdBook instead of Hugo, so I add the repo's link and online address in README.md.


4ldas (approved):
Don't speak the chinese language myself but ill assume that its fine, also no need to merge from this repo, github does that for you :)

robjtede:
So this is definitely interesting, nice work doing the translation part!

My apprehension at this time is that this is a separate repo outside our control but displayed in a very prominent position on the readme, for a couple reasons:

From an approval POV, I don't know if anyone else in the team can understand Chinese but it feels like a necessary step to having this be featured officially.

From a maintenance POV, I'd like any documentation we have to stay reasonably up-to-date. It's going to be difficult to track when changes are needed and such in this form.

I'm wondering about the possibility of integrating this translation work directly into the website repo somehow.

Hopefully you can understand our position. What are you're thoughts on this?

zzy:
@robjtede @faith Hi, thank you. Okay, Let it stay in my repo, and I'll keep on updating from actix-website repo. 🤝

PR #1829, opened 2020-12-15
remove boxed future for Option<T> and Result<T, E> extract type

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

We don't have to box these two extract type and hand written future would do.
Also replace one futures::future::ok type with futures::future::ready.(Which already landed in std lib)

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -4,7 +4,7 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 
 use actix_http::error::Error;
-use futures_util::future::{ok, FutureExt, LocalBoxFuture, Ready};
+use futures_util::future::{ready, Ready};
 
 use crate::dev::Payload;
 use crate::request::HttpRequest;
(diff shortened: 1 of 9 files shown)


src/extract.rs
```diff
@@ -95,21 +95,40 @@ where
...
+
+impl<Fut, T, E> Future for FromRequestOptFuture<Fut>
+where
+    Fut: Future<Output = Result<T, E>>,
+    E: Into<Error>,
+{
+    type Output = Result<Option<T>, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project().fut.poll(cx) {
```
robjtede: little clearer with ready! macro

src/extract.rs
```diff
@@ -165,29 +184,47 @@ where
...
+}
+
+impl<Fut, T, E> Future for FromRequestResFuture<Fut>
+where
+    Fut: Future<Output = Result<T, E>>,
+{
+    type Output = Result<Result<T, E>, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project().fut.poll(cx) {
```
robjtede: same here

robjtede (approved):

src/extract.rs
fakeshadow: Trying to reduce the dependent on futures crate. At some point in the future it would be obsolete for actix-web crate.
We can already get rid of it if we bump MSRV to 1.48. 

I personally don't mind write some more code for matching polls. What do you think of the idea?

robjtede: ready macro is in futures-core

robjtede (approved):

PR #1831, opened 2020-12-16
optimize bytes and string payload extractors

robjtede:
## PR Type
Refactor

## PR Checklist

## Overview
Removes some boxes in the basic payload extractors.

### DIFF
--- a/src/types/payload.rs
+++ b/src/types/payload.rs
@@ -7,10 +7,15 @@ use std::task::{Context, Poll};
 use actix_http::error::{Error, ErrorBadRequest, PayloadError};
 use actix_http::HttpMessage;
 use bytes::{Bytes, BytesMut};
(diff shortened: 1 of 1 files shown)


src/types/payload.rs
```diff
@@ -205,25 +205,43 @@ impl FromRequest for String {
...
-            }
-            .boxed_local(),
-        )
+#[pin_project]
+pub struct StringExtractFut {
+    #[pin]
+    body_fut: HttpMessageBody,
+    encoding: &'static Encoding,
+}
+
```
fakeshadow: This could be working without pin_project.

fakeshadow (approved):
LGTM

src/types/payload.rs
robjtede: Good spot, I blame my auto-pilot.

PR #1339, opened 2020-02-04
export handler::Handler

sergeysova:
I want to implement a custom Factory to strictly define arguments for handlers

```rust
impl<F, B, T, R, O> Handler<F, T, R, O>
where
    F: BoundFactory<B, T, R, O>,
    R: Future<Output = O>,
    O: Responder,
{
    pub fn new(hnd: F) -> Self {
        Handler {
            hnd,
            _t: PhantomData,
        }
    }
}


pub trait BoundFactory<B, T, R, O>: Clone + 'static
where
    R: std::future::Future<Output = O>,
    O: actix_web::Responder,
{
    fn call(&self, bound: B, param: T) -> R;
}

impl<Func, B0, R, O> BoundFactory<(B0,), (), R, O> for Func
where
    Func: Fn(B0) -> R + Clone + 'static,
    R: std::future::Future<Output = O>,
    O: actix_web::Responder,
{
    fn call(&self, bound: (B0,), _: ()) -> R {
        (self)(bound.0)
    }
}
impl<Func, B0, A0, R, O> BoundFactory<(B0,), (A0,), R, O> for Func
where
    Func: Fn(B0, A0) -> R + Clone + 'static,
    R: std::future::Future<Output = O>,
    O: actix_web::Responder,
{
    fn call(&self, bound: (B0,), args: (A0,)) -> R {
        (self)(bound.0, args.0)
    }
}

impl<Func, B0, B1, R, O> BoundFactory<(B0, B1), (), R, O> for Func
where
    Func: Fn(B0, B1) -> R + Clone + 'static,
    R: std::future::Future<Output = O>,
    O: actix_web::Responder,
{
    fn call(&self, bound: (B0, B1), _: ()) -> R {
        (self)(bound.0, bound.1)
    }
}

impl<Func, B0, B1, A0, A1, R, O> BoundFactory<(B0, B1), (A0, A1), R, O> for Func
where
    Func: Fn(B0, B1, A0, A1) -> R + Clone + 'static,
    R: std::future::Future<Output = O>,
    O: actix_web::Responder,
{
    fn call(&self, bound: (B0, B1), args: (A0, A1)) -> R {
        (self)(bound.0, bound.1, args.0, args.1)
    }
}
```

PR #1843, opened 2020-12-20
Simplify handler

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

This is a follow PR of https://github.com/actix/actix-web/pull/1841.

`handler::Extract` was a wrapper Service around `handler::Handler`. It would extract types `Handler` want with `FromRequest` and move the extract types to `Handler` as `(T, HttpRequest)`.
This stops us from borrowing stuff from `HttpRequest` and use it in or as `T` because the move invalid the borrow.

This PR unify `Extract` and `Handler` as a single Service so `HttpRequest` and `T` are handled in the same context without move.(They are still moved in futures because of lack of async/await for now). 
Potentially enabling return references(Data<&T> for example) with `FromRequest` trait in the future.


robjtede:
https://github.com/actix/actix-web/pull/1339

Just remembered there's this long-standing PR that would conflict with this. I haven't looked into it much but seems as odds with this. The question is: Is having them be separate actually a useful abstraction for building on?

fakeshadow:
The only thing `Extract` does is call `FromRequest` and pass the result in a tuple with `HttpRequest` to `Handler`.
If we factor in aysnc/await and GAT it would be actually doing harm.

This would be what async/await looks like with this PR. 
This would enable use borrow stuff in HttpRequest , extract it with FromRequest and pass it to user's handler function. With the current abstration this is mostly impossible.
```rust
let (req, mut payload) = req.into_parts();
        let handle = self.hnd.clone();
        async move {
            // extract items from request.
            match T::from_request(&req, &mut payload).await {
                Ok(item) => {
                    // pass item to handler function and call Responder on the return type.
                    let res = handle.call(item).await.respond_to(&req).await;

                    // always return a ServiceResponse type and Error type is a placeholder type
                    // that never used.
                    match res {
                        Ok(res) => Ok(ServiceResponse::new(req, res)),
                        Err(e) => {
                            let res: Response = e.into().into();
                            Ok(ServiceResponse::new(req, res))
                        }
                    }
                }
                Err(e) => {
                    let req = ServiceRequest::new(req);
                    Ok(req.error_response(e))
                }
            }
        }
```

robjtede (approved):

PR #1852, opened 2020-12-25
Rename factory to handler

ibraheemdev:
## PR Type
Refactor

## PR Checklist
- [x] A changelog entry  has been made for the appropriate packages.

## Overview
This PR renames the `Factory` trait to `Handler`, and the `Handler` struct to `HandlerService`. This change better reflects the purpose of these types.

`The trait Factory is not satisfied` is a common error that has caused a lot of confusion as it is pretty cryptic. `The trait Handler is not satisfied` makes more sense. I also added documentation listing the requirements of a valid handler.

This is a breaking change (even users aren't really meant to implement `Factory` manually), but is important for ergonomics.

I also removed the third type parameter (`O`) from `Handler` which simplifies the type definition overall.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,8 +9,7 @@
 * added the actual parsing error to `test::read_body_json` [#1812]
 
 [#1812]: https://github.com/actix/actix-web/pull/1812
-
-
+[#1852]: https://github.com/actix/actix-web/pull/1852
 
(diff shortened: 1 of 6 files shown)


robjtede:
Add changelog entry.

> Rename `Handler` to `HandlerService` and rename `Factory` to `Handler`.

robjtede (approved):

PR #1813, opened 2020-12-05
Tokio 1.0

fakeshadow:
## PR Type

Update

## PR Checklist
Check your PR fulfills the following:

- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview

Update all cates in actix-web repo to tokio 1.0.

PR #1864, opened 2020-12-31
use ByteString as container for websocket text message

### LINKED ISSUE #1846: Sending shared text messages to Websocket clients
Currently Websocket message enum is defined as follows:

```rust
pub enum Message {
    Text(String),
    Binary(Bytes),
    Continuation(Item),
    Ping(Bytes),
    Pong(Bytes),
    Close(Option<CloseReason>),
    Nop,
}
```

The `Binary` type contains `Bytes` which essentially is a shared pointer to the actual data. But at the same time `Text` member contains an owned `String`, which makes it impossible to send the same `String` to multiple Websocket connections without actually copying string bytes.

## Expected Behavior
I would expect Text messages to be consistent with Binary messages and for it to be possible to make Websocket text messages with shared strings.

## Current Behavior
With text Websocket messages it is only possible to move a string with it's data.

## Possible Solution
A breaking change in the API, which changes String to an Arc<String> in

```rust
pub enum Message {
    ...
    Text(String),
    ...
}
```

## Context
Image a scenario where many Websocket clients are connected to a server, and a server occasionaly wants to send the same (potentially huge) text payload to all of the clients. A server would have to clone the same String for each client, which is a huge overhead.

## Your Environment

* rustc 1.50.0-nightly (d9a105fdd 2020-11-23)
* Actix Web Version: 3.3.2

robjtede:
## PR Type
Improvement

## PR Checklist

## Overview
Use `ByteString` as the contained format for websocket text messages so that the underlying memory can be re-used when sending to many clients; eliminating need for more expensive String cloning.

Blocked on #1813 for bytestring v1.0 upgrade.

Breaking change to public interface.

Closes #1846.

### DIFF
--- a/actix-web-actors/src/ws.rs
+++ b/actix-web-actors/src/ws.rs
@@ -1,9 +1,10 @@
-//! Websocket integration
-use std::collections::VecDeque;
+//! Websocket integration.
+
 use std::future::Future;
 use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
+use std::{collections::VecDeque, convert::TryFrom};
 
 use actix::dev::{
     AsyncContextParts, ContextFut, ContextParts, Envelope, Mailbox, StreamHandler,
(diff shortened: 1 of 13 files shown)


actix-web-actors/src/ws.rs
```diff
@@ -528,16 +530,15 @@ where
...
-                            .map_err(|e| {
-                                ProtocolError::Io(io::Error::new(
-                                    io::ErrorKind::Other,
-                                    format!("{}", e),
-                                ))
-                            })?
-                            .to_string(),
-                    ),
+                    Frame::Text(data) => {
+                        // TODO: remove the as_ref when bytes v1 is used in actix-http.
```
therealprof: I would really hope, the move to bytes v1 would happen first. 😅

fakeshadow:
LGTM. 
Btw well you are at it can we make Codec::new a const function? It would be useful if we make websocket extract type style where a const default config type would be ideal

fakeshadow (approved):

PR #1868, opened 2021-01-03
optimize write_camel_case in h1 encoder

robjtede:
## PR Type
Optimization

## PR Checklist

## Overview
New method reduces number of bounds checks when assigning to buffer.

2021-01-03_03 38 31-788dec03


fakeshadow (approved):
LGTM

actix-http/Cargo.toml
```diff
@@ -97,7 +97,7 @@ open-ssl = { version="0.10", package = "openssl" }
 rust-tls = { version="0.18", package = "rustls" }
 
 [[bench]]
-name = "content-length"
+name = "write-camel-case"
```
otavio, line 100: Why remove the old one instead of just adding the new bench?

robjtede: Since we switched to `itoa` crate it's not really a useful thing to keep track of now.

PR #1869, opened 2021-01-03
response header rework

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
This PR normalizes header methods used throughout the crates and enables header K/V pairs to use the same methods on `Response` as the typed headers (and give more attention to typed headers in the process).

In general, the signatures have changed from `insert_header(key: K, value: V)` to `insert_header(header: H)`, where H implements a new trait `IntoHeaderPair`.

The easiest migration is simply to use `insert_header((K, V))`, noting the tuple which implements the new trait for all previously accepted types.


fakeshadow (approved):
LGTM

actix-http/CHANGES.md
```diff
@@ -1,8 +1,26 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-* `Response::content_type` now takes an `impl IntoHeaderValue` to support `mime` types. [#1894]
+### Added
+* `IntoHeaderPair` trait that allows using types and non-typed headers in the same methods. [#1869]
```
jplatte:
```suggestion
* `IntoHeaderPair` trait that allows using typed and untyped headers in the same methods. [#1869]
```

jplatte (commented):
I don't really like `.insert_header((name, value))` but I guess it's better than having separate methods for typed and untyped headers and having to come up with distinct names for them. The only other solution I could think of is having the same method on two different traits and users importing the one they prefer (I doubt many people are using both typed and untyped headers in the same module).

Also, I see actix-http has its own typed headers; do these predate [the headers crate](https://docs.rs/headers/0.3.2/headers/) or is there some other reason they are separate?

robjtede:
@jplatte 

> I don't really like .insert_header((name, value)) but

In my view, an increased focus on typed headers alleviates this enough to make it worth doing. In most cases you're adding common headers to responses, if we don't have the header someone's after we can add it. Plus typed headers are theoretically cheaper than (string, V) since the HeaderName conversion is free. Do you agree that this trade off becomes more worth it when typed headers are used more liberally?

You've pointed out the primary issue with the trait approach; we already have folks getting confused when trying to use extensions_mut for example.

There's some foreign trait rules things to workaround to use an external crate. It's come up and we're looking into the viability of it. Either way this PR would be largely similar.

jplatte:
> In my view, an increased focus on typed headers alleviates this enough to make it worth doing. In most cases you're adding common headers to responses, if we don't have the header someone's after we can add it. Plus typed headers are theoretically cheaper than (string, V) since the HeaderName conversion is free. Do you agree that this trade off becomes more worth it when typed headers are used more liberally?

Yeah, I agree. Still it will be some annoying churn that has to be handled as part of transitioning to actix-web 4.0. Maybe you should consider having these new methods before the next major version and deprecate the old ones so upgrading can be done in independent steps. If it's too much work to backport these, maybe you could keep the old ones around as deprecated for the next major version.

robjtede:
I'll have a think about keeping old ones, deprecated, to make migration easier. Maybe just on ResponseBuilder where most churn would occur.

therealprof (approved):
So much more ergonomic. I like it.

robjtede:
I've added back the response builder methods `set_header` and `header` as deprecated.

omid:
I get the following error when I want to build a simple project based on a local instance of actix in the master branch:

```
error[E0432]: unresolved import `actix_http::http::header::IntoHeaderPair`
 --> ../actix-web/src/responder.rs:5:12
  |
5 |     http::{header::IntoHeaderPair, Error as HttpError, HeaderMap, StatusCode},
  |            ^^^^^^^^--------------
  |            |       |
  |            |       help: a similar name exists in the module: `IntoHeaderName`
  |            no `IntoHeaderPair` in `http::header`

error[E0432]: unresolved import `actix_http::http::header::IntoHeaderPair`
 --> ../actix-web/src/test.rs:9:45
  |
9 | use actix_http::http::header::{ContentType, IntoHeaderPair};
  |                                             ^^^^^^^^^^^^^^
  |                                             |
  |                                             no `IntoHeaderPair` in `http::header`
  |                                             help: a similar name exists in the module: `IntoHeaderName`
```

omid:
Sorry for this false alarm, it's fixed by adding to Cargo.toml:

```
[patch.crates-io]
actix-http = { path = "../actix-web/actix-http" }
awc = { path = "../actix-web/awc" }
```

PR #1870, opened 2021-01-03
optimize ServiceRequest methods

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Optimize `ServiceRequest::from_parts` by remove duplicate ref counter check.
Optimize `ServiceRequest::from_request` by ignore weak pointer check.
Add comment on `HttpRequest` for the non weak pionter nature of `Rc<HttpRequestInner>`


src/service.rs
```diff
@@ -73,11 +73,12 @@ impl ServiceRequest {
         mut req: HttpRequest,
         pl: Payload,
     ) -> Result<Self, (HttpRequest, Payload)> {
-        if Rc::strong_count(&req.0) == 1 && Rc::weak_count(&req.0) == 0 {
```
robjtede, line 76: keep the weak count as a debug assertion here and on from_request

src/request.rs
```diff
@@ -17,6 +17,9 @@ use crate::rmap::ResourceMap;
 #[derive(Clone)]
 /// An HTTP Request
 pub struct HttpRequest(pub(crate) Rc<HttpRequestInner>);
+// *. Rc<HttpRequestInner> is used exclusively and NO Weak<HttpRequestInner>
+// is allowed anywhere in the code. Weak pointer is purposely ignored when
+// doing Rc's ref counter check.
```
robjtede: make this a proper doc comment on HttpRequestInner

robjtede (changes requested):

src/service.rs
fakeshadow: `Rc::get_mut` only work on this condition. There is no need to double check.

fakeshadow: `Rc::get_mut` only work on this condition. `from_request` indeed worth a debug assert

src/request.rs
fakeshadow: HttpRequestInner type does not know the outer Rc wrapper. 
I feel the comment should be at HttpRequest level. Could be worth make HttpRequest a struct with named field for a clear view of doc.

robjtede: yep i like this too

robjtede (approved):

PR #1878, opened 2021-01-06
remove actix-threadpool.use actix_rt::task::spawn_blocking

fakeshadow:
## PR Type

Dependency

## PR Checklist

## Overview

Remove `actix-threadpool` from `actix-http` and `actix-web`.

Use `actix_rt::task::spawn_blocking` as the new blocking thread pool.

`actix_threadpool::BlockingError` has been moved into `actix_http::error` module and can also be import from `actix_web::error` module.

This is not a break change from `actix-web` and `actix-http` as the public API and types related stays the same. It's only when factoring people may use `actix-threadpool` together in their app and import types(BlockingError for example) from there and that would be breaking.


src/web.rs
```diff
@@ -280,5 +280,11 @@ where
     I: Send + 'static,
     E: Send + std::fmt::Debug + 'static,
 {
-    actix_threadpool::run(f).await
+    // map error to BlockingError. this is for not breaking the existing api.
+    // preferably this should just return actix_rt::task::JoinHandle for a more
+    // flexible control of when the task is awaited and/or the abort(if needed).
```
robjtede: we should make an issue to discuss this

actix-http/src/error.rs
```diff
@@ -304,6 +300,20 @@ impl From<httparse::Error> for ParseError {
...
+#[derive(Debug, Display)]
+/// A set of errors that can occur running blocking tasks in thread pool.
+pub enum BlockingError<E: fmt::Debug> {
+    #[display(fmt = "{:?}", _0)]
+    Error(E),
+    #[display(fmt = "Thread pool is gone")]
+    Canceled,
+}
+
+impl<E: fmt::Debug> std::error::Error for BlockingError<E> {}
```
robjtede, line 308: derive_more Error would delcare E the source

actix-http/src/error.rs
```diff
@@ -304,6 +300,20 @@ impl From<httparse::Error> for ParseError {
     }
 }
 
+#[derive(Debug, Display)]
+/// A set of errors that can occur running blocking tasks in thread pool.
```
robjtede: nit: docs then attributes

src/web.rs
fakeshadow: After a check from the tokio doc. It seems there is no canceling of blocking tasks. So I could be wrong on the abort of blocking task.

actix-http/src/error.rs
fakeshadow: This is copied from the actix-threadpool code and I believe it's explictly added by a PR.

src/web.rs
fakeshadow: I'm removing this comment and keep the code as it is for now. A unabortable task handle is not of much use.

fakeshadow:
I've migrate `actix-files` to use `spawn_blocking` making it runs on typed future with every chunk. It's a semver-norelease change on it

actix-http/src/error.rs
robjtede: okay can do it later if needed

Cargo.toml
```diff
@@ -133,6 +132,9 @@ actix-multipart = { path = "actix-multipart" }
 actix-files = { path = "actix-files" }
 awc = { path = "awc" }
 
+# TODO: remove override
+actix-rt = { git = "https://github.com/actix/actix-net.git" }
+
```
robjtede: 🗑️

actix-http/src/encoding/decoder.rs
```diff
@@ -3,14 +3,14 @@ use std::io::{self, Write};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use actix_threadpool::{run, CpuFuture};
+use actix_rt::task::{spawn_blocking, JoinHandle};
```
robjtede, line 6: import from actix-web

robjtede (approved):

PR #1879, opened 2021-01-06
More refactor of app_service

fakeshadow:
## PR Type

Refactor 

## PR Checklist

## Overview

Remove `AppRoutingFactoryResponse` type and use boxed async block for construct `AppRouting` service.

Make `AppRouting`'s default service non optional as `AppRoutingFactory` always pass a default service to it.

Remove the unreachable `Response::NotFound` in `AppRouting`'s service call due to the above change.

### DIFF
--- a/src/app_service.rs
+++ b/src/app_service.rs
@@ -1,6 +1,4 @@
 use std::cell::RefCell;
-use std::future::Future;
-use std::pin::Pin;
 use std::rc::Rc;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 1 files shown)


robjtede (approved):
join_all is known to be bad but given it's a one-time startup futures i'm not concerned

fakeshadow:
> join_all is known to be bad but given it's a one-time startup futures i'm not concerned

The original future though is typed. It's function just like a worse version of join_all where every new_service is wrapped in a box.

This new boxed async looks allocation more but in fact it's only two(or three if default service is included) no matter how many services the app factory hold(join_all only allocate one box). 

So in theory we could have a little faster routing init if App carries a lot of services.

robjtede (approved):

PR #1880, opened 2021-01-07
fix bug where upgrade future is not reset properly

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Fix two instances of bug where upgrade future is not reset properly and future could be polled after resolve.

Remove the indirection method `State::poll`.

Remove some import from `futures_util` and offload some to `futures_core`.


robjtede (commented):
LGTM

robjtede (approved):

PR #1881, opened 2021-01-07
prepare v4 beta releases

robjtede:


fakeshadow (approved):
LGTM

tyranron:
@robjtede hi, any ETAs for releasing stable 4.0 with the ecosystem bump? Any help for that?

Actix with its ecosystem is the only thing which makes us sit on tokio 0.2 still. Would love to move forward.

Thanks!

robjtede:
Here's the v4 milestone but this is by no means an exhaustive list. There are many TODOs in the codebase and open issues that will be added also.
https://github.com/actix/actix-web/milestone/6

At the moment I am concentrating on getting actix-rt stable so that some libs (eg. sqlx) can update.

I estimate 3 weeks until v4 is possible.

tyranron:
@robjtede thanks!

PR #1883, opened 2021-01-07
refactor Resource

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Rename `Resource` and `ResourceService` data field to app_data to keep consistent with app_service types.

Use `MethodNotAllowed` as default service of `Resource` that can be override by user defined default service.
This saves 1 or 2 heap allocation and simplify the logic. As when no user defined default service the resource would fall back to `MethodNotAllowed` anyway.

Simplify construct method of `ResourceService` with boxed async block. Improve code readability.

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -412,7 +412,7 @@ where
     fn into_factory(self) -> T {
         *self.factory_ref.borrow_mut() = Some(ResourceFactory {
             routes: self.routes,
-            data: self.app_data.map(Rc::new),
+            app_data: self.app_data.map(Rc::new),
             default: self.default,
         });
 
@@ -422,7 +422,7 @@ where
 
 pub struct ResourceFactory {
     routes: Vec<Route>,
-    data: Option<Rc<Extensions>>,
+    app_data: Option<Rc<Extensions>>,
     default: HttpNewService,
 }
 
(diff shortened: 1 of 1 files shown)


src/resource.rs
```diff
@@ -432,126 +432,60 @@ impl ServiceFactory<ServiceRequest> for ResourceFactory {
...
-                        done = false;
-                    }
-                },
-                CreateRouteServiceItem::Service(_) => continue,
-            };
-        }
+        Box::pin(async move {
+            let default = default_fut.await?;
+            let routes = factory_fut
+                .await
```
robjtede, line 450: potential join! here

src/resource.rs
```diff
@@ -423,7 +423,7 @@ where
 pub struct ResourceFactory {
     routes: Vec<Route>,
     data: Option<Rc<Extensions>>,
```
robjtede: app_data normalization

robjtede (changes requested):

src/resource.rs
fakeshadow: I don't think we need eager poll in any of the service factory futures. join acts like join_all it has no idea which variant is waked up and has to check both. 

Either way the startup is not an issue in most cases. I personally would prefer a homebrew join_all like the one used in actix-server to get rid of futures_util or just serialize every future.

robjtede: > Either way the startup is not an issue in most cases

fair enough

> to get rid of futures_util

removing futures_util in this repo is a non-goal

robjtede (approved):

src/resource.rs
fakeshadow: It construbites to near the amount of compile time as actix-web or awc and actually only used on unncessary places. It would be a good thing to remove it from actix-web and actix-http.(which is pretty easy too)

robjtede (approved):

PR #1887, opened 2021-01-08
actix-files: Fix If-(Un)Modified to not consider sub-seconds

rschoon:
## PR Type
Bug Fix

## PR Checklist

## Overview

Fix If-Modified-Since and If-Unmodified-Since to not compare using sub-second values.

If you have a filesystem with sub-second precision on modified time, If-Modified-Since will fail to work properly.  The browser will be providing a timestamp that is truncated (since http dates don't use microseconds so it gets lost in the conversion), but actix-files will compare using a timestamp that is less than a second later, since it is checking the timestamp untruncated.

This bug doesn't typically come up with default behavior because If-None-Match has precedence over If-Modified-Since.  I'm not sure when If-Unmodified-Since would be used in practice, but I updated it too.

I looked at changing HttpDate instead (since it shouldn't contain sub-second values anyway), but the underlying type doesn't have a good way to truncate it that I could see.

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2021-xx-xx
 
+* Fix If-Modified-Since and If-Unmodified-Since to not compare using sub-second timestamps.
 
 ## 0.6.0-beta.1 - 2021-01-07
 * `HttpRange::parse` now has its own error type.
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -1,8 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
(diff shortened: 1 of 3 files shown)


actix-files/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2021-xx-xx
 
+* Fix If-Modified-Since and If-Unmodified-Since to not compare using sub-second timestamps.
```
robjtede:
```suggestion
* Fix If-Modified-Since and If-Unmodified-Since to not compare using sub-second timestamps. [#1887]

[#1887]: https://github.com/actix/actix-web/pull/1887
```

robjtede (approved):

PR #1889, opened 2021-01-08
use a non leak pool for HttpRequestInner

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Use a non leak `HttpRequestPool` for storing recycled `Rc<HttpRequestInner>`.

In general a leakd pool is not a big issue but it's kinda mess lifetime wise and could confuse people when profiling the memory usage.

This PR use a pool respect the lifetime at the cost of one extra `RefCell::borror_mut`. It has a little bit overhead in theory but I doubt it's observable.

A detailed flamegraph analyze with 10gb data samples shows around 0.13% stack increase on drop HttpRequest object with hello world bench. So the regression should be around there.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@ use actix_service::{boxed, IntoServiceFactory, ServiceFactory};
 use crate::data::{Data, DataFactory};
 use crate::error::Error;
 use crate::guard::Guard;
+use crate::request::HttpRequestPool;
 use crate::resource::Resource;
 use crate::rmap::ResourceMap;
 use crate::route::Route;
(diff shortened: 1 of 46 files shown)


fakeshadow:
added a test for `HttpRequestPool`'s default capactiy and explicity drop to make sure there is not circular reference between it and `AppInitService`

src/config.rs
```diff
@@ -131,11 +132,17 @@ struct AppConfigInner {
...
 }
 
 impl AppConfig {
     pub(crate) fn new(secure: bool, addr: SocketAddr, host: String) -> Self {
-        AppConfig(Rc::new(AppConfigInner { secure, addr, host }))
+        AppConfig(Rc::new(AppConfigInner {
+            secure,
+            addr,
+            host,
+            pool: HttpRequestPool::default(),
```
robjtede: AppConfig is per connection. Pool belongs in AppInitService

fakeshadow: AppConfig is passed as HttpService's factory config. It's constructed only once on every worker thread.(Unless worker is faulted and get reconstruct on a newly spawned thread.)

fakeshadow: That said it would get constructed based on service which means bind to different port would result in duplicate app config on each thread. But this is true for the old pool too.

robjtede: you're right, i've traced it through and it is indeed one per worker

robjtede (approved):

PR #1891, opened 2021-01-08
use sync method on Responder trait

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Through out `actix-web` `actix-extras` and `examples` repo there isn't a single instance where `Responder` returns a truely future type. They are all types wrapped in future types for an illusion of async.

This adds a little bit overhead(almost non measurable though) and complicated `Responder` trait with two associated types and future wrappers to deal with the difference of various future types.

This PR address this by making the trait sync and returns only `Response<Body>` type. It makes the trait more simple and performant(in theory).

This is a breaking change for people using futures inside custom `Responder` trait but it would be fairly easy to move these async code into handler function where async code should be ran.

### DIFF
--- a/src/responder.rs
+++ b/src/responder.rs
@@ -1,8 +1,4 @@
 use std::convert::TryFrom;
-use std::future::Future;
-use std::marker::PhantomData;
-use std::pin::Pin;
-use std::task::{Context, Poll};
 
 use actix_http::error::InternalError;
 use actix_http::http::{
@@ -10,9 +6,6 @@ use actix_http::http::{
 };
 use actix_http::{Error, Response, ResponseBuilder};
 use bytes::{Bytes, BytesMut};
-use futures_util::future::{err, ok, Either as EitherFuture, Ready};
-use futures_util::ready;
-use pin_project::pin_project;
 
 use crate::request::HttpRequest;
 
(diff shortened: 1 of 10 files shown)


src/responder.rs
```diff
@@ -1,33 +1,20 @@
...
-    /// The associated error which can be returned.
-    type Error: Into<Error>;
-
-    /// The future response value.
-    type Future: Future<Output = Result<Response, Self::Error>>;
-
-    /// Convert itself to `AsyncResult` or `Error`.
-    fn respond_to(self, req: &HttpRequest) -> Self::Future;
+    /// Convert self to `Response`.
+    fn respond_to(self, req: &HttpRequest) -> Response;
```
robjtede, line 17: I think it would be better to have `-> Result<Response, impl Into<Error>>` so that pre-made responses from the `error` module can be used.

fakeshadow: After `respond_to` is called it's immediately matched by `HandlerService` and converted to `Response` anyway.
Return result here is kinda pointless by adding one additional branch and make the type larger by 8 bytes.
I believe a macro like this for error type would be a better alternative
```rust
macro_rules! err_respond {
    ($ty: ty) => (
        impl Responder for $ty {
            fn respond_to(self, _: &HttpRequest) -> Response {
                Response::from_error(self.into())
            }
        }
    )
}
```

robjtede: hard no on the macro approach tbh

fakeshadow: impl Responder for error type is mostly not useful anyway. as impl T would still be one type making a handler can return error and only error. It's part of the reason why not all crate error types impl Responder

robjtede (approved):

PR #1892, opened 2021-01-09
pretty printing internal server error

DominicWrege:
## PR Type

Update

## PR Checklist

## Overview

Internal server errors would be more readable

### DIFF
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -54,7 +54,7 @@ impl Response<Body> {
     pub fn from_error(error: Error) -> Response {
         let mut resp = error.as_response_error().error_response();
         if resp.head.status == StatusCode::INTERNAL_SERVER_ERROR {
-            error!("Internal Server Error: {:?}", error);
+            error!("Internal Server Error: {:#?}", error);
         }
         resp.error = Some(error);
         resp


otavio (commented):
I have mixed feelings about this. This will increase the reading of the error for sure, but I also think some people don't want this as it becomes harder to process as it becomes a multi-line error.

Maybe we could have a type to let the user choose if he/she wants the `PrettyPrintError` or not?

DominicWrege:
I do not see any benefits in an error which its unreadable. Currently the user has no option either but maybe we can find a good way between?!

otavio:
> I do not see any benefits in an error which its unreadable. Currently the user has no option either but maybe we can find a good way between?!

As said, it is more readable but I am unsure all users will prefer it as you are proposing. Maybe @robjtede or @fakeshadow can share their view on this?

DominicWrege:
Yeah good point. Maybe a compromise would be to leave it as it currently is and let the user override the default with a Formatter.

cetra3:
This is something that can be handled by your own function rather than embedded in the framework:

```rust
pub fn log_error<E: Into<Error>>(err: E) -> actix_web::Error {
    let err: Error = err.into();
    error!("{:?}", err);
    err.chain()
        .skip(1)
        .for_each(|cause| error!("because: {}", cause));
    ErrorInternalServerError(err)
}
```

I don't feel it's worthy of being changed in actix-web itself for that reason.

robjtede:
Consensus seems like a no on this one. Many logging systems assume one log per line and user-land solutions are easy enough.

PR #1893, opened 2021-01-09
Refactor/service request

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

`SerivceRequest` woluld take ownership of `Payload` type.
This save the runtime cost whenever call into_parts/from_parts/from_request on `ServiceRequest`. 
The return types would be simpler and smaller as no Error variant exists anymore.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -537,7 +537,7 @@ mod tests {
     use crate::{guard, http, web, App, HttpResponse};
     use actix_service::Service;
     use futures_util::future::ok;
-    
+
     #[actix_rt::test]
     async fn test_service() {
         let mut srv = init_service(
--- a/src/service.rs
+++ b/src/service.rs
@@ -70,8 +70,6 @@ impl ServiceRequest {
     }
 
     /// Construct request from parts.
-    ///
-    /// `ServiceRequest` can be re-constructed only if `req` hasn't been cloned.
     pub fn from_parts(req: HttpRequest, payload: Payload) -> Self {
         Self { req, payload }
     }
--- a/src/service.rs
+++ b/src/service.rs
@@ -537,28 +537,7 @@ mod tests {
     use crate::{guard, http, web, App, HttpResponse};
     use actix_service::Service;
     use futures_util::future::ok;
-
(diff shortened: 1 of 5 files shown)


src/service.rs
```diff
@@ -554,23 +540,23 @@ mod tests {
...
+        // assert!(ServiceRequest::from_parts(r, pl).is_err());
+
+        // let req = TestRequest::default().to_srv_request();
+        // let (r, _pl) = req.into_parts();
+        // assert!(ServiceRequest::from_request(r).is_ok());
+
+        // let req = TestRequest::default().to_srv_request();
+        // let (r, _pl) = req.into_parts();
+        // let _r2 = r.clone();
+        // assert!(ServiceRequest::from_request(r).is_err());
```
otavio: Why is this commented out?

otavio (changes requested):

src/service.rs
fakeshadow: With this change these calls would not fail. They lost the purpose of testing the failure case.

It's temporary disabled until the change proven to be working as intended.

We don't need rush to remove tests. It can be left there until everything settled down and if things goes south it can be revived very quickly.

otavio: I'd keep it enabled or drop it. We risk possibly merging commented code, and there is no need for this as if we want to revert this, Git is our good friend. 👍🏼

fakeshadow: Merging comment code is nothing wrong. There are a lot of commented code through out tests and actix repos. You can make a PR to clean them up if you like to.

robjtede: At least add a todo marker.

otavio: I think it is wrong to merge commented code as we have an SCM which allows the revert if need.

otavio (approved):

src/service.rs
robjtede: OTOH, a TODO marker is searchable for outstanding items without making an issue in all commits until it is sorted. I find them very valuable.

robjtede (approved):

PR #1899, opened 2021-01-10
simplify h1 dispatcher

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Reduce branches in hot path of `poll_flush`. `BytesMut::advance` internally would check on 0 len case.
`BytesMut::set_len(0)` only reachable when `written == len`.

Reomve a redendent `Option::is_none` check for `InnerDispatcher.io`. The only case when io field is none is when the `DispatcherState` enters `Upgrade` variant. Where it would never switch to `Normal` variant until it's resolved.

Simplify `poll_shutdown`.

Remove one time usage free function `read`.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -287,42 +287,38 @@ where
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Result<bool, DispatchError> {
-        if self.write_buf.is_empty() {
+        let len = self.write_buf.len();
+        if len == 0 {
             return Ok(false);
         }
 
-        let len = self.write_buf.len();
-        let mut written = 0;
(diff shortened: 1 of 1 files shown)


actix-http/src/h1/dispatcher.rs
```diff
@@ -287,42 +287,38 @@ where
...
                 Poll::Ready(Ok(n)) => {
-                    written += n;
+                    if n == 0 {
+                        return Err(DispatchError::Io(io::Error::new(
+                            io::ErrorKind::WriteZero,
+                            "",
+                        )));
+                    } else {
+                        written += n;
+                    }
```
robjtede: this is better as it was

robjtede (changes requested):
overall nice tweaks

robjtede (approved):

PR #1900, opened 2021-01-11
Refactor h1 encoder

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Use a closure to generic over both the cases of have extra headers or not.
This would save unnecessary stack calls on iterator of an empty hashmap when there is no extra headers.

### DIFF
--- a/actix-http/src/h1/encoder.rs
+++ b/actix-http/src/h1/encoder.rs
@@ -8,7 +8,7 @@ use bytes::{BufMut, BytesMut};
 
 use crate::body::BodySize;
 use crate::config::ServiceConfig;
-use crate::header::map;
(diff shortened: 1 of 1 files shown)


actix-http/src/h1/encoder.rs
```diff
@@ -273,6 +263,25 @@ pub(crate) trait MessageType: Sized {
 
         Ok(())
     }
+
+    #[inline]
```
robjtede: inlining happens inside crate boundaries by default (at compilers discretion), so either remove or `#[inline(always)]` if you think it's appropirate

robjtede (approved):

PR #1901, opened 2021-01-13
Refactor poll_keepalive for readability

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Use match on `poll_keepalive`. This removes extra if Let legs for double checking the timer.
Add comment to clearify the logic of every branch.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -287,42 +287,35 @@ where
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Result<bool, DispatchError> {
-        if self.write_buf.is_empty() {
+        let len = self.write_buf.len();
+        if len == 0 {
             return Ok(false);
         }
 
-        let len = self.write_buf.len();
-        let mut written = 0;
         let InnerDispatcherProj { io, write_buf, .. } = self.project();
         let mut io = Pin::new(io.as_mut().unwrap());
+
+        let mut written = 0;
         while written < len {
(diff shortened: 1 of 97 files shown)


actix-http/src/h1/dispatcher.rs
```diff
@@ -645,83 +645,82 @@ where
...
+        match this.ka_timer.as_mut().as_pin_mut() {
+            None => {
+                // conditionally go into shutdown timeout
+                if this.flags.contains(Flags::SHUTDOWN) {
+                    if let Some(deadline) = this.codec.config().client_disconnect_timer()
+                    {
+                        // write client disconnect time out and poll again to
+                        // go into Some<Pin<&mut Sleep>> branch
+                        this.ka_timer.set(Some(sleep_until(deadline)));
+                        return self.poll_keepalive(cx);
```
robjtede, line 660: Why recurse? We wouldnt expect the keepalive to proc instantly. Seems to be the only flow control difference.

robjtede (approved):

actix-http/src/h1/dispatcher.rs
fakeshadow: timer has to be polled once for registering waker. It's either a short recurse to make it happen or write this
```rust
this.ka_timer.as_mut().set(Some(sleep_until(deadline)));
let _ = this.ka_timer.as_mut().as_pin_mut().unwrap().poll(cx);
```

robjtede: true, and it's not on the hot path since processing is paused on socket idle, i agree with your choice

PR #1902, opened 2021-01-13
optimize ErrorHandler middleware

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Use static future for service call.

### DIFF
--- a/src/middleware/err_handlers.rs
+++ b/src/middleware/err_handlers.rs
@@ -152,26 +152,22 @@ where
     type Output = Fut::Output;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        loop {
-            match self.as_mut().project() {
-                ErrorHandlersProj::ServiceFuture { fut, handlers } => {
-                    let res = ready!(fut.poll(cx))?;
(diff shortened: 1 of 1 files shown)


src/middleware/err_handlers.rs
```diff
@@ -114,35 +119,67 @@ where
...
+                ErrorHandlersProj::ServiceFuture { fut, handlers } => {
+                    let res = ready!(fut.poll(cx))?;
+                    match handlers.get(&res.status()) {
+                        Some(handler) => match handler(res)? {
+                            ErrorHandlerResponse::Response(res) => {
+                                return Poll::Ready(Ok(res))
+                            }
+                            ErrorHandlerResponse::Future(fut) => self
+                                .as_mut()
+                                .set(ErrorHandlersFuture::HandlerFuture { fut }),
```
robjtede: from a quick skim seems like this is the only branch that breaks out if the loop and would be a bit clearer if it could be unwrapped from the loop and call self.poll() instead

fakeshadow: I feel loop is always clearer than recursive call but it could be just me.

robjtede (approved):

PR #1903, opened 2021-01-13
Remove HttpResponseBuilder::json2()

lberrymage:
## PR Type
Refactor

## PR Checklist

## Overview
It's not necessary to keep both json() and json2() around since the
former reduces the ownership of its parameter to a borrow only to pass
the reference to the latter. Users can instead borrow themselves when
passing an owned value: there doesn't need to be two separate functions.

This change also makes HttpResponseBuilder::json() take a value by
reference since it's cheaper to convert from an owned value to a
borrowed one rather than the other way around.

**This is a breaking change**

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -14,6 +14,8 @@
 * Renamed `IntoHeaderValue::{try_into => try_into_value}` to avoid ambiguity with std
   `TryInto` trait. [#1894]
 * `Extensions::insert` returns Option of replaced item. [#1904]
+* Remove `HttpResponseBuilder::json2()` and make `HttpResponseBuilder::json()` take a value by
+  reference. [#1903]
 
 ### Removed
 * `ResponseBuilder::set`; use `ResponseBuilder::insert_header`. [#1869]
@@ -24,6 +26,7 @@
 
 [#1869]: https://github.com/actix/actix-web/pull/1869
 [#1894]: https://github.com/actix/actix-web/pull/1894
+[#1903]: https://github.com/actix/actix-web/pull/1903
 [#1904]: https://github.com/actix/actix-web/pull/1904
 [#1912]: https://github.com/actix/actix-web/pull/1912
 
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -5,6 +5,7 @@ use std::{
     convert::TryInto,
     fmt,
     future::Future,
+    ops,
     pin::Pin,
     str,
     task::{Context, Poll},
@@ -651,19 +652,15 @@ impl ResponseBuilder {
         self.body(Body::from_message(BodyStream::new(stream)))
     }
 
-    #[inline]
     /// Set a json body and generate `Response`
     ///
     /// `ResponseBuilder` can not be used after this call.
-    pub fn json<T: Serialize>(&mut self, value: T) -> Response {
-        self.json2(&value)
-    }
-
-    /// Set a json body and generate `Response`
-    ///
-    /// `ResponseBuilder` can not be used after this call.
-    pub fn json2<T: Serialize>(&mut self, value: &T) -> Response {
-        match serde_json::to_string(value) {
+    pub fn json<T>(&mut self, value: T) -> Response
+    where
+        T: ops::Deref,
+        T::Target: Serialize,
+    {
+        match serde_json::to_string(&*value) {
             Ok(body) => {
                 let contains = if let Some(parts) = parts(&mut self.head, &self.err) {
                     parts.headers.contains_key(header::CONTENT_TYPE)
@@ -979,25 +976,7 @@ mod tests {
     fn test_json_ct() {
         let resp = Response::build(StatusCode::OK)
             .insert_header((CONTENT_TYPE, "text/json"))
-            .json(vec!["v1", "v2", "v3"]);
-        let ct = resp.headers().get(CONTENT_TYPE).unwrap();
-        assert_eq!(ct, HeaderValue::from_static("text/json"));
-        assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
-    }
-
-    #[test]
-    fn test_json2() {
-        let resp = Response::build(StatusCode::OK).json2(&vec!["v1", "v2", "v3"]);
-        let ct = resp.headers().get(CONTENT_TYPE).unwrap();
-        assert_eq!(ct, HeaderValue::from_static("application/json"));
-        assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
-    }
-
-    #[test]
-    fn test_json2_ct() {
-        let resp = Response::build(StatusCode::OK)
-            .insert_header((CONTENT_TYPE, "text/json"))
-            .json2(&vec!["v1", "v2", "v3"]);
+            .json(&vec!["v1", "v2", "v3"]);
         let ct = resp.headers().get(CONTENT_TYPE).unwrap();
         assert_eq!(ct, HeaderValue::from_static("text/json"));
         assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
--- a/src/test.rs
+++ b/src/test.rs
@@ -239,7 +239,7 @@ where
 ///             web::resource("/people")
 ///                 .route(web::post().to(|person: web::Json<Person>| async {
 ///                     HttpResponse::Ok()
-///                         .json(person.into_inner())})
+///                         .json(person)})
 ///                     ))
 ///     ).await;
 ///
@@ -299,7 +299,7 @@ where
 ///             web::resource("/people")
 ///                 .route(web::post().to(|person: web::Json<Person>| async {
 ///                     HttpResponse::Ok()
-///                         .json(person.into_inner())})
+///                         .json(person)})
 ///                     ))
 ///     ).await;
 ///
@@ -1121,9 +1121,7 @@ mod tests {
     #[actix_rt::test]
     async fn test_response_json() {
         let mut app = init_service(App::new().service(web::resource("/people").route(
-            web::post().to(|person: web::Json<Person>| {
-                HttpResponse::Ok().json(person.into_inner())
-            }),
+            web::post().to(|person: web::Json<Person>| HttpResponse::Ok().json(person)),
         )))
         .await;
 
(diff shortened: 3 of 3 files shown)


actix-http/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2021-xx-xx
 * `Response::content_type` now takes an `impl IntoHeaderValue` to support `mime` types. [#1894]
+* Remove `HttpResponseBuilder::json2()` and make `HttpResponseBuilder::json()` take a value by reference
```
robjtede:
```suggestion
* Remove `HttpResponseBuilder::json2()` and make `HttpResponseBuilder::json()` take a value by reference. [#1903]

[#1903]: https://github.com/actix/actix-web/pull/1903
```

actix-http/src/response.rs
```diff
@@ -639,18 +639,10 @@ impl ResponseBuilder {
         self.body(Body::from_message(BodyStream::new(stream)))
     }
 
-    #[inline]
```
robjtede, line 654: keep the inline annotation

robjtede (changes requested):

robjtede:
Looks like it's possible to accept both in the same method actually: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=44c2a332a117191974c302839d22d155

actix-http/src/response.rs
lberrymage: The reason I "removed" it is because it never existed for `json2()`: I just renamed `json2()` to `json()` and removed the original `json()` function, and since `#[inline]` was never on `json2()`, "keeping" `#[inline]` would be more like adding it.

If you want me to inline the new `json()` I can, but I want to clarify that I'm not really removing it.

actix-http/CHANGES.md
lberrymage: Done

actix-http/src/response.rs
robjtede: Sorry you're right I read the diff wrong.

lberrymage:
> Looks like it's possible to accept both in the same method actually: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=44c2a332a117191974c302839d22d155

Interesting. I still think it's better to expose a function which takes its argument by reference rather than by value if it only needs a reference though since it's potentially expensive for a user to convert a borrowed value to an owned one (if they only have access to a borrowed value).

robjtede:
I agree that only exposing one of the two should be the by-ref variant.

robjtede:
Not able to solve the merge conflicts but here's the best solution I came up with:

See that it directly supports owned, ref, mut ref, box, rc, web::Json directly as arguments.

```diff
diff --git a/actix-http/src/response.rs b/actix-http/src/response.rs
index 880f93c6..4c19e00e 100644
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -5,6 +5,7 @@ use std::{
     convert::TryInto,
     fmt,
     future::Future,
+    ops,
     pin::Pin,
     str,
     task::{Context, Poll},
@@ -651,19 +652,15 @@ impl ResponseBuilder {
         self.body(Body::from_message(BodyStream::new(stream)))
     }
 
-    #[inline]
-    /// Set a json body and generate `Response`
+    /// Set a JSON body and generate `Response`.
     ///
     /// `ResponseBuilder` can not be used after this call.
-    pub fn json<T: Serialize>(&mut self, value: T) -> Response {
-        self.json2(&value)
-    }
-
-    /// Set a json body and generate `Response`
-    ///
-    /// `ResponseBuilder` can not be used after this call.
-    pub fn json2<T: Serialize>(&mut self, value: &T) -> Response {
-        match serde_json::to_string(value) {
+    pub fn json<T>(&mut self, value: T) -> Response
+    where
+        T: ops::Deref,
+        T::Target: Serialize,
+    {
+        match serde_json::to_string(value.deref()) {
             Ok(body) => {
                 let contains = if let Some(parts) = parts(&mut self.head, &self.err) {
                     parts.headers.contains_key(header::CONTENT_TYPE)
@@ -861,6 +858,8 @@ impl From<BytesMut> for Response {
 
 #[cfg(test)]
 mod tests {
+    use std::{borrow::Cow, rc::Rc};
+
     use serde_json::json;
 
     use super::*;
@@ -969,35 +968,43 @@ mod tests {
 
     #[test]
     fn test_json() {
-        let resp = Response::build(StatusCode::OK).json(vec!["v1", "v2", "v3"]);
+        let resp = Response::Ok().json(vec!["v1", "v2", "v3"]);
         let ct = resp.headers().get(CONTENT_TYPE).unwrap();
         assert_eq!(ct, HeaderValue::from_static("application/json"));
         assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
-    }
 
-    #[test]
-    fn test_json_ct() {
-        let resp = Response::build(StatusCode::OK)
-            .insert_header((CONTENT_TYPE, "text/json"))
-            .json(vec!["v1", "v2", "v3"]);
+        let resp = Response::Ok().json(&vec!["v1", "v2", "v3"]);
         let ct = resp.headers().get(CONTENT_TYPE).unwrap();
-        assert_eq!(ct, HeaderValue::from_static("text/json"));
+        assert_eq!(ct, HeaderValue::from_static("application/json"));
         assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
-    }
 
-    #[test]
-    fn test_json2() {
-        let resp = Response::build(StatusCode::OK).json2(&vec!["v1", "v2", "v3"]);
+        let resp = Response::Ok().json(&mut vec!["v1", "v2", "v3"]);
+        let ct = resp.headers().get(CONTENT_TYPE).unwrap();
+        assert_eq!(ct, HeaderValue::from_static("application/json"));
+        assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
+
+        let resp = Response::Ok().json(Box::new(vec!["v1", "v2", "v3"]));
+        let ct = resp.headers().get(CONTENT_TYPE).unwrap();
+        assert_eq!(ct, HeaderValue::from_static("application/json"));
+        assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
+
+        let resp = Response::Ok().json(Rc::new(vec!["v1", "v2", "v3"]));
+        let ct = resp.headers().get(CONTENT_TYPE).unwrap();
+        assert_eq!(ct, HeaderValue::from_static("application/json"));
+        assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
+
+        let data: Cow<'_, Vec<&str>> = Cow::Owned(vec!["v1", "v2", "v3"]);
+        let resp = Response::Ok().json(data);
         let ct = resp.headers().get(CONTENT_TYPE).unwrap();
         assert_eq!(ct, HeaderValue::from_static("application/json"));
         assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
     }
 
     #[test]
-    fn test_json2_ct() {
+    fn test_json_custom_ct() {
         let resp = Response::build(StatusCode::OK)
             .insert_header((CONTENT_TYPE, "text/json"))
-            .json2(&vec!["v1", "v2", "v3"]);
+            .json(&vec!["v1", "v2", "v3"]);
         let ct = resp.headers().get(CONTENT_TYPE).unwrap();
         assert_eq!(ct, HeaderValue::from_static("text/json"));
         assert_eq!(resp.body().get_ref(), b"[\"v1\",\"v2\",\"v3\"]");
diff --git a/src/test.rs b/src/test.rs
index 55b43608..7c93c2c4 100644
--- a/src/test.rs
+++ b/src/test.rs
@@ -239,7 +239,7 @@ where
 ///             web::resource("/people")
 ///                 .route(web::post().to(|person: web::Json<Person>| async {
 ///                     HttpResponse::Ok()
-///                         .json(person.into_inner())})
+///                         .json(person)})
 ///                     ))
 ///     ).await;
 ///
@@ -298,9 +298,8 @@ where
 ///         App::new().service(
 ///             web::resource("/people")
 ///                 .route(web::post().to(|person: web::Json<Person>| async {
-///                     HttpResponse::Ok()
-///                         .json(person.into_inner())})
-///                     ))
+///                     HttpResponse::Ok().json(person)
+///                  })))
 ///     ).await;
 ///
 ///     let payload = r#"{"id":"12345","name":"User name"}"#.as_bytes();
@@ -1121,9 +1120,7 @@ mod tests {
     #[actix_rt::test]
     async fn test_response_json() {
         let mut app = init_service(App::new().service(web::resource("/people").route(
-            web::post().to(|person: web::Json<Person>| {
-                HttpResponse::Ok().json(person.into_inner())
-            }),
+            web::post().to(|person: web::Json<Person>| HttpResponse::Ok().json(person)),
         )))
         .await;
 
@@ -1142,9 +1139,7 @@ mod tests {
     #[actix_rt::test]
     async fn test_body_json() {
         let mut app = init_service(App::new().service(web::resource("/people").route(
-            web::post().to(|person: web::Json<Person>| {
-                HttpResponse::Ok().json(person.into_inner())
-            }),
+            web::post().to(|person: web::Json<Person>| HttpResponse::Ok().json(person)),
         )))
         .await;
 
@@ -1164,9 +1159,7 @@ mod tests {
     #[actix_rt::test]
     async fn test_request_response_form() {
         let mut app = init_service(App::new().service(web::resource("/people").route(
-            web::post().to(|person: web::Form<Person>| {
-                HttpResponse::Ok().json(person.into_inner())
-            }),
+            web::post().to(|person: web::Form<Person>| HttpResponse::Ok().json(person)),
         )))
         .await;
 
@@ -1190,9 +1183,7 @@ mod tests {
     #[actix_rt::test]
     async fn test_request_response_json() {
         let mut app = init_service(App::new().service(web::resource("/people").route(
-            web::post().to(|person: web::Json<Person>| {
-                HttpResponse::Ok().json(person.into_inner())
-            }),
+            web::post().to(|person: web::Json<Person>| HttpResponse::Ok().json(person)),
         )))
         .await;
```

lberrymage:
Rebased to fix the merge conflict. I'll get on utilizing `Deref` to support `web::Json` etc. shortly.

lberrymage:
Everything _should_ be dandy now provided checks pass.

robjtede (approved):
Great. Thank you 👍

robjtede:
Seems to need a rebase though. I'm not able to merge master in from here for some reason.

lberrymage:
That should do it.

frol:
This change broke my code which previously passed (1) a custom struct with `Serialize` implemented on it passed by value, (2) `serde_json::Value` passed also by value. Now I have to either wrap it with `actix_web::web::Json()` or pass by reference. Is it intensional?

image

The fix in my code is:

```diff
-Ok(HttpResponse::Ok().json(message))
+Ok(HttpResponse::Ok().json(&message))
```

lberrymage:
If you're asking whether this is intentionally a breaking change, yes.

PR #1907, opened 2021-01-15
Rename `HttpResponse` httpcode methods using snake_case

lberrymage:
## PR Type
Code Style

## PR Checklist

## Overview
It's Rust convention for method names to be in snake case, so much so that the compiler emits a warning by default for non-snake-case methods. Reading something like `HttpResponse::Ok()` leads me to believe that I'm creating an instance of the `Ok` variant of a `HttpResponse` enum, but for some reason there's no associated data (which _is_ possible, there's just no reason to define an enum that way. [Here's what I'm talking about](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a6aecea976b9b5f342cdf32583b0d2bc)).

Note that `HttpResponse::Continue()` was renamed to `HttpResponse::kontinue()` rather than `HttpResponse::r#continue()`. This is because naming the function `r#continue()` requires that _users_ call it as `r#continue()` rather than simply `continue()` as can be seen [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e08a1fecedc278ffc7a6d2aa2500529a). I assume `r#continue()` is a bit too ugly to be in the public API, but `kontinue` doesn't look fantastic either, so I'm open to suggestions in this area.

There are other cases of `#[allow(non_snake_case)]` in the actix codebase, but this commit only addresses the `HttpResponse::XXX()` methods.

**This is a breaking change**

### DIFF
--- a/actix-files/src/directory.rs
+++ b/actix-files/src/directory.rs
@@ -107,7 +107,7 @@ pub(crate) fn directory_listing(
     );
     Ok(ServiceResponse::new(
         req.clone(),
-        HttpResponse::Ok()
+        HttpResponse::ok()
             .content_type("text/html; charset=utf-8")
             .body(html),
     ))
--- a/actix-files/src/lib.rs
+++ b/actix-files/src/lib.rs
@@ -674,7 +674,7 @@ mod tests {
     async fn test_default_handler_file_missing() {
         let mut st = Files::new("/", ".")
             .default_handler(|req: ServiceRequest| {
-                ok(req.into_response(HttpResponse::Ok().body("default content")))
+                ok(req.into_response(HttpResponse::ok().body("default content")))
             })
             .new_service(())
             .await
--- a/actix-files/src/service.rs
+++ b/actix-files/src/service.rs
@@ -77,7 +77,7 @@ impl Service<ServiceRequest> for FilesService {
 
         if !is_method_valid {
             return Either::Left(ok(req.into_response(
-                actix_web::HttpResponse::MethodNotAllowed()
+                actix_web::HttpResponse::method_not_allowed()
                     .insert_header(header::ContentType(mime::TEXT_PLAIN_UTF_8))
                     .body("Request did not meet this resource's requirements."),
             )));
@@ -101,7 +101,7 @@ impl Service<ServiceRequest> for FilesService {
                     let redirect_to = format!("{}/", req.path());
 
                     return Either::Left(ok(req.into_response(
-                        HttpResponse::Found()
+                        HttpResponse::found()
                             .insert_header((header::LOCATION, redirect_to))
                             .body("")
                             .into_body(),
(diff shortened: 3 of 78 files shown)


realaravinth:
It's like you said, `kontinue` is a little weird. 

Following convention is important as it brings a feeling of familiarity but I think we should also consider what the other frameworks are doing when going ahead with this change. In my opinion, familiarity within domains is more important.

All rust frameworks are re-exporting `StatusCode` from the `http` crate with the help of some hand-written macro. The only difference between `actix` and the rest is that `actix`'s macro creates functions that returns `ResponseBuilder` wrapped `StatusCodes`. Maybe we can tweak [this macro](https://github.com/actix/actix-web/blob/1c95fc26547043376a665cbd5f2a35aebc6297e0/actix-http/src/httpcodes.rs#L9) produce structs rather than functions?

robjtede:
Maybe I'm just too used to the methods being this way but this change just feels like unnecessary churn for no real gain.

Yes, conventions exist to help beginners find their way more quickly but I can't honestly say I know any beginners who've had a problem with it. `r#continue` and `kontinue` are both pretty bad, non-obvious APIs, even it is a lesser used status code. The enum-ish look to the methods is helpful because that's almost what it is.

lberrymage:
> All rust frameworks are re-exporting StatusCode from the http crate with the help of some hand-written macro. The only difference between actix and the rest is that actix's macro creates functions that returns ResponseBuilder wrapped StatusCodes. Maybe we can tweak this macro produce structs rather than functions?

I've looked at doing this... I don't think it's possible because `ResponseBuilder` includes a `BoxedResponseHead` (which contains `Box`), so it can't be evaluated at compile time.

> Maybe I'm just too used to the methods being this way but this change just feels like unnecessary churn for no real gain.
>
> Yes, conventions exist to help beginners find their way more quickly but I can't honestly say I know any beginners who've had a problem with it. r#continue and kontinue are both pretty bad, non-obvious APIs, even it is a lesser used status code. The enum-ish look to the methods is helpful because that's almost what it is.

That's a fair argument. And maybe I should have mentioned this in my original comment, but I meant this PR as more of a "Can we make this better? Here's what I have so far" rather than "We should make this change because it is strictly superior to the current state of things." I agree that `r#continue` and `kontinue` are pretty bad APIs, but if there was a way to change them to be good-looking and obvious (e.g. `const` structs if they would work), I would personally would prefer using them over camel case functions. But like you said, the enum-ish look to the methods is almost helpful, so perhaps most others wouldn't.

I understand that this isn't (and shouldn't be) a priority and might not even be considered, so feel free to close this PR if you like. But if you think it's worth trying to find a better solution (again, I understand if you don't), I'm open to working on that.

robjtede:
Always good to hear views on improvements. I think in this case the consensus is to keep it the same.

I'd encourage future readers to comment if they feel strongly about this and we can consider it for the next set of breaking releases.

PR #1911, opened 2021-01-17
optimize Route guards

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Add `Guard::clone_guard` method for cloning a `Guard` trait object. 
Internal types that impl `Guard` trait are all constrant to `Clone`. This is a break change for `FnGuard` type.
In return `Route`'s guards are not wrapped in `Rc` therefore reduce indirection when checking guard.


robjtede:
really not a fan of the extra method, i don't think it's valuable enough for dirtying the API

PR #1914, opened 2021-01-18
optimize actix-http messages

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `Clone` impl from `Message<T>` type. This would eliminate all unintentional clone of said type.

Remove strong ref count check when dropping `Message<T>` as it's not clonable anymore and it's only the `T `type that are exposed to public so there is no possible of cloning `Rc<T>`. Therefore strong ref count should always be one when dropping.


robjtede:
why keep a panicking clone at all?

fakeshadow:
`Message` is a export public type and I guess future maintainers may find re add Clone impl useful.
Maybe it would be more proper to use a mark to mark it as non Clone.

robjtede (approved):

PR #1929, opened 2021-01-25
fix memory usage for h1 and read bug on buffer size.

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

This is an atempt to fix the memory issue related to h1 dispatcher.

Reduce the size of read/write buffer to 1024 low and 8192 high in bytes.
Add max hw buffer size matching the current behavior of actix-web.
http parse on buffer exceed max hw buffer would return as a parse error.(Result in 400 error to client).
Only check buffer length on successful read.

### DIFF
--- a/actix-http/src/h1/decoder.rs
+++ b/actix-http/src/h1/decoder.rs
@@ -208,8 +208,9 @@ impl MessageType for Request {
                         trace!("MAX_BUFFER_SIZE unprocessed data reached, closing");
                         Err(ParseError::TooLarge)
                     } else {
+                        // Return None to notify more read are needed for parsing request
                         Ok(None)
-                    }
+                    };
                 }
             }
         };
@@ -277,14 +278,16 @@ impl MessageType for ResponseHead {
 
                     (len, version, status, res.headers.len())
                 }
-                httparse::Status::Partial => return {
-                    if src.len() >= MAX_BUFFER_SIZE {
-                        error!("MAX_BUFFER_SIZE unprocessed data reached, closing");
-                        Err(ParseError::TooLarge)
-                    } else {
-                        Ok(None)
+                httparse::Status::Partial => {
+                    return {
+                        if src.len() >= MAX_BUFFER_SIZE {
+                            error!("MAX_BUFFER_SIZE unprocessed data reached, closing");
+                            Err(ParseError::TooLarge)
+                        } else {
+                            Ok(None)
+                        }
                     }
-                },
+                }
             }
         };
 
--- a/actix-http/src/h1/decoder.rs
+++ b/actix-http/src/h1/decoder.rs
@@ -279,13 +279,11 @@ impl MessageType for ResponseHead {
                     (len, version, status, res.headers.len())
                 }
                 httparse::Status::Partial => {
-                    return {
-                        if src.len() >= MAX_BUFFER_SIZE {
-                            error!("MAX_BUFFER_SIZE unprocessed data reached, closing");
-                            Err(ParseError::TooLarge)
-                        } else {
-                            Ok(None)
-                        }
+                    return if src.len() >= MAX_BUFFER_SIZE {
+                        error!("MAX_BUFFER_SIZE unprocessed data reached, closing");
+                        Err(ParseError::TooLarge)
+                    } else {
+                        Ok(None)
                     }
                 }
             }
--- a/actix-http/src/h1/decoder.rs
+++ b/actix-http/src/h1/decoder.rs
@@ -203,7 +203,13 @@ impl MessageType for Request {
 
                     (len, method, uri, version, req.headers.len())
                 }
-                httparse::Status::Partial => return Ok(None),
+                httparse::Status::Partial => {
+                    return if src.len() >= super::dispatcher::MAX_HW_BUFFER_SIZE {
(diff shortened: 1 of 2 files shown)


robjtede:
At the least, reducing the HW amount is a breaking change since header blocks also use this buffer and cannot parse correctly if it is full. That's to say, now requests with headers more than 8kB will not parse. Haven't worked through the other changes yet but this, at least, should be reverted.

fakeshadow:
let's see if it actually reduce memory usage. It's not final and only changed to better understand the problem.

actix-http/src/h1/dispatcher.rs
```diff
@@ -913,6 +908,12 @@ where
                 if n == 0 {
                     return Ok(Some(true));
                 } else {
+                    // If buf is full return but do not disconnect since
+                    // there is more reading to be done
+                    if buf.len() >= HW_BUFFER_SIZE {
+                        return Ok(Some(false));
```
therealprof, line 928: Shouldn't this return/signal an `Error` instead?  This should ideally somehow turn into a "413 Entity Too Large error" sent to remote.

therealprof:
Using a 32Kb limit is a bit weird, most webservers use <= 8Kb as default, some allow configuration of higher values.

actix-http/src/h1/dispatcher.rs
fakeshadow: Yep could be.
 
But I'm not unstanding enough about the entire logic of dispatcher and this code is added by a specific PR. So I intend to leave it be from this PR in case I make mistake on this and mess things up.

fakeshadow:
> some allow configuration of higher values.

A user configuration sounds good. It's weird but like rob pointed out people using actix-web would expect that size to work. An additional configuration to lower it would be a good alternative.

fakeshadow:
revert to use 128kb as max buffer but the starting hw buffer size would be 8kb.
http parse on decode request would be able to catch buffer overflow the max buffer size and return error.

This change would have some overhead for more buffer reserve when the payload is large. (8kb per read/write iteration instead of 32kb) I believe it can be overcome with some change to `poll_read_buf` in the future. But for now I believe it's a good balance between save people memory on small buffer and keep the ability for a 128k read max buffer size.

actix-http/src/h1/dispatcher.rs
fakeshadow: With latest change after this return decoder would give it a try on parse and if it's still a fail a TooBig error is returned. It's a 400 error though which I consider is a change beyond this PR

actix-http/src/h1/decoder.rs
```diff
@@ -203,7 +203,14 @@ impl MessageType for Request {
...
                     (len, method, uri, version, req.headers.len())
                 }
-                httparse::Status::Partial => return Ok(None),
+                httparse::Status::Partial => {
+                    return if src.len() >= MAX_BUFFER_SIZE {
+                        trace!("MAX_BUFFER_SIZE unprocessed data reached, closing");
+                        Err(ParseError::TooLarge)
+                    } else {
+                        Ok(None)
+                    }
```
therealprof: Can't the `if` be moved directly into the match arm?

fakeshadow: you can but you need another match on the condition which would lead to another variant and return. So you end up more code doing the same thing with the same logic.

therealprof: Huh? I find 
```suggestion
                    httparse::Status::Partial if src.len() >= MAX_BUFFER_SIZE => {
                       error!("MAX_BUFFER_SIZE unprocessed data reached, closing");
                       return Err(ParseError::TooLarge);
                    }
                    httparse::Status::Partial => return Ok(None),
```

much clearer and simpler.

therealprof: It's also rather weird to use explicit `return` here while a in another arm `?` is used. I'd probably just create the `Result` and have one `?` at the end of the `match` block.

therealprof: ... and probably add a note that `httparse::Status::Partial => return Ok(None)` short-circuits the `decode` function in case we don't have enough data yet. Easy to miss... 😅

fakeshadow: It really is a style choice. Your write more code for fewer lines and I don't find it's simple or clear as it's the same thing.

therealprof: Having all conditions directly on match arm allows to mentally filter all possible cases on the same level, also it allows to group match arms with the same outcome which is nice. But okay, I take your point about style choice...

robjtede: Since the condition has nothing to do with the variant matched I feel that its appropriate to use an inner if statement.

fakeshadow:
Fix a possible stream hang when reading.
The early return force on reader buffer failed to check if the io is scheduled to be wake up or not.
A force self wake up is added to make sure dispatcher can wake itself up no matter the stream io's state.

actix-http/src/h1/dispatcher.rs
fakeshadow: A proper 413 error is returned with latest change. I would mark this as resolved.

robjtede (approved):

PR #1931, opened 2021-01-26
add ClientResponse::timeout

### LINKED ISSUE #1572: Inability to configure full request timeout leads to DoS / resource leak in HTTP client
awc currently does not allow specifying a timeout for the entire request (i.e. until the request body is finished transferring), which means an established connection will hang indefinitely if the remote host doesn't stops replying.

While it's not necessary to set such timeout by default (most other HTTP clients don't), it must be possible to configure it for awc to be usable for e.g. building reliable services that query a remiote API over HTTP, or for defending against DoS attacks where the remote host sends 1 byte per second, keeping the connection open indefinitely.

Consider an image hosting service like imgur.com: users enter URLs of images they want to host, and the HTTP client fetches the images and puts them into storage. If a user enters a URL to a malicious host that replies more often than the read timeout, but keeps replying indefinitely, this means that the connection will be open indefinitely. If an attacker keeps getting the service to open such connections, this will eventually lead to the exhaustion of the thread pool or network ports on the client.

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Add conditional timeout setting on `ClientResponse`.

Closes https://github.com/actix/actix-web/issues/1572


awc/src/response.rs
```diff
@@ -21,6 +24,43 @@ use crate::error::JsonPayloadError;
 pub struct ClientResponse<S = PayloadStream> {
     pub(crate) head: ResponseHead,
     pub(crate) payload: Payload<S>,
+    pub(crate) timeout: ResponseTimeout,
+}
+
+// a helper enum for response timeout for reusing the boxed sleep.
+// It's pass from `SendClientRequest`.
```
robjtede: despite it being private, please make all possible struct/enum/fn comments into doc comments so they show up on symbol hover hints

fakeshadow: removed. It's just a hint to people really want to looking at the code why a boxed timer is passed around. I guess it's easy enough to understand so I removed them.

robjtede: I would say (having not looked at the code yet) it's not particularly obvious why the Disbled variant contains a timer too. I can guess why but maybe some can't.

robjtede (approved):
👍🏻

awc/tests/test_client.rs
```diff
@@ -157,6 +157,79 @@ async fn test_timeout_override() {
...
+    });
+
+    let client = awc::Client::new();
+
+    let res = client
+        .get(srv.url("/"))
+        .send()
+        .await
+        .unwrap()
+        .timeout(Duration::from_millis(300))
```
robjtede: apparently this isn't long enough for ci :(

PR #1934, opened 2021-01-26
Add "name" attribute to route macro

stdrc:
## PR Type

Feature

## PR Checklist

## Overview

This PR adds a new `name` attribute to the `route` macro, so that user can specify a resource name, which by default is the name of handler function since #1178, and call `HttpRequest.url_for` on the name later.

Example:

```rust
#[get("/static", name = "static")]
async fn static_file() -> impl Responder {
    // ...
}
```

### DIFF
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -78,6 +78,7 @@ impl TryFrom<&syn::LitStr> for MethodType {
 
 struct Args {
     path: syn::LitStr,
+    resource_name: Option<syn::LitStr>,
     guards: Vec<Ident>,
     wrappers: Vec<syn::Type>,
     methods: HashSet<MethodType>,
@@ -86,6 +87,7 @@ struct Args {
 impl Args {
     fn new(args: AttributeArgs, method: Option<MethodType>) -> syn::Result<Self> {
         let mut path = None;
+        let mut resource_name = None;
         let mut guards = Vec::new();
         let mut wrappers = Vec::new();
         let mut methods = HashSet::new();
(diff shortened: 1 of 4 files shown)


robjtede:
Why do you feel using the handler name is not sufficient? Doesn using raw identifiers `r#static` not work?

stdrc:
Thanks for reminding me the `r#static` solution, but I think this PR is still usefull if we want some special characters in the resource name. BTW, it's just a counterpart to Flask's [`endpoint` parameter of `route` decorator](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.route).

actix-web-codegen/src/route.rs
```diff
@@ -260,13 +272,18 @@ impl ToTokens for Route {
...
                     methods,
                 },
             resource_type,
         } = self;
-        let resource_name = name.to_string();
+        let resource_name = if let Some(resource_name) = resource_name {
+            resource_name.value()
+        } else {
+            name.to_string()
+        };
```
otavio: Could it be a `unwrap_or_else`?

stdrc: ok, changed

robjtede:
would be okay to move ahead with this if you can get it rebased @richardchien

stdrc:
> would be okay to move ahead with this if you can get it rebased @richardchien

Ok, I've rebased

therealprof:
@richardchien Needs conflict resolution now. ;)

stdrc:
@therealprof Ok, resolved

robjtede:
I can usually rebase master branch into PRs in order to coordinate the merge but you must have that option disabled. I'm sorry but I'm going to have to ask you to rebase again because another PR just got merged that fixed CI.

stdrc:
@robjtede Rebased again:)

robjtede:
@richardchien thanks, i shall merge this one next

robjtede (approved):

robjtede:
😅 there we go... thanks for you contribution @richardchien

PR #1935, opened 2021-01-27
Fix bug of client_unread_response test

stdrc:
## PR Type

Bug Fix

## PR Checklist

- [x] (No change) Documentation comments have been added / updated.
- [x] (No change) A changelog entry has been made for the appropriate packages.

## Overview

Currently, [unittest `client_unread_response` of `awc` package](https://github.com/actix/actix-web/blob/c201c15f8c19d0056f27b8a7e1fc675bbdfca9e8/awc/tests/test_client.rs#L722) may break due to timing issue, this can be observed [here](https://github.com/actix/actix-web/runs/1770087396?check_suite_focus=true#step:7:1096) and [here](https://github.com/actix/actix-web/pull/1929/checks?check_run_id=1774185124#step:9:43). This PR fixes the bug by ensuring that the client won't send request until the server is ready.

### DIFF
--- a/awc/tests/test_client.rs
+++ b/awc/tests/test_client.rs
@@ -722,8 +722,12 @@ async fn test_client_cookie_handling() {
 async fn client_unread_response() {
     let addr = test::unused_addr();
 
+    let ready = Arc::new(std::sync::atomic::AtomicBool::new(false));
+    let ready_clone = ready.clone();
+
     std::thread::spawn(move || {
         let lst = std::net::TcpListener::bind(addr).unwrap();
+        ready_clone.store(true, Ordering::Relaxed);
 
         for stream in lst.incoming() {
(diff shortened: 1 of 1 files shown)


stdrc:
Seems that #1931 fixes this bug too, by moving the `bind(addr)` out of new thread, which may be a better approach.

fakeshadow:
yea it's already fixed in that pr and it's just not merged yet. 

In general you don't want to spin on an atomic for anything. It's a very bad pattern. 
You can use channel or other sync prmitive like condvar and mutex to properly block a thread and wait something happen.

stdrc:
> In general you don't want to spin on an atomic for anything. It's a very bad pattern.
> You can use channel or other sync prmitive like condvar and mutex to properly block a thread and wait something happen.

Thanks for advice.

PR #1834, opened 2020-12-17
Remove boxed future from HttpMessage

fakeshadow:
## PR Type

Refactor 

## PR Checklist
Check your PR fulfills the following:

- [x] Documentation comments have been added / updated.
- [x] Format code with the latest stable rustfmt

## Overview

Remove boxed future from HttpMessage. It's now static dispatched.

PR #1938, opened 2021-01-27
fix limit not working on HttpMessageBody::limit

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Fix the bug introduced by https://github.com/actix/actix-web/pull/1834.
When content length overflow the default limit overflow error is generated. The error is not removed when calling limit and set a new limit that is bigger than content length


robjtede:
There wasn't a test to catch this?

fakeshadow:
The old code would check limit against content-length everytime HttpMessageBody is polled as a future. It's a bit overhead but that made sure no tests are needed for the logic of catching overflow error.

I feel API for various `XXXBody::new().limit()` style is overhead and complex in general. An extra match and mutation is needed in limit method.

PR #1939, opened 2021-01-27
Improve `impl ResponseError` documentation

### LINKED ISSUE #1886: Documentation for `impl ResponseError for io::Error` doesn't match behaviour
The [implementation](https://github.com/actix/actix-web/blob/188ee44f819670944c0af92730f863e5683e7045/actix-http/src/error.rs#L207) of `ResponseError` for `io::Error` is documented to return an 500 Internal Server Error, but in actuality it can also return 404 Not Found (if the file doesn't exist) and 403 Forbidden (if the program does not have permission to access the file). I'm not sure which is the intended behaviour (although I would much prefer the documented behaviour over the actual behaviour).

alexander-jackson:
## PR Type
Other - Documentation

## PR Checklist

## Overview
Improve the detail of documentation for different implementations of `ResponseError` on other errors. Most of these return the basic `StatusCode::INTERNAL_SERVER_ERROR` but some depend on the contents of the error, so these are explained as well.

No breaking changes.

Closes #1886.

### DIFF
--- a/actix-http/src/error.rs
+++ b/actix-http/src/error.rs
@@ -147,7 +147,10 @@ impl From<ResponseBuilder> for Error {
     }
 }
 
-/// Return `GATEWAY_TIMEOUT` for `TimeoutError`
+/// Inspects the underlying enum and returns an appropriate status code.
+///
+/// If the variant is [`TimeoutError::Service`], the error code of the service is returned.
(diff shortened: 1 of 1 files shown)


dunnock (approved):
LGTM, docs improvements

PR #1953, opened 2021-02-04
Use askama-escape for html escaping

Gelbpunkt:
## PR Type

Refactor/Other

## PR Checklist

## Overview

This replaces [v_htmlescape](https://docs.rs/v_htmlescape/0.12.0/v_htmlescape/) in actix-files with [askama_escape](https://docs.rs/askama_escape/0.10.1/askama_escape/), which has zero dependencies instead of 3 like v_htmlescape and does not pull in outdated `bytes` versions. 

I was not able to run tests or even build it because actix-server seems to be broken right now, but both the old and new method take a `&str` and the return types are only used for Display , which it still implements (see [here](https://github.com/djc/askama/blob/main/askama_escape/src/lib.rs#L179-L184)).

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -2,8 +2,10 @@
 
 ## Unreleased - 2021-xx-xx
 * Fix If-Modified-Since and If-Unmodified-Since to not compare using sub-second timestamps. [#1887]
+* Replace `v_htmlescape` with `askama_escape`. [#1953]
 
 [#1887]: https://github.com/actix/actix-web/pull/1887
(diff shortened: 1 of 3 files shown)


robjtede:
runtime perf differnce is negligible for the short inputs we use escaping for

compile time give askama_escape the edge, even ignoring duplicated deps with the rest of actix:

| v_escape | askama |
| - | - |
| 2021-02-07_04 42 21-15c034d5 | 2021-02-07_04 42 28-31fc4177 |

robjtede (approved):

PR #1957, opened 2021-02-06
fix master branch build. change web::block output type.

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Change `web::block` to output any type the closure gives in Ok variant.(Send + 'static bound).

Simplify `BlockingError` type to only triggered by blocking threadpool dead.

Reduce size of `actix_files::ChunkReadFile` by 8 bytes.

Temoprary fix git build by pin dep version.

### DIFF
--- a/actix-files/src/chunked.rs
+++ b/actix-files/src/chunked.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use actix_web::{
-    error::{Error, ErrorInternalServerError},
+    error::{BlockingError, Error},
     rt::task::{spawn_blocking, JoinHandle},
 };
 use bytes::Bytes;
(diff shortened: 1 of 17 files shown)


robjtede (approved):

actix-http/src/encoding/decoder.rs
```diff
@@ -79,15 +79,8 @@ where
...
-                    Ok(Ok(item)) => item,
-                    Ok(Err(e)) => {
-                        return Poll::Ready(Some(Err(BlockingError::Error(e).into())))
-                    }
-                    Err(_) => {
-                        return Poll::Ready(Some(Err(BlockingError::Canceled.into())))
-                    }
-                };
+                let (chunk, decoder) =
+                    ready!(Pin::new(fut).poll(cx)).map_err(|_| BlockingError)??;
```
dunnock, line 83: That seems will clear error, which might be fs error too.. odds are low but might complicate debugging I guess

robjtede: the fs error will be propagated by the second ?

fakeshadow: `std::io::Error` would be converted to `PayloadError::Incomplete` variant.
The old behavior is to wrap `io::Error` in `BlockingError::Error` and then extract it as `PayloadError::Io` variant.
It's kinda inconsistent as all other io::Error are treated as Incomplete variant.

dunnock: Yeah I got confused, sorry

actix-files/src/chunked.rs
```diff
@@ -38,49 +53,52 @@ impl Stream for ChunkedReadFile {
...
+                        Ok((file, Bytes::from(buf)))
+                    });
+                    this.state = ChunkedReadFileState::Future(fut);
+                    self.poll_next(cx)
                 }
+            }
+            ChunkedReadFileState::Future(ref mut fut) => {
+                let (file, bytes) =
+                    ready!(Pin::new(fut).poll(cx)).map_err(|_| BlockingError)??;
+                this.state = ChunkedReadFileState::File(Some(file));
```
dunnock, line 95: I wonder would it decrease number of polls spawning next chunk future right here?

dunnock (approved):
LGTM, left a question if we can skip some polls in ChunkedFile Stream impl

actix-files/src/chunked.rs
fakeshadow: From the structure of this future. The caller is responsible to actively poll an extra round when he get Some(Bytes) from the stream to go into the file branch again. Otherwise this stream would be stuck.
And I believe downstream API would be doing this. This is an internal stream chain that I think it's better to leave this way. Using Context::waker::wake_by_ref here certainlly is a good way to self wakeup but I'm afraid it would have a liitle bit more overhead than the downstream caller handling it(or result in more polling than needed as it does not aware this change)

dunnock: I see, makes sense in that case

PR #1962, opened 2021-02-07
reduce branch.remove deadcode for h1 dispatcher

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove dead code of reserve `write_buf` in dispatcher's poll. It's not used for the following reason:
`write_buf` is only populated by `h1::codec::Codec::encode` and it's done with `Message` enum.
`Message::Item` variant would call `MessageType::encode_status/encode_headers` methods where thye reserve enough space for buffer.
`Message::Chunk` variant would call `MessageType::encode_chunk/encode_eof` methods where they reserve enough space for buffer.

Simplify `read_available` function to return boolean in Ok variant. As dispatcher's poll only check on Some(true) condition for it so it could be simplify to true for Some(true) conditon and false to everything else.

Simplify match on `PollResponse` enum. `PollResponse::DrainWriteBuf` is mapped to true for draining write buffer. `PollResponse::Upgrade` would return directly and goes to upgrade variant of dispatcher.

Move upgrade to a separate method to make dispatcher's poll method more compat.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -734,6 +734,78 @@ where
         }
         Ok(())
     }
+
+    /// Returns true when io stream can be disconnected after write to it.
+    ///
+    /// It covers these conditions:
+    ///
+    /// - `std::io::ErrorKind::ConnectionReset` after partial read.
+    /// - all data read done.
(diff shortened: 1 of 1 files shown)


robjtede (approved):
LGTM

PR #1982, opened 2021-02-12
rework ci

robjtede:
- Use single file for CI.
- Reduce OpenSSL install time on Windows.


robjtede:
@JohnTitor I'm going crazy with this windows CI thing. Any advice?

Goal is to use single CI file and reduce the install openssl time to <1 min. Seems to work until link time.

JohnTitor:
Seems we don't have some necessary libs in the proper path(s). And rocket-rs' package is old therefore the current script doesn't work well (https://github.com/actix/actix-web/runs/1886098868?check_suite_focus=true). I'd check some related dirs (e.g. `VCPKG_INSTALLATION_ROOT\installed\x64-windows\lib`) with `Get-ChildItem` then use the move command if we find `libcrypto.lib`.

robjtede:
What I don't understand is why (on most recent incantation) the failure is limited to stable runner.

JohnTitor:
Oh, that's weird indeed. Hmm, does it work when specifying 1.50.0, not stable? If it works unless it's `stable`, something is misconfigured, I guess.

robjtede:
Same error using 1.50.0 explicity. How odd.

JohnTitor:
Hmmm, 1.50 fails here but the config on master successes?? What I can think of is the difference between the OpenSSL version. Rocket's seems 1.0.2s (just extracted and checked on local) but rustc uses 1.1.x internally, this might cause the failure?

robjtede:
Ok could be. It's a shame because I tried copying config from a bunch of places that built OpenSSL and Rocket's was the one that worked at compile-openssl-sys time.

Using the "vendored" feature of the create seemed to work. Can you think of any downsides of using it?

JohnTitor:
I think the downside is that the feature is less configurable (https://docs.rs/openssl/0.10.32/openssl/#vendored). I'm not sure how many people use the `openssl` feature on Windows but I think it's worth trying it out in this beta period.

JohnTitor (approved):
Looks good, nice cleanup 👍

PR #1983, opened 2021-02-12
refactor DateService

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Refactor `DateService` to a infinite running async task that would update Date and Instant at 500 milli seconds.
This change save one branch on every request when getting date and instant. And would also save 1 additional spawn + 1 new async timer every 500 milli seconds when server is busy.
It would also simplify the logic of this service and reduce code amount by a bit.

Add notify_on_drop util mod for testing despawn async task with associated handle.

### DIFF
--- a/actix-http/src/config.rs
+++ b/actix-http/src/config.rs
@@ -249,7 +249,7 @@ impl DateService {
     fn new() -> Self {
         // shared date and timer for DateService and update async task.
         let current = Rc::new(Cell::new((Date::new(), Instant::now())));
-        let current_clone = current.clone();
+        let current_clone = Rc::clone(&current);
         // spawn an async task sleep for 500 milli and update current date/timer in a loop.
         // handle is used to stop the task on DateService drop.
         let handle = actix_rt::spawn(async move {
@@ -258,11 +258,8 @@ impl DateService {
 
             let mut interval = interval(Duration::from_millis(500));
             loop {
-                // TODO: consider use the Instant returned by interval stream
-                // (If it's accurate enough)
-                let _now = interval.tick().await;
+                let now = interval.tick().await;
                 let date = Date::new();
-                let now = Instant::now();
                 current_clone.set((date, now));
             }
         });
--- a/actix-http/src/config.rs
+++ b/actix-http/src/config.rs
@@ -4,9 +4,11 @@ use std::rc::Rc;
 use std::time::Duration;
 use std::{fmt, net};
 
-use actix_rt::time::{sleep, sleep_until, Instant, Sleep};
+use actix_rt::{
+    task::JoinHandle,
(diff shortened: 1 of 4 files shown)


actix-http/src/config.rs
```diff
@@ -323,13 +323,14 @@ mod tests {
         settings.set_date(&mut buf1);
         let now1 = settings.now();
 
-        sleep_until(Instant::now() + Duration::from_secs(1)).await;
+        sleep_until(Instant::now() + Duration::from_secs(2)).await;
+        yield_now().await;
```
robjtede, line 324: required to let the bg task update clock due to single threaded test runtime? do i have that right?

fakeshadow: I find ci handles async delay poorly. (like tests fail when it never would be on local)
I just hope a 2 second delay and a yield would actually make the date service timer updated for once.

actix-http/src/config.rs
```diff
@@ -230,72 +232,132 @@ impl fmt::Write for Date {
...
 
-impl DateServiceInner {
+impl DateService {
     fn new() -> Self {
-        DateServiceInner {
-            current: Cell::new(None),
-        }
+        // shared date and timer for DateService and update async task.
+        let current = Rc::new(Cell::new((Date::new(), Instant::now())));
+        let current_clone = current.clone();
```
robjtede: prefer Rc::clone call

actix-http/src/config.rs
```diff
@@ -230,72 +232,132 @@ impl fmt::Write for Date {
...
+        // handle is used to stop the task on DateService drop.
+        let handle = actix_rt::spawn(async move {
+            #[cfg(test)]
+            let _notify = notify_on_drop::NotifyOnDrop::new();
+
+            let mut interval = interval(Duration::from_millis(500));
+            loop {
+                // TODO: consider use the Instant returned by interval stream
+                // (If it's accurate enough)
+                let _now = interval.tick().await;
```
robjtede: it does not appear there would be any lag to using this returned value

robjtede (approved):
LGTM

actix-http/src/config.rs
fakeshadow: Could be happen when server is blocked by user code for a long time or there are too much polling on other futures like FuturesUnordered or JoinAll and delay the interval. The interval return the Instant from the timer's deadline in timer wheel rather than generate a new Instant.

I'm just guessing and have not test this.

robjtede: should not occur with join all or futures unordered
that said, if user is blocking thread or has written a poll busy loop they got bigger problems than an outdated Date header; could even be a useful signal that something is wrong

robjtede (approved):

PR #1984, opened 2021-02-12
do not self wake up when have a payload

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Do not self wake up when there is a payload. This would prevent the active consumption of request payload when payload is dropped intentionally.


robjtede:
For future reference, can you add to the description what test cases to perform to verify the behavior and what we'd expect for each.

fakeshadow:
I'd like to leave it like this for now as a temporary fix. 
This is a long standing problem and was not shown beacuse the 32kb buffer bug accidently block it from happening.

We have to determin what to do on a active drop of payload and make further refactor to this.

robjtede (approved):

PR #1994, opened 2021-02-15
rework client connection pool

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Rework actix-http client connection pool. This pool is only used in awc and is fairly complicated.

This PR use a more heavy but simpler approach. A `tokio::sync::Semaphore` is used to limit concurrent access to pool and connection's lifecycle is only checked when they are popped from pool. 

In theory this would make the pool slower and more memory hungry but it's arguable noticeable in real world.
And follow up changes can made into the new pool to adjust these problems if they show.


robjtede (approved):
LGTM!

PR #2005, opened 2021-02-17
return poll in poll_flush

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Return `Poll` directly. This would make the logic more clear when the dispatcher would branch to different path base on the poll state.


robjtede (approved):
LGTM

PR #2008, opened 2021-02-18
refactor awc::ClientBuilder

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Refactor `awc::ClientBuilder` to hold a generic connector type.
This change would make it easier to add middleware API to `ClientBuilder` as we would want middlewares to wrap around the generic connector type.
With the old API it would force inner connector type be a box and would also complicate the consturction method.

Remove redirect fields from `ClientBuilder`. They are no actual use for now and can be re add as middleware API for redirect,

This PR would also fix the issue where `max_http_version` ,`initial_window_size` and `initial_connection_window_size` have no effect on self defined connector type.


robjtede (changes requested):
add redirect method removal to changelog

robjtede (approved):

PR #2010, opened 2021-02-18
Fix Json extractor to be 32kB by default

SuperFluffy:
## PR Type
Other

## PR Checklist
- [x] Tests for the changes have been added / updated. (**not applicable**)

## Overview

While technically the default limit for `JsonBody` is indeed 256kB when called through `JsonBody::new`, in practice the limit is always set to 32kB for every context where `JsonBody` is used (via `JsonConfig`). This is also consistent with the documentation of the `JsonError::Overflow`, which mentions a default maximum of `32kB`.

This commit changes the default to also be `32kB` instead of `256kB`.

## Alternatives

The alternative to this commit is to instead change the default of `JsonConfig` to have a limit of `256kB` instead of currently `32kB`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,8 +3,11 @@
 ## Unreleased - 2021-xx-xx
 ### Changed
 * Feature `cookies` is now optional and enabled by default. [#1981]
+* `JsonBody::new` returns a default limit of 32kB to be consistent with `JsonConfig` and the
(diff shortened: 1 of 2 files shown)


robjtede (approved):
Needs a changelog entry but it's a reasonable change.

SuperFluffy:
> Needs a changelog entry but it's a reasonable change.

Done.

robjtede (approved):

PR #2013, opened 2021-02-19
add client middleware

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Add `awc::ClientBuilder::wrap` method and `awc::middleware::Transform` trait for building up middlewares for awc.
Middleware would wrap around `awc::Connector`'s `Service` type(Constructed by Connector::finish()) and can be used to mutate the input and output of connector service.

Add a sample redirect middleware as `awc::middleware::RedirectMiddleware`to test the middleware API. 
It only works on very limit case and is meant to be a placeholder middleware for test.


fakeshadow:
Sadly due to the separation of `actix-http` and `awc` currently multiple boxes are needed on connector service and dynamic dispatched futures are used. 
This can be improved in the future by moving more connector related service to wrap API and use as middleware.

awc/src/middleware/redirect.rs
```diff
@@ -0,0 +1,267 @@
...
+                connector,
+            } => {
+                match ready!(fut.poll(cx))? {
+                    ConnectResponse::Client(res) => match res.head().status {
+                        StatusCode::MOVED_PERMANENTLY
+                        | StatusCode::FOUND
+                        | StatusCode::SEE_OTHER
+                        | StatusCode::TEMPORARY_REDIRECT
+                        | StatusCode::PERMANENT_REDIRECT
+                            if *max_redirect_times > 0 =>
```
dunnock: Could be simplified to single match
 
```
ConnectResponse::Client(res) if res.head().status.is_redirection() && *max_redirect_times > 0 => {
```

fakeshadow: I'm not familiar with redirect but it seems there should be different treatment on all these status codes.(303, 307 for example) 

Your suggestion is a good one but I left them group together because I have not figured out details like how to reuse body for different cases.

awc/src/middleware/redirect.rs
```diff
@@ -0,0 +1,267 @@
...
+
+                            // rebuild uri from the location header value.
+                            let uri = res
+                                .headers()
+                                .get(header::LOCATION)
+                                .map(|value| {
+                                    Uri::builder()
+                                        .scheme(uri.scheme().cloned().unwrap())
+                                        .authority(uri.authority().cloned().unwrap())
+                                        .path_and_query(value.as_bytes())
```
dunnock: Just wondering of why are we reusing scheme and authority if via location we would get full blown URL if so can be replaced with Url::parse:
```
        .map(Url::parse).transpose().map_err(SendRequestError::Url)?
```

fakeshadow: I failed to parse location to Uri type so this is what I end up with.

fakeshadow: Changed to parse the location to uri directly. And fallback to builder when the new uri lack schema and authority.

awc/src/middleware/redirect.rs
fakeshadow: Added special handle for 307 and 308

robjtede (commented):
LGTM

robjtede (approved):

PR #2017, opened 2021-02-20
Inner field of web::Query is public again (#2016)

### LINKED ISSUE #2016: Consider making `web::Path` and `web::Query` fields public again
Hello! First of all thanks for the awesome work being done in the libraries!

While updating our company's codebase to tokio v1 (and thus actix-web v4) we noticed that #1894 made the field of `web::Path` and `web::Query` private. This change alone broke almost all our services' routes.

Given that the `web::Json` and `web::Form` still have their field public, it would be possible to rollback the change made in the former types? Having the fields private is a big burden for us, having to call `.into_inner()` everywhere is pretty annoying.

I can send a PR if this change is desired by the maintainers.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): `nightly-2021-01-25`
* Actix Web Version: `4.0.0-beta.3`

adamchalmers:
## PR Type

Feature

## PR Checklist

## Overview
The v4 beta made Query's inner field private, but the discussion in the linked issue said that we should revert this change, as it doesn't prevent any bad behaviour.

Because none of the documentation or examples have mentioned using .0 syntax, I didn't think any documentation changes were necessary. I didn't think any tests were necessary either, as this doesn't add any logic, it merely exports a field. If you'd like me to add an example or test, I'd be happy to!

Closes #2016

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,6 @@
 * Feature `cookies` is now optional and enabled by default. [#1981]
 * `JsonBody::new` returns a default limit of 32kB to be consistent with `JsonConfig` and the
   default behaviour of the `web::Json<T>` extractor. [#2010] 
-* `web::Query` has a public inner field again [#2016]
 
 [#1981]: https://github.com/actix/actix-web/pull/1981
 [#2010]: https://github.com/actix/actix-web/pull/2010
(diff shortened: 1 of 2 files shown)


robjtede:
Would appreciate a doc comment/test. You can edit the current one to use the public field if you like.

And no need for a changelog entry since this is a beta revert.

adamchalmers:
OK, added doctests to demonstrate how to use .0 or .into_inner.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 * Feature `cookies` is now optional and enabled by default. [#1981]
 * `JsonBody::new` returns a default limit of 32kB to be consistent with `JsonConfig` and the
   default behaviour of the `web::Json<T>` extractor. [#2010] 
+* `web::Query` has a public inner field again [#2016]
```
robjtede:
```suggestion
```

robjtede (approved):
LGTM

PR #2019, opened 2021-02-21
actix-http-test: minimize features of dependencies

alpire:
## PR Type

Other

## PR Checklist

## Overview

This PR disables default features from `awc` and `actix-web` in `actix-http-test`. This speeds up compilation by removing some optional dependencies like `brotli2` when they are not needed.

Helps with #1561


robjtede (approved):

therealprof (approved):
LGTM

PR #2021, opened 2021-02-22
Implement getters for named file fields

SuperFluffy:
## PR Type

Feature

## PR Checklist

## Overview

I wanted to change `ContentDisposition` to always be `attachment` no matter the underlying file type. But since there is currently no way to retrieve the current content disposition of `NamedFile`, I would have to guess the `filename` parameter for this header.

And since the other fields might also potentially be interesting to users, getter methods for those were added as well.

## Open questions

I did not add `pub fn flags(&self) -> &Flags`, because having the bit mask is quite meaningless without knowing what the bits stand for. Since there is no `set_flags` setter either, should we add methods like `pub fn is_FLAG_set(&self) -> bool`? For example, `pub fn is_etag_set(&self) -> bool`?


SuperFluffy:
Are you fine with the `tests` being doc tests that just check for compilation, similar to the `path` and `file` getters?

Also, should I add some methods to check if flags are set?

actix-files/src/named.rs
```diff
@@ -171,6 +171,42 @@ impl NamedFile {
...
 
+    /// Retrieve the time the file was modified.
+    #[inline]
+    pub fn modified(&self) -> Option<&SystemTime> {
+        self.modified.as_ref()
+    }
+
+    /// Retrieve the metadata associated with this file.
+    #[inline]
+    pub fn md(&self) -> &Metadata {
```
robjtede, line 273: would call this `metadata`

robjtede (changes requested):
otherwise, looks okay. for is_*_set, we can omit it for now unless someone brings a use case

yes, doc tests would be enough

robjtede:
superseded by #2689 due to restricted branch

thanks for getting this added 👍🏻

PR #2022, opened 2021-02-23
Port over doc comments in route macros.

### LINKED ISSUE #1389: proc-macro docstrings issue
As discussed on gitter (https://gitter.im/actix/actix?at=5e58817aca2f5a558d544542), it appears that docstrings used with the proc-macros gets wrapped into the inner function.

An example of the problem:
```rust
// Index function
///
/// Arguements:
///
/// * None
///
/// Returns:
/// * Responder with the string `Hello World!` in it.
#[get("/")]
pub async fn index() -> impl Responder {
    "Hello World!".to_string()
}
```

Then when running `cargo doc` on the project, the above docstring does not get rendered.

erikjohnston:
## PR Type

Bug Fix / Feature

## PR Checklist

## Overview

This allows documentation on the functions to appear in the generated
docs, e.g.:

```rust
/// The index page
async fn index() -> &'static str {
    "Hello!"
}
```

Closes #1389

--- 

I'm afraid I have no idea how to add tests for this beyond making sure this doesn't break when using docstrings.

### DIFF
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -229,13 +229,12 @@ impl Route {
         //
         // Note that multi line doc comments are converted to multiple doc
         // attributes.
-        let mut doc_attributes = Vec::new();
-        for attr in &ast.attrs {
-            if attr.path.get_ident().is_ident("doc") {
-                    doc_attribute.push(attr.clone());
-                }
-            }
-        }
+        let doc_attributes = ast
+            .attrs
+            .iter()
+            .filter(|attr| attr.path.is_ident("doc"))
+            .cloned()
+            .collect();
 
         let args = Args::new(args, method)?;
         if args.methods.is_empty() {
@@ -264,7 +263,7 @@ impl Route {
             args,
             ast,
             resource_type,
-            doc_attributes: doc_attribute,
+            doc_attributes,
         })
     }
 }
@@ -304,13 +303,8 @@ impl ToTokens for Route {
             }
         };
 
-        let doc_comment_token_stream: TokenStream2 = doc_attribute
-            .iter()
-            .map(ToTokens::to_token_stream)
-            .collect();
-
         let stream = quote! {
-            #(#doc_comments)*
+            #(#doc_attributes)*
             #[allow(non_camel_case_types, missing_docs)]
             pub struct #name;
 
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -177,8 +177,8 @@ pub struct Route {
     ast: syn::ItemFn,
     resource_type: ResourceType,
 
-    /// The doc comment attribute to copy to generated struct, if any.
-    doc_attribute: Option<syn::Attribute>,
+    /// The doc comment attributes to copy to generated struct, if any.
+    doc_attributes: Vec<syn::Attribute>,
 }
 
 fn guess_resource_type(typ: &syn::Type) -> ResourceType {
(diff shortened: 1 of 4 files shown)


robjtede:
Does this handle multi line doc-comments? I don't recall if they're converted to one doc attribute with line breaks or multiple doc attrs.

erikjohnston:
> Does this handle multi line doc-comments? I don't recall if they're converted to one doc attribute with lie breaks or multiple doc attrs.

Huh, you're right, had no idea that is multi line comments worked. I've pushed a fix and manually tested.

Tangentially, I am wondering whether it'd be better to just copy all attributes over? It's probably not *that* useful, but would mean e.g. that people could have manually added `#[allow(missing_docs)]` to the structs.

robjtede:
> Tangentially, I am wondering whether it'd be better to just copy all attributes over?

I'm apprehensive about that because it may seem strange that attributes meant for functions will not work on the generated struct. If you don't feel strongly that it should be done I'd like to wait until it is raised specifically.

actix-web-codegen/src/route.rs
```diff
@@ -221,6 +224,20 @@ impl Route {
...
 
+        // Try and pull out the doc comments so that we can reapply them to the
+        // generated struct.
+        //
+        // Note that multi line doc comments are converted to multiple doc
+        // attributes.
+        let mut doc_attribute = Vec::new();
+        for attr in &ast.attrs {
+            if let Some(ident) = attr.path.get_ident() {
+                if ident == "doc" {
```
jplatte:
```suggestion
            if attr.path.get_ident().is_ident("doc") {
```

actix-web-codegen/src/route.rs
```diff
@@ -221,6 +224,20 @@ impl Route {
         let ast: syn::ItemFn = syn::parse(input)?;
         let name = ast.sig.ident.clone();
 
+        // Try and pull out the doc comments so that we can reapply them to the
+        // generated struct.
+        //
+        // Note that multi line doc comments are converted to multiple doc
+        // attributes.
+        let mut doc_attribute = Vec::new();
```
jplatte:
```suggestion
        let mut doc_attributes = Vec::new();
```

then the return value can also use struct init shorthand for this.

jplatte: Actually I think this + the whole loop can be simplified to

```rust
let doc_attributes = ast.attrs.iter().filter(|attr| attr.path.is_ident("doc")).cloned().collect();
```

actix-web-codegen/src/route.rs
```diff
@@ -265,6 +283,7 @@ impl ToTokens for Route {
                     methods,
                 },
             resource_type,
+            doc_attributes: doc_attribute,
```
jplatte:
```suggestion
            doc_attributes,
```

actix-web-codegen/src/route.rs
```diff
@@ -286,7 +305,13 @@ impl ToTokens for Route {
             }
         };
 
+        let doc_comment_token_stream: TokenStream2 = doc_attribute
+            .iter()
+            .map(ToTokens::to_token_stream)
+            .collect();
+
```
jplatte: This is not needed, you can just use a repetition in the output. Suggestion in next review comment.

actix-web-codegen/src/route.rs
```diff
@@ -286,7 +305,13 @@ impl ToTokens for Route {
...
         let stream = quote! {
+            #doc_comment_token_stream
```
jplatte:
```suggestion
            #(#doc_comments)*
```

actix-web-codegen/src/route.rs
robjtede: https://docs.rs/syn/1.0.60/syn/struct.Path.html#method.is_ident

erikjohnston:
Thank you both, my proc macro foo is still quite weak :heart:

jplatte (approved):

robjtede (approved):

PR #2032, opened 2021-02-27
add interval check for payload pending

### LINKED ISSUE #1313: A Stream object in a pending state won't be dropped after the client disconnect

```console
```

```console
```

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Add interval check(1 second) on Io stream's read status when response payload stream enters pending.
When interval check detect Io is not readable anymore it would assume remote client have disconnect and end dispatcher. All unfinished responses would be lost.

Closes #1313


fakeshadow:
Draft this for now until finding a way to detect client disconnect

fakeshadow:
Should this active across aboard or put behind a feature gate?

robjtede:
What's the rationale for wanting a feature flag? You expect a perf hit?

fakeshadow:
I suspect this could conflict with user keep alive code like a longer heartbeat they define with extra logic on heartbeat expire. This could end the connection before their heartbeat logic kick in hense breaking.

I'm not sure about this though just a guess.

kmyers608:
Is there anything I can help with to move this forward? We believe that this issue is the root cause behind an issue we have been running into running actix behind envoy.

fakeshadow:
> Is there anything I can help with to move this forward? We believe that this issue is the root cause behind an issue we have been running into running actix behind envoy.

Thanks for asking.

I want to make a bit clearify that this PR is not an issue fix but mostly a feature(a hacky one).

The problem of this feature is it goes against convensional handing of Rust future/stream type and may cause problem for people expecting otherwise. Like this https://github.com/actix/actix-web/pull/2032#issuecomment-809375785 discussed.

It would be pretty hard to test the possibility of break people code so I tend to give this PR sometime and potentially make it a gated feature that is opt-in. 

Edit: It would be appreciated to help test and/or help putting this behind a feature. I would try to keep this PR in sync with master

fakeshadow:
I don't have time to keep up the sync with master branch anymore.
This PR would be left where it's now and anyone feel like to can pick it up or make a new one for the same feature.

PR #2046, opened 2021-03-05
add support of filtering guards in Files of actix-files

### LINKED ISSUE #2045: actix-files Files use_guards produce errors instead of moving to the next possible path match
When using the Files use_guards is not possible to use them to filter the call, is possible to use them only for error checking.
what I'm trying to solve is a something like:

```
        App::new()
            .service(Files::new("/", "/var/www/example").use_guard(Any(Host("example.com"))))
            .service(Files::new("/", "/var/www/example2").use_guard(Any(Host("example2.com"))))
```

What happen  in this case, is that as soon as the first guard is not matched the request goes in error, instead to progress to the 
other service.

digging inside the code I found the reason of the problem, the guards are checked here:
https://github.com/actix/actix-web/blob/78384c3ff57f2fee1820cdb13a720cbf75de54e9/actix-files/src/service.rs#L63
instead of been sent forward as parameter in this call:
https://github.com/actix/actix-web/blob/78384c3ff57f2fee1820cdb13a720cbf75de54e9/actix-files/src/files.rs#L239

I can fix this myself and do a pull request, anyway this is a braking change, because the request instead of failing it will progress to the next service, and people may rely on the failure.

An alternative solution would be just add a new method called only 'guard' as most of actix use with the "filtering" behaviour and  then leave `use_guards` with the current behaviour.

tglman:
## PR Type

Feature

## PR Checklist

## Overview

This PR fix the issue https://github.com/actix/actix-web/issues/2045, this though at the moment has a braking changed, i had to remove the implementation of Clone for Files, I'm investigating if the clone is actually needed in the actix flow.

closes #2045


actix-files/src/files.rs
```diff
@@ -46,24 +47,6 @@ impl fmt::Debug for Files {
     }
 }
 
-impl Clone for Files {
```
ibraheemdev: Is removing the `Clone` impl necessary?

actix-files/src/files.rs
```diff
@@ -225,7 +217,13 @@ impl Files {
 }
 
 impl HttpServiceFactory for Files {
-    fn register(self, config: &mut AppService) {
+    fn register(mut self, config: &mut AppService) {
+        let guards = if self.guards.is_empty() {
+            None
+        } else {
+            Some(std::mem::take(&mut self.guards))
+        };
```
ibraheemdev:
```suggestion
        let guards = self.guards.take();
```

actix-files/src/files.rs
```diff
@@ -184,7 +168,15 @@ impl Files {
     /// Default behaviour allows GET and HEAD.
     #[inline]
     pub fn use_guards<G: Guard + 'static>(mut self, guards: G) -> Self {
-        self.guards = Some(Rc::new(guards));
+        self.use_guards = Some(Rc::new(guards));
+        self
+    }
+
+    /// Specifies custom guards to use for directory listings and files.
+    ///
```
ibraheemdev, line 176:
```suggestion
    /// Specifies custom guards to use for directory listings and files.
```

ibraheemdev (changes requested):

actix-files/src/files.rs
tglman: the issue is that for support the clone the guards need to be wrapped by an `Rc`, or need to be able to implement the clone, 
the previous implementation had the `Rc` but the `config.register_service(rdef, guards, self, None)` do not support `Rc` guards.

A  way around this would be introduce a new `impl <T:Guard> Guard for Rc<T>`  so then the guards vec can become a 
` Vec<Box<Rc<Guard>>>`  that become clonable and implement itself `Vec<Box<Guard>>`

If this make sense i can add that implementation myself and than re-introduce the clone.
The question is also if the `Clone` is needed other structure like `Resource` do not implement it.

tglman: Ok Could manage to do that, now the changes looks even smaller then before.

ibraheemdev (approved):
Looks good now :+1:

tglman:
Hi, 

Since the last changes all the braking changes were fixed, any chance to get this merged ?

Regards

robjtede (approved):
i will tweak some docs outside this PR but overall it is good change and needed, thanks 👍🏻

PR #2049, opened 2021-03-07
fix h1 client for handling expect header request

### LINKED ISSUE #1775: actix client struggles with 'Expect: 100-continue' header
First of all, thanks for this great tool.

## Expected Behavior
When the client send a put request with a `Expect: 100-continue` header, it could receive a 100 (Continue) intermediary response. (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect, https://tools.ietf.org/html/rfc7231#section-5.1.1). But the client must not consider it as the final server answer.

## Current Behavior
The client immediatly returns the intermediary response `HTTP/1.1 100 Continue`.

```
use actix_web::Error;

#[actix_web::main]
async fn main() -> Result<(), Error> {

    let client = actix_web::client::Client::new();

    let response = client
        .put("http://httpbin.org/put")
        .header("Expect", "100-continue")
        .send_body("hello")
        .await?;

    println!("Is success: {:?} ", response.status().is_success()); // is_success : false ; status code: 100

    Ok(())
}
```

## Context
I was hit by this pb while trying to debug an integration test of an proxy (https://github.com/betagouv/ds_proxy/blob/master/tests/end_to_end.rs#L91) . It could also be reproduce using the vanilla example https://github.com/actix/examples/tree/master/http-proxy . And I suppose it should re open https://github.com/actix/examples/issues/273 .

## Your Environment
ubuntu 18.04
rustc 1.46.0
actix-web 3.2.0

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Properly handle h1 EXPECT client request. Would contribute to solve #1775

### DIFF
--- a/actix-http/src/client/h1proto.rs
+++ b/actix-http/src/client/h1proto.rs
@@ -7,13 +7,15 @@ use actix_codec::{AsyncRead, AsyncWrite, Framed, ReadBuf};
 use bytes::buf::BufMut;
 use bytes::{Bytes, BytesMut};
 use futures_core::Stream;
-use futures_util::future::poll_fn;
-use futures_util::{SinkExt, StreamExt};
+use futures_util::{future::poll_fn, SinkExt, StreamExt};
 
 use crate::error::PayloadError;
 use crate::h1;
 use crate::header::HeaderMap;
-use crate::http::header::{IntoHeaderValue, HOST};
+use crate::http::{
+    header::{IntoHeaderValue, EXPECT, HOST},
+    StatusCode,
+};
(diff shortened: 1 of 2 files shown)


robjtede:
https://tools.ietf.org/html/rfc7231#section-5.1.1

In particular, "Requirements for clients":
> A client MUST NOT generate a 100-continue expectation in a request that does not include a message body.

So is_expect condition should also check if body exists. Also suggest referencing the spec in comments here.

robjtede (approved):
LGTM

PR #2050, opened 2021-03-07
use iota for more content-length insertions

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Use `itoa` in more places.


fakeshadow (approved):
LGTM.

It would be great if None can still be used as password input. I feel it make more sense when None means no password other than ""

robjtede:
The motives to change:
- password is used more often than it isn't when doing basic auth
  - techincally username would be optional too; this gives it symmetry
- the actual output is a blank password so input reflects this
- impl Display opens it up to more types
  - in particular Zeroize wrapped types for a password field are something i'd like to support
- using Option<impl Display> requires type annotations when using None since default types are not supported in that position

PR #2052, opened 2021-03-07
json method receives plain serialize

### LINKED ISSUE #2047: `ResponseBuilder::json` requiring the argument to impl `Deref` is confusing (4.0.0-beta.3)
## Expected Behavior

It should be possible to use an owned value for `HttpResponse::Ok().json(<here>)`. Currently, this doesn't work, and the error message is not very helpful:

```
error[E0277]: the trait bound `MySerializableType: Deref` is not satified
```

## Current Behavior

Currently, only types implementing `Deref` can be passed to `Json`.

## Possible Solution

Only require `Serialize` again. `&T where T: Serialize` implements `Serialize` itself so that would work, otherwise requiring a reference like `serde_json::to_string` would work too (it would mean churn when you moved the value into `.json()` before, but at least give a helpful diagnostic). `AsRef<T> where T: Serialize` would work too if you want to be more generic (i.e. supporting smart pointers that don't implement `Serialize`).

## Context

Upgrading from actix-web 3.0 to 4.0.0-beta.3.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.50.0 (cb75ad5db 2021-02-10)
* Actix Web Version: `4.0.0-beta.3`

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Following on from discussion in issue (tl;dr we want to support T and &T with a single method and generall support as many variants of Serialize as possible), this might be a suitable replacement for the weird signature .json has now while supporting all it's current use cases and more; plus better error messages.

closes #2047

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -4,6 +4,7 @@
 ### Changed
 * Feature `cookies` is now optional and disabled by default. [#1981]
 * `ws::hash_key` now returns array. [#2035]
+* `ResponseBuilder::json` now takes `impl Serialize`. [#2052]
 
 ### Removed
 * re-export of `futures_channel::oneshot::Canceled` is removed from `error` mod. [#1994]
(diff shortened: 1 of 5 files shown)


ibraheemdev (approved):
:+1:

jplatte (approved):

PR #2056, opened 2021-03-08
Print response body in read_response panic messages

### LINKED ISSUE #2055: test::read_response_json: log response body when deserialization fails
## Expected Behavior
When json deserialization fails, not only should it panic with an error, but the panic message should include the json body to make debugging easier.

## Current Behavior
It panics with an error message like `thread 'integration_tests::get_asset_route' panicked at 'read_response_json failed during deserialization', /home/jonas/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-web-4.0.0-beta.3/src/test.rs:326:29` with no further information.

## Possible Solution
Pretty sure you could just add `body: {:?}", body` to the `panic!` invocation in read_response_json; that prints something like `b"my json string"` for me.

Call read_response_json with a request to any route that doesn't return json or returns a different format from what you're trying to deserialize.

## Context
Debugging would be made easier.

* Rust Version (I.e, output of `rustc -V`): rustc 1.52.0-nightly (4f20caa62 2021-03-01)
* Actix Web Version: 4.0.0-beta.3

ibraheemdev:
## PR Type
DX

## PR Checklist

## Overview

resolves #2055

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -162,7 +162,7 @@ where
     let mut resp = app
         .call(req)
         .await
-        .unwrap_or_else(|_| panic!("read_response failed at application call"));
+        .unwrap_or_else(|e| panic!("read_response failed at application call: {}", e));
 
     let mut body = resp.take_body();
     let mut bytes = BytesMut::new();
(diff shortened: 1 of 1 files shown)


robjtede:
I think a separate line for the bad json would be better, in particular because currently it outputs an list of decimal bytes.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4e6b37dafd5764275d496143af4e0d16

robjtede:
partly merged in with the headers PR but i'll finish it up before the beta release

PR #2059, opened 2021-03-09
fix bug where URIs could traverse server filesystem

### LINKED ISSUE #2001: Path tail extractor does not protect against directory traversal attacks
## Expected Behavior
When matching against the url tail any path segment that is equal to .. should be ignored

## Current Behavior
When using this code and the url `http://localhost:8080/..%5c/..%5c/test.txt` the decoded path results in `..\\/..\\/test.txt`  
This allows anyone to read the entire filesystem

```
async fn index(req: HttpRequest) -> Result<NamedFile> {
    let tail: PathBuf = req.match_info().query("filename").parse()?;
    println!("{:?}", tail);
    Ok(NamedFile::open(Path::new("public/").join(tail))?)
}

#[actix_web::main]
    use actix_web::{web, App, HttpServer};

    HttpServer::new(|| App::new().route("/{filename:.*}", web::get().to(index)))
        .bind("127.0.0.1:8080")?
        .run()
        .await
}
```

## Context
I just copied the example file server from the documentation. 
According to the documentation this should produce a path that is safe to use as a suffix but it is clearly not.
`As a result of these conditions, a PathBuf parsed from request path parameter is safe to interpolate within, or use as a suffix of, a path without additional checks.`

## Your Environment
rustc 1.52.0-nightly (5fa22fe6f 2021-02-14)
actix-web = "3.3.2"
actix-files = "0.5.0"
Windows 10

tomharmon:
## PR Type
Bug Fix

## PR Checklist

## Overview

#2001 has a good summary. This PR makes it so `actix_files` will return a Forbidden response if the request tries to access files that are outside of the directory the server is supposed to be served from.

Closes #2001


robjtede:
I think the fix should be more like: traversal is okay but going higher than the base directory is not allowed.

tomharmon:
@robjtede I updated the PR with that approach

tomharmon:
thanks for the fixes! I don't know what happened there. I ran them, I thought it said all the tests had passed 🤔 .

robjtede:
I don't think I fixed anything, just moved some things.

actix-files/src/service.rs
```diff
@@ -88,6 +88,18 @@ impl Service<ServiceRequest> for FilesService {
             Err(e) => return self.handle_err(e, req),
         };
 
+        let is_path_descendant = path
+            .as_path()
+            .ancestors()
+            .any(|ancestor| ancestor == self.directory.as_path());
+        if !is_path_descendant {
```
ibraheemdev, line 108:
```suggestion
        let is_path_descendant = path
            .as_path()
            .ancestors()
            .any(|ancestor| ancestor == self.directory.as_path());
            
        if !is_path_descendant {
```

ibraheemdev (approved):
:+1:

robjtede:
from CI

```
thread 'test_directory_traversal_prevention' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidUri(InvalidAuthority)', actix-http/src/test.rs:99:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test test_directory_traversal_prevention ... FAILED
```

tomharmon:
my bad, I was running the tests in the wrong folder.... 🤦 . However I don't think the test I wrote will work, because I don't think you can make a `Uri` with `..` in it anyways. Any suggestions on how to test?

robjtede:
could you just start it with / ?

aliemjay:
I have a couple problems with this:
* First, the linked issue reports a vulnerability in the documentation code, which uses `NamedFile` directly. So this PR doesn't fix the linked issue!
* Second, this implementation would prevent serving a symlink that points outside the `serve_from` path, which is a totally valid case.

Currently, `FilesService` behave as expected: the test url `..%5c/README.md` already failes and I've just tested it with https://github.com/xmendez/wfuzz/blob/master/wordlist/Injections/Traversal.txt

aliemjay:
@robjtede Can you have another look at this? the linked issue was closed in actix/actix-website#230

robjtede:
on review, i havent found an obvious path traversal vuln in the files service
added some tests here: 56ee97f722292cfc7c9dabd6c9039b1d9cee218f

if you can produce a reproduction case of being able to traverse parent dirs, please do get in touch

PR #2060, opened 2021-03-09
style change to h1 dispatcher and address todo

fakeshadow:
## PR Type

Code Style

## PR Checklist

## Overview

Address todo in `actix_http::h1::dispatcher`.
Style change to the code to make the logic more clear.

I'm not sure if this is a good change. It's mostly style change with no real meaning other than try to make the exit condition of poll_response loop more clear.


robjtede:
can you clarify the answer to the TODO? it had references to a macos specific behavior it seemed. idk what nikolay meant by it

fakeshadow:
> can you clarify the answer to the TODO? it had references to a macos specific behavior it seemed. idk what nikolay meant by it

I believe it's to make sure IO is fully read to Pending state. Where it would register the io event and schedule dispatcher to wake up for sure.

There are only 3 outcome from this loop (exclude the upgrade request):

1. `poll_response` pending. `poll_flush` ready/pending. In this case there would at least one source of event that wake up the dispatcher which are `poll_response` become ready and/or `poll_flush` ready.(If it returns pending in last poll.)

2. `poll_response` ready. `poll_flush` pending. So there is no point keep on poll next response as nothing can be written for now. In this case there would be one source of event that wake up dispatcher which is `poll_flush` become ready.

3. `poll_response` and `poll_flush` both ready. This is the case we want to keep loop because at this point there is no pending state which mean no event would register the dispatcher waker and wake it up. Hense the comment about hang the dispatcher. 

In short we only get a guarantee wake up and not stall dispatcher if we get pending state from either of `poll_respone` or `poll_flush` (This statement exclude keep_alive wake up and new request wake up. As they are not related to this loop and can happen at the time we want or not.)

therealprof:
Makes sense to me. I could actually test this if someone had a test case...

fakeshadow:
Some of the tests would fail if the condition is relaxed on `poll_flush`. Like we give `poll_response and poll_flush both ready.` case a pass.

Whenever `poll_flush` is the only waker for dispatcher if it's not write till pending(or finish) we get a dead dispatcher on some platforms.

There is a related talk happen recently on mio's discord channel for this and it points to this piece of doc. https://docs.rs/mio/0.7.11/mio/struct.Poll.html#draining-readiness

PR #2063, opened 2021-03-10
simplify connector generic type

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Simplify `actix_http::client::Connector`'s generic type. Remove Stream type from the signature.

Call `ConnectionPoll as Service<Req>::poll_ready` when tls connection pool is enabled. The current tls services use no op `poll_ready` impl but this change would make sure future changes to it would not inferior the ready state check.


robjtede (approved):
LGTM

PR #2070, opened 2021-03-13
enable fn index(web::Path((id, name)): web::Path<(u32, String)>)

rich-murphey:
## PR Type
Bug Fix

## PR Checklist

## Overview
Use of the Path extractor is documented in README.md, which says this is valid:
`async fn index(web::Path((id, name)): web::Path<(u32, String)>) -> impl Responder {`

The above fails to compile.  Earlier versions of actix-web supported Path de-structuring.  The example in README.md and any users' code that uses the Path de-structuring will fail to compile without this PR.

One of the motivations for using the path de-structuring, is to make it easier to review code and ensure that the mapping of path items to corresponding variables, for example:
```
#[get("/foo/{limit}/{offset}")]
pub async fn foo(
    web::Path((limit, offset)): web::Path<(i64, i64)>,
) -> HttpResponse {
```
I fee this is easier to maintain.

I'd be glad to add more tests, documentation or whatever else is needed.

### DIFF
--- a/src/types/path.rs
+++ b/src/types/path.rs
@@ -46,7 +46,7 @@ use crate::{dev::Payload, error::PathError, FromRequest, HttpRequest};
 /// }
 /// ```
 #[derive(PartialEq, Eq, PartialOrd, Ord)]
-pub struct Path<T>(T);
+pub struct Path<T>(pub T);
 
 impl<T> Path<T> {
     /// Unwrap into inner `T` value.


robjtede:
change is intentional, see this comment
https://github.com/actix/actix-web/issues/2016#issuecomment-782347086

rich-murphey:
The code in README.md does not compile, user's may not understand the change without some mention.

robjtede:
the code in readme is suitable for the current stable version, v3.3, and has been kept as such to reduce confusion during the beta period

brianbruggeman:
Where can I find documentation for the beta updates?

PR #2072, opened 2021-03-13
Remove ConnectionLifetime trait. Simplify Acquired handling

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `actix_http::client::ConnectionLifetime` trait. This trait is not generic over different types and is used to pass Frame's Io type in `actix_http::client::h1proto::PlStream`. Instead of this trait a `H1Connection<Io>` can be passed to `PlStream` as frame Io.

Remove the optional field requirement on multiple types and functions use `actix_http::client::pool::Acquired`.

Handle write error in `actix_http::client::h1proto::send_request` function.

### DIFF
--- a/actix-http/src/client/pool.rs
+++ b/actix-http/src/client/pool.rs
@@ -445,7 +445,7 @@ mod test {
     where
         T: AsyncRead + AsyncWrite + Unpin + 'static,
     {
-        let (conn, created, mut acquired) = conn.into_parts();
+        let (conn, created, acquired) = conn.into_parts();
         acquired.release(conn, created);
     }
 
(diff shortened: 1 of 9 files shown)


actix-http/src/client/pool.rs
```diff
@@ -448,7 +446,7 @@ mod test {
         T: AsyncRead + AsyncWrite + Unpin + 'static,
     {
         let (conn, created, mut acquired) = conn.into_parts();
```
robjtede: clippy

robjtede (approved):

PR #2081, opened 2021-03-17
refactor actix_http connection types and connector services

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `actix_utils::TimeoutService` usage in `actix_http::client::connector`. In exchange with some hand written timeout handling it's now possible to handle tcp connection and tls handshake timeout separately. `actix_http::client::Connector::handshake_timeout` method is added for customize the timeout independently.

Rename `actix_http::client::connector::InnerConnector` to `ConnectorServicePriv` and export it's type alias as `ConnectorService`. This would enable downstream dep using this named type of service directly instead of box a impl trait opaque service type. This would reduce indirection.

Remove `actix_http::client::connection::Connection` trait. It's not generic over different types and only impl for `EitherIoConnection`. Instead `EitherIoConnection` is renamed to `Connection` type that unfily all Http1/2 connection types. This would make future http3 variant type easier to added into the type and reduce unnecssary generics.

Refactor `actix_http::client::h1protp::H1Connection` to make them work better with above `Connection` type and move it to `actix_http::client::connection` module.

Refactor `actix_http::client::connection::H2Connection` for the same reason.

Remove `actix_http::client::connection::EitherIo` enum. Use `Connection` type directly in `Framed` and tunnel API. This reduce duplicate types.
 
Rename `awc::connect::ConnectorService` to `BoxConnectorService` to prevent naming conflict with `actix_http::client::connector::ConnectorService`.

Internal change to `awc::ClientConfig` to reduce indirection.

In general this PR is aiming for reduce indirection of awc::Client object(3 less boxes for accessing connector service), add tls handshake timeout and make future http3 implementation easier.

### DIFF
--- a/actix-http/src/client/connection.rs
+++ b/actix-http/src/client/connection.rs
@@ -1,14 +1,16 @@
-use std::ops::{Deref, DerefMut};
-use std::pin::Pin;
-use std::task::{Context, Poll};
-use std::{fmt, io, time};
+use std::{
+    io,
+    ops::{Deref, DerefMut},
+    pin::Pin,
+    task::{Context, Poll},
+    time,
+};
 
 use actix_codec::{AsyncRead, AsyncWrite, Framed, ReadBuf};
 use actix_rt::task::JoinHandle;
 use bytes::Bytes;
 use futures_core::future::LocalBoxFuture;
 use h2::client::SendRequest;
-use pin_project::pin_project;
 
 use crate::body::MessageBody;
 use crate::h1::ClientCodec;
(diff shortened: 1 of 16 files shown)


actix-http/src/client/connection.rs
```diff
@@ -19,177 +21,230 @@ use super::error::SendRequestError;
...
-    }
+    Tcp(ConnectionType<A>),
+    Tls(ConnectionType<B>),
 }
 
-impl<T: AsyncRead + AsyncWrite + Unpin> IoConnection<T> {
-    pub(crate) fn new(
-        io: ConnectionType<T>,
+/// Unified connection type cover Http1/2 protocols
+pub enum ConnectionType<Io: ConnectionIo> {
```
robjtede, line 193: should def be marked non_exhaustive if it's public

fakeshadow: There is no public mod in client module and this type is not exported. It's marked as public because some public service impl return it as output.

The only connection type export is `Connection` which can only be the two variant it has(I guess)

robjtede: as an output, if you could match on the result then it should be non_exhaustive

fakeshadow: How can it be matched if the type itself is not importable? The output type is not used in public either. It's a forward typed future that expose as generic param. To make most of the future chain static.

robjtede: truth

robjtede (approved):

Issue #2039, opened 2021-03-01
AppEntry inaccessible due to app_service being a private module


philip-peterson:
Hi there,

I was hoping to make a function that returns an `App`, but because it takes two generic types, one of which defaults to `AppEntry`, that was impossible. It would be nice if the `app_service` module was public so we can do this.

fakeshadow:
It's possible and you don't need the `AppEntry` type. As soon as you introduce some middleware it would transform to opaque type.

```rust
use actix_service::ServiceFactory;
use actix_web::dev::{Body, ServiceRequest, ServiceResponse};
use actix_web::middleware::{Compat, Logger};
use actix_web::{get, App, Error, HttpServer};

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| app())
        .bind("127.0.0.1:8080")?
        .run()
        .await
}

fn app() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Config = (),
        Response = ServiceResponse,
        Error = Error,
        InitError = (),
    >,
    Body,
> {
    App::new()
        .wrap(Compat::new(Logger::default()))
        .service(index)
}

#[get("/hi")]
async fn index() -> &'static str {
    "Hello World!"
}
```

In general do not return an App instance unless you have a very good understanding of how to describe the complicated types App returns. It wants an opaque type that impl `ServiceFactory` trait and a `Body` generic type of response body
Instead you can use `App::configure` API to customize your app build.

simon-an:
Your code does not compile with the latest stable release

robjtede:
[App::configure](https://docs.rs/actix-web/3/actix_web/struct.App.html#method.configure) is the recommended solution. `app_service` is intentionally private.

fakeshadow:
> Your code does not compile with the latest stable release

Sorry the code is for actix-web v4. For actix-web v3 there is no Compat middleware to help you figure out the body type but you can try this if you don't wrap any middleware. 
```rust
use actix_service::ServiceFactory;
use actix_web::dev::{Body, ServiceRequest, ServiceResponse};
use actix_web::{get, App, Error, HttpServer};

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| app())
        .bind("127.0.0.1:8080")?
        .run()
        .await
}

fn app() -> App<
    impl ServiceFactory<
        Request = ServiceRequest,
        Config = (),
        Response = ServiceResponse,
        Error = Error,
[...]

PR #2082, opened 2021-03-17
Publicly use actix_web::app_service::AppEntry as actix_web::AppEntry

kotovalexarian:
This allows Actix users to move application to separate function. This may be useful when application is big.

```rust
use actix_web::{get, App, AppEntry, HttpResponse, HttpServer, Responder};

#[actix_web::main]
    HttpServer::new(|| app())
        .bind("127.0.0.1:8000")?
        .run()
        .await
}

fn app() -> actix_web::App<AppEntry, actix_web::body::Body> {
    App::new()
        .service(index)
}

#[get("/")]
fn index() -> impl Responder {
    HttpResponse::Ok().body("Hello, World!")
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,6 +107,7 @@ pub use actix_rt as rt;
 pub use actix_web_codegen::*;
 
 pub use crate::app::App;
+pub use crate::app_service::AppEntry;
 pub use crate::extract::FromRequest;
 pub use crate::request::HttpRequest;
 pub use crate::resource::Resource;


robjtede:
App is not designed to be used directly.

Search AppEntry in closed issues to find more info.

fakeshadow:
No this can not. You can check the reason and how to properly return an App instance here
https://github.com/actix/actix-web/issues/2039#issuecomment-788102575

AppEntry is only the default type of App and it would mutate to some opaque type in your code. Exposing it would be not useful at all.

PR #2084, opened 2021-03-18
fix routes in Path documentation

obayemi:
## PR Type
Documentation

## PR Checklist

## Overview
The example in the documentation for `actix_web::web::Path` uses the `#[get()]` to set the url, but with wrong urls
I don't really know how relevant it is to ad such a insignificant doc change, should it be added too?


robjtede (approved):
thanks 👍🏻

PR #2089, opened 2021-03-19
reduce branches in h1 dispatcher poll_keepalive

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Else block starts on L651 is no use. When flag `SHUTDOWN` is enabled after `poll_keepalive` return dispatcher future would enter this branch where no stream read and payload handling happens https://github.com/actix/actix-web/blob/78fcd0237a4033e9802313de0299ce9cbb0a614c/actix-http/src/h1/dispatcher.rs#L852.

If condition on L685 is unnecessary as the Else condition on L683 itself is `!this.flags.contains(Flags::STARTED)`.

L696 is no op as `send_response` method would set dispatcher state to None when the response body is None or Empty. https://github.com/actix/actix-web/blob/78fcd0237a4033e9802313de0299ce9cbb0a614c/actix-http/src/h1/dispatcher.rs#L288


therealprof (approved):
LGTM!

robjtede (approved):

PR #2090, opened 2021-03-19
Ensure the content-disposition header is included in every part of a multipart message

### LINKED ISSUE #1947: `Content-Type` must be present in `multipart/form-data`
## Expected Behavior

`actix_multipart::Field::content_disposition` should return `&ContentDisposition`, because the `ContentDisposition` is mandatory on multipart form subparts.

## Current Behavior

At the moment, we have the following function signature:`actix_multipart::Field::content_disposition(&self) -> Option<ContentDisposition>`.

Interestingly, the code actually notes this: https://github.com/actix/actix-web/blob/master/actix-multipart/src/server.rs#L417-L426:

```rust
    /// Get the content disposition of the field, if it exists
    pub fn content_disposition(&self) -> Option<ContentDisposition> {
        // RFC 7578: 'Each part MUST contain a Content-Disposition header field
        // where the disposition type is "form-data".'
        if let Some(content_disposition) = self.headers.get(&header::CONTENT_DISPOSITION)
        {
            ContentDisposition::from_raw(content_disposition).ok()
        } else {
            None
        }
    }
```

## Possible Solution

`Field`s are constructed by `multipart.try_next().await`. A `Field` should be a valid subpart of the multipart stream, so `try_next` should fail if it is not.

## Context

RFC 7578 states the following in https://tools.ietf.org/html/rfc7578#section-4.2:

> Each part MUST contain a Content-Disposition header field [RFC2183]
   where the disposition type is "form-data".  The Content-Disposition
   header field MUST also contain an additional parameter of "name";

So modelling `Field::content_disposition` as an `Option` is actually too lenient.

* Actix Web Version: 3.0

craigpastro:
## PR Type

Bug Fix

## PR Checklist

## Overview

Content-Disposition header is required in every part of a multipart message.

I hope I am doing this correctly. Please give me your comments!

Closes #1947.

### DIFF
--- a/actix-http/src/header/common/content_disposition.rs
+++ b/actix-http/src/header/common/content_disposition.rs
@@ -72,7 +72,7 @@ impl<'a> From<&'a str> for DispositionType {
 #[derive(Clone, Debug, PartialEq)]
 #[allow(clippy::large_enum_variant)]
 pub enum DispositionParam {
-    /// For [`DispositionType::FormData`] (i.e. *multipart/form-data*), the name of an field from
+    /// For [`DispositionType::FormData`] (i.e. *multipart/form-data*), the name of a field from
     /// the form.
     Name(String),
     /// A plain file name.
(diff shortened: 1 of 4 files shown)


robjtede:
resumed work on this in #2451. thanks for getting it going

PR #2094, opened 2021-03-19
move typed headers and implement FromRequest

### LINKED ISSUE #1837: Implement FromRequest for header types
Currently, there are no extractors for HTTP headers, and application code has to extract the headers itself, which can be a little verbose, and can [annoy users](https://stackoverflow.com/questions/52919494/is-there-simpler-method-to-get-the-string-value-of-an-actix-web-http-header).

## Expected Behavior
```rust
#[post("mypost")]
async fn post_handler(
    req_body: web::Bytes,
    content_type: Option<http::header::ContentType>
) -> impl Responder {
    /* do something with the content type*/
    req_body
}
```

## Current Behavior
Compiling the code above raises the following error : 

```
the trait `actix_web::FromRequest` is not implemented for `actix_web::http::header::ContentType`
```

## Possible Solution

Maybe `FromRequest` could be implemented for the various types in `actix_web::http::header::*` ?

## Context

Sometimes an application has to handle a request differently based on an HTTP header. Having to parse the header in code is a slight friction that could be resolved. 

## Your Environment
* Actix Web Version: 3.3.2

ibraheemdev:
## PR Type

Refactor

## PR Checklist

## Overview

- Move common typed headers from `actix-http` into `actix-web`. 
- Add a `Header` extractor.
- Refactor `awc::TestResponse` to use `IntoHeaderPair`. This is needed now that the typed headers have been moved from `-http`.

Closes #1837.


src/types/header.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+impl<T> ops::DerefMut for Header<T> {
+    fn deref_mut(&mut self) -> &mut T {
+        &mut self.0
+    }
+}
+
+impl<T> fmt::Debug for Header<T>
```
robjtede, line 50: can be derived

src/types/header.rs
```diff
@@ -0,0 +1,113 @@
...
+    use crate::test::TestRequest;
+
+    #[actix_rt::test]
+    async fn test_header_extract() {
+        let (req, mut pl) = TestRequest::default()
+            .insert_header((header::CONTENT_TYPE, mime::APPLICATION_JSON))
+            .insert_header((header::ALLOW, header::Allow(vec![Method::GET])))
+            .to_http_parts();
+
+        let s = Header::<header::ContentType>::from_request(&req, &mut pl)
```
robjtede, line 98: nit: simplify tests a bit using `::extract(&req)`

awc/src/test.rs
```diff
@@ -58,6 +58,19 @@ impl TestResponse {
         panic!("Can not set header");
     }
 
+    // TODO: Should `set` be removed and replaced with this instead?
```
robjtede: Yes, please. Looking below, also header => append_header, with the same signature as this.

robjtede (commented):
low langing fruit

not delved in too deep yet, but are the parsing functions (from_one_raw_str etc) still part of -http?

ibraheemdev:
@robjtede I moved the parsing (`from_one_raw_str`) functions into -web as they were only used once in -http (for `ContentEncoding`). I can easily change it if you want though.

robjtede:
My thoughts are just that common parsing and formatting functions still belong in -http as long as their inputs or return types are std or HeaderValue or Vec\<HeaderValue\> etc.

src/types/header.rs
ibraheemdev: oh, I just copied that from `Json`

robjtede: fair, looks like many of these could be derived; don't mind if you leave this be then

ibraheemdev: I'll leave it for now and maybe create a separate PR fixing them.

actix-http/Cargo.toml
```diff
@@ -66,7 +66,6 @@ http = "0.2.2"
 httparse = "1.3"
 itoa = "0.4"
 language-tags = "0.2"
```
robjtede: i gather this isnt needed any more?

actix-http/src/header/mod.rs
```diff
@@ -41,17 +39,18 @@ pub trait Header: IntoHeaderValue {
     fn parse<T: HttpMessage>(msg: &T) -> Result<Self, ParseError>;
 }
 
+#[doc(hidden)]
 #[derive(Debug, Default)]
-pub(crate) struct Writer {
+pub struct Writer {
```
robjtede: needed in public API? if possible a local copy in web would be fine to reduce api surface

actix-http/src/header/shared/mod.rs
```diff
@@ -1,13 +1,15 @@
...
 mod charset;
+mod content_encoding;
 mod encoding;
 mod entity;
 mod extended;
 mod httpdate;
 mod quality_item;
 
 pub use self::charset::Charset;
+pub use self::content_encoding::ContentEncoding;
```
robjtede, line 10: whats the issue with ContentEncoding?

robjtede (changes requested):

actix-http/src/header/shared/mod.rs
ibraheemdev: It is exposed in -http's public api for the encoder and decoder.

actix-http/Cargo.toml
ibraheemdev: Its still being used by `Extended` , one of -http's shared headers, which I am not sure if I should move to web?

actix-http/src/header/mod.rs
ibraheemdev: It's actually only used once in -http for the `EntityTag` shared header. If that is moved to -web `Writer` can be removed from -http completely.

ibraheemdev:
So which headers from `shared` should be moved into -web?

robjtede:
> So which headers from `shared` should be moved into -web?

i get the impression entity tag is the only one that could move

actix-http/Cargo.toml
ibraheemdev: So, still being used then.

src/types/header.rs
ibraheemdev: so.... that doesn't work. You can't pass a `TestRequest` as a `HttpRequest` :)

robjtede (approved):
great. i think this is in a state to get merged after CI passes

ibraheemdev:
oh, there should also be a new changelog entry for awc.

lovasoa:
Great ! Thank you for implementing this !

PR #2097, opened 2021-03-20
added `TestServer::client_headers`

sfisol:
## PR Type
Feature (used in testing)

## PR Checklist

## Overview
TestServer methods `.ws()` and `.ws_at()` disallow any altering of request they perform.

~~This PR adds `.ws_at_with_config()` method that takes a _closure_ used for configuring the request.~~

This PR adds `TestServer::client_headers` method that exposes mutable headers of client used to perform requests. This allows to add custom headers.

For example, this allows to add authorization headers if WebSocket handler needs it.

Changes were added to both `actix-web/src/test.rs` and `actix-http-test/src/lib.rs` files

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+### Added
+* Added `TestServer::client_headers` method. [#2097]
+
 ### Fixed
 * Double ampersand in Logger format is escaped correctly. [#2067]
 
--- a/actix-http-test/src/lib.rs
+++ b/actix-http-test/src/lib.rs
@@ -13,7 +13,9 @@ use std::{net, thread, time};
 use actix_codec::{AsyncRead, AsyncWrite, Framed};
 use actix_rt::{net::TcpStream, System};
 use actix_server::{Server, ServiceFactory};
-use awc::{error::PayloadError, ws, Client, ClientRequest, ClientResponse, Connector};
+use awc::{
+    error::PayloadError, http::HeaderMap, ws, Client, ClientRequest, ClientResponse, Connector,
+};
 use bytes::Bytes;
 use futures_core::stream::Stream;
 use http::Method;
@@ -26,7 +28,7 @@ use socket2::{Domain, Protocol, Socket, Type};
 ///
 /// # Examples
 ///
-/// ```rust
+/// ```
 /// use actix_http::HttpService;
 /// use actix_http_test::TestServer;
 /// use actix_web::{web, App, HttpResponse, Error};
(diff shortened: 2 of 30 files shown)


actix-web-actors/tests/test_ws.rs
```diff
@@ -1,8 +1,9 @@
 use actix::prelude::*;
-use actix_web::{test, web, App, HttpRequest};
+use actix_web::{test, web, App, HttpRequest, HttpResponse};
 use actix_web_actors::*;
 use bytes::Bytes;
 use futures_util::{SinkExt, StreamExt};
+use http::StatusCode;
```
robjtede: status code is re-exported though actix_web::http::StatusCode and awc::http::StatusCode so extra dependency in cargo.toml is not needed

CHANGES.md
```diff
@@ -1,6 +1,8 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+* Added TestServer::ws_at_with_config method
```
robjtede: standard changelog format:

```suggestion
### Added
* Added `TestServer::ws_at_with_config` method. [#2097]

```

...with PR referenced as others are at end of version section. Similar change needed in other changelog.

actix-http-test/src/lib.rs
```diff
@@ -241,14 +241,27 @@ impl TestServer {
...
     }
 
+    /// Connect to WebSocket server at given path and configure WebsocketsRequest using provided function.
+    pub async fn ws_at_with_config<ConfFunc>(
+        &mut self,
+        path: &str,
+        config_func: ConfFunc,
+    ) -> Result<Framed<impl AsyncRead + AsyncWrite, ws::Codec>, awc::error::WsClientError>
+    where
+        ConfFunc: FnOnce(ws::WebsocketsRequest) -> ws::WebsocketsRequest,
```
robjtede: would prefer signature to receive `&mut WebsocketRequest` and return nothing

sfisol: Most methods on `WebsocketsRequest` like `.set_header` needs owned `self`.
Is it possible to perform such methods on `&mut` ?

robjtede: So they do, that's unfortunate. We'll leave this be for now.

fakeshadow:
It would be better to just expose headers field from `awc::Client` object.
`TestServer` object would be able to mutate it directly.

sfisol:
I rebased on #2114 and amended my changes to use mutable headers in TestServer's client.

robjtede (approved):

sfisol:
@robjtede I've amended your last commit because somehow HeaderMap was still not imported.

PR #2111, opened 2021-03-24
fix parsing uri from Location header value

### LINKED ISSUE #2102: awc: "Invalid URL: URL parse error: invalid format" error on some websites
## Expected Behavior
awc being able to successfully download the websites that work with curl and Firefox.

## Current Behavior
On some websites, e.g. http://viralnugget.com, awc fails with the following error:

> Invalid URL: URL parse error: invalid format

1620 out of the top million websites are affected. List of all affected websites: [invalid-format-domain-list.txt.gz](https://github.com/actix/actix-web/files/6179185/invalid-format-domain-list.txt.gz)

1. `git clone https://github.com/Shnatsel/rust-http-clients-smoke-test`
2. `cd rust-http-clients-smoke-test/awc`
3. `cargo build --release`
4. `target/release/awc-smoke-test viralnugget.com`

## Context
I am testing a bunch of HTTP clients on the top million websites according to the [Feb 3 Tranco list](https://tranco-list.eu/list/3G6L).

There is a very similar issue in another HTTP client: https://github.com/adamreichold/zeptohttpc/issues/6

## Your Environment
Linux, Ubuntu 20.04 LTS

* Actix Web Version: awc 3.0.0-beta.3

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Fix bug in awc redirect middleware where uri parse does not properly fall through to fall back parse when Location header value does not contain full length url.

Closes #2102

### DIFF
--- a/awc/src/middleware/redirect.rs
+++ b/awc/src/middleware/redirect.rs
@@ -257,28 +257,31 @@ where
 }
 
 fn rebuild_uri(res: &ClientResponse, org_uri: Uri) -> Result<Uri, SendRequestError> {
-    let uri = res
-        .headers()
+    res.headers()
         .get(header::LOCATION)
-        .map(|value| {
-            // try to parse the location to a full uri
-            let uri = Uri::try_from(value.as_bytes())
-                .map_err(|e| SendRequestError::Url(InvalidUrl::HttpError(e.into())))?;
+        // TODO: this error type is wrong.
+        .ok_or(SendRequestError::Url(InvalidUrl::MissingScheme))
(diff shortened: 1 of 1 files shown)


robjtede:
quick test case would be good so it's obvious what case this is fixing

fakeshadow:
It's just a simple fall through when Location returns "/abc" instead of "https://abc.com/abc".

There is also a following fix need this that have to introduce `url` crate to handle another layer of fallback. So any test would be better added later.

PR #2113, opened 2021-03-25
fix broken pipe for h2 when client is instantly dropped

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

```rust
    let mut res = awc::Client::new()
        .get("https://www.youtube.com/")
        .send()
        .await
        .unwrap();

    let body = res.body().await.unwrap();
```
This is a broken pipe error because `Client` object is dropped in place and h2 connection is dropped together with it.
This PR would try to send a end of stream in case like this and only force drop h2 connection when `SendRequest` object is broken.

But in general we should encourage reuse client object instead of drop it immediately. This would not cause error even without this patch

```rust
    let client = awc::Client::new();
    let mut res = client
        .get("https://www.youtube.com/")
        .send()
        .await
        .unwrap();

    let body = res.body().await.unwrap();
```

### DIFF
--- a/actix-http/src/client/connection.rs
+++ b/actix-http/src/client/connection.rs
@@ -174,7 +174,13 @@ impl H2ConnectionInner {
 /// Cancel spawned connection task on drop.
 impl Drop for H2ConnectionInner {
     fn drop(&mut self) {
-        self.handle.abort();
+        if self
+            .sender
+            .send_request(http::Request::new(()), true)
+            .is_err()
+        {
+            self.handle.abort();
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


robjtede:
if this is a problem in stable, then a changelog entry is appropriate

fakeshadow:
> if this is a problem in stable, then a changelog entry is appropriate

This is a fix for v4 beta I believe. stable would leak h2 connection and never end them. It does make sense to me now since by leaking the connection it would never trigger broken pipe.

fakeshadow:
sadly ci is too slow to pass test for this.

robjtede (approved):

PR #2114, opened 2021-03-25
add Client::headers

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Expose Client default headermap. This would help case like https://github.com/actix/actix-web/pull/2097 instead of adding specific APIs and offer a general way of mutating default headers at fly.


sfisol:
This should work for me. If so, then I will amend #2097

robjtede (approved):

fakeshadow:
> This should work for me. If so, then I will amend #2097

I believe client field is not public in TestServer so we would want an API to let you access the client default headers. But it should be relative easy to add.

PR #2117, opened 2021-03-26
Refactor actix_http::h1::service

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `H1ServiceResponse` type. This would reduce loc and simplify `<H1Service as ServiceFactory>::new_service` method.

### DIFF
--- a/actix-http/src/h1/service.rs
+++ b/actix-http/src/h1/service.rs
@@ -7,6 +7,7 @@ use actix_codec::{AsyncRead, AsyncWrite, Framed};
 use actix_rt::net::TcpStream;
 use actix_service::{pipeline_factory, IntoServiceFactory, Service, ServiceFactory};
 use futures_core::{future::LocalBoxFuture, ready};
+use futures_util::future::ready;
 
 use crate::body::MessageBody;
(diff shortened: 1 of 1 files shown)


robjtede:
maybe lets revert the async blocks here too

actix-http/src/h1/service.rs
```diff
@@ -417,47 +372,27 @@ where
...
-        if ready {
-            Poll::Ready(Ok(()))
-        } else {
-            Poll::Pending
-        }
+        ready!(self.flow.service.poll_ready(cx)).map_err(|e| {
+            let e = e.into();
+            log::error!("Http service readiness error: {:?}", e);
+            DispatchError::Service(e)
+        })?;
```
robjtede, line 393: old order: expect => service => upgrade  
new order: expect => upgrade => service

do you expect that to have any impact?

fakeshadow: I don't think it make any difference. All service type in HttpFlow must be ready(If upgrade service is Some) before accepting a request.

robjtede (approved):

PR #2126, opened 2021-03-29
merge H1ServiceHandler requests into HttpServiceHandler

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `actix_http::h1::service::H1ServiceHandler` type. This type share the same state with `actix_http::service::HttpServiceHandler` and the same poll_ready logic. A type alias for it is used to keep backward type compat.

Remove `actix_http::service::HttpServiceResponse` future type and use boxed future for init `HttpServiceHandler`

### DIFF
--- a/actix-http/src/h1/service.rs
+++ b/actix-http/src/h1/service.rs
@@ -6,7 +6,7 @@ use std::{fmt, net};
 use actix_codec::{AsyncRead, AsyncWrite, Framed};
 use actix_rt::net::TcpStream;
 use actix_service::{pipeline_factory, IntoServiceFactory, Service, ServiceFactory};
-use futures_core::{future::LocalBoxFuture, ready};
+use futures_core::future::LocalBoxFuture;
 use futures_util::future::ready;
 
 use crate::body::MessageBody;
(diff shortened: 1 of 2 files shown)


actix-http/src/h1/service.rs
```diff
@@ -315,47 +315,10 @@ where
...
-    ) -> H1ServiceHandler<T, S, B, X, U> {
-        H1ServiceHandler {
-            flow: HttpFlow::new(service, expect, upgrade),
-            cfg,
-            on_connect_ext,
-            _phantom: PhantomData,
-        }
-    }
-}
+pub type H1ServiceHandler<T, S, B, X, U> = HttpServiceHandler<T, S, B, X, U>;
```
otavio, line 318: As this is a breaking change, wouldn't be nice to drop this as well?

fakeshadow: I don't think this is a break change. The changes only effect internal logic and the dropped public type is only public because it's a leaf future exposed as future response. There is no export of either the types.

The type alias can indeed be dropped though. 

Correction: H1ServiceHandler is exported so I would like to keep this alias. So people import the type can still have back compat without changing it to HttpServiceHandler which could cause confusion

otavio: I meant the new release, in beta, will include breaking changes.

robjtede (commented):
nice, glad that service request type change is proving useful

robjtede (approved):

PR #2142, opened 2021-04-08
derive debug where possible

ibraheemdev:
## PR Type
Refactor

## PR Checklist

## Overview
Replace manual `Debug` impls with derives where possible

### DIFF
--- a/src/types/form.rs
+++ b/src/types/form.rs
@@ -80,7 +80,7 @@ use crate::{
 ///     })
 /// }
 /// ```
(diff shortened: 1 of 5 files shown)


ibraheemdev:
I'm not sure if debug impls like `Path` or `Query` that call `self.0.fmt(f)` in should be replaced with a derive, which would wrap `self.0` in the struct name. @robjtede ?

robjtede (approved):

robjtede:
wrapper names as part of the debug impl are useful. i'd say update the tests that care to match the new format

PR #2148, opened 2021-04-09
remove json and url encoded form support from -http

robjtede:
## PR Type
Dep Reduction

## PR Checklist

## Overview
Further reduce the number of dependencies on -http, making it simpler.

Removes some `Into<Error>` and `ResponseError` impls for each lib's error types. I think this makes sense anyway but need to look at the error system before v4 anyhow, it's possible the response error trait can be moved to -web, too, which would make it possible to re-impl for these types.

### DIFF
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -2,7 +2,6 @@
 
 use std::{
     cell::{Ref, RefMut},
-    convert::TryInto,
     fmt,
     future::Future,
     pin::Pin,
@@ -12,17 +11,13 @@ use std::{
 
 use bytes::{Bytes, BytesMut};
 use futures_core::Stream;
-use serde::Serialize;
 
 use crate::{
     body::{Body, BodyStream, MessageBody, ResponseBody},
     error::Error,
     extensions::Extensions,
     header::{IntoHeaderPair, IntoHeaderValue},
-    http::{
-        header::{self, HeaderName},
-        Error as HttpError, HeaderMap, StatusCode,
-    },
+    http::{header, Error as HttpError, HeaderMap, StatusCode},
     message::{BoxedResponseHead, ConnectionType, ResponseHead},
 };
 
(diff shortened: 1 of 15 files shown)


awc/src/sender.rs
```diff
@@ -209,7 +209,8 @@ impl RequestSender {
     ) -> SendClientRequest {
         let body = match serde_json::to_string(value) {
             Ok(body) => body,
-            Err(e) => return Error::from(e).into(),
+            // TODO: own error type
+            Err(e) => return Error::from(io::Error::new(io::ErrorKind::Other, e)).into(),
```
robjtede, line 213: Thoughts on what do do here @fakeshadow, maybe a wrapped error that can also hold serialization errors as well as client errors from -http. Might leave it for a separate PR if you want to handle it.

actix-http/src/response.rs
```diff
@@ -533,32 +454,10 @@ impl ResponseBuilder {
         self.body(Body::from_message(BodyStream::new(stream)))
     }
 
-    /// Set a json body and generate `Response`
-    ///
-    /// `ResponseBuilder` can not be used after this call.
-    pub fn json(&mut self, value: impl Serialize) -> Response {
```
popzxc, line 539: Won't it make sense to leave this method under an optional feature `json`? If the project aims to get rid of it completely, this feature can be declared deprecated from the start.

There may be code that relies on this method, thus removing it completely will make migration harder; I think it will be a bit more friendly to users to provide some way to start migrating without introducing way too many changes.

popzxc (approved):
In general LGTM.

actix-http/src/response.rs
robjtede: In terms of migration, it should be a non-issue for most: `actix_web::HttpResponse` was a re-export from `actix_http::Response`. Now actix-web has it's own HttpResponse object that keeps the json method.

popzxc: Ah, I see, it totally makes sense 👍

tjkirch:
Howdy!  Bit of feedback on this, particularly the error changes, since it wasn't clear to me how to operate without those helpers.  Up through beta.5, I was able to do things like `HttpResponse::BadRequest().body(err.to_string()).into()` in an `error_handler`, but now there would be more manual conversion required, since the `Into` is gone.  I hope that if the error changes stay in place for v4 final, there can be some wording in the docs/changelog about how to adapt existing code.  (I had to update from v3 to v4 beta to be able to update tokio.)

PR #2152, opened 2021-04-11
remove unpin from body types

robjtede:
## PR Type

Improvement

## PR Checklist

## Overview
- `impl<T: MessageBody> MessageBody for Pin<Box<T>>`
- The type parameter of `Response` no longer has a default.
- The `Message` variant of `body::Body` is now `Pin<Box<dyn MessageBody>>`.
- `BodyStream` and `SizedStream` are no longer restricted to Unpin types.


actix-http/src/lib.rs
```diff
@@ -6,7 +6,6 @@
 //! | `openssl`        | TLS support via [OpenSSL].                            |
 //! | `rustls`         | TLS support via [rustls].                             |
 //! | `compress`       | Payload compression support. (Deflate, Gzip & Brotli) |
-//! | `secure-cookies` | Adds for secure cookies. Enables `cookies` feature.   |
```
therealprof, line 9: Doesn't look like it belongs here?

robjtede: was missed from previous PR

therealprof:
Sweet!

Any indication of the performance impact? It sounds like the ability to use `mem::replace` might be a performance improvement but magic behind the scenes could negate that. 😅

actix-http/src/lib.rs
therealprof: I suspected so but I'm not a big fan of throwing random stuff into an otherwise clean PR. 😅

therealprof (approved):

robjtede:
> Any indication of the performance impact? It sounds like the ability to use mem::replace might be a performance improvement...

The main impact of this change is to accept more stream types and reduce boxes needed which is a small improvement.

therealprof:
> The main impact of this change is to accept more stream types and reduce boxes needed which is a small improvement.

Whether removing boxes is an improvement always depends on the data contained within and how it is used. A `Box` can also provide a performance improvement for large datastructures otherwise passed around by value.

I guess we will see the overall performance and memory use impact when we run it through the paces. 😅

CC @thalesfragoso

PR #2156, opened 2021-04-12
files: Don't use canonical path when serving file

### LINKED ISSUE #2128: actix-files::Files: symlinks are served with the filename of the original file.
## Expected Behavior

When serving a symbolic link with `actix-files::Files` service, the `content-disposition` header should be set to the filename of the symlink.

## Current Behavior

The filename of the original file is used.

## Possible Solution

Don't use the canonical path when serving a file?
https://github.com/actix/actix-web/blob/f66774e30b239546991fb6c0d54c8885171d1b8f/actix-files/src/service.rs#L86

## Context

Originally reported in svenstaro/miniserve#462

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.50.0
* Actix Web Version: 3
* actix-files version: 0.5

aliemjay:
## PR Type

Bug Fix

## PR Checklist

## Overview

Otherwise, for symlink files, the content-disposition header would be set to the filename of the pointed-to file.

Closes #2128

### DIFF
--- a/actix-files/src/service.rs
+++ b/actix-files/src/service.rs
@@ -83,10 +83,10 @@ impl Service<ServiceRequest> for FilesService {
             };
 
         // full file path
-        let path = match self.directory.join(&real_path).canonicalize() {
-            Ok(path) => path,
-            Err(err) => return Box::pin(self.handle_err(err, req)),
-        };
+        let path = self.directory.join(&real_path);
+        if let Err(err) = path.canonicalize() {
+            return Box::pin(self.handle_err(err, req));
+        }
 
         if path.is_dir() {
             if let Some(ref redir_index) = self.index {
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+* For symbolic links, `Content-Disposition` header no longer shows the filename of the original file. [#2156]
+
+[#2156]: https://github.com/actix/actix-web/pull/2156
 
 
 ## 0.6.0-beta.4 - 2021-04-02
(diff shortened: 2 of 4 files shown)


aliemjay:
Is it necessary to add a test case and a changelog entry for such change?

ibraheemdev:
Adding a test case for the bug fix is a good idea.

aliemjay:
Just added a test case. I don't know if a changelog entry is required?

robjtede:
> I don't know if a changelog entry is required?

Yes please.

actix-files/src/service.rs
```diff
@@ -83,10 +83,10 @@ impl Service<ServiceRequest> for FilesService {
...
 
         // full file path
-        let path = match self.directory.join(&real_path).canonicalize() {
-            Ok(path) => path,
-            Err(err) => return Box::pin(self.handle_err(err, req)),
-        };
+        let path = self.directory.join(&real_path);
+        if let Err(err) = path.canonicalize() {
+            return Box::pin(self.handle_err(err, req));
+        }
```
robjtede, line 89: My concern about this is that we probably want to normalize the path in the way that canonicalize does just without following symlinks. Have you done any testing with paths that includes '../' and such? I'm not sure we have test coverage for it.

aliemjay: I think `real_path` is already normalized by `PathBufWrap::parse_path()` and `self.directory` is also a canonical path (see `Files::new()`) , so I think this simple join should pose no problem.

I've just tested it with '../' in both url path and `self.directory`. If you wish I can modify the testcase for this?

Alternatively, we can keep the old code and use `NamedFile::from_file()` to pass the correct filename.

robjtede: It's okay then, I don't see any immediate problems with the change made.

robjtede (approved):

PR #2160, opened 2021-04-13
`Path`: make content field `pub` again

abonander:
## PR Type

Refactor

## PR Checklist

## Overview
Upgrading to the beta and this seems to have changed from `3.3.5` but no reason is given in #1894 for the change. The content field of `Json` is still public which confuses me as to why `Path` and `Query` got this treatment but `Json` did not.

I recently switched from calling `.into_inner()` to destructuring in the function arguments and it's a lot nicer, especially for singular path arguments such as:

```rust
#[get("/user/{id}")]
async fn get_user(Path(user_id): Path<Uuid>) -> Option<User> {
    // ...
}
```

I really don't want to have to go back to writing `.into_inner()` again. It's just noise.

### DIFF
--- a/src/types/path.rs
+++ b/src/types/path.rs
@@ -46,7 +46,7 @@ use crate::{dev::Payload, error::PathError, FromRequest, HttpRequest};
 /// }
 /// ```
 #[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
-pub struct Path<T>(T);
+pub struct Path<T>(pub T);
 
 impl<T> Path<T> {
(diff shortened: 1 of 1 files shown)


abonander:
@robjtede I see the discussion in #2016 but I don't think making the field private was necessary, it can just be a style suggestion in the docs to not access tuple fields directly. Paths with multiple parameters should probably switch to using structs anyway.

robjtede:
Sorry I'm not considering this change at this time. The discussion in #2016 remains true.

PR #2162, opened 2021-04-14
Adjust JSON limit to 2MB and report on sizes

peter-formlogic:
## PR Type

Other

## PR Checklist

## Overview

This adjusts the default limit of JSON payloads to 2mb from 32kb and adjusts the error to include what the limits are when they're hit.  As this changes the error enum & has possibly knock on effects to memory usage by allowing more through this is probably considered a breaking change.

To justify the increase, I have looked at other defaults in other software for what their limits are:

*  in .Net it's [set to 2097152](https://docs.microsoft.com/en-us/dotnet/api/system.web.script.serialization.javascriptserializer.maxjsonlength?view=netframework-4.8) (which they say is 4mb? not sure why)
* In Tomcat [the max post size is set to 2mb](https://tomcat.apache.org/tomcat-8.5-doc/config/http.html)
* In PHP the [default upload size is 2mb](https://www.php.net/manual/en/ini.core.php#ini.sect.file-uploads)

(*Honestly, I'd much prefer having no limit at all by default as these things seem to have a way of coming back to bite at unrealistic times, like 2am on a saturday morning!*)


fakeshadow:
I think it's fine to relax the limit in general. We need a changelog for limit and error changes

peter-formlogic:
Done!

src/error.rs
```diff
@@ -86,9 +86,13 @@ impl ResponseError for UrlencodedError {
...
 #[non_exhaustive]
 pub enum JsonPayloadError {
-    /// Payload size is bigger than allowed. (default: 32kB)
-    #[display(fmt = "Json payload size is bigger than allowed")]
-    Overflow,
+    /// Payload size is bigger than allowed. (default: 2MB)
+    #[display(
+        fmt = "JSON payload ({} bytes) is larger than allowed (limit: {} bytes).",
+        size,
+        limit
```
robjtede: I suspect trying to report the size would be confusing because when read chunk-by-chunk it's going to get triggered as soon as the limit is exceeded. We don't necessarily know the final payload size.

peter-formlogic: Yep, so the [behaviour here](https://github.com/actix/actix-web/pull/2162/files#diff-7b229864a63b2f9e26f81d52e49931eb3f81f995a3ddf547a03db3800302f8a4R416-R418) is to show the content length header value if present, if not it will be the chunk size that tilts it over the edge.

I tried to make this field an optional but had issues trying to format the error in a nice way, since we don't impose a requirement on content length being present.

peter-formlogic: I've just broken out the enums so that there are two scenarios, one where there is a known length (via the content length header) and one when there isn't.

thalesfragoso:
Thanks for the PR, is this just waiting for the conflicts to be resolved ?

peter-formlogic:
@thalesfragoso sorry didn't notice that it was blocked on conflicts.  I've merged master into this PR

robjtede (approved):

PR #2172, opened 2021-04-19
accept owned strings in TestRequest::param

### LINKED ISSUE #1921: Make TestRequest#param not use static lifetimes.
This is the current method definition:
```rust
pub fn param(mut self, name: &'static str, value: &'static str) -> Self
```

https://docs.rs/actix-web/3.3.2/actix_web/test/struct.TestRequest.html#method.param

## Expected Behavior

Make the param method parameters (or at least the value parameter) not need a static lifetime.

One may want to use a non-static parameter in the current situation:

Your test makes a request that gets x info, you then want to make another request using that info (e.g. the id) as a parameter, but it won't have a static lifetime.

## Current Behavior

The current behavior is that the param method parameters have a static lifetime.

## Possible Solution

Remove the static lifetime.

Example:
```rust
    #[actix_rt::test]
    async fn test_app() {
        let mut app = test::init_service(
            App::new()
                .route("/", web::post().to(add_func))
                .route("/{id}", web::get().to(get_func)),
        )
        .await;

        let req = test::TestRequest::post().to_request();
        let resp = test::call_service(&mut app, req).await;
        assert!(resp.status().is_success());

        let body = test::read_body(resp).await;
        let id: &str = std::str::from_utf8(&body).unwrap();

        // Won't work, id doesn't live long enough.
        let req = test::TestRequest::get().param("id", id).to_request();
        let resp = test::call_service(&mut app, req).await;
        assert!(resp.status().is_success());
    }
```

## Context

I think I already explained a real world issue, so I'll repeat it here:

Your test makes a request that gets x info, you then want to make another request using that info (e.g. the id) as a parameter, but it won't have a static lifetime.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.49.0 (e1884a8e3 2020-12-29)
* Actix Web Version: 3.3.2

ibraheemdev:
## PR Type
Feature

## PR Checklist

## Overview

Blocked upstream on actix/actix-net#345
Closes #1921


robjtede:
needs rebase, doc test addition of `&'static str` and `String` examples and changelog

MavenRain (approved):

MavenRain:
I just realized that this PR would unblock one on which I am currently working, which involves refactoring a unit test to use params calculated within the existing test.

robjtede (approved):

ibraheemdev:
@robjtede thanks for picking this up, sorry I've been a bit busy lately.

PR #2176, opened 2021-04-19
move logical error handling to -web 

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
The rationale is that -http should be concerned with Request -> Response only (except it's own errors) and delegate higher level error handling to user or downstream lib.

Therefore, -http's service builders and dispatchers now expect `Service::Error` types to be `Into<Response<Body>>` instead of `Into<Error>` (which `Error` is). These are almost the same from perspective of this lib.

The upside is that it will reduce API surface of -http by moving Error struct and ResponseError trait to -web, where more ergonomic errors can be produced. As seen in the fallout from -web beta.6, there are many people who implement ResponseError trait and requiring it to build a -http Response was too burdensome.

### DIFF
--- a/actix-http/examples/hello-world.rs
+++ b/actix-http/examples/hello-world.rs
@@ -1,6 +1,6 @@
 use std::{env, io};
 
-use actix_http::{http::StatusCode, HttpService, Response};
+use actix_http::{http::StatusCode, Error, HttpService, Response};
 use actix_server::Server;
 use actix_utils::future;
 use http::header::HeaderValue;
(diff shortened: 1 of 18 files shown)


therealprof:
A definite big thumbs up from me!

robjtede:
This is definitely a tricky change and is requiring a large refactor or errors in general. We're so tied into `actix_http::Error` being the sole error type and it's making this very painful. To ignore this set of changes would be harmful and would require reverting a lot of progress in order to have a sane API, namely the split out `HttpResponse` types used in -web.

therealprof:
I agree it would be painful once, but at the moment it's frequently painful because it also ties versions of crates very tightly together so every time some version of some crate moves it's a timeconsuming puzzle to determine a set of compatible versions.

This can be seen with the betas at the moment: I'm spending 1-2 hours per week pinning/unpinning crates and solving (sometimes unsolvable) dependency puzzles just so we can stay up to date with the latest improvements and do some proper testing.

Decoupled dependencies is certainly a very high up topic on my wishlist...

PR #2183, opened 2021-04-21
add associated error type to MessageBody

robjtede:
## PR Type

Feature

## PR Checklist

## Overview
Step 1. Split out from the error remodelling, MessageBody was a large part of the pain because the error type was assumed to be `actix_http::Error` all the time.


actix-http/src/body/body.rs
```diff
@@ -25,7 +26,7 @@ pub enum Body {
     Bytes(Bytes),
 
     /// Generic message body.
-    Message(Pin<Box<dyn MessageBody>>),
+    Message(BoxAnyBody),
```
robjtede, line 29: `Pin<Box<dyn MessageBody<Error = Error>>>` is the alternative and locks in that error type. This way, converting to and from StdError is shown to be viable.

actix-http/src/body/message_body.rs
```diff
@@ -7,75 +7,99 @@ use std::{
...
     /// Body size hint.
     fn size(&self) -> BodySize;
 
     /// Attempt to pull out the next chunk of body bytes.
     fn poll_next(
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
-    ) -> Poll<Option<Result<Bytes, Error>>>;
-
-    downcast_get_type_id!();
```
robjtede, line 26: With some change to the macros (or inlining) it's possible to write `impl<E: 'static> dyn MessageBody<Error = E> + 'static` but i removed this at some point during writing this to fix an error.

I think it is possible to just use `std::any::Any` trait for any downcasting needs anyway (see altered test).

actix-http/src/service.rs
```diff
@@ -59,6 +59,7 @@ where
     S::Response: Into<Response<B>> + 'static,
     <S::Service as Service<Request>>::Future: 'static,
     B: MessageBody + 'static,
+    B::Error: Into<Error>,
```
robjtede, line 62: having bounds look like this (for now) is a pretty well compatible change since trivially: `Error: Into<Error>`

fakeshadow:
LGTM. 

What is wrong with concrete to `actix_http::Error` though. It's just a trait object that help generating a Response in dispatcher.

fakeshadow (approved):

robjtede:
> What is wrong with concrete to actix_http::Error.

We need to move that type to -web along with `ResponseError` trait in order to make it so `error_response()` can return a `HttpResponse` instead of `Response`. Then introduce a new error type in -http that's not relying on `ResponseError` and mostly change these bounds to `B::Error: Into<Response<Body>>`.

It's an all or nothing situation tbh, if this isn't done then the `HttpResponse` changes will need to be reverted.

PR #2186, opened 2021-04-22
Static middleware futures

ibraheemdev:
## PR Type
Refactor

## PR Checklist

## Overview
Replace boxed future with a custom future type in `Compat`, `Condition`, and `ErrHandler` middlewares. All middlewares now use static future types.

### DIFF
--- a/src/middleware/err_handlers.rs
+++ b/src/middleware/err_handlers.rs
@@ -9,7 +9,10 @@ use std::{
 
 use actix_service::{Service, Transform};
 use ahash::AHashMap;
-use futures_core::{future::LocalBoxFuture, ready};
+use futures_util::{
+    future::{ok, LocalBoxFuture, Ready},
+    ready,
+};
 
 use crate::{
     dev::{ServiceRequest, ServiceResponse},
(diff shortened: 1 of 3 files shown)


src/middleware/err_handlers.rs
```diff
@@ -9,7 +9,10 @@ use std::{
 
 use actix_service::{Service, Transform};
 use ahash::AHashMap;
-use futures_core::{future::LocalBoxFuture, ready};
+use futures_util::{
+    future::{ok, LocalBoxFuture, Ready},
```
robjtede, line 13: Ready futures must come from actix-utils

robjtede (changes requested):
It's a bit unnecessary to have custom futures for the transforms given they're only called once. I think there's value in the simpler code in such cases.

ibraheemdev:
Oh, yeah, I guess. You can close it if you think it's unnecessary.

robjtede:
The err_handlers change is okay but we've put pin boxes in a bunch of other places including AppService ServiceFactory impl in order to reduce complexity in single-call situations.

fakeshadow:
I personally would prefer `LocalBoxFuture` over `Ready` for every one time cost future.
Eventually it can be replaced with `impl Future`

PR #2190, opened 2021-04-27
set timeout value for connection

klim-iv:
## PR Type

PR_TYPE

## PR Checklist

### DIFF
--- a/awc/src/builder.rs
+++ b/awc/src/builder.rs
@@ -266,7 +266,10 @@ where
         };
         if let Some(val) = self.local_address {
             connector = connector.local_address(val);
-        }
+        };
+        if let Some(val) = self.timeout {
+            connector = connector.timeout(val);
+        };
 
         let connector = DefaultConnector::new(connector.finish());
         let connector = boxed::rc_service(self.middleware.new_transform(connector));


fakeshadow:
Connector's timeout is used for the first establish of a tcp connection (include tls handshake).

awc's timeout is used for actual http request timeout using the tcp connection connector gives you. They have different meaning and you should not mix them as a single timeout.

If you want to modify the connector's timeout setting you should construct a Connector and pass it to awc's Builder with this API:
https://github.com/actix/actix-web/blob/6a29a50f25fd103c61e93d97dbee5745640bb911/awc/src/builder.rs#L69
You can set a single duration if you want to but you should aware that they are different timeout for different usage. (Normally you don't want a large duration for first establish of connection so give a relative lower timeout to connector is prefered.)

klim-iv:
> Connector's timeout is used for the first establish of a tcp connection (include tls handshake).
> awc's timeout is used for actual http request timeout using the tcp connection connector gives you. They have different meaning and you should not mix them as a single timeout.

yes, I understand. In my case, dns-resolver is working slow , 1s is not enough for **connect** timeout (may be need to increase default value?)

> If you want to modify the connector's timeout setting you should construct a Connector and pass it to awc's Builder with this API:
> 
> https://github.com/actix/actix-web/blob/6a29a50f25fd103c61e93d97dbee5745640bb911/awc/src/builder.rs#L69
> 
> You can set a single duration if you want to but you should aware that they are different timeout for different usage. 

Thanks, I'll try, but maybe better to little bit increase connect-timeout in default values? (1s-> 3s)

fakeshadow:
> Thanks, I'll try, but maybe better to little bit increase connect-timeout in default values? (1s-> 3s)

The default tcp connection and tls handshake timeout all has been relaxed to 5 seconds default in recent releases. You would get the relaxed timeout from 3.0.0-beta.2 and later versions.

klim-iv:
> > Thanks, I'll try, but maybe better to little bit increase connect-timeout in default values? (1s-> 3s)
> 
> The default tcp connection and tls handshake timeout all has been relaxed to 5 seconds default in recent releases. You would get the relaxed timeout from 3.0.0-beta.2 and later versions.

Oh, I see - fresh commit :-)
Thank you!

PR #2201, opened 2021-05-07
remove responsebody indirection from response

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
The `ResponseBody` abstraction is a kind of annoying way to get around the problem of `B` vs `Body` in dispatcher-land as well as in fallible functions. This PR attempts to remove this indirection by providing different paths for error responses and hopefully provides a more ergonomic interface especially in the area of `map_body`.

## Further Work
- Remove ResponseBody from encoding mod.
- I suspect some issues would surface in user code due to removing the old `into_body` method(s) but I have an idea for it. (We might see it occur in -extras.)

### DIFF
--- a/actix-http/src/body/mod.rs
+++ b/actix-http/src/body/mod.rs
@@ -86,16 +86,6 @@ mod tests {
         }
     }
 
-    impl ResponseBody<Body> {
-        #[allow(dead_code)]
-        pub(crate) fn get_ref(&self) -> &[u8] {
-            match *self {
-                ResponseBody::Body(ref b) => b.get_ref(),
-                ResponseBody::Other(ref b) => b.get_ref(),
-            }
-        }
-    }
-
     #[actix_rt::test]
     async fn test_static_str() {
         assert_eq!(Body::from("").size(), BodySize::Sized(0));
--- a/actix-http/src/body/mod.rs
+++ b/actix-http/src/body/mod.rs
@@ -87,6 +87,7 @@ mod tests {
     }
 
     impl ResponseBody<Body> {
+        #[allow(dead_code)]
         pub(crate) fn get_ref(&self) -> &[u8] {
             match *self {
                 ResponseBody::Body(ref b) => b.get_ref(),
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -296,11 +296,11 @@ where
         io.poll_flush(cx)
     }
 
-    fn send_response(
+    fn send_response_inner(
         self: Pin<&mut Self>,
         message: Response<()>,
-        body: B,
-    ) -> Result<(), DispatchError> {
+        body: &impl MessageBody,
+    ) -> Result<BodySize, DispatchError> {
         let size = body.size();
         let mut this = self.project();
         this.codec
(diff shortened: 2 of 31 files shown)


actix-http/src/body/mod.rs
```diff
@@ -87,6 +87,7 @@ mod tests {
     }
 
     impl ResponseBody<Body> {
+        #[allow(dead_code)]
```
popzxc: Wouldn't it be better to leave a comment on why is it kept? Otherwise, it has chances to either be removed earlier than needed or not removed when it's time.

actix-http/src/h1/dispatcher.rs
```diff
@@ -413,7 +438,42 @@ where
...
+                                this.codec
+                                    .encode(Message::Chunk(None), &mut this.write_buf)?;
+                                // payload stream finished.
+                                // set state to None and handle next message
+                                this.state.set(State::None);
+                                continue 'res;
+                            }
+
+                            Poll::Ready(Some(Err(err))) => {
+                                return Err(DispatchError::Service(err.into()))
```
popzxc: It would be nice to take care of `clippy` warnings.

actix-http/src/response.rs
```diff
@@ -83,162 +79,142 @@ impl Response<Body> {
...
     pub fn with_body(status: StatusCode, body: B) -> Response<B> {
         Response {
             head: BoxedResponseHead::new(status),
-            body: ResponseBody::Body(body),
+            body: body,
             error: None,
         }
     }
 
+    /// Return a reference to the head of this response.
```
popzxc: `Returns` (here and in a couple of methods below)?

actix-http/src/response_builder.rs
```diff
@@ -235,24 +236,27 @@ impl ResponseBuilder {
     /// This `ResponseBuilder` will be left in a useless state.
     #[inline]
     pub fn body<B: Into<Body>>(&mut self, body: B) -> Response<Body> {
-        self.message_body(body.into())
+        match self.message_body(body.into()) {
```
popzxc: `self.message_body(body.into()).unwrap_or_else(Response::from_error)` ?

src/test.rs
```diff
@@ -153,16 +154,17 @@ where
...
 
-    let mut body = resp.take_body();
+    let body = resp.into_body();
     let mut bytes = BytesMut::new();
 
-    while let Some(item) = body.next().await {
-        bytes.extend_from_slice(&item.unwrap());
+    actix_rt::pin!(body);
+    while let Some(item) = poll_fn(|cx| body.as_mut().poll_next(cx)).await {
+        bytes.extend_from_slice(&item.map_err(Into::into).unwrap());
```
popzxc, line 167: Cant `map_err(Into::into).unwrap()` be replaced with `unwrap_or_else(Into::into)` (here and in a method below)?

popzxc (approved):
Aside from minor nitpicks, LGTM.

src/test.rs
robjtede: nah, the panic behavior from before should stay but we need to convert the `Into<Error>` type into `Error` since we're not declaring a Debug bound necessary for unwrap.

PR #2214, opened 2021-05-13
Response::from_error take impl Into<Error>

fakeshadow:
## PR Type

PR_TYPE

## PR Checklist

## Overview

Response::from_error take generic type so this would save all the manual into.


JohnTitor (approved):
Nice improvement, LGTM.

PR #2215, opened 2021-05-13
move internalerror to actix web

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Removes `InternalError` from actix-http; a rational step in the error migration.

Aliases `Body` to `AnyBody`. It's a more appropriate name for this enum and idea being that in v5 the `MessageBody` trait could be renamed to simply `Body`.


Thomasdezeeuw:
I have a question about this pr, specifically: https://github.com/actix/actix-web/commit/2a8c650f2c4ccc5cafcfba01e85109a0388e604f#diff-9bce557d772a63c48169e7eaa216c68164495d524c6d4df4c8467572aa3ec271L125-L129. We used that `From<Response<Body>> for Error` impl in things like `QueryConfig::error_handler` to return our own JSON responses, rather then the default (non-JSON) response. Is there still a way to create `actix_web::error::Error` from a `HttpResponse`, or control the returned response?

robjtede:
@Thomasdezeeuw sure, using [this method](https://docs.rs/actix-web/3.3.2/actix_web/error/struct.InternalError.html#method.from_response):

```rust
let res = HttpResponse::Ok().finish();
let err: Error = InternalError::from_response("", res).into();
```

Thomasdezeeuw:
Thanks @robjtede that worked.

PR #2216, opened 2021-05-13
Route handler macro accepts idents for path

realaravinth:
## PR Type

Feature

## PR Checklist

## Overview

This PR enables route handler macros(`actix_web::get`, etc.) to accept variable names for path. It does so by adding a `path` attribute to the macro.

- This is possible:
```rust
const PATH: &str = "/path";
#[get(path = "PATH")]
async fn path() -> impl Responder {
    HttpResponse::Ok()
}
```
- and when both `path` and literal string is used, compilation fails with the following error:
```rust
error: Multiple paths specified! Should be only one!
  --> $DIR/simple-fail.rs:32:20
   |
32 | #[get("/six", path="SIX")]
   |                    ^^^^^
```


robjtede:
At first glance, `path = "{IDENT}"` is kinda confusing to look at imo. One could reasonably try to write `path = "/index"` and expect that to work.

What's the motivation for wanting this?

realaravinth:
I understand I found using `const` for denoting paths improved maintainability. I could use the same variable in to mark handlers, in tests and in templates.

---
While working on [mCaptcha](https://github.com/mCaptcha/guard), I was seeing way too many 404s simoly becaise dumb me couldn't type properly😂

popzxc:
What do you guys think about the following syntax? 

```rust
const PATH: &str = "/path";
#[get(path = "{PATH}")]
async fn path() -> impl Responder {
    HttpResponse::Ok()
}
```

On the one hand, it looks consistent with the [implicit format arguments RFC](https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html), on the other hand, it explicitly states that we are not specifying the actual path, but rather referring to some variable.

realaravinth:
> What do you guys think about the following syntax?
> 
> ```rust
> const PATH: &str = "/path";
> #[get(path = "{PATH}")]
> async fn path() -> impl Responder {
>     HttpResponse::Ok()
> }
> ```
> 
> On the one hand, it looks consistent with the [implicit format arguments RFC](https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html), on the other hand, it explicitly states that we are not specifying the actual path, but rather referring to some variable.

Thanks for the suggestion, my latest commits add support for expressions as well. If you had a `struct` describing all the routes and an instance of it at the crate root, you could do something like:

```rust
#[get(path="crate::ROUTE.auth.login")]
async too() -> impl Responder{
// <---snip--->
}
```

If I understood implicit arguments correctly, they are used to capture stuff from same scope, isn't?

robjtede:
I don't think the use case for this feature is strong enough to implement this in the main codegen lib. Unless there's some signals from others, I'm going to close this in a week or so.

PR #2225, opened 2021-05-19
files: Fix `redirect_to_slash_directory()` when used with `show_files_listing()`

aliemjay:
## PR Type

Bug fix

## PR Checklist

- [x] Documentation comments have been added / updated (There is none for such change).

## Overview

`Files::redirect_to_slash_directory()` currently redirects only when an index file is set with `.index_file(_)`. However, the rationale behind this feature (see #1134) is also valid when showing directory listings instead of index files. 

I considered this a bug because the current behavior is totally not expected according to the documents.


aliemjay:
CI tests failed consistently on nightly due to a seemingly unrelated reason

JohnTitor (approved):
Makes sense for me, LGTM.

robjtede (approved):

PR #2226, opened 2021-05-19
Extends Rustls ALPN protocols instead of replacing them when creating Rustls based services

### LINKED ISSUE #1631: rustls ALPN protocols are ignored
## Expected Behavior
- When creating a `rustls` config:
  `let tls_config = ServerConfig::new(NoClientAuth::new())`
- When setting ALPN protocols:
  `tls_config.set_protocols(&["h2".to_string().into(), "http/1.1".to_string().into(), "acme-tls/1".to_string().into()])`
- And when binding with `rustls`:
  `HttpServer::new(...).bind_rustls("0.0.0.0:443", tls_config)`
- I expect my server to handle the ALPN protocols `h2`, `http/1.1` and `acme-tls/1`

## Current Behavior
`bind_rustls` override the ALPN protocols to `h2` and `http/1.1`. Thus, I loose the `acme-tls/1` ALPN protocol and can't handle the `TLS-ALPN-01` ACME challenge.

## Possible Solution
The problem is here: https://github.com/actix/actix-web/blob/5aad8e24c74beb4c5c2b0df118a67f5005b5cc23/actix-http/src/service.rs#L294-L295

I'm new to `rust` and `actix` so I don't see a correct way to handle this yet, but we may allow user to provide custom protocols? Or we may delegate the fix to `rustls` to let user append protocols instead of overwriting them? (https://github.com/ctz/rustls/blob/a5e148825b807202db7b1ce352ac33517c5a1bb0/rustls/src/server/mod.rs#L306-L308)

## Context

I was trying to handle the `TLS-ALPN-01` ACME challenge with `actix-web` and `rustls`. To do so I've implemented a custom `ResolvesServerCert`. But to be valid, the `ServerHello` responding to the `ClientHello` coming from the ACME server needs to have an ALPN protocol set to `acme-tls/1`.

I've tested to modify the `actix` sources and manually add the `acme-tls/1` ALPN protocol and it works!

## Your Environment

* Rust Version: `rustc 1.45.0 (5c1f21c3b 2020-07-13)`
* Actix Web Version: `3.0.0-beta.1`

01101101:
## PR Type
Feature

## PR Checklist

## Overview
When creating a Rustls based services, the ALPN protocols defined in the Rustls config are overwritten.
This PR changes this behavior, and instead of replacing the Rustls config ALPN protocols, it extends them.
Thus, a user can defines some custom ALPN protocols without them being replaced when creating Rustls based services.

An example use-case is when a user wants to implement the `TLS-ALPN-01` ACME challenge. To do so, the user server has to handle the `acme-tls/1` ALPN protocol.

Closes #1631


actix-http/tests/test_rustls.rs
```diff
@@ -460,3 +460,107 @@ async fn test_h1_service_error() {
...
+    assert_eq!(
+        srv.get_negotiated_alpn_protocol(CUSTOM_ALPN_PROTOCOL),
+        Some(CUSTOM_ALPN_PROTOCOL.to_vec())
+    );
+
+    let response = srv.sget("/").send().await.unwrap();
+    assert!(response.status().is_success());
+
+    Ok(())
+}
```
thalesfragoso: Should this test also be testing for the h1.1 ?

thalesfragoso (commented):
Thanks, I think this change makes sense, I just left a comment in a test case.

actix-http/tests/test_rustls.rs
01101101: The client and server only handle `custom` ALPN protocol. If we test for the `http/1.1` ALPN protocol there will be no overlap, so the test will fails.

Note: Rustls will not fail the handshake if there is no overlap, it may be a violation of the RFC7301. Go seems to have the same kind of problems to solve https://github.com/golang/go/issues/46310

thalesfragoso: What is the difference between this test and `test_alpn_h1()` ?

actix-http/src/service.rs
```diff
@@ -305,7 +305,9 @@ mod rustls {
             Error = TlsError<io::Error, DispatchError>,
             InitError = (),
         > {
-            let protos = vec!["h2".to_string().into(), "http/1.1".to_string().into()];
+            let mut protos =
+                vec!["h2".to_string().into(), "http/1.1".to_string().into()];
```
aliemjay:
```suggestion
            let mut protos = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
```

actix-http-test/src/lib.rs
```diff
@@ -223,6 +232,24 @@ impl TestServer {
         self.client.request(method, path.as_ref())
     }
 
+    #[cfg(feature = "rustls")]
+    /// Get the negotiated ALPN protocol with the server
+    pub fn get_negotiated_alpn_protocol(&self, client_alpn_protocol: &[u8]) -> Option<Vec<u8>> {
```
aliemjay: I think this doesn't belong here. `TestServer` is not concerned with such low-level specific use case. I think it is better to move it to the test module as a helper function.

actix-http/src/h2/service.rs
```diff
@@ -171,7 +171,8 @@ mod rustls {
             Error = TlsError<io::Error, DispatchError>,
             InitError = S::InitError,
         > {
-            let protos = vec!["h2".to_string().into()];
+            let mut protos = vec!["h2".to_string().into()];
```
aliemjay:
```suggestion
            let mut protos = vec![b"h2".to_vec()];
```

CHANGES.md
```diff
@@ -3,13 +3,15 @@
...
 * `HttpServer::worker_max_blocking_threads` for setting block thread pool. [#2200]
+* Added `TestServer::get_negotiated_alpn_protocol` method.
 
 ### Changed
 * `ServiceResponse::error_response` now uses body type of `Body`. [#2201]
 * `ServiceResponse::checked_expr` now returns a `Result`. [#2201]
 * Update `language-tags` to `0.3`.
 * `ServiceResponse::take_body`. [#2201]
 * `ServiceResponse::map_body` closure receives and returns `B` instead of `ResponseBody<B>` types. [#2201]
+* Extends Rustls ALPN protocols instead of replacing them when creating Rustls based services
```
aliemjay:
```suggestion
* `HttpServer::{listen_rustls(), bind_rustls()}` now honor the ALPN protocols in the configuation parameter. [#2226]
```

Also, their docs needs an update to reflect this and, preferably, with an example on `acme-tls/1` as I seen no other use case for this.

actix-http/CHANGES.md
```diff
@@ -19,6 +19,7 @@
 * Update `language-tags` to `0.3`.
 * Reduce the level from `error` to `debug` for the log line that is emitted when a `500 Internal Server Error` is built using `HttpResponse::from_error`. [#2201]
 * `ResponseBuilder::message_body` now returns a `Result`. [#2201]
+* Extends Rustls ALPN protocols instead of replacing them when creating Rustls based services
```
aliemjay: You may want to mention `HttpService::rustls()` and link to the PR

aliemjay (changes requested):
Thanks for the PR! 

This makes sense to me overall. Here is my review...

aliemjay:
Does openssl have the same problem?

actix-http/tests/test_rustls.rs
01101101: I blindly copy-pasted the HTTP tests above, so yes `test_alpn_h1` and `test_alpn_h1_1` are the same, I will remove one of them.

01101101:
> Does openssl have the same problem?

I don't know, I didn't try with OpenSSL. But looking at the code it didn't seems to have the same problem.

aliemjay (approved):
Looks good now, thanks.

Could you update  the docs for `HttpServer::{listen_rustls, bind_rustls}`? They currently state:
> This method sets alpn protocols to “h2” and “http/1.1”

aliemjay (approved):

robjtede (approved):

robjtede:
lgtm

would be interested to see an implementation of an ACME acceptable server

actix-http/CHANGES.md
```diff
@@ -19,6 +19,7 @@
 * Update `language-tags` to `0.3`.
 * Reduce the level from `error` to `debug` for the log line that is emitted when a `500 Internal Server Error` is built using `HttpResponse::from_error`. [#2201]
 * `ResponseBuilder::message_body` now returns a `Result`. [#2201]
+* `HttpServer::{listen_rustls(), bind_rustls()}` now honor the ALPN protocols in the configuation parameter. [#2226]
```
pickfire, line 22: Is the `()` a typo?

PR #2228, opened 2021-05-20
files: allow `show_files_listing()` with `index_file()`

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

Currently, `Files::show_files_listing()` has no effect when `index_file(_)` is set - if the index file is not found, an error response is always returned regardless. 

This PR changes this behavior to allow showing file listing when an index file is not found

### DIFF
--- a/actix-files/src/service.rs
+++ b/actix-files/src/service.rs
@@ -129,6 +129,7 @@ impl Service<ServiceRequest> for FilesService {
             match self.index {
                 Some(ref index) => match NamedFile::open(path.join(index)) {
                     Ok(named_file) => serve_named_file(req, named_file),
+                    Err(_) if self.show_index => show_index(req),
                     Err(err) => self.handle_err(err, req),
                 },
                 None if self.show_index => show_index(req),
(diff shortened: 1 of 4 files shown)


aliemjay:
The bulk of the diff is a refactor that was necessary to avoid duplicate code. Please see individual commits!

actix-files/src/service.rs
```diff
@@ -127,11 +121,19 @@ impl Service<ServiceRequest> for FilesService {
...
+            match self.index {
+                Some(ref index) => match NamedFile::open(path.join(index)) {
+                    Ok(named_file) => serve_named_file(req, named_file),
+                    Err(_) if self.show_index => show_index(req),
+                    Err(err) => self.handle_err(err, req),
+                },
+                None if self.show_index => show_index(req),
+                _ => Box::pin(ok(ServiceResponse::from_err(
                     FilesError::IsDirectory,
                     req.into_parts().0,
```
aliemjay, line 138: Although unrelated, but I think that this should be `self.handle_err(FilesError::IsDirectory, req)` to give a chance for the `::default_handler(_)`, if set, to run?

robjtede (approved):
nice change, thanks 👍🏻

bitemyapp (approved):

PR #2233, opened 2021-05-24
Remove `FromRequest::Config`

### LINKED ISSUE #2221: Shall we remove `type Config` of `FromRequest`?
Most of the `type Config` are `()`, and the others can be replaced by `()` too.

In fact, the `Config` belongs to implementation, which should not be exposed in `FromRequest` trait.

arniu:
## PR Type

Refactor

## PR Checklist

## Overview

closes #2221


robjtede:
Aside from `String` extractor respecting `PayloadConfig` becoming non-obvious with this change, I don't have a strong opinion and would like to hear from @actix/contributors.

fakeshadow:
This is a breaking change for nothing.

jplatte:
What was the point in the first place if this associated type isn't actually needed (as #2221 says)?

> Aside from `String` extractor respecting `PayloadConfig` becoming non-obvious with this change, …

IMHO "respects some server config" should just be plain documentation (as it already is).

ibraheemdev:
I like this change, it simplifies the public API.

fakeshadow:
Arguing it's needed or not is pointless. You should do that when the trait is first introduced.

The point of having it is people don't have to make break change. It's a long standing trait and it works fine and that's all.

ibraheemdev:
Yeah, it's obvious that the type is not *needed*. I guess the decision is just whether or not it's worth a breaking change. You could argue both ways.

popzxc:
> You should do that when the trait is first introduced.

IMHO it doesn't mean that things may never be improved. Having breaking changes is OK, that's why we have semver in the first place.
In that case, the migration procedure is simple, and it actually improves the interface: having less associated types/generic  parameters is usually better (and in that case, it *is* better).

I personally like this change.

fakeshadow:
> > You should do that when the trait is first introduced.
> 
> IMHO it doesn't mean that things may never be improved. Having breaking changes is OK, that's why we have semver in the first place.
> In that case, the migration procedure is simple, and it actually improves the interface: having less associated types/generic parameters is usually better (and in that case, it _is_ better).
> 
> I personally like this change.

What improvement? Reduce one loc when you importing the trait? Does it really make a difference?
I would rather people actually export this trait and expose it in their apps/libs not be breaking for this meaningless change.

You have to remember any lib using this trait would be affected and they need a new release just for this change alone.

Look at the current betas. You already have enough people can not figuring out how to make actix-xxx be compat with their beta releases.

neoeinstein:
Having this coupled with a SemVer major release makes this the right time to make this change. It's vestigial and confusing to consumers. After the change is made, the API will be cleaner and the better for it.

neoeinstein (approved):

popzxc:
> What improvement? Reduce one loc when you importing the trait? Does it really make a difference?

I'd say that it's not about writing/removing one line of code, but rather figuring out what the heck is it and do you actually need it (spoiler: you don't).

The point of upgrades is not only to care about users who already use the library but also to care about ones who will use it in the future. Giving the fact that Rust's popularity constantly grows and `actix-web` is a pretty important framework for its infrastructure, I think that simplifying things is always a good thing.

Otherwise, we can get stuck forever with "it works for current users, they already know how it works and there is no need to introduce the inconvenience".

MIGRATION.md
```diff
@@ -10,6 +10,7 @@
 
   Alternatively, explicitly require trailing slashes: `NormalizePath::new(TrailingSlash::Always)`.
 
+* The `type Config` of `FromRequest` was removed.
```
popzxc, line 14: This should be also mentioned in the changelog, right?

popzxc (approved):

fakeshadow:
There are so many dead code in current code base and you are like nah they are fine and I want this public API change for nothing to happen.
I already said in previous reply it's pointless to argue if it's need or not. You can remove what you need or keep what you don't. It all comes down to what you actually gain from the change and this is the hard fact of this PR.

You save one loc for using this trait at the cost of breaking a public API.

I won't make further reply in this thread as there is no point to argue for nothing. This PR does nothing break something and real changes for meaningful PRs are ignored. We should put effort on those.

popzxc:
> There are so many dead code in current code base

It also doesn't sound good to me, to be honest. I think it's a good idea to remove all the dead code, because why keep it.

arniu:
Yes, the change is trivial, and it is a break change.

But, it make the public API cleaner and simple and it does make sense.

Now we are marching toward a new, groundbreaking version.

It's time to make some break change!

robjtede:
I'm not entirely convinced about this yet. If it is going to be merged I'd like to see adequate docs for String and Bytes extractors regarding their config type.

popzxc:
> I'd like to see adequate docs for String and Bytes extractors regarding their config type.

👍 on this. Also, it will be really nice to extend `FromRequest` trait docs with information on how you *can* configure extractor, ideally with a doc-test example. Like, while most users don't need it, it'd be helpful for ones who in fact do.

arniu:
I am not a native English speaker. Any one can help to improve the words?

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
 
 use crate::{dev::Payload, Error, HttpRequest};
 
-/// Trait implemented by types that can be extracted from request.
+/// [`FromRequest`] is one of the most commonly used trait, although some may not realize it.
 ///
-/// Types that implement this trait can be used with `Route` handlers.
+/// A type that implements [`FromRequest`] is called an **extractor** and can be extracted
```
jplatte:
```suggestion
/// A type that implements [`FromRequest`] is called an **extractor** and can extract data
```

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
...
+/// from the request, for example [`Json`], [`Form`], [`Path`] and so on.
+///
+/// An extractor can be customized by injecting the corresponding configuration with:
+///
+/// - [`App::data()`](`crate::App::data`)
+/// - [`App::app_data()`](`crate::App::app_data`)
```
jplatte: Only one of `data()` / `app_data()` works, no?

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
 
 use crate::{dev::Payload, Error, HttpRequest};
 
-/// Trait implemented by types that can be extracted from request.
+/// [`FromRequest`] is one of the most commonly used trait, although some may not realize it.
 ///
-/// Types that implement this trait can be used with `Route` handlers.
+/// A type that implements [`FromRequest`] is called an **extractor** and can be extracted
+/// from the request, for example [`Json`], [`Form`], [`Path`] and so on.
```
popzxc:
```suggestion
/// from the request. Examples of types that implement this trait are [`Json`], [`Form`], [`Path`].
```

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
...
+///
+/// An extractor can be customized by injecting the corresponding configuration with:
```
popzxc: I'd suggest putting this part below the list of the built-in extractors, so the reader can actually look at the configs, and then see how he can inject their own.

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
 
 use crate::{dev::Payload, Error, HttpRequest};
 
-/// Trait implemented by types that can be extracted from request.
+/// [`FromRequest`] is one of the most commonly used trait, although some may not realize it.
```
popzxc: IMHO this line is a bit bogus since it doesn't provide a user with any meaningful information. 
I'd suggest removing it completely.

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
...
+/// [`FormConfig`]: crate::web::FormConfig
+/// [`Path`]: crate::web::Path
+/// [`PathConfig`]: crate::web::PathConfig
+/// [`Query`]: crate::web::Query
+/// [`QueryConfig`]: crate::web::QueryConfig
+/// [`Payload`]: crate::web::Payload
+/// [`PayloadConfig`]: crate::web::PayloadConfig
+/// [`String`]: FromRequest#impl-FromRequest-for-String
+/// [`Bytes`]: crate::web::Bytes#impl-FromRequest
+#[doc(alias = "Extractor")]
```
popzxc: I think this is not needed (and experimental).

jplatte: To make this work on `docs.rs` without breaking builds on stable you can add

```toml
[package.metadata.docs.rs]
rustdoc-args = ["--cfg", "docsrs"]
```

to Cargo.toml and change this to

```suggestion
#[cfg_attr(docsrs, doc(alias = "Extractor"))]
```

src/types/form.rs
```diff
@@ -236,14 +228,25 @@ impl FormConfig {
         self.err_handler = Some(Rc::new(f));
         self
     }
+
+    /// Extract payload config from app data. Check both `T` and `Data<T>`, in that order, and fall
+    /// back to the default payload config.
```
jplatte:
```suggestion
    /// Extract payload config from app data.
    ///
    /// Checks both `T` and `Data<T>`, in that order, and falls back to the default payload config.
```

popzxc (commented):
In general, I'm happy with the added docs. Great job.

src/types/form.rs
```diff
@@ -236,14 +228,25 @@ impl FormConfig {
...
+    err_handler: None,
+};
+
 impl Default for FormConfig {
     fn default() -> Self {
-        FormConfig {
-            limit: 16_384, // 2^14 bytes (~16kB)
-            err_handler: None,
-        }
+        DEFAULT_CONFIG.clone()
```
jplatte: This should work without `.clone()`.

```suggestion
        DEFAULT_CONFIG
```

src/extract.rs
```diff
@@ -11,13 +11,47 @@ use futures_core::ready;
...
+///
+/// - [`App::data()`](`crate::App::data`)
+/// - [`App::app_data()`](`crate::App::app_data`)
+/// - [`Scope::data()`](`crate::Scope::data`)
+/// - [`Scope::app_data()`](`crate::Scope::app_data`)
+/// - [`Resource::data()`](`crate::Resource::data`)
+/// - [`Resource::app_data()`](`crate::Resource::app_data`)
+///
+/// Here are some built-in extractors and their corresponding configuration.
+/// Please refer to the respective documents for details.
```
jplatte:
```suggestion
/// Please refer to the respective documentation for details.
```

src/extract.rs
otavio: Not really; it is stable since 1.48. 

See: https://blog.guillaume-gomez.fr/articles/2020-12-04+doc%28alias%29+is+stable+and+it%27s+gonna+be+super+useful%21

src/extract.rs
```diff
@@ -11,13 +11,45 @@ use futures_core::ready;
...
 use crate::{dev::Payload, Error, HttpRequest};
 
-/// Trait implemented by types that can be extracted from request.
+/// A type that implements [`FromRequest`] is called an **extractor** and can extract data
+/// from the request. Examples of types that implement this trait are [`Json`], [`Form`], [`Path`].
 ///
-/// Types that implement this trait can be used with `Route` handlers.
+/// An extractor can be customized by injecting the corresponding configuration with one of:
+///
+/// - [`App::data()`](`crate::App::data`)
```
LukeMathWalker: Given that `::data` has been marked as deprecated in favour of `::app_data` in the main branch it probably shouldn't be recommended here as an option.

src/types/payload.rs
```diff
@@ -197,14 +193,15 @@ fn bytes_to_string(body: Bytes, encoding: &'static Encoding) -> Result<String, E
...
+/// Note that the [`Payload`] extractor does not automatically check
+/// conformance with this configuration to allow more flexibility when
+/// building extractors on top of [`Payload`].
 ///
 /// By default, the payload size limit is 256kB and there is no mime type condition.
 ///
-/// To use this, add an instance of it to your app or service through one of the
-/// `.app_data()` methods.
+/// To use this, add an instance of it to your [`app`](crate::App), [`scope`](crate::Scope)
+/// or [`resource`](crate::Resource) through the associated `.app_data()` or `.data()` method.
```
LukeMathWalker: Same here, let's suggest just `app_data`

LukeMathWalker:
2c from a long-time `actix-web` user (+library author, both public and private): the simpler you can make traits that are used often by application users, the better. Each generic parameter/associated type adds a non-trivial amount of complexity, especially in the hands of non-very-experienced Rust users.
E.g. writing middlewares for `actix-web` is not easy at all. 

If you have a chance to simplify without losing expressiveness or performance, I'd take it (of course providing good docs on how to move forward).

arniu:
I really hope it can be merged before 4.0.0. And what's the problem for the moment?

popzxc:
@robjtede ?

omid:
OK, this breaks serde_qs: https://github.com/samscott89/serde_qs/blob/main/src/actix.rs#L100
What is the workaround to pass a config?

jplatte:
@omid There is no workaround required. `req.app_data::<QsQueryConfig>()` will still work without the associated type.

omid:
@jplatte thanks. So in this code here: https://github.com/samscott89/serde_qs/blob/main/src/actix.rs#L151
We need to just set app_data with the QsQueryConfig, right?

jplatte:
I don't really understand what you mean or what the `QsQuery::<Info>::configure` in the docs you ilnked is about. I think it would be good to open a Q&A [discussion](https://github.com/actix/actix-web/discussions) for this rather than discussing on this merged PR.

PR #2243, opened 2021-06-01
Additional test coverage and tidyup (middleware::normalize)

Digital-Chaos:
## PR Type
Other

## PR Checklist

## Overview

Added missing test coverage for normalization of uri paths with a query string and some tidyup.

### DIFF
--- a/src/middleware/normalize.rs
+++ b/src/middleware/normalize.rs
@@ -189,6 +189,7 @@ mod tests {
     use super::*;
     use crate::{
         dev::ServiceRequest,
+        guard::fn_guard,
         test::{call_service, init_service, TestRequest},
         web, App, HttpResponse,
     };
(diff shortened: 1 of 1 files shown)


src/middleware/normalize.rs
```diff
@@ -199,77 +200,150 @@ mod tests {
...
+            "//v2//something//?query=test",
+        ];
 
-        let req = TestRequest::with_uri("/v1/something////").to_request();
-        let res = call_service(&app, req).await;
-        assert!(res.status().is_success());
+        for uri in test_uris {
+            let req = TestRequest::with_uri(uri).to_request();
+            let res = call_service(&app, req).await;
+            assert!(res.status().is_success());
```
otavio: On those asserts, it'd be nice to provide a more informative error when the assert fails so it provides some more information as it'll act on a loop.

Digital-Chaos: Thanks for the suggestion, I've added assertion messages including the specific uri under test.

robjtede (approved):
thanks 👍🏻

PR #2244, opened 2021-06-02
Add zstd ContentEncoding support

arthurlm:
## PR Type

Feature

## PR Checklist

## Overview

Add support zstd compression using `Content-Encoding: zstd`.

Zstd compression algorithm for content encoding is part of IANA references formats: https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding

This PR add `zstd` as a direct actix dependency if feature `compress` is enabled.
On current `actix_web::middleware::Compress` architecture, it is not possible to add this as a _"plugin"_.
One solution might be to add a new feature like `compress-zstd` if this change is a problem.

By default this PR will not change current behavior since Q-factor is set to 0.0 for zstd.


otavio (approved):
The change seems simple enough for addition. 

I'd argue we shouldn't add an extra feature for `zstd` only. We could rework it to add all compression support code under specific feature flags, thus reducing the dependencies and binary size.

So my vote goes for merge this and then, friendly ask you to prepare a new PR moving compression support code under `compression-<format>` specific feature flags.

arthurlm:
Thanks for your feedback.

I have start working on the change your suggest: https://github.com/arthurlm/actix-web/compare/zstd...arthurlm:compress-features
I agree with you: it would be nice to be able to select compression algorithm using features flags.
Please note work about algorithm selection is sill ongoing.
I will provide a PR when this one is accepted, merged and my work is completed.

I also have fix lint errors + downgrade zstd to make it works with rustc 1.46.0+.
So CI should now work better.

otavio:
@arthurlm, please squash all the changes, so we have a unified commit.

arthurlm:
Sure ! It is done.

otavio (approved):

otavio:
@arthurlm, thanks for your fast response. For me, it is good to go. Let's see what other @actix/contributors have to say.

robjtede:
fyi @otavio, we do squash commits when merging PRs and there are some minor benefits when reviewing to keeping the history during PR iterations. Not a big deal just to note.

tests/test_server.rs
```diff
@@ -476,6 +477,125 @@ async fn test_body_brotli() {
...
+        .take(320_000)
+        .map(char::from)
+        .collect::<String>();
+
+    let srv = actix_test::start_with(actix_test::config().h1(), || {
+        App::new().service(
+            web::resource("/")
+                .app_data(web::PayloadConfig::new(320_000))
+                .route(web::to(move |body: Bytes| {
+                    HttpResponse::Ok().streaming(TestBody::new(body, 10240))
```
robjtede, line 577: to avoid any self-inversion bugs can we add an assertion of the Bytes content in the handlers

actix-http/src/encoding/decoder.rs
```diff
@@ -45,6 +46,10 @@ where
             ContentEncoding::Gzip => Some(ContentDecoder::Gzip(Box::new(
                 GzDecoder::new(Writer::new()),
             ))),
+            ContentEncoding::Zstd => match ZstdDecoder::new(Writer::new()) {
+                Ok(decoder) => Some(ContentDecoder::Zstd(Box::new(decoder))),
+                Err(_) => None,
```
popzxc: Is this OK to ignore an error here?
What happens if the `zstd` decoder was not created and then `Decoder` is used to decode `zstd`-compressed data?

arthurlm: This is a good question. Actually my first implementation was to just unwrap the `ZstdDecoder::new` result.
Looking at the code I do not really see how this can be an `Err`:
- [stream::write::Decoder::new](https://docs.rs/zstd/0.8.2+zstd.1.5.0/src/zstd/stream/write/mod.rs.html#245-247)
- calls [stream::write::Decoder::with_dictionary](https://docs.rs/zstd/0.8.2+zstd.1.5.0/src/zstd/stream/write/mod.rs.html#253-257)
- calls [raw::Decoder::with_dictionary](https://docs.rs/zstd/0.8.2+zstd.1.5.0/src/zstd/stream/raw.rs.html#142-149)
- which panic if dictionary is invalid. But in our case, I do not understand how the default dictionary can be invalid.

I always "fear" unwrapping and prefer handle error case (even if they are not really possible) in case of library changes.
But indeed in this case, behavior will be "strange".
So it is up to you, I can update this to just unwrap the `ZstdDecoder::new` result or keep this as it.

tests/test_server.rs
arthurlm: Sure ! Do you have any example 😄 ?
I have base this tests on `test_brotli_encoding_large` and `test_brotli_encoding`.
But they do not contains such asserts. I will be happy to add this asserts in this tests too.

actix-http/src/encoding/decoder.rs
popzxc: I'd go with `expect`. It *shouldn't* panic, thus it must be safe in theory, but just for the case it will *ever* panic, we should provide a helpful error message telling user that it's a bug in either `actix-web` or `zstd` and this issue must be reported.

arthurlm: I just update the code.
I resolve this conversation but feel free to reopen it if there is still something missing.

actix-http/src/encoding/decoder.rs
```diff
@@ -45,6 +46,9 @@ where
             ContentEncoding::Gzip => Some(ContentDecoder::Gzip(Box::new(
                 GzDecoder::new(Writer::new()),
             ))),
+            ContentEncoding::Zstd => Some(ContentDecoder::Zstd(Box::new(
+                ZstdDecoder::new(Writer::new()).expect("Fail to create zstd decoder"),
```
popzxc:
```suggestion
                ZstdDecoder::new(Writer::new())
                    .expect(
                        "Failed to create zstd decoder. This is a bug and should normally happen. \
                        Please report it to the actix-web repository."
                    ),
```

popzxc (approved):

actix-http/src/encoding/decoder.rs
jplatte: Should normally happen? 😄

popzxc: Oops 😅

tests/test_server.rs
robjtede: indeed they don't (yet) but this new test could lead the way

robjtede: actually i'm reading it wrong, the current asserts are fine

arthurlm:
Thanks for your feedback everyone !
I now see that your reviews and comments have been fixed :smile:.

I have take a look at why CI failed and I am not sure this is related to the changes I have made.
Unfortunately, I do not have a macOS to run tests myself and check it.
If anyone has macOS, can you please check on your side ?
Otherwise is it possible to relaunch CI in order to test if the bug is not related to CI itself ?

If everyone is ok with this change, we may merge this PR and I will start working on enabling compression algorithm using features flags.

robjtede:
nice job on this @arthurlm, thank you

PR #2257, opened 2021-06-09
files: inline disposition for common web app file types

### LINKED ISSUE #610: Javascript Files Content-Disposition Incorrectly Detected as Attachement
Current the mime detection from mime_guess correctly identifies javascript files as "application/javascript" but that doesn't match any of the specially named Mime types `mime::IMAGE`, `mime::TEXT`, or `mime::VIDEO` which is currently marked as inline.

This can be worked around through the StaticFileConfig, but the default for these files should be correct for a webserver.

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

Closes #610

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -4,10 +4,12 @@
 * `NamedFile` now implements `ServiceFactory` and `HttpServiceFactory` making it much more useful in routing. For example, it can be used directly as a default service. [#2135]
 * For symbolic links, `Content-Disposition` header no longer shows the filename of the original file. [#2156]
 * `Files::redirect_to_slash_directory()` now works as expected when used with `Files::show_files_listing()`. [#2225]
+* `application/{javascript, json, wasm}` mime type now have `inline` disposition by default. [#2257]
 
 [#2135]: https://github.com/actix/actix-web/pull/2135
 [#2156]: https://github.com/actix/actix-web/pull/2156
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):
LGTM! Could you also add a changelog entry?

ekrem95 (approved):

robjtede:
Are there other mime types that should be considered here?

Ones that come to mind:
- application/javascript (utf8 variant)
- application/wasm
- application/json
- application/x-www-form-urlencoded

aliemjay:
>     * application/javascript (utf8 variant)
already covered

>     * application/wasm
>     * application/json
sure!

>     * application/x-www-form-urlencoded
This doesn't occur in static files. right?

robjtede:
utf8 => right, `.subtype` doesn't care about parameters
urlencoded => true, leave it out

aliemjay:
How about?:

audio/*
application/graphql
application/xml

aliemjay:
On the other side, I don;t want this to be bloated. The user can easily override the defaults.

robjtede:
We'd like to have sensible defaults for common web activities where static files are used. I think WASM and JSON make sense in addition to javascript but the others mentioned are a stretch, as you say other use cases can be supported with custom configs.

robjtede (approved):

PR #2261, opened 2021-06-10
[RFC] fix app_data in middleware of scope

### LINKED ISSUE #1790: Cannot access app_data in middleware of Scope
## Simple Use Case
```rust
...
web::scope("/api")
    .wrap_fn(|req, srv| {
        let fut = srv.call(req);

        async move {
            let res = fut.await?;
            Ok(res)
        }
    })
...
```
## Expected Behavior

## Current Behavior

## Possible Solution

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

This is my take on #1790

This simply wraps the scope endpoint with a middleware that indroduces `app_data` early in the middleware chain! I'm unsure about the performance, though.

`Resource` has the same problem.

Closes #1790

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -424,7 +424,6 @@ where
 
         // complete scope pipeline creation
         *self.factory_ref.borrow_mut() = Some(ScopeFactory {
-            app_data: self.app_data.take().map(Rc::new),
             default,
             services: cfg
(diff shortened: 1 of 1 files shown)


aliemjay:
Closing in favor of #2267

PR #2262, opened 2021-06-12
`Route::service`

ibraheemdev:
## PR Type
Feature

## PR Checklist

## Overview
Add a method to register service factories under a route directly.


src/route.rs
```diff
@@ -126,6 +126,55 @@ impl Route {
         self
     }
 
+    /// Set the service to be called when this route is hit.
+    ///
+    /// ```
+    /// # use actix_web::{*, dev::*, http::header};
```
popzxc: > `use actix_web::dev::*`

I am a bit concerned about the combination of the following facts:
- `Service` lies in the `actix_web::dev::*`
- `actix_web::dev` has a description that states "prelude for library developers" (i.e. it's not for *everyone*)
- `Route` is a very widely used entity.

It may lead to the fact that `actix_web::dev` will have to be used not only by the library developers but also by the end application developers, which is inconsistent.

Doesn't it make more sense now to re-export `Service` from `actix_web` top module itself and extend the description of it to explain more precisely when it should and should not be used directly (for example, with a link to this method).

src/route.rs
```diff
@@ -126,6 +126,55 @@ impl Route {
         self
     }
 
+    /// Set the service to be called when this route is hit.
```
popzxc: I think `service` must be a link to a structure. Otherwise, this comment doesn't explain much.

src/route.rs
```diff
@@ -126,6 +126,55 @@ impl Route {
...
+    ///
+    /// ```
+    /// # use actix_web::{*, dev::*, http::header};
+    /// # use std::task::{Context, Poll};
```
popzxc: Minor nit: nested imports are used above, but not here, which disturbs my perfectionism 😅

src/lib.rs
```diff
@@ -135,6 +135,7 @@ pub mod dev {
     pub use crate::info::ConnectionInfo;
     pub use crate::rmap::ResourceMap;
     pub use crate::service::{HttpServiceFactory, ServiceRequest, ServiceResponse, WebService};
+    pub use actix_service::fn_factory;
```
popzxc: `fn_factory` docs use the `fn_service` function which looks convenient but is not re-exported. Shouldn't it be re-exported too?

src/route.rs
ibraheemdev: Good points, but it is already inconsistent. Anyone writing a middleware imports `dev::ServiceRequest`. I think `dev` is more for people diving deeper than writing regular function handlers and extractors. Middleware and services are on the same level in that regard.

src/route.rs
ibraheemdev: Where would nested imports be used here?

src/lib.rs
ibraheemdev: It's not really useful because it serves the same purpose as a hander.

src/route.rs
popzxc: `use std::{task::{Context, Poll}, pin::Pin, future::Future};`

src/route.rs
popzxc: Then I guess the description of the `dev` module should be reworked, as now it provides an expression of something you normally shouldn't really want to use 😓 

Not a subject for this PR though, I guess.

src/route.rs
ibraheemdev: Eh, the comment is more of a header, the code explains more. If it linked to `Service` it would be weird because the method takes a `ServiceFactory`, and if it linked to `ServiceFactory` it would be weird because it isn't really meant for service factories, it's meant for custom services and not many people even know what a service factory is. The comment is consistent with other `Route` methods docs so :man_shrugging:

src/lib.rs
popzxc: I mean that it's used in the example of this function's documentation and it **looks** convenient, so some folks who'll bump into it may want to use it according to the provided example, only to find out that it's not available through `actix-web`.

Like, it's not about actual helpfulness, but rather about avoiding a little point of frustration.

ibraheemdev: I see your point, I'll let @robjtede chime in on the export situtation here and with the `dev` module.

popzxc (approved):
Despite my nitpicks on docs, LGTM.

src/route.rs
robjtede: I'm (barely) on the side of exporting `fn_service`. Also re-exporting the `always_ready!` and `forward_ready!` macros would help here and in other places.

src/lib.rs
robjtede: I'm (barely) on the side of exporting `fn_service`. Also re-exporting the `always_ready!` and `forward_ready!` macros would help here and in other places.

src/route.rs
```diff
@@ -126,6 +126,53 @@ impl Route {
...
+    /// struct HelloWorld;
+    ///
+    /// impl Service<ServiceRequest> for HelloWorld {
+    ///     type Response = ServiceResponse;
+    ///     type Error = actix_web::Error;
+    ///     type Future = LocalBoxFuture<'static, Result<Self::Response, Self::Error>>;
+    ///
+    ///     fn poll_ready(&self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+    ///         Poll::Ready(Ok(()))
+    ///     }
```
robjtede: prefer a re-export of always_ready! macro and use here

ibraheemdev: it was already being re-exported :)

aliemjay:
This is the only `service()` method that takes `impl ServiceFactory`. All others expect `impl HttpServiceFactory`, which is expected to define routing information. I think the name and docs should reflect this.

The docs for other service methods lists few implementations for "Service": https://docs.rs/actix-web/4.0.0-beta.7/actix_web/struct.Scope.html#method.service

ibraheemdev:
The naming of this method makes sense I think. What else would it be called? If `service_factory`, that's weird because the other methods take `HttpServiceFactory` but are called `service`? I agree the docs for this method could be better, but I am not sure about renaming. If anything, those method should be renamed to `http_service`.

robjtede:
They're all services. We could distinguish this one with with `raw_service()` maybe. I thought adding "raw" into the doc comment would be enough to hint this was a more advanced method. Personally, I am happy with as it is but I can see where it could be confusing still

ibraheemdev:
Are http factories not "raw" services even more so than a request handler service?

I think it's fine, maybe add a note that `to` and function handlers cover most use cases.

aliemjay:
From the examples here, I see that `impl Handler for T where T: Service` would be sufficient to make it work with `to()` method. What do you think?

`ServiceFactory::new_service()` can be used to create a new service if the user have an instance of it.

ibraheemdev:
`impl Handler for T where T: Service` would be conflicting.

robjtede:
i think this is safe enough to merge, if a rename is needed after stabilizing it's not too intrusive to pick a new one and deprecate this given it's probably rare use

robjtede (approved):

aliemjay (approved):

PR #2263, opened 2021-06-12
Implement `FromRequest` for request parts

ibraheemdev:
## PR Type
Feature

## PR Checklist

## Overview
Implement `FromRequest` for request parts (Method, Uri, Peer address, etc.)

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -225,9 +225,8 @@ where
 /// ```
 /// use actix_web::{web, App, HttpRequest, http::Uri};
 ///
-/// async fn index(uri: Uri) -> &'static str {
-///     log::info!("Incoming request to: {}", uri.path());
-///     "Welcome!"
+/// async fn index(uri: Uri) -> String {
+///     format!("Requested path: {}", uri.path())
 /// }
 ///
 /// fn main() {
@@ -251,9 +250,8 @@ impl FromRequest for Uri {
 /// ```
 /// use actix_web::{web, App, HttpRequest, http::Method};
 ///
-/// async fn index(method: Method) -> &'static str {
-///     assert_eq!(method, Method::GET);
-///     "Welcome!"
+/// async fn index(method: Method) -> String {
+///     format!("Request method: {}", method)
 /// }
 ///
 /// fn main() {
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -2,21 +2,16 @@
 
 use std::{
     convert::Infallible,
-    fmt,
     future::Future,
-    net::SocketAddr,
     pin::Pin,
     task::{Context, Poll},
 };
 
 use actix_http::http::{Method, Uri, Version};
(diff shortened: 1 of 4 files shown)


popzxc:
While code itself LGTM, I'm not sure if understand what's the use case of that. Can you elaborate a bit (and probably add comments to the code similarly to `impl<T> FromRequest for Result<T, T::Error>`; though maybe less verbose)?

ibraheemdev:
Do you think it needs comments? I thought it's pretty self explanatory... If you want the request Uri, now you can just use the `Uri` extractor instead of extracting `HttpRequest` and then calling `.uri()`. `Result` and `Option` have comments and examples because they do some more logic. That said, examples wouldn't hurt I guess. Plus we would have to add these to the list on the website.

ibraheemdev:
I guess extracting a `Version` is probably uncommon. There are definitely use cases for `Uri`, `Method`, `ConnectionInfo`, and `SocketAddr` though.

popzxc:
> That said, examples wouldn't hurt I guess.

Yeah, I think it'd be more than enough.

src/extract.rs
```diff
@@ -218,14 +225,81 @@ where
...
+
+impl fmt::Display for MissingPeerAddr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Missing peer address.")
+    }
+}
+
+impl ResponseError for MissingPeerAddr {}
+
+impl FromRequest for SocketAddr {
```
robjtede: I think i'd prefer a newtype struct for this `PeerAddr(SocketAddr)` since it _could_ be non-obvious what socket addr this refers to. Plus in general, it's easier to find these sorts of things in docs.

robjtede (changes requested):

src/info.rs
```diff
@@ -187,6 +193,64 @@ impl ConnectionInfo {
...
+impl PeerAddr {
+    /// Unwrap into inner `SocketAddr` value.
+    pub fn into_inner(self) -> SocketAddr {
+        self.0
+    }
+}
+
+#[derive(Debug, Display, Error)]
+#[display(fmt = "Missing peer address")]
+pub struct MissingPeerAddr;
```
jplatte, line 252: Is there any point in allowing users to construct this type? If not, I'd suggest making it `#[non_exhaustive]`.

robjtede: this is such a hot tip, i will remember this forever; i'd missed the nuance of non_exhaustive that it made structs non contstructible outside the crate

jplatte: Related tip: You can add `#![warn(clippy::exhaustive_structs)]` to the crate attributes to catch all instances of structs being constructible by users using struct literals and selectively add `#[allow(clippy::exhaustive_structs)]` where appropriate.

robjtede (approved):
👍🏻

PR #2264, opened 2021-06-16
Introduce `Group`

### LINKED ISSUE #414: Apply middleware(s) to group of routes without prefix
I have set of routes that need to pass through group of middlewares. I tried to use route scopes but it requires prefix to be applied as well. I'm looking for feature like in laravel [route groups ](https://laravel.com/docs/5.6/routing#route-groups)

eg.
public routes:
/login

private routes: 
/dashboard
/users/**
/activities/**
/logout

### LINKED ISSUE #2096: Extract path arguments in middleware before processing the request
As per https://github.com/actix/actix-web/issues/1784#issue-739668855, the path arguments in middleware are not accessible until after the request is processed.

## Expected Behavior
`actix_web::web::Path::<String>::from_request(request, payload)` in a service call returns path argument(s) at any point in the service call.

## Current Behavior
Output of request.match_info(), request.match_pattern() and actix_web::web::Path::<String>::from_request(request, payload) from inside the service call: "wrong number of parameters: 0 expected 1".

1. Implement a middleware service, with the Path extractor
2. Call the middleware on a route which takes a path argument

## Context
I am trying to implement guards on the URL path to ensure that users can only access what they are allowed to access. Only being able to access the patch arguments after the request means that it is not possible to use this method for endpoints that update information

## Your Environment

    Rust Version: 1.50.0
    Actix Web Version: 3.3.2

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

This tries to implement #414 . I implemented these changes on `Scope` temporarily to use existing tests and for smaller diff.

The new behavior for scope is equivalent to registering component services individually. This allows the scope to fallthrough when no matching route is found (instead of calling default service). A typical use case would be:
```rust
app
.service(
    web::scope("/users")
        .wrap(no_auth_needed)
        .route("u1", web::get().to(handle_u1))
        .route("u2", web::get().to(handle_u2)),
)
.service(
    web::scope("/users")
        .wrap(need_auth)
        .route("{user}", handle_other_users),
)
````

Setting `Scope::default_service()`, would make it mimic the original behavior.

As a side-effect, this will also make #2096 possible.

# Open questions
This introduces new API. So I'm looking for input before finishing it. Specifically, I'm looking for the naming (eg. `EndpointConstructor` trait) and  what mehods to have for it.

Closes #414
Closes #2096

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -775,7 +775,7 @@ mod tests {
                         },
                     )))
                     .default_service(web::to(move |req: HttpRequest| {
-                        assert!(req.match_pattern().is_none());
+                        assert_eq!(req.match_pattern().as_deref(), Some("/user/{id}"));
                         HttpResponse::Ok().finish()
                     })),
             ),
--- a/src/rmap.rs
+++ b/src/rmap.rs
@@ -26,6 +26,10 @@ impl ResourceMap {
         }
     }
 
+    pub fn set_root_prefix(&mut self, prefix: &str) {
+        rdef_set_root_prefix(&mut self.root, prefix);
+    }
+
     pub fn add(&mut self, pattern: &mut ResourceDef, nested: Option<Rc<ResourceMap>>) {
         pattern.set_id(self.patterns.len() as u16);
         self.patterns.push((pattern.clone(), nested));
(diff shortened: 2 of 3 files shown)


robjtede:
closing as stale, and i'm having a crack at this with Astra's help

PR #2267, opened 2021-06-18
Fix app_data accessiblity in middlewares

### LINKED ISSUE #1790: Cannot access app_data in middleware of Scope

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

An alternative implementation for #2261. This delegates the responsility for setting `app_data` to upstream routing services (`App` and `Scope`). Since `Resource` is not a router, it doesn't have to set `app_data` itself. 

Closes #1790 
Closes #2261


aliemjay:
Giving it more thought, I think #2261 still have some merit as less invasive change.

src/scope.rs
```diff
@@ -537,11 +542,10 @@ impl Service<ServiceRequest> for ScopeService {
...
-        }
-
-        if let Some((srv, _info)) = res {
+        if let Some(((srv, app_data), _info)) = res {
+            if let Some(ref app_data) = app_data {
+                req.add_data_container(app_data.clone());
+            }
             srv.call(req)
         } else {
             self.default.call(req)
```
robjtede, line 553: there's a regression here in that default services are losing access to it's layer's data

aliemjay: Its layer's data are already set by  the parent layer before calling Service::call

aliemjay: At leat this is what I think. I'll write a test case for this once I get to my dev pc.

aliemjay: There are already some unit tests for such case: `resource::tests::test_data_default_service` and `scope::tests::test_override_data_default_service`.

So I can confirm that this behaves as I expect: any layer's app_data are registered to and set by the parent layer's routing service.  In other words, `req.add_data_container(app_data.clone());`, in the quoted diff, sets app_data for lower layers, not the current one.

aliemjay:
Closing in favor of #2288

PR #2269, opened 2021-06-18
Fix bug where guards can be escaped

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

`ScopeFactory` and `AppInit` move the guards only to the first service they create.

I was able to have a POST request to GET route.


aliemjay:
Sorry, I think this is a false alarm. Investigating it further...

PR #2274, opened 2021-06-20
files: file path filtering closure

### LINKED ISSUE #1916: enable file middleware filtering closure

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

Closes #1916


aliemjay:
Should we provide the absolute or the relative path to the closure? 

What about `&HttpRequest` parameter?

robjtede:
> What about `&HttpRequest` parameter?

There's precedent in the CORS crate for providing the request head to a closure (too) like this.

https://docs.rs/actix-cors/0.5.4/actix_cors/struct.Cors.html#method.allowed_origin_fn

Thoughts?

aliemjay:
> Thoughts?

I think it is reasonable to have it here too.

aliemjay:
I'm still unsure about whether to provide the relative path as it is in the request url, or to provide the absolute path (i.e. after joining the root path with the request path).

I can see a use case for both of these: If you want to filter paths by depth or file names then a relative path is  easy to use here, but if the intention is to filter out symlinks, for example, then the the absolute path is better (it would prevent an otherwise necessary allocation with  `root_path.join(request_path).is_symlink()???`)

aliemjay:
Giving it more thought, I think it is better to have the the relative path since it would be easy to convert to absolute (though with a necessary allocation). But the converse is not easy since `actix-files` stores "cannonical" root path, which may be different than the one provided by the user, the root path, and it would then be inconvenient to strip the root prefix.

robjtede:
We can provide both with a 2-ary function call, `|relative_path, req_head|` similar to CORS.

actix-files/src/lib.rs
```diff
@@ -901,4 +904,40 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let req = TestRequest::with_uri("/Cargo.toml").to_srv_request();
+        let resp = test::call_service(&st, req).await;
+        assert_eq!(resp.status(), StatusCode::OK);
+
+        let req = TestRequest::with_uri("/tests/test.png").to_srv_request();
+        let resp = test::call_service(&st, req).await;
+        assert_eq!(resp.status(), StatusCode::NOT_FOUND);
```
robjtede: This test doesn't really prove that the filter works, it would be the same outcome if the file really wasnt there. ~If we set a `.default_handler()` that returns a different status code then the test would be valid.~

aliemjay: The default handler is calso called when the file doesn't exist. maybe use 'src/lib.rs' as a test path?

robjtede: true; that would be a better test case at least

robjtede (approved):

PR #2281, opened 2021-06-22
Various refactorings

popzxc:
## PR Type

Refactor

## PR Checklist

## Overview
A lot of minor refactorings here and there.

I guess it must be `semver-major` because of many `#[must_use]` attributes (which can be breaking with `-D warnings`).


ibraheemdev:
`must_use` is really only meant for functions that can cause problems if ignored, like `Result` or `Future`.

jplatte:
> I guess it must be `semver-major` because of many `#[must_use]` attributes (which can be breaking with `-D warnings`).

The warnings emitted are not usually considered to be tracked by semver. Otherwise you couldn't even deprecate things in a minor release.

popzxc:
> `must_use` is really only meant for functions that can cause problems if ignored, like `Result` or `Future`.

I googled for that and found mention of that in the original [rfc](https://github.com/rust-lang/rfcs/blob/161ce8a26e70226a88e0d4d43c7914a714050330/text/1940-must-use-functions.md).

However, even in this RFC there is a following statement:

> The same thing for `ne`, and also `lt`, `gt`, `ge`, `le` in `PartialOrd`. There is no reason to discard the results of those operations. 

While `eq` *may* cause problems, other ones in general cannot; and my motivation is the same as in the latter statement: if it makes no sense to discard return value, it's totally worth a warning.

---

On the other hand, why `must_use` in these cases is bad? When can it make things worse that without it?

ibraheemdev:
Not bad, just unnecessary I think. `std` doesn't mark everything `must_use`, only a select few types.

popzxc:
> `std` doesn't mark everything `must_use`, only a select few types.

AFAIK it's a hot topic over there, and currently the core team is in progress forming guidelines on how to use it.
For example, [std::Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) is not `must_use` yet, even though discarding the result may result in a needless and resource-heavy operation.

I personally find `must_use` a good midground between not having any linting at all and using [`unused-results`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-results) lint which is [very prone to false-positives](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=509c3d66b46d37a6b1c34b79a90a2673).

robjtede:
I'd like to remove the must_use annotations for now / move them to a separate PR for further bikeshedding.

popzxc:
Lowered to `semver-patch`, as now it contains internal things only.

actix-http/src/h1/dispatcher.rs
```diff
@@ -515,14 +515,13 @@ where
         cx: &mut Context<'_>,
     ) -> Result<(), DispatchError> {
         // Handle `EXPECT: 100-Continue` header
+        // set dispatcher state so the future is pinned.
+        let mut this = self.as_mut().project();
         if req.head().expect() {
```
robjtede:
```suggestion
        let mut this = self.as_mut().project();
        if req.head().expect() {
            // set dispatcher state so the future is pinned.
```

src/request.rs
```diff
@@ -286,7 +282,7 @@ impl HttpRequest {
         if let Ok(cookies) = self.cookies() {
             for cookie in cookies.iter() {
                 if cookie.name() == name {
-                    return Some(cookie.to_owned());
+                    return Some(cookie.clone());
```
robjtede:
```suggestion
                    return Some(cookie.to_owned());
```

src/response/response.rs
```diff
@@ -132,10 +132,7 @@ impl<B> HttpResponse<B> {
     pub fn del_cookie(&mut self, name: &str) -> usize {
         let headers = self.headers_mut();
 
-        let vals: Vec<HeaderValue> = headers
-            .get_all(header::SET_COOKIE)
-            .map(|v| v.to_owned())
-            .collect();
+        let vals: Vec<HeaderValue> = headers.get_all(header::SET_COOKIE).cloned().collect();
```
robjtede:
```suggestion
        let vals: Vec<HeaderValue> = headers
            .get_all(header::SET_COOKIE)
            .map(|v| v.to_owned())
            .collect();
```

robjtede (approved):

otavio (approved):

src/response/response.rs
popzxc: Is there a reason for that? As far as aI can see, `ToOwned` for both this type and `Cookie` have a blanket implementation which simply calls `Clone`.

While for the `Cookie` example there is no much difference, here we introduce an additional closure.

popzxc: I also see a method [`into_owned`](https://docs.rs/actix-web/3.3.2/actix_web/http/struct.Cookie.html#method.into_owned) for `Cookie`, maybe it's what was supposed to be called instead of `to_owned`?

PR #2282, opened 2021-06-22
simplify `ConnectionInfo::new`

ibraheemdev:
## PR Type
Refactor

## PR Checklist
- [x] tests
- [x] changelog

## Overview
Simplify `ConnectionInfo::new` and cover spec better + more tests.


src/info.rs
```diff
@@ -48,105 +51,73 @@ impl ConnectionInfo {
...
+                    "proto" => {
+                        scheme.get_or_insert_with(|| val.trim());
+                    }
+                    "host" => {
+                        host.get_or_insert_with(|| val.trim());
+                    }
+                    _ => {}
                 }
             }
         }
```
aliemjay:
```suggestion
        req
            .headers
            .get_all(&header::FORWARDED)
            .into_iter()
            .filter_map(|hdr| hdr.to_str().ok())
            .flat_map(|val| val.split(';'))
            .flat_map(|pair| pair.split(','))
            .filter_map(|pair|
                let mut items = pair.trim().splitn(2, '=');
                Some((items.next()?, items.next()?))
            })
            .for_each(|(name, val)| {
                match name.to_lowercase().as_ref() {
                    "for" => &mut realip_remote_addr,
                    "proto" => &mut scheme,
                    "host" => &mut host,
                    _ => &mut Some(""),
                }
                .get_or_insert_with(|| val.trim());
            });
```

aliemjay (approved):
LGTM. Very elegant. Less code and much easier to read.

src/info.rs
aliemjay: for more functional style

ibraheemdev: `for` loops are generally preferred over `for_each`, and that last match won't compile because of `_ => &mut Some(""),`. It's less about enforcing a function style and more about maintainability.

aliemjay: Oh, That really doesn't compile. I though the temporary would last until the first semicolon `;` - forgot about `}`.  But I still see `filter_map` would be nicer here ;-)

robjtede: i've hijacked this PR and incorporated the extra filter_map

aliemjay (approved):

PR #2283, opened 2021-06-23
Note that Form cannot require data ordering

jonasmalacofilho:
## PR Type

Docs

## PR Checklist

Note: I was unsure whether this should go into the changelog.

## Overview

URL encoded forms are always unordered `key=value` pairs.  Similarly to `Query`, note that trying to decode it into something like a tuple will fail at runtime.


robjtede (approved):
nice tweak, thanks 👍🏻

PR #2288, opened 2021-06-25
fix scope and resource middleware data access

### LINKED ISSUE #1790: Cannot access app_data in middleware of Scope

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
Fixes middleware's access to self-layer data on `Scope`s and `Resource`s using the approach outlined in #2261. Tests ensure that both routes and default services are wrapped with data injection in both cases.

Co-authored by: @aliemjay. Thanks for their work exploring both approaches to this; this one was chosen over #2267 because it more closely matches how data injection works for `App` and does not change public API.

Also technically removes an `IntoServiceFactory` impl for `Resource` which looks like a mistake because it allowed `Resource` to be used as a `default_service` which doesn't make sense. Intentionally not included in changelog.

supersedes #2267
supersedes #2261
closes #1790


src/resource.rs
```diff
@@ -400,34 +400,28 @@ where
...
             routes: self.routes,
-            app_data: self.app_data.map(Rc::new),
             default: self.default,
         });
 
-        self.endpoint
+        let resource_data = self.app_data.map(Rc::new);
+
+        // wraps endpoint service (including middleware) call and injects app data for this scope
+        let endpoint = apply_fn_factory(self.endpoint, move |mut req: ServiceRequest, srv| {
```
ibraheemdev, line 411: Is the type annotation (on `req`) needed?

robjtede: yep

```
454 |         let endpoint = apply_fn_factory(self.endpoint, move |mut req, srv| {
    |                                                              ^^^^^^^ consider giving this closure parameter a type
```

aliemjay (approved):
LGTM.

Have you considered not to add a middleware in the common case where `app_data == None`? I have this in #2264. Maybe micro-optimization?

robjtede:
> Have you considered not to add a middleware in the common case where app_data == None? I have this in #2264. Maybe micro-optimization?

Not concerned about that now, one branch vs one nested stack call is not worth the added complexity. If we want it can be added later without problems.

aliemjay:
> Also technically removes an `IntoServiceFactory` impl for `Resource` which looks like a mistake because it allowed `Resource` to be used as a `default_service` which doesn't make sense. Intentionally not included in changelog.

I already had this in mind. But my suggested solution is more radical: `Route` shouldn't have `impl ServiceFactory` because it makes no use of the defined guards. `Route::to` and `::default_service` should have the same signature with a new super trait with a blanket impl for `IntoServiceFactory` and `Handler`. This would also revert #2262.

PR #2294, opened 2021-06-28
add lazy data extractor

### LINKED ISSUE #2285: lazy data type
## Motivation
The ~now deprecated~ `App::data_factory` method ~was is not very useful because it ran the~ runs futures to completion before starting the server up. For semi-expensive constructions it can be beneficial to delay their creation until a request needs it for the first time.

For example: especially in development when using debug builds, operations such as key derivation are very slow and are often only needed for a small set of endpoints. If you can afford a single slow op (per-worker) at app startup then such a type is useful. Otherwise, constructing the type before server creation is necessary.

## Solution
A way to lazily initalize data maybe built in Tokio's new OnceCell type: `LazyData`.

ibraheemdev:
## PR Type
Feature

## PR Checklist

## Overview
Closes #2285

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ smallvec = "1.6"
 socket2 = "0.4.0"
 time = { version = "0.2.23", default-features = false, features = ["std"] }
 url = "2.1"
+tokio = { version = "1.7.1", features = ["sync"] }
 
 [dev-dependencies]
 actix-test = { version = "0.1.0-beta.3", features = ["openssl", "rustls"] }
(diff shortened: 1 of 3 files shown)


robjtede:
@ibraheemdev do you mind if I throw this in `actix-web-lab` for a bit?

ibraheemdev:
Sure, go ahead.

robjtede:
A preview of this is in `actix-web-lab`: https://docs.rs/actix-web-lab/0.2.2/actix_web_lab/web/struct.LazyData.html

PR #2296, opened 2021-06-28
clippy fixes

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Clippy fixes.

Also makes sure that `client_disconnect` is set for all types of server. Havent dug too deep on why they weren't set before but seems like they should be, they only feed into ServiceConfig.

### DIFF
--- a/actix-http/src/error.rs
+++ b/actix-http/src/error.rs
@@ -55,6 +55,8 @@ impl Error {
         Self::new(Kind::Io)
     }
 
+    // used in encoder behind feature flag so ignore unused warning
+    #[allow(unused)]
     pub(crate) fn new_encoder() -> Self {
         Self::new(Kind::Encoder)
     }
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,6 @@
 [alias]
 chk = "check --workspace --all-features --tests --examples --bins"
-lint = "clippy --workspace --tests --examples"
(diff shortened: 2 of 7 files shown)


popzxc (approved):
👍

actix-http/src/error.rs
```diff
@@ -55,6 +55,8 @@ impl Error {
         Self::new(Kind::Io)
     }
 
+    // used in encoder behind feature flag so ignore unused warning
+    #[allow(unused)]
     pub(crate) fn new_encoder() -> Self {
```
JohnTitor, line 60: Is there any chance this will be used in the future? If so that's fine, but if not, then I'd prefer `#[cfg(..)]` or `#[cfg_attr()]` to it (but yeah, the current may be a simple way).

JohnTitor (approved):
Looks like there're still some warnings (`clippy::needless_return` and `clippy::inconsistent_struct_constructor`), it'd be great if we could address them while we're here. Otherwise overall LGTM.

actix-http/src/error.rs
robjtede: To be clear, it is used in current code but only by modules behind "__compress" feature.

I don't really like the idea of changing this error type based on the feature flags, potentially makes other things more difficult in downstream crates. It's free to keep all error variants in place even if their feature is disabled given how it works.

PR #2301, opened 2021-07-05
Expose common types for App<T, _>

chinedufn:
Before this commit, many of the common monomorphizations of `App<T, B>`
where impossible to return from a function.

For example:

```rust
/// Create the actix-web application that can be used to start an HttpServer.
pub fn create_app() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Config = (),
        Response = ServiceResponse<StreamLog<Body>>,
        Error = actix_web::error::Error,
        InitError = (),
    >,
    StreamLog<Body>,
> {
    unimplemented!()
}
```

Would fail to compile since the `ServiceFactory` a a couple of other types were not publicly exported from actix-web.

---

In this commit, a few types that are commonly used when constructing an App are now exported.

This allows you to have a `fn create_app () -> App<T, B>` that can be re-used across your test suite.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,8 +96,10 @@ pub mod test;
 pub(crate) mod types;
 pub mod web;
 
+pub use actix_service::ServiceFactory;
+
 pub use actix_http::Response as BaseHttpResponse;
-pub use actix_http::{body, HttpMessage};
+pub use actix_http::{body, HttpMessage, Request};
 #[doc(inline)]
 pub use actix_rt as rt;
 pub use actix_web_codegen::*;
@@ -105,6 +107,7 @@ pub use actix_web_codegen::*;
 pub use cookie;
 
 pub use crate::app::App;
+pub use crate::app_service::AppEntry;
 pub use crate::error::{Error, ResponseError, Result};
 pub use crate::extract::FromRequest;
 pub use crate::request::HttpRequest;
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -11,7 +11,7 @@ pub use self::compat::Compat;
 pub use self::condition::Condition;
 pub use self::default_headers::DefaultHeaders;
 pub use self::err_handlers::{ErrorHandlerResponse, ErrorHandlers};
-pub use self::logger::Logger;
+pub use self::logger::{Logger, StreamLog};
 pub use self::normalize::{NormalizePath, TrailingSlash};
 
 #[cfg(feature = "__compress")]


robjtede:
we can see about exposing `ServiceFactory` through the `dev` module but the others are intentionally private types

robjtede:
```rust
pub fn create_app() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Response = ServiceResponse<impl MessageBody>,
        Config = (),
        InitError = (),
        Error = Error,
    >,
    impl MessageBody,
> {
    App::new()
        .wrap(Logger::default()) // <-- can be removed without affecting the return type
        .route("/", web::to(|| async { "hello" }))
}
```

chinedufn:
Thanks for the `impl MessageBody` bit.

I think your example still doesn't work currently since `ServiceFactory` isn't exposed.

Can we re-open this PR and I'll change it to re-export `ServiceFactory` from dev?

chinedufn:
Ah, one other issue. Your signature works fine, but it doesn't pass the trait bounds for `HttpServer::new()`

```
error[E0277]: the trait bound `App<impl ServiceFactory<ServiceRequest>, impl MessageBody>: actix_service::IntoServiceFactory<_, actix_http::request::Request>` is not satisfied
  --> crates/api-server/src/main.rs:20:5
   |
20 |     HttpServer::new(move || {
   |     ^^^^^^^^^^^^^^^ the trait `actix_service::IntoServiceFactory<_, actix_http::request::Request>` is not implemented for `App<impl ServiceFactory<ServiceRequest>, impl MessageBody>`
   |
   = help: the following implementations were found:
             <App<T, B> as actix_service::IntoServiceFactory<actix_web::app_service::AppInit<T, B>, actix_http::request::Request>>
   = note: required by `HttpServer::<F, I, S, B>::new`
```

```rust
    HttpServer::new(|| {
        create_app()
    })
```

I'm reading through the different trait bounds now to get myself familiar with how to get this working, but if you can suggest any tips here that would be awesome.

---

Also, I can't re-open the PR so I believe you'll have to.

---

Thanks!

---

**EDIT** - I think the issue is that the two `impl MessageBody` aren't guaranteed to be the same type, which is why this doesn't pass the `HttpServer::new()` trait bounds.

chinedufn:
Ok got the signature down

```rust
pub fn create_app<B: MessageBody>() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Config = (),
        Response = ServiceResponse<B>,
        Error = actix_web::error::Error,
        InitError = (),
    >,
    B,
> {
```

---

Now just need to re-export `ServiceFactory`. I've updated this PR branch with that change to re-export it from dev and I've removed the other exports.

robjtede:
there's a problem with reopening, you'll have to create a new pr

robjtede:
indeed it seems like there's no nice way to express the link between those two `impl MessageBody` types (yet) so the last middleware would need putting in `Compat` and use this:

```rust
pub fn create_app() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Response = ServiceResponse<AnyBody>,
        Config = (),
        InitError = (),
        Error = Error,
    >,
    AnyBody,
> {
    App::new()
        .wrap(Logger::default())
        .wrap(Compat::new(Logger::default()))
        .route("/", web::to(|| async { "hello" }))
}

fn main() {
    let _ = HttpServer::new(|| create_app());
}
```

robjtede:
Alternatively, (and this is what I did in my own projects) using `macro_rules!` is simple and easy way to share a create_app fn with main and tests.

```rust
macro_rules! create_app {
    () => {
        App::new()
            .wrap(Logger::default())
            .route("/", web::to(|| async { "hello" }))
    };
}

fn main() {
    let _ = HttpServer::new(|| create_app!());
}
```

chinedufn:
Yeah the macro was going to be my last resort.

Awesome, thanks for teaching me about Compat. Worked like a charm.

---

I'll open a new PR thanks!

PR #2302, opened 2021-07-07
Export actix_web::dev::ServiceFactory

chinedufn:
## PR Type

Re-export

## PR Checklist

## Overview

Enables:

```rust
pub fn create_app() -> App<
    impl ServiceFactory<
        ServiceRequest,
        Response = ServiceResponse<AnyBody>,
        Config = (),
        InitError = (),
        Error = Error,
    >,
    AnyBody,
> {
    App::new()
        .wrap(Compat::new(Logger::default())
        .route("/", web::to(|| async { "hello" }))
}
```

Related: #2301

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -29,7 +29,7 @@ pub use actix_http::{Extensions, Payload, PayloadStream, RequestHead, ResponseHe
 pub use actix_router::{Path, ResourceDef, ResourcePath, Url};
 pub use actix_server::Server;
 pub use actix_service::{
-    always_ready, fn_factory, fn_service, forward_ready, Service, Transform,
+    always_ready, fn_factory, fn_service, forward_ready, Service, ServiceFactory, Transform
 };
 
 pub(crate) fn insert_slash(mut patterns: Vec<String>) -> Vec<String> {


chinedufn:
The last thing that I still can't do:

```rust
fn create_request(body: MyType) -> Request<PayloadStream> {
    TestRequest::post()
        .uri("/index")
        .set_json(&body)
        .to_request()
}
```

This is useful when I have multiple tests all testing the same route.

Not a huge deal as I can just `-> TestRequest`, but I'm curious as to why re-exporting `Request` is an issue? It seems like it's public in `actix-http` so anyone could just depend on that if they really wanted it?

Still new to actix-web so apologies if I'm missing something.

Thanks!

chinedufn:
Mmm actually I'm not sure if I can just use `TestRequest` to handle all of my cases. 

Here's a situation where I (think?) I need `Request`.

```rust
/// The type returned from actix_web::test::init_service
pub type TestAppService = impl actix_web::dev::Service<
    actix_web::Request,
    Response = ServiceResponse,
    Error = actix_web::Error,
>;

pub async fn create_app_test_service_with_pool(pool: Pool) -> TestAppService {
    actix_web::test::init_service(api_test_app(pool)).await
}
```

Let me know if I should open a new issue for this instead of roping it into this PR.

`Request` is exported by `actix-http` so I guess if it's decided that it can't be re-exported by `actix-web` I can just depend on `actix-http`.

I don't know enough about these types to know that's supposed to be private and what isn't.

robjtede:
I'd like to hold off on re-exporting Request for now. You should import it through actix_http.

PR #2309, opened 2021-07-08
don't flush headers if no expect present

polachok:
## PR Type
Other

## PR Checklist

## Overview
Currently request headers are sent separately from the body (`.send()` method on framed flushes), which leads to a considerable slowdown in my [test](https://github.com/polachok/awc-test/tree/master). I changed it to only flush if there's no body, or otherwise flush with the body. See [test link](https://github.com/polachok/awc-test/tree/master) for wireshark screenshots and code.

First found while comparing hyper & awc client performance.

### DIFF
--- a/actix-http/src/client/h1proto.rs
+++ b/actix-http/src/client/h1proto.rs
@@ -83,12 +83,11 @@ where
         false
     };
 
-    framed.send((head, body.size()).into()).await?;
-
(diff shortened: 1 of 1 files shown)


robjtede:
GitHub said the conflict was too complex to resolve so I applied the diff into #2684. Thanks for the fix! 👍🏻

PR #2310, opened 2021-07-08
Send headers within the redirect requests.

### LINKED ISSUE #2308: awc: Doesn't send headers for redirect requests
When I set some headers for my request, I don't see them if the server redirects to another address (301, 302, 307, 308 and ... status codes)

## Expected Behavior
See the defined headers on all request redirects.

## Current Behavior
I just see the default headers (host, date and content-length)

To reproduce this we need a server and a client. The server code is a simple web server with one redirect and also to print request headers:
```
use actix_web::{web, App, Error, HttpRequest, HttpResponse, HttpServer, Responder};

async fn a(req: HttpRequest) -> impl Responder {
    HttpResponse::TemporaryRedirect()
        .body(format!("{:?}", req.headers()))
        .with_header(("Location", "/b"))
}

async fn b(req: HttpRequest) -> Result<impl Responder, Error> {
    Ok(format!("{:?}", req.headers()))
}

#[actix_rt::main]
    HttpServer::new(|| {
        App::new()
            .service(web::resource("/a").to(a))
            .service(web::resource("/b").to(b))
    })
    .bind("127.0.0.1:8081")?
    .run()
    .await
}
```

and the client is:
```
use awc::ClientBuilder;

#[actix_rt::main]
async fn main() {
    let link = "http://127.0.0.1:8081/a";
    let res = ClientBuilder::new()
        .max_redirects(0)
        .header("custom", "aaa")
        .finish()
        .get(link)
        .insert_header(("custom", "bbb"))
        .send()
        .await
        .unwrap()
        .body()
        .await
        .unwrap();

    let res = String::from_utf8_lossy(&res).to_string();

    println!("{}", res);
}
```
When you set `max_redirects` to `0` the output is similar to this:
```
HeaderMap { inner: {"date": Value { inner: ["DATE"] }, "content-length": Value { inner: ["0"] }, "custom": Value { inner: ["bbb"] }, "accept-encoding": Value { inner: ["br, gzip, deflate, zstd"] }, "host": Value { inner: ["127.0.0.1:8081"] }} }
```
When you set `max_redirects` to `10` (for example) the output is similar to this, which seems wrong and I EXPECT to see the one above:
```
HeaderMap { inner: {"content-length": Value { inner: ["0"] }, "host": Value { inner: ["127.0.0.1:8081"] }, "date": Value { inner: ["DATE"] }} }
```

## Your Environment
* Rust Version: 1.53
* Actix Web Version: 4.0.0-beta.8
* awc: 3.0.0-beta.7

omid:
## PR Type
Bug Fix

## PR Checklist

## Overview
Please check #2308

Closes #2308


awc/src/middleware/redirect.rs
```diff
@@ -347,4 +361,61 @@ mod tests {
...
+    #[actix_rt::test]
+    async fn test_redirect_headers() {
+        let srv = actix_test::start(|| {
+            async fn root(req: HttpRequest) -> HttpResponse {
+                if req
+                    .headers()
+                    .get("custom")
+                    .unwrap_or(&HeaderValue::from_str("").unwrap())
+                    == "value"
+                {
```
aliemjay, line 417:
```suggestion
                    .get("custom")
                    == Some("value")
                {
```

aliemjay (approved):
LGTM

awc/src/middleware/redirect.rs
omid: The output of get is `Option<&HeaderValue>`.
So it can be like: `== Some(&HeaderValue::from_str("value").unwrap())` if you think is better 🙈 
I can have this, but I prefer not, it's not safe enough: `== HeaderValue::from_str("value").ok().as_ref()`

aliemjay: If it is `Result`, then you can write  `== Ok("value")` because `HeaderValue: PartialEq<str>`. Anyhow this may be pedantic, you can ignore it. Thanks for the PR.

omid: Oh sry I wrote wrong, it's an `Option`. (fixed the comment above)
Anyway, for `== Some("value")` shows this error:

```
372 |                     == Some("value")
    |                             ^^^^^^^ expected struct `HeaderValue`, found `str`
```

aliemjay: Oh, sorry, then just ignore it.

robjtede (changes requested):
For security reasons, we should detect cross-origin redirects and omit sensitive headers.

- `Cookie`
- `Authorization`
- `Proxy-Authorization`
- `WWW-Authenticate`

omid:
@robjtede nice. let me try that :)
Does a logic like reqwest suit? https://github.com/seanmonstar/reqwest/blob/c6eb2c4fcbc3934b119772e0997d91671ff68dc7/src/redirect.rs#L234

What is "cookie2" in reqwest? 🤔

omid:
We have `PROXY_AUTHORIZATION` and `PROXY_AUTHENTICATE`.
@robjtede you referred to the second one, but in reqwest they remove the first one.

Which one is correct? Where can I find an official documentation/spec about this, couldn't find anything useful :/

awc/src/middleware/redirect.rs
```diff
@@ -148,87 +152,70 @@ where
...
                 uri,
                 method,
+                headers,
                 body,
                 addr,
                 connector,
             } => match ready!(fut.poll(cx))? {
                 ConnectResponse::Client(res) => match res.head().status {
-                    StatusCode::MOVED_PERMANENTLY
+                    status @ (StatusCode::MOVED_PERMANENTLY
```
omid, line 161: I also merged these two almost-the-same match branches.

awc/src/middleware/redirect.rs
```diff
@@ -148,87 +152,70 @@ where
...
-                        // reset method
-                        let method = method.take().unwrap();
-                        let method = match method {
-                            Method::GET | Method::HEAD => method,
-                            _ => Method::GET,
-                        };
+                        let prev_uri = uri.take().unwrap();
+
+                        // rebuild uri from the location header value.
+                        let next_uri = build_next_uri(&res, &prev_uri)?;
```
omid, line 174: used prev and next to avoid confusion.

omid:
> What is "cookie2" in reqwest?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie2
Should we remove it also? 🤔

awc/src/middleware/redirect.rs
```diff
@@ -156,79 +161,59 @@ where
                     StatusCode::MOVED_PERMANENTLY
                     | StatusCode::FOUND
                     | StatusCode::SEE_OTHER
+                    | StatusCode::TEMPORARY_REDIRECT
+                    | StatusCode::PERMANENT_REDIRECT
```
omid, line 165: I also merged the two almost-the-same match branches.

robjtede:
> We have `PROXY_AUTHORIZATION` and `PROXY_AUTHENTICATE`.
> @robjtede you referred to the second one, but in reqwest they remove the first one.
> 
> Which one is correct? Where can I find an official documentation/spec about this, couldn't find anything useful :/

Updated the list... `PROXY_AUTHENTICATE` is not a request header.

---

> What is "cookie2" in reqwest?

> developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie2
> Should we remove it also? 🤔

I would vote no.

robjtede:
Also looks like WWW-Authenticate is a response header too and won't need stripping.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

awc/src/middleware/redirect.rs
```diff
@@ -156,79 +161,59 @@ where
...
                         if *max_redirect_times > 0 =>
                     {
-                        let org_uri = uri.take().unwrap();
-                        // rebuild uri from the location header value.
-                        let uri = rebuild_uri(&res, org_uri)?;
+                        let is_redirect = res.head().status == StatusCode::TEMPORARY_REDIRECT
```
robjtede, line 168: is_redirect => should_reset_method

awc/src/middleware/redirect.rs
```diff
@@ -156,79 +161,59 @@ where
...
-                        // use a new request head.
-                        let mut head = RequestHead::default();
-                        head.uri = uri.clone();
-                        head.method = method.clone();
 
-                        let head = RequestHeadType::Owned(head);
-
-                        max_redirect_times -= 1;
+                        // reset method
+                        let method = if is_redirect {
```
robjtede, line 181: maybe i'm reading this wrong (and i think there isn't yet a test for this) but the 307 and 308 status codes should preserve the method, this looks like it's flipped the logic

omid: Added some more test cases 👍🏼

omid:
~~Some of the tests randomly fail, I couldn't find the main reason.~~
~~The error is coming from https://github.com/actix/actix-web/blob/master/actix-http/src/client/h1proto.rs#L113 and randomly appears for tests which are calling ".send_body()".~~

~~On Github, they fail most of the time, but on my local, about 25%.~~

Fixed in 8b89bb2 and d1c06eb

robjtede (approved):
lgtm now, nice work @omid

PR #2327, opened 2021-07-12
fix connection data on keep alive connections

### LINKED ISSUE #1770: Add on_connect extension data to all subsequent calls made through the connection
## Expected Behavior
I'm using on_connect to extract a TLS client certificate and add it to the Extensions data structure to expose it to API handlers via ReqData. I expect this data "attached" to extensions to be retrievable during all subsequent calls made to the API through the same TLS session.

## Current Behavior
The data attached to the Extensions structure is discarded after the first HTTP call. Subsequent HTTP calls yield calls to the rust API handlers I implemented, but without the TLS client certificate available through ReqData anymore.

## Possible Solution
Since one TLS connection can be reused multiple times, the Extensions data structure should be preserved for the duration of the TLS session, or there should be an alternative mechanism to attach data to the TLS session.

Call .on_connect on an HttpServer and have the function attach some data to the Extensions structure, like this:

    struct HelloWorldExt;
    fn attach_test(connection: &dyn Any, data: &mut Extensions) {
        data.insert(HelloWorldExt);
    }
    http_server.on_connect(attach_test);

Then implement an API handler like this:

    async fn api_handler(hwext: ReqData<HelloWorldExt>) { eprintln!("Success!"); ... }

This call will fail after the first time when made through the same TLS connection.

## Context
This issue prevents me from implementing authentication via TLS client certificates for my REST API.

## Your Environment
* Actix-web Version: 3.2.0

robjtede:
## PR Type

Fix

## PR Checklist

	- [x] Justify new unsafe code.

## Overview
A new CloneableExtensions object is introduced to that the 1-to-many relationship of connections-to-request can co-operate with `on_connect` handlers. In this way any on_connect data is cloned into the request extension map for each request.

Unfortunately this required a small amount of unsafe code to cast between CloneAny and Any traits so extra scrutiny should be given to this code. I've have done some tests with Miri and it did not complain.

Test that this fix is effective: `cargo run --example=on_connect` before and after, loading web browser to `localhost:8080`, observe difference between refresh (keep alive) and _force refresh_ (new connection). With this patch, refreshing normally displays connection info each time.

closes #1770


actix-http/src/extensions.rs
```diff
@@ -140,6 +142,107 @@ fn downcast_owned<T: 'static>(boxed: Box<dyn Any>) -> Option<T> {
...
+}
+
+impl UncheckedAnyExt for dyn CloneAny {}
+
+fn downcast_cloneable<T: 'static>(boxed: Box<dyn CloneAny>) -> T {
+    // Safety:
+    // Box is owned and `T` is known to be true type from map containing TypeId as key.
+    *unsafe { UncheckedAnyExt::downcast_unchecked::<T>(boxed) }
+}
+
```
ibraheemdev:
```suggestion
```
I think this method should be removed and the unsafe downcast should be moved to the place it is used and it is guaranteed that `T` is the true type. Right now this method could be used unsoundly, and even though it's not exposed, I think it's better to limit the scope of the unsafety.

actix-http/src/extensions.rs
```diff
@@ -140,6 +142,107 @@ fn downcast_owned<T: 'static>(boxed: Box<dyn Any>) -> Option<T> {
...
+    /// let mut map = Extensions::new();
+    /// assert_eq!(map.insert(""), None);
+    /// assert_eq!(map.insert(1u32), None);
+    /// assert_eq!(map.insert(2u32), Some(1u32));
+    /// assert_eq!(*map.get::<u32>().unwrap(), 2u32);
+    /// ```
+    pub fn insert<T: CloneAny>(&mut self, val: T) -> Option<T> {
+        self.map
+            .insert(TypeId::of::<T>(), Box::new(val))
+            .and_then(downcast_cloneable)
```
ibraheemdev:
```suggestion
        if let Some(old) = self.map.insert(TypeId::of::<T>(), Box::new(val)) {
            // Safety:
            // Box is owned and `T` is known to be the true type from the map containing TypeId as key.
            return Some(*unsafe { UncheckedAnyExt::downcast_unchecked::<T>(old) });
        }
        None
```
Per above.

ibraheemdev (changes requested):

robjtede:
Could do with an opinion on whether this is a good enough fix for v4 or wether we should seriously look at an alterative solution like connection-level state.

PR #2333, opened 2021-07-15
Remove concurrency from `FromRequest` tuples

ibraheemdev:
## PR Type
Refactor

## PR Checklist.
- [ ] Not sure about changelog and doc comments, it wasn't documented before..

## Overview
The current behavior for `FromRequest` is problematic, see https://github.com/rust-lang/futures-rs/issues/2201. This PR updates it to match the equivalent hand-written code `(a.await, b.await, c.await, ...)`. We lose concurrency this way, but it's better than the current solution. Per @fakeshadow, in the long run, we might want to look into a specialized `FuturesUnordered`.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -10,6 +10,7 @@ use std::{
 use actix_http::http::{Method, Uri};
 use actix_utils::future::{ok, Ready};
 use futures_core::ready;
+use futures_util::future::{maybe_done, MaybeDone};
 
 use crate::{dev::Payload, Error, HttpRequest};
 
@@ -297,7 +298,7 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
         /// A helper struct to allow us to pin-project through
         /// to individual fields
         #[pin_project::pin_project]
-        struct FutWrapper<$($T: FromRequest),+>($(#[pin] $T::Future),+);
+        struct FutWrapper<$($T: FromRequest),+>($(#[pin] MaybeDone<$T::Future>),+);
 
         /// FromRequest implementation for tuple
         #[doc(hidden)]
@@ -310,8 +311,7 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
 
             fn from_request(req: &HttpRequest, payload: &mut Payload) -> Self::Future {
                 $fut_type {
-                    items: <($(Option<$T>,)+)>::default(),
-                    futs: FutWrapper($($T::from_request(req, payload),)+),
+                    futs: FutWrapper($(maybe_done($T::from_request(req, payload)),)+),
                 }
             }
         }
@@ -319,7 +319,6 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
         #[doc(hidden)]
         #[pin_project::pin_project]
         pub struct $fut_type<$($T: FromRequest),+> {
-            items: ($(Option<$T>,)+),
             #[pin]
             futs: FutWrapper<$($T,)+>,
         }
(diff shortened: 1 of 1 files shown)


src/extract.rs
```diff
@@ -329,28 +328,28 @@ macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {
...
-                    if this.items.$n.is_none() {
-                        match this.futs.as_mut().project().$n.poll(cx) {
-                            Poll::Ready(Ok(item)) => {
-                                this.items.$n = Some(item);
+                    let mut fut = futs.$n.as_mut();
+
+                    if fut.as_mut().output_mut().is_none() {
+                        ready!(fut.as_mut().poll(cx));
+
+                        // this is a bit hacky, but we want to first check if there was an error
```
popzxc, line 340:
```suggestion
                        // This is a bit hacky, but we want to first check if there was an error
```

popzxc (approved):

robjtede:
I'd like to see some benchmark results to know if/how much of a hit removing concurrency has.

fakeshadow:
A bench would not show anything difference and it would just as fast.

But in real world it would make very big difference considering situations like this:
User extract 3 types of states according to request. The first one make an external api call with http client. The second one make a db query and let's say it's a remote db. The third would read the request body.

These 3 states are all from different IO source and can be handled concurrently. But since concurrency is removed they would take place in sequence. The user would find it's individual api latency spike up big time especially when server is not under heavy load.

robjtede:
> But in real world it would make very big difference

This is my concern, I don't think the quadtatic nature is so bad given we're extracting at most 10 (and far fewer on average) things so I'd want to see a real benefit to making the change.

fakeshadow:
I feel an alternative could be spawn every type with `tokio::task::spawn_local`(or accept https://github.com/actix/actix-net/pull/369 and use `actix_rt::spawn`) and sequentially await on the join handle.

A better solution would be something like this:
```rust
macro_rules! tuple_from_req ({$fut_type:ident, $(($n:tt, $T:ident)),+} => {

    // This module is a trick to get around the inability of
    // `macro_rules!` macros to make new idents. We want to make
    // a new `FutWrapper` struct for each distinct invocation of
    // this macro. Ideally, we would name it something like
    // `FutWrapper_$fut_type`, but this can't be done in a macro_rules
    // macro.
    //
    // Instead, we put everything in a module named `$fut_type`, thus allowing
    // us to use the name `FutWrapper` without worrying about conflicts.
    // This macro only exists to generate trait impls for tuples - these
    // are inherently global, so users don't have to care about this
    // weird trick.
    #[allow(non_snake_case)]
    mod $fut_type {

        // Bring everything into scope, so we don't need
        // redundant imports
        use super::*;

        /// FromRequest implementation for tuple
        #[doc(hidden)]
        #[allow(unused_parens)]
        impl<$($T: FromRequest + 'static),+> FromRequest for ($($T,)+)
        {
            type Error = Error;
            type Future = $fut_type<$($T),+>;
            type Config = ($($T::Config),+);

            fn from_request(req: &HttpRequest, payload: &mut Payload) -> Self::Future {
                let flag = Arc::new(AtomicUsize::new(!0));
                $fut_type {
                    items: ($((FutItem::Fut($T::from_request(req, payload)), Arc::new(ExtractWaker {
                        idx: 1 << $n,
                        flag: flag.clone(),
                        waker: AtomicWaker::new()
                    })),) +),
                    flag,
                }
            }
        }

        #[doc(hidden)]
        #[pin_project::pin_project]
        pub struct $fut_type<$($T: FromRequest),+> {
            /// FromRequest::Future and it's extract waker.
            items: ($((FutItem<$T>, Arc<ExtractWaker>),) +),
            /// A shared state between extract waker and FromRequest future.
            flag: Arc<AtomicUsize>,
        }

        /// A helper enum for storing individual FromRequest::Future and it's Output.
        #[pin_project::pin_project(project = FutItemProj, project_replace = FutItemReplace)]
        enum FutItem<T: FromRequest> {
            Fut(#[pin] T::Future),
            Output(T),
            Done,
        }

        impl<$($T: FromRequest),+> Future for $fut_type<$($T),+>
        {
            type Output = Result<($($T,)+), Error>;

            fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
                let this = self.project();

                let mut ready = true;
                let mut flag = this.flag.load(Ordering::SeqCst);

                // keep a record for the flag. it would be used for cas operation later.
                let mut cur = flag;

                $(
                    let (fut, waker) = &mut this.items.$n;
                    match unsafe { Pin::new_unchecked(&mut *fut) }.project() {
                        FutItemProj::Fut(f) => {
                            // check the waker index's bit in flag state.
                            // A from request future is only ready to be polled when it's bit in flag is 1.
                            if flag & waker.idx == waker.idx {
                                flag &= !waker.idx;
                                // update extract waker's origin waker from current context.
                                waker.waker.register(cx.waker());

                                // poll from request future with extract waker only.    
                                let waker = Arc::clone(waker).into();
                                let cx = &mut Context::from_waker(&waker);
                                match f.poll(cx) {
                                    Poll::Ready(Ok(item)) => {
                                        let _ = unsafe { Pin::new_unchecked(fut) }.project_replace(FutItem::Output(item));
                                    }
                                    Poll::Pending => ready = false,
                                    Poll::Ready(Err(e)) => return Poll::Ready(Err(e.into())),
                                }
                            } else {
                                ready = false
                            }
                        }
                        _ => {}
                    }
                )+

                if ready {
                    Poll::Ready(Ok(
                        ($(
                            match unsafe { Pin::new_unchecked(&mut this.items.$n.0) }.project_replace(FutItem::Done) {
                                FutItemReplace::Output(item) => item,
                                _ => unreachable!()
                            }
                            
                        
                        ,)+)
                    ))
                } else {
                    // This loop should be very short and it should be at worst case iter for max items size times.
                    while let Err(c) = this.flag.compare_exchange_weak(cur, flag, Ordering::SeqCst, Ordering::SeqCst) {
                        // one fut item is about to wake up from request future.
                        // add it's state to flag and try update flag again.
                        cur = c;
                        flag |= c;
                    };

                    Poll::Pending
                }
            }
        }
    }
});

use futures_core::task::__internal::AtomicWaker;
use std::sync::{
    atomic::{AtomicUsize, Ordering},
    Arc,
};
use std::task::Wake;

pub struct ExtractWaker {
    idx: usize,
    flag: Arc<AtomicUsize>,
    waker: AtomicWaker,
}

impl Wake for ExtractWaker {
    fn wake(self: Arc<Self>) {
        self.flag.fetch_or(self.idx, Ordering::Relaxed);
        self.waker.wake();
    }

    fn wake_by_ref(self: &Arc<Self>) {
        self.flag.fetch_or(self.idx, Ordering::Relaxed);
        self.waker.wake();
    }
}
```

Benefit of this is it would be able to be embedded in async block and runs futures that are not 'static lifetime concurrently like FuturesUnordered

robjtede:
actix/actix-net#369 did indeed get merged so could have another crack at this idea if it still seems necessary (i don't think so yet)

running extractors sequentially, however, is a non-starter for the reasons indicated above

PR #2337, opened 2021-07-19
rework `ResourceMap` internals

aliemjay:
## PR Type

Refactor

## PR Checklist

## Overview

Restructure `ResourceMap` to a simpler tree where each node is a `ResourceMap` instance even for terminal resources, which used to be stored as `ResourceDef`s in a vec).

This also clones the hash map for named resources from lower to higher layer when constructing the tree, which makes finding named resource possible in a single hash map search instead of recursively (rough 10% speedup in `url_for` benches).

Possible behavior change:
`ResourceMap::has_resource` used to return true when called on a external resource. However, this behavior is not documented (the docs for `external_resource()` mention only `HttpResquest::url_for`) and is broken when external resource is defined in `ServiceConfg::external_resource` vs `App::external_resource`, so I deemed it unimportant.


aliemjay:
Breaking changes TODO:

1. Make `::{new, add}` private. The user cannot currently construct ResourceMap anyway because `finish()` is not exposed.

2. ~~Make `::finish()` an associated function.~~

3. Replace `::add()` with more specific methods (e.g. `add_external_resource`, `add_nested`, `add_resource`) 

4. `::add()` should receive owned `ResourceMap` instead of `Rc<_>`. Otherwise we could have the same node in different trees.

aliemjay:
awaiting update to router 0.5 in #2339

src/rmap.rs
```diff
@@ -350,14 +285,14 @@ mod tests {
...
-        root.finish(Rc::clone(&root));
+        ResourceMap::finish(&root);
 
         // sanity check resource map setup
 
         assert!(root.has_resource("/info"));
         assert!(!root.has_resource("/bar"));
 
-        assert!(root.has_resource("/user/22"));
+        assert!(!root.has_resource("/user/22"));
```
aliemjay, line 293: router doesn't match this path. has_resource should be consistent with this.

robjtede (approved):

PR #2338, opened 2021-07-20
Add `ServiceConfig::default_service`

lily-mara:
## PR Type

Feature

## PR Checklist

## Overview

The `App` struct supports a "default" service which will be called when there
are no other matching services available. I propose that we support adding this
default service from the `ServiceConfig` struct, which will enable it to be used
from `App::configure`. Currently if you want to test a default service (if
you're creating a proxy for example) you must re-specify this default service
each time you construct your app. This change will allow users to put their
default services into the shared `App::configure` helper.


src/app.rs
```diff
@@ -205,6 +205,10 @@ where
         self.services.extend(cfg.services);
         self.external.extend(cfg.external);
         self.extensions.extend(cfg.app_data);
+        if let Some(default) = cfg.default {
+            self.default = Some(default);
+        }
+
```
aliemjay, line 211: We should have this for `Scope::configure` as well.

aliemjay (changes requested):

robjtede:
The source branch does not allow maintainer edits so i've carried on the work in #2743 with attribution back here.

Thanks for the contribution.

PR #2344, opened 2021-07-23
Fix quality parse error in Accept-Encoding HTTP header

### LINKED ISSUE #2322: middleware::Compress does not handle accept-encoding header correctly
The middleware to compress the response body interprets the `accept-encoding` headers incorrectly. It should choose the compression algorithm with the highest quality (`q`) value from the algorithms it understands. If it understands none of them, it should fall back to an uncompressed response, unless that has a quality value of zero. It should never use an algorithm that has a quality value of zero.

## Steps to Reproduce

Run the following program:

```rust
use actix_web::{middleware, web, App, HttpResponse, HttpServer};

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .route("/", web::get().to(|| HttpResponse::Ok().body("Test\r\n")))
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

Send requests with different `accept-encoding` headers (see below) to the server. I used the Linux command line to send requests and observe the response:

```sh
printf 'GET / HTTP/1.1\r\naccept-encoding: gzip;q=0.4, br;q=0.6\r\n\r\n' | nc -N localhost 8080
```

## Expected and Current Behavior

The following cases demonstrate the issue:

* Send header `accept-encoding: gzip;q=0.4, br;q=0.6`
  * Expected: Response is `br` compressed
  * Actual: Response is `gzip` compressed
* Send header `accept-encoding: gzip;q=0`
  * Expected: Response is uncompressed
  * Actual: Response is `gzip` compressed
* Send header `accept-encoding: compress, identity;q=0`
  * Expected: Request is rejected with a `406 Not Acceptable` status (uncompressed response is not acceptable to the client)
  * Actual: Response is uncompressed
* Send header `accept-encoding: compress, *;q=0`
  * Expected: Request is rejected with a `406 Not Acceptable` status (any encoding other than `compress`, including an uncompressed response, is not acceptable to the client)
  * Actual: Response is uncompressed

## Environment

* Rust Version: 1.53.0
* Actix Web Version: 4.0.0-beta.8

arthurlm:
## PR Type

Bug Fix

## PR Checklist

## Overview

HTTP header `Accept-Encoding` was not correctly parsed.

Other changes:

- Better respect of client algorithms choices.
- Follow MDN spec about default qfactor value.
- It client need compressed response, now server send 406 HTTP error if there is no available algorithms.
- Make `ContentEncoding::parse` fail instead of returning default value.

### Issue open to discussion

I have change `CompressMiddleware` trait requirements.
```diff
- B: MessageBody
+ B: MessageBody + From<String>
```

Indeed, I have not been able to use the `Either` trait correctly to answer available compress algorithm when we have to answer NotAcceptable status code.

So the only way I have been able to work with is adding `From<String>` requirement to  `B`.

This make `test common_combinations` fail (see: https://github.com/actix/actix-web/commit/dade818ebaab441e8cc5d359068209daa002b488). So I have disable it ...
If anyone has a good idea to fix this, I will be happy to edit my PR.

Closes #2322

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,13 +5,11 @@
 * Re-export actix-service `ServiceFactory` in `dev` module. [#2325]
 
 ### Changes
+* Fix quality parse error in Accept-Encoding HTTP header. [#2344]
 * Minimum supported Rust version (MSRV) is now 1.51.
 
 [#2325]: https://github.com/actix/actix-web/pull/2325
-
-### Changed
-* Fix quality parse error in Accept-Encoding HTTP header. [#2344]
-
+[#2344]: https://github.com/actix/actix-web/pull/2344
 
 ## 4.0.0-beta.8 - 2021-06-26
 ### Added
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,12 +4,16 @@
 ### Added
 * Re-export actix-service `ServiceFactory` in `dev` module. [#2325]
 
-### Changes
-* Fix quality parse error in Accept-Encoding HTTP header. [#2344]
+### Changed
 * Minimum supported Rust version (MSRV) is now 1.51.
+* Compress middleware will return 406 Not Acceptable when no content encoding is acceptable to the client. [#2344]
+
+### Fixed
+* Fix quality parse error in Accept-Encoding header. [#2344]
 
 [#2325]: https://github.com/actix/actix-web/pull/2325
 [#2344]: https://github.com/actix/actix-web/pull/2344
+
 
 ## 4.0.0-beta.8 - 2021-06-26
 ### Added
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@
 
 [#2325]: https://github.com/actix/actix-web/pull/2325
 
+### Changed
+* Fix quality parse error in Accept-Encoding HTTP header. [#2344]
+
 
 ## 4.0.0-beta.8 - 2021-06-26
 ### Added
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -16,6 +16,8 @@
 ### Fixed
 * Potential HTTP request smuggling vulnerabilities. [RUSTSEC-2021-0081](https://github.com/rustsec/advisory-db/pull/977)
 
+### Changed
+* Fix quality parse error in Accept-Encoding HTTP header. [#2344]
 
 ## 3.0.0-beta.8 - 2021-06-26
 ### Changed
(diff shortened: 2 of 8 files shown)


robjtede (approved):
nice change, thanks for the contribution

robjtede:
I was able to sort out the problems you outlined in the description. Indeed adding the `From<String>` bound would have been unacceptable.

istr:
While https://github.com/actix/actix-web/issues/2322 is a valid bug report, I would tend to say that some of the _additional_ changes made with this PR are a _regression_ over the previous version.

- Neither [the HTTP RFC](https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding) nor the (only informative) [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) "spec" state that the _order_ of equally weighted options has any significance.
- Hence it is still _up to the server_ to decide on the best option for equally weighted preferences. `gzip,br,zstd`, `br,zstd,gzip`, and `zstd,gzip,br` are _completely equivalent_ (in each of the cases the client gives an implict `q=1` for each option).
- Given this, the test cases https://github.com/actix/actix-web/blob/06c3513bc0da212d87c6fae5332e845146991027/actix-web/tests/compression.rs#L94 https://github.com/actix/actix-web/blob/06c3513bc0da212d87c6fae5332e845146991027/actix-web/tests/compression.rs#L123 https://github.com/actix/actix-web/blob/06c3513bc0da212d87c6fae5332e845146991027/actix-web/tests/compression.rs#L152 are not only _misleading_. I would say they are plain wrong (ymmv).
- Instead of this, the _server_ should have a deterministic preference. These tests should verify that the server determines the preference of equally weighted options, _regardless of the order_ in which the client sent them.
- You could still argue that the server's preference is based on the order of equally weighted options. Still, like with TLS handshake it is desirable that the server will go for the _best_ option at hand.
- It was a good idea to have a built-in server preference, however the implementation was questionable:
```rust
    /// Default Q-factor (quality) value.
    #[inline]
    pub fn quality(self) -> f64 {
        match self {
            ContentEncoding::Br => 1.1,
            ContentEncoding::Gzip => 1.0,
            ContentEncoding::Deflate => 0.9,
            ContentEncoding::Identity | ContentEncoding::Auto => 0.1,
            ContentEncoding::Zstd => 0.0,
        }
    }
```

@robjtede could we try to recover the old server-side preference behavior? Or add an option to pass the preferred order to `middleware::Compress` somehow?

PR #2353, opened 2021-08-01
`actix-files` Do not send a message body in the NOT_MODIFIED case

max-m:
## PR Type

Bug Fix

## PR Checklist

I did not touch the changelog file because I was unsure whether this PR should have been made against the master branch.
This issue probably affects the current beta version as well.

## Overview

As per RFC7230 section 3.3.2 <https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2>:

> A server MAY send a Content-Length header field in a 304 (Not
> Modified) response to a conditional GET request (Section 4.1 of
> [RFC7232]);
> a server MUST NOT send Content-Length in such a response
> unless its field-value equals the decimal number of octets that would
> have been sent in the payload body of a 200 (OK) response to the same
> request.

And per RFC7232 section 4.1 <https://datatracker.ietf.org/doc/html/rfc7232#section-4.1>:

> A 304 response cannot contain a message-body; it is always terminated
> by the first empty line after the header fields.

Previously a `Content-Length: 0` header was sent because `.finish()`
sets the body to `Body::Empty`, which indicates a zero-sized response.

### DIFF
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -8,7 +8,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use std::os::unix::fs::MetadataExt;
 
 use actix_web::{
-    dev::{BodyEncoding, SizedStream},
+    dev::{Body, BodyEncoding, SizedStream},
     http::{
         header::{
             self, Charset, ContentDisposition, DispositionParam, DispositionType,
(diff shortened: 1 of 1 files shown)


aliemjay (approved):
Can you add tests and update changelog?

max-m:
Of course.
How would you like me to update the PR? Should I amend the current commit or make a commit for the `CHANGES.md` and another for the test?
I assume tests for this should be added to a new file in `actix-files/tests` because currently there’s only a single test in its `encoding.rs`.

robjtede:
we're not accepting patches into the 3.0 branch, please target master

robjtede:
superceded by #2453
thanks for the submission

PR #2355, opened 2021-08-07
ResourceDef: define behavior for prefix with trailing slash

aliemjay:
## PR Type

Fix

## PR Checklist

## Overview

Redefine so that prefixes with a trailing slash are no longer a special case. See the docs for details.

This allows us to have a simpler solution to actix/actix-net#376

### DIFF
--- a/actix-files/src/files.rs
+++ b/actix-files/src/files.rs
@@ -106,7 +106,7 @@ impl Files {
         };
 
         Files {
-            path: mount_path.to_owned(),
+            path: mount_path.trim_end_matches('/').to_owned(),
             directory: dir,
             index: None,
             show_index: false,
--- a/actix-router/CHANGES.md
+++ b/actix-router/CHANGES.md
@@ -5,11 +5,14 @@
 * Disallow prefix routes with tail segments. [#379]
 * Enforce path separators on dynamic prefixes. [#378]
 * Improve malformed path error message. [#384]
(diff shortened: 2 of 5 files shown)


actix-files/src/files.rs
```diff
@@ -106,7 +106,7 @@ impl Files {
         };
 
         Files {
-            path: mount_path.to_owned(),
+            path: mount_path.trim_end_matches('/').to_owned(),
```
aliemjay, line 109: Many tests in `files` failed because they use `File::new("/", _)`. A similar thing is needed for `Scope` if we want to keep this behavior.

robjtede: it's definitely reasonable to support this format where we're defining a service at the root path

aliemjay: yes, trimming a trailing slash would perfectly emulate the old behavior.  note that scope is arguably already broken by the update to router 0.5: `scope("/").route("/")` wouldn't match the root path.

aliemjay: Ideally I want the users to use empty prefixes but that is not what most users expect. So the solution I propose is to just "normalize" prefix pattern (by removing trailing slash). This way we get the benefit of greatly simplifying ResourceDef logic and not breaking user expectations.

The question is whether to normalize it in acix-web or in ResourceDef constructors?

robjtede (approved):

PR #2391, opened 2021-09-30
add responder for unit

axos88:
## PR Type

Feature

## PR Checklist

## Overview

Added a responder for () that will respond with 204 no content

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased - 2021-xx-xx
 ### Added
 * Option to allow `Json` extractor to work without a `Content-Type` header present. [#2362]
+* Implemented responder for `()` to respond with `204 No Content`
 
 ### Changed
 * Associated type `FromRequest::Config` was removed. [#2233]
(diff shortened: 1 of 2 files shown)


robjtede:
See the following for rationale on why this is a dangerous impl.
- https://github.com/actix/actix-web/pull/1167
- https://github.com/actix/actix-web/issues/1108

axos88:
@robjtede That makes sense. How would you feel about implementing it for `Result<(), E: Into<Error>>` though?

Usually one would keep the ability to return errors anyways, and this way one would have to explicitly return an `Ok(())`, a stray semicolon would not create an issue.

And maybe discourage the usage of `-> impl Responder`? Feels like trying to force dynamic typing in a statically typed language to me. It was a huge help before `async fn`, because of all the futures wrapped and wrapped and wrapped, but now that we can simply `.await` on them I'm not sure there's a rationale of not explicitly stating the return type of a handler.

```
//Note see https://github.com/actix/actix-web/issues/1108 on why Responder is not implemented for ()
impl<E: Into<Error>> Responder for Result<(), E> {
    fn respond_to(self, req: &HttpRequest) -> HttpResponse {
        match self {
            Ok(()) => HttpResponse::new(StatusCode::NO_CONTENT),
            Err(e) => HttpResponse::from_error(e.into())
        }
    }
}
```

PR #2396, opened 2021-10-10
Bump dependencies to rustls 0.20

### LINKED ISSUE #2304: Bump Rustls to 0.20 for v4.0.0 release
Hey actix team!

Rustls is near approaching a 0.20 release (with two betas already released), which has significant breaking changes. Since actix is also close to releasing a major version release as well, I thought It was worth requesting if actix can update its dependency for Rustls to 0.20 before the 4.0.0 release is made.

As for reasons why actix should consider using Rustls 0.20 to begin with:
 - Obligatory mention of "standard best practice" to keep security libraries up to date in general.
 - If #2054 is implemented for v4 (which appears to be on the roadmap), then updating Rustls to 0.20 after a 4.0.0 release is cut would require a major version increment as the API changes are not backwards compatible.
 - No major breaking changes are envisioned after the 0.20 release, so updating this dependency before 4.0.0 is released is perfect timing.
 - There's an huge performance benefit for actix-web servers using runtime-selected certificates that removes a clone operation on a cert chain for every TLS connection, which I've personally run into. Unfortunately, without an dependency update to 0.20, this isn't resolvable by an end user from my investigation.

I'd be happy to try and create a PR for this as well since I'm advocating for this change, if the actix team deems this a good idea but does not have the time to implement it.

Thanks for your consideration.

edward-shen:
## PR Type

Feature

## PR Checklist

## Overview

This draft PR is a WIP meant for tracking progress on bumping `rustls` to 0.20. Note that this will almost certainly require a blocking PR in the `actix-net` repo, which will almost certainly be blocked on something upstream.

Closes #2304.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,7 @@ zstd = "0.7"
 rand = "0.8"
 rcgen = "0.8"
 tls-openssl = { package = "openssl", version = "0.10.9" }
-tls-rustls = { package = "rustls", version = "0.19.0" }
+tls-rustls = { package = "rustls", version = "0.20" }
 
 [profile.dev]
 # Disabling debug info speeds up builds a bunch and we don't rely on it for debugging that much.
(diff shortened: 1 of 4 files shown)


edward-shen:
Update: actix/actix-net#396 is out which should address the primary blocker of `actix-test` not correctly compiling. This branch will be rebased when a beta for `actix-tls` is cut with those changes.

edward-shen:
Closing in favor of #2416.

PR #2398, opened 2021-10-11
files: percent-decode url path

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

`re.match_info().path()` may still have some percent-encoded chars.

This is partly a regression after upgrading actix-router because of actix/actix-net#357


actix-files/src/service.rs
```diff
@@ -77,11 +77,13 @@ impl Service<ServiceRequest> for FilesService {
             )));
         }
 
-        let real_path =
-            match PathBufWrap::parse_path(req.match_info().path(), self.hidden_files) {
-                Ok(item) => item,
-                Err(e) => return Box::pin(ok(req.error_response(e))),
-            };
+        let path_decoded =
+            percent_encoding::percent_decode_str(req.match_info().path()).decode_utf8_lossy();
```
robjtede: This is a potential security problem. Would prefer a failed decode to return a 400 response.

aliemjay: Will do! But anyway lossy decoding is unreachable here anyway because only `b"%/+"` are left encoded and I can't see how decoding these could corrupt a utf8 string.

Also. router already does lossy decoding (actix-router/src/url.rs), you may want to re-evaluate this!

aliemjay: More importantly, now I think we maybe should disallow decoding `%2f` into `/`?

svenstaro:
Needs a rebase!

robjtede:
> Needs a rebase!

The question has been brought up around how we normalize paths in other places, so this is on hold for now.

I'm following some conversations [like this one](https://github.com/whatwg/url/issues/606) to get a bigger picture of how we can correctly and consistently handle escaped reserved and unreserved characters.

actix-files/src/path_buf.rs
```diff
@@ -26,6 +26,10 @@ impl PathBufWrap {
     pub fn parse_path(path: &str, hidden_files: bool) -> Result<Self, UriSegmentError> {
         let mut buf = PathBuf::new();
 
+        let path = percent_encoding::percent_decode_str(path)
```
karwa, line 34: Paths in URLs are lists of percent-encoded components, not percent-encoded lists of components. Decoding an entire path string can introduce new components by decoding forward/back-slashes.

For example: `https://example.com/music/bands/AC%2FDC`

Decode then split results in the components: `["music", "bands", "AC", "DC"]`
Split then decode results in the correct components: `["music", "bands", "AC/DC"]`

If you have a component like `foo%2F..%2F..%2F..`, decode-then-split would introduce new ".." components, which can be a security risk as it defeats any previous validation of the path.

(This is just a drive-by comment because I saw this issue linked from somewhere, and recognised this as a common misunderstanding)

aliemjay: Indeed I'm aware of this https://github.com/actix/actix-web/pull/2398#discussion_r726706854.

What is arguable though is the assumption that there is a one-to-one mapping between url path components and filesystem path components. In my opinion, a file server can safely interpret a single url path component to be multiple filesystem  path components as long as this transformation is done before any validation step (which is the case here). So I don't really see a problem here.

karwa: Ah - I missed your comment about it. There's no standard describing how this conversion should be done, and all kinds of funky edge-cases, so validating the result is definitely a good idea 👍.

FWIW, Chromium [will reject](https://github.com/chromium/chromium/blob/87ca5380f1d05f7218edd5659ed857b16a0bb2d0/net/base/filename_util.cc#L59) such URLs from being turned in to file paths. In my own library (for the Swift language), I also reject percent-encoded NULL bytes, because ultimately you're likely to be talking to the OS using null-terminated strings. Windows paths are significantly more complex with all kinds of additional edge-cases.

aliemjay: I followed [the discussion](https://crbug.com/585422) for chrome and couldn't see any security argument being made to justify the behavior. Meanwhile, firefox, nginx and `python -m http.server` do decode `%2F` into `/` for file paths. So I guess we're on the safe side!

For NULL chars, we rely on the std lib to reject it.

aliemjay:
@robjtede Is this semver-major because of the behavior change? I ask because UriSegmentError is not public so it shouldn't be breaking.

robjtede:
After some thought about this, I think the best way to move this forward is to just decode the remaining chars we want to, excluding `/`. I just don't see a good reason to allow decoding this character and the security doubts make it a good choice to exclude in my opinion.

Since `percent-encoding` can't target character sets for decoding, we can use the `Quoter` from `actix-router`.

Thoughts on this patch @aliemjay ?

```diff
diff --git a/actix-files/Cargo.toml b/actix-files/Cargo.toml
index 745e3afee..8b7a83f96 100644
--- a/actix-files/Cargo.toml
+++ b/actix-files/Cargo.toml
@@ -25,6 +25,7 @@ experimental-io-uring = ["actix-web/experimental-io-uring", "tokio-uring"]
 actix-http = "3.0.0-beta.17"
 actix-service = "2"
 actix-utils = "3"
+actix-router = "0.5.0-beta.3"
 actix-web = { version = "4.0.0-beta.18", default-features = false }
 
 askama_escape = "0.10"
diff --git a/actix-files/src/error.rs b/actix-files/src/error.rs
index 242eda4be..a7da6452e 100644
--- a/actix-files/src/error.rs
+++ b/actix-files/src/error.rs
@@ -27,15 +27,14 @@ pub enum UriSegmentError {
     /// The segment started with the wrapped invalid character.
     #[display(fmt = "The segment started with the wrapped invalid character")]
     BadStart(char),
+
     /// The segment contained the wrapped invalid character.
     #[display(fmt = "The segment contained the wrapped invalid character")]
     BadChar(char),
+
     /// The segment ended with the wrapped invalid character.
     #[display(fmt = "The segment ended with the wrapped invalid character")]
     BadEnd(char),
-    /// The path is not a valid UTF-8 string after doing percent decoding.
-    #[display(fmt = "The path is not a valif UTF-8 string after percent-decoding")]
-    NotValidUtf8,
 }
 
 /// Return `BadRequest` for `UriSegmentError`
diff --git a/actix-files/src/lib.rs b/actix-files/src/lib.rs
index 5d141d6ea..a32e19d2a 100644
--- a/actix-files/src/lib.rs
+++ b/actix-files/src/lib.rs
@@ -809,7 +809,7 @@ mod tests {
             .uri("/test/%2F..%2F..%2Ftests%2Ftest.binary")
             .to_request();
         let res = test::call_service(&srv, req).await;
-        assert_eq!(res.status(), StatusCode::OK);
+        assert_eq!(res.status(), StatusCode::NOT_FOUND);
 
         let req = TestRequest::get().uri("/test/Cargo.toml%00").to_request();
         let res = test::call_service(&srv, req).await;
diff --git a/actix-files/src/path_buf.rs b/actix-files/src/path_buf.rs
index e46d3f232..2e6a68ab0 100644
--- a/actix-files/src/path_buf.rs
+++ b/actix-files/src/path_buf.rs
@@ -3,11 +3,17 @@ use std::{
     str::FromStr,
 };
 
+use actix_router::Quoter;
 use actix_utils::future::{ready, Ready};
 use actix_web::{dev::Payload, FromRequest, HttpRequest};
 
 use crate::error::UriSegmentError;
 
+thread_local! {
+    /// Replaces remaining reserved characters except `%2F` (forward slash "/").
+    static FILENAME_QUOTER: Quoter = Quoter::new(b"%+", b"/");
+}
+
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) struct PathBufWrap(PathBuf);
 
@@ -26,10 +32,6 @@ impl PathBufWrap {
     pub fn parse_path(path: &str, hidden_files: bool) -> Result<Self, UriSegmentError> {
         let mut buf = PathBuf::new();
 
-        let path = percent_encoding::percent_decode_str(path)
-            .decode_utf8()
-            .map_err(|_| UriSegmentError::NotValidUtf8)?;
-
         for segment in path.split('/') {
             if segment == ".." {
                 buf.pop();
@@ -48,7 +50,11 @@ impl PathBufWrap {
             } else if cfg!(windows) && segment.contains('\\') {
                 return Err(UriSegmentError::BadChar('\\'));
             } else {
-                buf.push(segment)
+                let decoded_segment = FILENAME_QUOTER
+                    .with(|q| q.requote(segment.as_bytes()))
+                    .unwrap_or_else(|| segment.to_owned());
+
+                buf.push(decoded_segment)
             }
         }
```

actix-files/src/path_buf.rs
robjtede: @karwa thanks for your insight on this 👍🏻

aliemjay:
We need then to ouright reject these paths. Leaving some encoded chars seems wrong to me. 

Will look into it later today.

robjtede:
They _are_ being rejected with the 404. The only edge case left is if the fs really does contain a file with `%2F` in the name. And if so, you're 1) asking for trouble 2) possibly really actually want to serve such a stupid file.

IMO, this is such a niche case that it is not worth the performance hit of re-scanning the decoded segment for `"%2F"`. Plus, it's not really a security problem to leave it.

actix-files/src/path_buf.rs
```diff
@@ -26,8 +26,21 @@ impl PathBufWrap {
     pub fn parse_path(path: &str, hidden_files: bool) -> Result<Self, UriSegmentError> {
         let mut buf = PathBuf::new();
 
+        // equivalent to `path.split('/').count()`
+        let mut segment_count = path.matches('/').count() + 1;
+
+        let path = percent_encoding::percent_decode_str(path)
```
robjtede, line 34: needs a comment in here about why the "full" decode is acceptable

robjtede (approved):
cool, this approach is fine

aliemjay:
Files with `%2F` in the name should be already encoded to `%252F`, so this shouldn't be a 
problem.

I also added some code to double-check the resulting path using `std::path::Path` because of Windows, which seems to have many edge cases.

One Idea I have now is to expose PathBufWrap (under a better name ofc).

PR #2402, opened 2021-10-11
Replace askama_escape in favor of v_htmlescape

zzau13:
## PR Type
Refactor/Other

## PR Checklist
* [x]  A changelog entry has been made for the appropriate packages.
* [x]  Format code with the latest stable rustfmt.
* [x]  (Team) Label with affected crates and semver status.
  
## Overview
I have already uncoupled the crates `bytes` from `v_escape`. This version is several times faster as it is optimized with SIMD instructions.

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -1,7 +1,7 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-
+* Replace `askama_escape` with `v_htmlescape`
 
 ## 0.6.0-beta.7 - 2021-09-09
 * Minimum supported Rust version (MSRV) is now 1.51.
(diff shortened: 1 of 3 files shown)


robjtede:
I decided on askama_escape in [this PR](https://github.com/actix/actix-web/pull/1953) because of it's better compile time, is v_htmlescape any closer to this goal now?

The rationale is that this is not called very often at all and as such it is better for consumers of actix-web to have improved compile times when using actix-files.

zzau13:
It is not so much and yes, I can reduce the compilation time a lot, there are many things that were for testing. 
image

I will notify you when I make the change.

robjtede:
just doing some clean up, feel free to re-open if things change

PR #2407, opened 2021-10-12
web: implement `into_inner` for `Data<T: ?Sized>`

### LINKED ISSUE #2403: Data::<T>::into_inner requires `T: Sized` unnecessarily
## Expected Behavior
I have some code that currently looks something like:

```rust
trait MyTrait {
   // ...
}

#[post("/some-path")]
pub async fn do_something(data: Data<dyn MyTrait>) {
    Container(data.into_inner()).do_something();
}

struct Container(Arc<dyn MyTrait>);

impl Container {
   // ...
}
```

However, this fails to compile since `into_inner` is only implemented for `Data<T>` (which has an implicit `Sized` bound on `T`).

## Current Behavior
The above example doesn't compile since  `into_inner` is only implemented for `Data<T>` (which has an implicit `Sized` bound on `T`). Instead, the handler has to call:

```rust
Container((*data).clone()).do_something()
```

to retrieve an owned `Arc<dyn MyTrait>`. This is both harder to read and unnecessary cloning.

## Possible Solution
`into_inner` should be in an `impl Data<T: ?Sized>` block instead of `impl Data`.

jarhodes314:
## PR Type
Feature

## PR Checklist

## Overview
Closes #2403

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 ### Changed
 * Associated type `FromRequest::Config` was removed. [#2233]
 * Inner field made private on `web::Payload`. [#2384]
+* `Data::into_inner` and `Data::get_ref` no longer require T: Sized. [#2403]
 
 ### Removed
 * `ServiceResponse::checked_expr` was a legacy and just removed. [#2401]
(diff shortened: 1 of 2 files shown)


robjtede:
Tests are relevant for this PR to prevent regression

robjtede (approved):
thanks 👍🏻

PR #2413, opened 2021-10-18
add HttpServer::new_async

### LINKED ISSUE #2286: async `HttpServer::new` as alternative to `data_factory`
Part of the apeal of `App::data_factory` is that it uses the local reactor when instantiating app data. It's equivalent and more natural to make `HttpServer::new` async.

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

Closes #2286

### DIFF
--- a/actix-web/src/server.rs
+++ b/actix-web/src/server.rs
@@ -2,9 +2,7 @@ use std::{
     any::Any,
     cmp, fmt,
     future::Future,
-    io,
-    marker::PhantomData,
-    net,
+    io, net,
     sync::{Arc, Mutex},
     time::Duration,
 };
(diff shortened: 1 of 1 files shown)


aliemjay:
Depends on actix/actix-net#397

FnOnce was chosen for the closure because otherwise it would be incovenient to use  - the users are required to clone captures before the async block:
```rust
HttpServer::new_async(move || {
    let data = data.clone();
    async move { App::new().data(data) }
})
```


`HttpServer::new` was relaxed to accept FnOnce too for consistency.

robjtede:
not pursuing async server builders this cycle so delaying this til post v4

PR #2418, opened 2021-10-20
Fix/rustls update

Drevoed:
## PR Type

Compilation fix


robjtede:
this was addressed in #2416 and #2417

PR #2436, opened 2021-11-05
Bump `actix-codec` version, Reexport ConnectionIo

brockelmore:
## PR Type

Other

## PR Checklist

## Overview

During the refactor of moving `actix_http::client` over to `awc`, `ConnectionIo` was not reexported. Users need this trait for websocket usage. Additionally, updates the `actix-codec` version so websocket clients don't silently drop/not send continuation frames when the data is large (see actix/actix#431 and actix/actix-net#409)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## Unreleased - 2021-xx-xx
 ### Changed
 * `ContentType::html` now returns `Content-Type: text/html; charset=utf-8` instead of `Content-Type: text/html`.
-
+* Updated `actix-codec` to v0.4.1 [#2436]
 
 ## 4.0.0-beta.10 - 2021-10-20
 ### Added
--- a/actix-test/CHANGES.md
+++ b/actix-test/CHANGES.md
@@ -1,7 +1,8 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-
+* Updated `actix-codec` to v0.4.1 
+[#2436]: https://github.com/actix/actix-web/pull/2436
 
 ## 0.1.0-beta.5 - 2021-10-20
 * Updated rustls to v0.20. [#2414]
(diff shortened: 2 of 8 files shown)


fakeshadow:
Can you show me how would you need the `ConnectionIo` trait? It's never intended to be exported as it's just a trait alias for writing bound easier.

brockelmore:
Ah - I see i should use BoxedSocket. Previously, I used it as so:
```rust
/// A wrapper around a raw Websocket writer
pub type WsWriterRaw = SplitSink<Framed<Box<dyn ConnectionIo>, awc::ws::Codec>, awc::ws::Message>;

/// A wrapper around a raw Websocket stream
pub type WsStreamRaw = SplitStream<Framed<Box<dyn ConnectionIo>, awc::ws::Codec>>;

// ...
fn() -> Result<(WsWriterRaw, WsStreamRaw), ()> {
    let (_response, framed) = client_clone
        .ws(url)
        .connect()
        .await
        .map_err(|e| {
            tracing::error!("Websocket connection error: {}", e);
        })?;

    let (sink, stream) = framed.split();
    Ok((sink, stream))
}
```
But that can be replaced by BoxSocket i believe? But when i originally wrote the above function, the compiler said to use ConnectionIo in older versions of awc & actix-http

fakeshadow:
yea I believe you can use `BoxedSocket`. It's a shortcut explicitly exported.
You can also write it in a generic way as `Framed<BoxedSocket, awc::ws::Codec>` can be described as `S: Stream + Sink<Message>` or `impl Stream + Sink<Message>` where  `Stream`  is used to receive and `Sink` is used to send message This way you don't have to import the concrete types and keep the code cleaner since the only type that is awc specific is `Message`, `Frame` and maybe `ProtocolError`.

fakeshadow:
As for the compiler error I believe we should do better to hide the internal traits and types to reduce confusion.

brockelmore:
gonna close this and allow someone else to bump codec versions whenever they want to release a new version

PR #2448, opened 2021-11-16
make `AnyBody` generic on `Body` type

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Big clean up. Altering `AnyBody` to be generic on the "other" body type and finally removing `ResponseBody`.

One more egonomic improvement after this will be a method on HttpResponseBuilder for matching the B parameter and another method for boxing the body. These should greatly reduce the number of map_body calls needed currently when handling polymorphic response bodies.


actix-http/src/body/body.rs
```diff
@@ -33,29 +34,60 @@ impl AnyBody {
...
-        B::Error: Into<Box<dyn StdError + 'static>>,
-    {
-        Self::Stream(BoxAnyBody::from_body(body))
+    pub fn new(body: B) -> Self {
+        Self::Body(body)
+    }
+
+    pub fn into_boxed(self) -> AnyBody {
+        match self {
+            AnyBody::None => AnyBody::new_boxed(()),
```
fakeshadow:
```rust
match self {
    Self::None => AnyBody::None,
    Self::Bytes(bytes) => AnyBody::Bytes(bytes)
}
```
This would keep all response with fixed in memory body unboxed.

actix-http/src/body/body.rs
```diff
@@ -33,29 +34,60 @@ impl AnyBody {
...
+            AnyBody::Bytes(body) => AnyBody::new_boxed(body),
+            AnyBody::Body(body) => AnyBody::new_boxed(body),
+        }
     }
 }
 
-impl MessageBody for AnyBody {
+impl<B> MessageBody for AnyBody<B>
+where
+    B: MessageBody + Unpin,
```
fakeshadow: This does not seems right to me. The `AnyBody<B>` API should serve the purpose of pinning the B type and only box it when necessary. In general I think the response body should be a type with pinned B type and the box of it happens should at the top level where body type erase is necessary(or a move must happen)

fakeshadow: An example:
```rust
pin_project_lite::pin_project! {
    struct Body<B = futures_util::stream::BoxStream<'static, ()>> {
        #[pin]
        body: B
    }
}

fn unpin<T: Unpin>(_t: &T) {}

fn test() {
    let body = Body {
        body: Box::pin(futures_util::stream::once(async {}))
    };

    unpin(&body); // infered correctly.

    let body = Body {
        body: futures_util::stream::once(async {})
    };

    unpin(&body); // compile error
}
```

actix-http/src/body/body.rs
```diff
@@ -16,15 +16,16 @@ use super::{BodySize, BodyStream, MessageBody, MessageBodyMapErr, SizedStream};
 pub type Body = AnyBody;
 
 /// Represents various types of HTTP message body.
-pub enum AnyBody {
+#[derive(Clone)]
+pub enum AnyBody<B = BoxBody> {
     /// Empty response. `Content-Length` header is not set.
     None,
 
     /// Specific response body.
```
fakeshadow: `AnyBody::Bytes` may better be worded as in memory response body.
Typically the generic body variant is for async streaming where the vaule is not able and/or desirable to be in memory as a whole. 
This would help consumer to construct the correct variant.

actix-http/src/body/body.rs
robjtede: well spotted, the bound has been removed

fakeshadow (approved):
LGTM

actix-http/src/body/body.rs
```diff
@@ -33,39 +36,70 @@ impl AnyBody {
         Self::Bytes(Bytes::new())
     }
 
-    /// Create body from slice (copy)
+    /// Create boxed body from generic message body.
+    pub fn new_boxed<B>(body: B) -> Self
```
ibraheemdev, line 40: Maybe just call this `boxed`?

robjtede: toyed wth the idea yea

the counter argument is that `boxed` is well known as a combinator-style method in futures-util: https://docs.rs/futures-util/0.3.17/futures_util/future/trait.FutureExt.html#method.boxed

robjtede: Im sure counter examples exist. This shouldn't get used too much i think. I just like the idea of being explicit about its function and that `new` and `new_boxed` would be together in the docs.rs sidebar.

PR #2449, opened 2021-11-17
fix actix_http::Error conversion.

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

This restore the conversion from `Error` to `AnyBody<B>`. As error type always converted to a in memory response body that is stored in Bytes variant. So whay B type actually is does not matter.

`body/mod.rs` has some issue to infer the default type so a type alias is introduced.

With this PR it would be possible to clean up all the messy code that is specialized for handling `Into<Response<AnyBody>>` in `actix-http`. As the `Into<Response<B>>` on service type is always `AnyBody<B>` from actix-web stand point. Of course for explicity they can all be unifed to `Response<AnyBody<B: MessageBody>>`

### DIFF
--- a/actix-http/src/body/body.rs
+++ b/actix-http/src/body/body.rs
@@ -139,56 +139,56 @@ impl<S: fmt::Debug> fmt::Debug for AnyBody<S> {
     }
 }
 
-impl From<&'static str> for AnyBody {
-    fn from(string: &'static str) -> AnyBody {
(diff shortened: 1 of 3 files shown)


robjtede:
> body/mod.rs has some issue to infer the default type so a type alias is introduced.

I've also run across this in tests. I don't think it will be a big problem in real code but it feels like it should be inferrable, regardless.

robjtede (approved):

PR #2450, opened 2021-11-17
simplify handler.rs

aliemjay:
## PR Type

Refactor

## PR Checklist

## Overview

Replace Service and Future impl's by fn_service.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -1,16 +1,13 @@
 use std::future::Future;
-use std::marker::PhantomData;
-use std::pin::Pin;
-use std::task::{Context, Poll};
 
-use actix_service::{Service, ServiceFactory};
-use actix_utils::future::{ready, Ready};
-use futures_core::ready;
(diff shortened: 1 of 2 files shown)


robjtede:
@fakeshadow would this approach be compatible with GAT if fn_service was changed to support it?

aliemjay:
> would this approach be compatible with GAT if fn_service was changed to support it?

I'm not experienced with GAT, but I don't see why not. I think it is just need to be changed simply to `fn_service(|req: &mut Req| {})`

fakeshadow:
Handler service before and after are both not compat with GAT. 
The key factor is GAT is not object safe so as long as a boxed factory is needed it would be not able to be migrate to GAT. 

To be specific. It's OK to have boxed factory type but it's not possible to return a `Box<dyn Service<Future<'fut>>>` from it.

robjtede (approved):

PR #2452, opened 2021-11-17
allow extractors to borrow request

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

`FromRequestX` is meant to replace `FromRequest`. It is implemented as a second trait for prototyping.

Builds on #2450

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -7,31 +7,47 @@ use actix_service::{
 
 use crate::{
     service::{ServiceRequest, ServiceResponse},
-    Error, FromRequest, HttpResponse, Responder,
+    Error, FromRequestX, HttpResponse, Responder,
 };
 
+// TODO inaccessible docs
 /// A request handler is an async function that accepts zero or more parameters that can be
 /// extracted from a request (i.e., [`impl FromRequest`](crate::FromRequest)) and returns a type
 /// that can be converted into an [`HttpResponse`] (that is, it impls the [`Responder`] trait).
 ///
 /// If you got the error `the trait Handler<_, _, _> is not implemented`, then your function is not
 /// a valid handler. See [Request Handlers](https://actix.rs/docs/handlers/) for more information.
-pub trait Handler<T, R>: Clone + 'static
-where
-    R: Future,
-    R::Output: Responder,
-{
-    fn call(&self, param: T) -> R;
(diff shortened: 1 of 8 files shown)


aliemjay:
TODO:
* remove `FromRequest::Output`
* have `trait FromRequestOwned: for<'a> FromRequest<'a>`
* have trait `ArgumentFromRequest` to mimic the function of Output and have a blanket impl for FromRequestOwned

src/handler.rs
```diff
@@ -7,69 +7,143 @@ use actix_service::{
...
+    }
+
+    #[derive(Deserialize)]
+    struct ParamsOwned {
+        name: String,
+    }
+
+    async fn handler(
+        req: &HttpRequest,
+        (data, params1): (Option<&web::Data<usize>>, web::Path<ParamsOwned>),
```
ibraheemdev, line 123: why does `web::Path<Params<'_>>` not work?

aliemjay:
We can think of this as two separate problem domains:
* support borrowed extractors when using FromRequest methods directly: This can be easily solved by intoducing the lifetime parameter to the trait.
* support borrowed extractors in argument position. See below.

In order to support borrowed extractors *in function arguments* we need to somehow construct `&'a Uri` from `&'b Uri` for  any covariant lifetimes `<'a, 'b>` that appears in the extracted type. This is exactly what `FromRequest<'a>::Output` is for: `for<'a, 'b> <&'a Uri as FromRequestX<'b>>::Output = &'b Uri`. and hence `FromRequest` trait currently serves two puposes: an extractor and a type constructor.

While this works fine for other extractors, a limitation emerges in `Path<_>` and `Query<_>`, we need the type constructor function for the inner type `T`. [This todo](https://github.com/actix/actix-web/pull/2452#issuecomment-971871809) tries to address this by having two separate traits for the two separate functions: `FromRequest<'a>` + `ArgumentPosExtractor<'a>`, so that any extractor with a lifetime parameter need to impl both in order to support being extracted as a handler argument.

Other option is to keep supporting only owned types for `Path` and `Query` in argument position, and for zero-copy deserializtion the user can use `FromRequest` directly in the handler function.

src/handler.rs
aliemjay: yeah, this is  known limitation.  See https://github.com/actix/actix-web/pull/2452#issuecomment-974599063

PR #2467, opened 2021-11-28
guarantee ordering of header map get_all

### LINKED ISSUE #2466: consider guaranteeing the iteration order of `HeaderMap::get_all`
It was brought up that header collapsing is difficult without being able to find out/guarantee ordering at some level.

> A recipient MAY combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value to the combined field value in order, separated by a comma.  The order in which header fields with the same field name are received is therefore significant to the interpretation of the combined field value; a proxy MUST NOT change the order of these field values when forwarding a message.
\- https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2

This idea might also extend to guaranteeing same-name header orders elsewhere.

robjtede:
## PR Type
Doc / Tests

## PR Checklist

## Overview
Effectively a doc change + tests.

closes #2466

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -1,6 +1,11 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+### Changed
+* Guarantee ordering of `header::GetAll` iterator to be same as insertion order. [#2467]
+* Expose `header::{GetAll, Removed}` iterators. [#2467]
+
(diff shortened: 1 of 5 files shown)


robjtede:
cc @mkantor

popzxc (approved):

PR #2469, opened 2021-11-28
re-work from request macro.

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Reduce size of future generated. Depend on the type can be up to 30%. 
Enable `pin_project_lite` in macro.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ log = "0.4"
 mime = "0.3"
 paste = "1"
 pin-project = "1.0.0"
+pin-project-lite = "0.2.7"
 regex = "1.4"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -97,6 +97,7 @@ log = "0.4"
 mime = "0.3"
 paste = "1"
 pin-project = "1.0.0"
+pin-project-lite = "0.2.7"
```
ibraheemdev, line 100: `pin-project` *and* `pin-project-lite`?

robjtede: we're migrating as we go, both crates are in the dep graph anyway

robjtede (approved):

Cargo.toml
fakeshadow: The only blocker of removing `pin-project`besides this macro was the pin drop attr used in logger middleware. 
`pin-project-lite 0.2.7` added this feature so at this point it's possible to remove `pin-project` from actix-web

PR #2480, opened 2021-12-01
wrap `LanguageTags` type in new `AnyOrSome` type to support wildcards

### LINKED ISSUE #2434: "any language" (`*` - wildcard) does not get parsed for the `Accept-Language` header
I'm using Actix Web to attempt to parse the Accept-Language header out of a request. However, when the request contains an asterisk (`*`) (to indicate any language, used as a wildcard), that does not get parsed at all: any other language and quality pair get parsed, but the wildcard is simply ignored and not reported at all.
The [spec](http://tools.ietf.org/html/rfc7231#section-5.3.5) seem to implicitly state that the language range [can be a wildcard](https://datatracker.ietf.org/doc/html/rfc4647#section-2.1).

Moreover, actix-web tests [don't seem to have one for wildcards](https://github.com/actix/actix-web/blob/2eacb735a4f2f284eeaf80244a119ca89621234d/src/http/header/accept_language.rs#L54).

## Expected Behavior

**Example 1**
Consider the header `Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5`. When parsed using the actix-web helpers, this should produce the following list of quality items for languages:

`[("fr-CH, fr", 900), ("en", 800), ("de", 700), ("*", 500)]`

**Example 2**
Moreover, the `Accept-Language: *` should produce a non-empty list:

`[("*", 1000)]`

Both examples were taken from the [header MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language).

## Current Behavior

With reference to the examples of the previous section, these are the outputs actix-web gives instead.

**Example 1**

`[("fr-CH, fr", 900), ("en", 800), ("de", 700)]` (note the lack of `("*", 500)`)

**Example 2**

`[]`

## Possible Solution

The implementation for the [Accept-Encoding header](https://github.com/actix/actix-web/blob/master/src/http/header/accept_encoding.rs) seem to account for the wildcard item, so the potential solution could be to use the same logic from that header? 

This is the sample code I use to parse my requests:

```rust
let langs = header::AcceptLanguage::parse(req).unwrap();
println!("**** DEBUG - header {:?} - langs {:?}", req.headers().get(header::ACCEPT_LANGUAGE), languages);
```

With `req` being a request object containing the examples.

## Context

I'm trying to leverage actix-web helpers for parsing the Accept-Language header instead of rolling up (and maintaining!) our own.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): rustc 1.55.0 (c8dfcfe04 2021-09-06)
* Actix Web Version: web-v4.0.0-beta.8

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
Introduce new `AnyOrSome` (bikeshedding welcome) wrapper type to support wildcard items in header types where the backing type has no concept of wildcard/any types.

Used initially in `AcceptLanguage` but for consistency it might make sense to re-work `Accept` (which _does_ know about wildcards).

fixes #2434


robjtede:
cc @Dexterp37

finnbear:
> Introduce new AnyOrSome (bikeshedding welcome)

I assume you are trying to avoid a compound-word name. Here are some ideas:
1. `Preference<T>`
```rust
enum Preference<T> {
    Any,
    Specific(T)
}
```
2. Just use `Option<T>` where `None` is taken to mean nothing specific i.e. any
3. Other considered `enum` names: `Specificity<T>`

robjtede:
1. I kinda like `Preference<T>`
2. Nah, the None variant would be confusing.
3. it's alright, just a bit harder to spell/type than preference

PR #2487, opened 2021-12-05
Remove extensions from head

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
- always seemed weird to me
- requesthead can be clone now
- and therefore can impl fromrequest
- also seeing some perf gains


fakeshadow:
It's in head for caching purpose. Every thread holds up to 128 RequestHead re-use their allocation.
Are you sure this improves perf. From what I see this would likely cause more allocation and potential memory fragmentation.

robjtede:
some benches, seems to support removing this from head

%diff in right-hand column are based on p99 latencies

2021-12-07_21 27 24-1823d5fb

PR #2491, opened 2021-12-06
add connection level data container

### LINKED ISSUE #1770: Add on_connect extension data to all subsequent calls made through the connection

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
Second attempt to solve #1770 again without introducing unsafe code and fewer API breaks.

An Rc<Extensions> is kept in the dispatcher and cloned into a Request each time one is created for a connection. The API exposes read-only access to this container through request methods `conn_data`.

See [on_connect example](https://github.com/actix/actix-web/pull/2491/files#diff-e29431a78f487c16502ba92644a5aa9079d3384a9ac194289c0c2101bdd56bec) for how this changes the API in on_connect use cases.

Supercedes #2327


robjtede:
@ibraheemdev Hope it's alright to call you in as someone who reviewed the original PR.

PR #2492, opened 2021-12-07
enable scope middleware with generic res body.

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Enable `Scope::wrap` and `Scope::wrap_fn` accepting middlware that transform response body without boxing.
The final output `Scope` must still have `BoxBody`. This can be achieved with `Compat` or another middleware does the same job without recieving any other middleware type

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -1,6 +1,6 @@
-use std::{cell::RefCell, fmt, future::Future, mem, rc::Rc};
+use std::{cell::RefCell, fmt, future::Future, marker::PhantomData, mem, rc::Rc};
 
-use actix_http::Extensions;
+use actix_http::{body::BoxBody, Extensions};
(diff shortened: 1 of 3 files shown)


src/scope.rs
```diff
@@ -304,23 +307,24 @@ where
     /// ServiceResponse.
```
aliemjay: Docs on wrap and wrap_fn need an update

aliemjay (changes requested):
I think we need the same thing for Resource.

I'll do it in a followup PR if you like.

aliemjay:
semver-major?

robjtede:
> semver-major

ye adding a type param to wrap makes it breaking, strictly, though i doubt anyone would be using it in a way that it would _actually_ break

robjtede (approved):

PR #2493, opened 2021-12-07
remove generic body type in App.

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Removing generic B type associated with potential middleware types.
This change is to lay ground work for adding generic body type for App's child service types.
(B type could be added back with it but for a total different purpose.)


aliemjay (approved):
A minor downside is that incorrectly implemented middleware now shows as `ServiceFactory is not implemented for App` and the error spans whole App, which may be confusing to beginners. 
Otherwise, LGTM.

robjtede (approved):

PR #2494, opened 2021-12-07
update actix-http readme example.

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Update the readme example to show proper way of embedding `actix_web::App` into `actix-http`

### DIFF
--- a/actix-http/README.md
+++ b/actix-http/README.md
@@ -19,34 +19,30 @@
 ## Example
 
 ```rust
-use std::{env, io};
-
-use actix_http::{HttpService, Response};
+use actix_http::HttpService;
 use actix_server::Server;
-use futures_util::future;
-use http::header::HeaderValue;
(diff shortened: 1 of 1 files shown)


robjtede (changes requested):
think i'd prefer readme example be kept as pure actix-http and this be included in a new example

fakeshadow:
All the questions I come across with have been asking how to use actix-http with actix-web.
This states a lot of what most users want from this crate.
The original example does not have any purpose and can be confusing of what it can do.

robjtede:
> The original example does not have any purpose

Sure, but stating this crate as just another way to run -web apps is undervaluing it. We should be able to push it as a standalone thing.

robjtede:
added in 9587261c2099ea46182147c7f8807f8b1055448e

PR #2497, opened 2021-12-08
provide optimisation path for single-chunk body types

robjtede:
## PR Type

Feature

## PR Checklist

## Overview

Two new methods on `MessageBody`:
- `fn is_complete_body(&self) -> bool`
- `fn take_complete_body(&mut self) -> Bytes`

Allows detecting body types that can be extracted in their entirety in one `poll_next` and therefore a shortcut for this synchronous operation. It is possible to build optimisations around these methods, as can be seen in the updated compression encoder.


aliemjay:
I quite like the trait solution and I think it can be improved but not sure how.

One idea does unfortunately introduces AnyBody again, but I hope the benefits overweighs this.

```rust
struct BoxStream(Pin<Box<dyn MessageBody>>); // <- BoxBody renamed

enum Body<S = BoxStream> {
  None,
  Bytes(Bytes),
  Stream(S),
}

trait MessageBody {
  // other items...
  fn into_body(self) -> Body<Self> where Self: Sized;
}
```

This way the trait methods are usable directly, since the trait protocol is more fail-safe than passing &mut Self, which may leave self in undefined state while still be usable as a body.

Perf-wise, it should eliminate boxing completely in the common case where None or Bytes is the body when passing services and the middleware chain.

I also imagine this is forward-compatible when we decide to use `Stream` for the streaming interface as it solves the problem of encoding size_hint for `None` bodies:
```rust
trait MessageBody {
  type Stream: Stream<Item=Bytes>;
  // remove streaming interface from message body:
  fn into_body(self) -> Body<Self::Stream> where Self: Sized;
}
```

Any ideas about this?

If you have no major objections, I'm happy to implement it. I wouldn't expect it to be merged eventually, but I think it is worth trying!

aliemjay (approved):
LGTM. Thanks!

robjtede:
> Any ideas about this?

I think this can be tried in a different way through a `MessageBodyExt`. I'm gonna create an `actix-lab` sort of testing crate that this kinda thing would be good for.

PR #2515, opened 2021-12-14
allow error handler middleware to return different body type

### LINKED ISSUE #2513: ErrorHandlers middleware can't mutate body type
The signature for error handler callback dicatates that the output body type is the same as the input.

After removing `ResponseBody`, it is impossible then to make any useful mutation to response body.

* Actix Web Version: v4.0.0-beta.14

robjtede:
## PR Type

Fix

## PR Checklist

- [ ] Docs are updated.

## Overview

fixes #2513

### DIFF
--- a/src/middleware/compat.rs
+++ b/src/middleware/compat.rs
@@ -6,12 +6,15 @@ use std::{
     task::{Context, Poll},
 };
 
-use actix_http::body::MessageBody;
-use actix_service::{Service, Transform};
 use futures_core::{future::LocalBoxFuture, ready};
 use pin_project_lite::pin_project;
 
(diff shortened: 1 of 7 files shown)


src/middleware/compat.rs
```diff
@@ -116,14 +119,15 @@ where
 
 /// Convert `ServiceResponse`'s `ResponseBody<B>` generic type to `ResponseBody<Body>`.
 pub trait MapServiceResponseBody {
```
aliemjay, line 121: I don't think this trait is useful without impl specialization!

aliemjay (approved):
LGTM. Thanks!

PR #2519, opened 2021-12-16
various fixes to MessageBody::complete_body

aliemjay:
## PR Type

Bug Fix

## PR Checklist

## Overview

1. `BoxBody.is_complete_body()` was always `false`, because of absent impl in `MapErr`
2. `None` shouldn't return bytes::new()
3. fix `impl MessageBody for Pin<Box<B>>`

### DIFF
--- a/actix-http/src/body/message_body.rs
+++ b/actix-http/src/body/message_body.rs
@@ -164,7 +164,7 @@ mod foreign_impls {
 
     impl<B> MessageBody for Pin<Box<B>>
     where
-        B: MessageBody,
+        B: MessageBody + ?Sized,
     {
         type Error = B::Error;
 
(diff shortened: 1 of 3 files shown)


actix-http/src/body/none.rs
```diff
@@ -40,14 +40,4 @@ impl MessageBody for None {
...
-
-    #[inline]
-    fn is_complete_body(&self) -> bool {
-        true
-    }
-
-    #[inline]
-    fn take_complete_body(&mut self) -> Bytes {
-        Bytes::new()
-    }
```
robjtede: I think I understand why you removed this; the use case for these methods also should go along with checking the size hint for `None` and bailing and, as such, these methods will/should never be called. So for me there are two options:
- Keep them and allow for some unexpected, exotic use case where they _are_ reasonably called.
- Use a custom panic message that indicates that size hint should be checked first.

actix-http/src/body/message_body.rs
```diff
@@ -164,7 +164,7 @@ mod foreign_impls {
 
     impl<B> MessageBody for Pin<Box<B>>
     where
-        B: MessageBody,
+        B: MessageBody + ?Sized,
```
robjtede, line 167: i guess ?Sized also applies to `Box<B>`

robjtede (changes requested):

PR #2520, opened 2021-12-16
MessageBody::boxed

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

`Either::right(Bytes::new().boxed()).boxed().boxed()` now allocate once.

The motivation behind this is not the perf as much as to enable zero-cost boxing to be more liberal with using boxed body.

### DIFF
--- a/actix-http/src/body/boxed.rs
+++ b/actix-http/src/body/boxed.rs
@@ -15,15 +15,11 @@ pub struct BoxBody(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>);
 
 impl BoxBody {
     /// Same as `MessageBody::boxed`.
+    ///
+    /// If the body type to wrap is unknown or generic it is better to use [`MessageBody::boxed`] to
+    /// avoid double boxing.
+    #[inline]
     pub fn new<B>(body: B) -> Self
-    where
-        B: MessageBody + 'static,
-    {
-        body.boxed()
-    }
-
-    /// Boxes a `MessageBody` and any errors it generates.
-    pub fn new_raw<B>(body: B) -> Self
     where
         B: MessageBody + 'static,
     {
@@ -32,6 +28,7 @@ impl BoxBody {
     }
 
     /// Returns a mutable pinned reference to the inner message body type.
+    #[inline]
     pub fn as_pin_mut(&mut self) -> Pin<&mut (dyn MessageBody<Error = Box<dyn StdError>>)> {
         self.0.as_mut()
     }
@@ -46,10 +43,12 @@ impl fmt::Debug for BoxBody {
 impl MessageBody for BoxBody {
     type Error = Error;
 
+    #[inline]
     fn size(&self) -> BodySize {
         self.0.size()
     }
 
+    #[inline]
     fn poll_next(
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
@@ -60,10 +59,12 @@ impl MessageBody for BoxBody {
             .map_err(|err| Error::new_body().with_cause(err))
     }
 
+    #[inline]
     fn is_complete_body(&self) -> bool {
         self.0.is_complete_body()
     }
 
+    #[inline]
     fn take_complete_body(&mut self) -> Bytes {
         self.0.take_complete_body()
     }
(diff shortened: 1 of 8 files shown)


aliemjay:
I though It might be better to keep `BoxBody::new()` as it is and instead use `MessageBody::boxed` whenever boxing is needed, but now I'm not sure. So I'll keep it as it is unless you have another opinion,

actix-http/src/body/boxed.rs
```diff
@@ -14,8 +14,16 @@ use crate::Error;
...
+    /// Same as `MessageBody::boxed`.
     pub fn new<B>(body: B) -> Self
+    where
+        B: MessageBody + 'static,
+    {
+        body.boxed()
+    }
+
+    /// Boxes a `MessageBody` and any errors it generates.
+    pub fn new_raw<B>(body: B) -> Self
```
robjtede: I don't get the new method, wouldn't it be okay to just use `BoxBody::new` in the default `.boxed` impl ?

robjtede:
seems to work:

```diff
diff --git a/actix-http/src/body/boxed.rs b/actix-http/src/body/boxed.rs
index 734df8404..611d80a9d 100644
--- a/actix-http/src/body/boxed.rs
+++ b/actix-http/src/body/boxed.rs
@@ -16,14 +16,6 @@ pub struct BoxBody(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>);
 impl BoxBody {
-    /// Same as `MessageBody::boxed`.
+    /// Boxes a `MessageBody` and any errors it generates.
     pub fn new<B>(body: B) -> Self
-    where
-        B: MessageBody + 'static,
-    {
-        body.boxed()
-    }
-
-    /// Boxes a `MessageBody` and any errors it generates.
-    pub fn new_raw<B>(body: B) -> Self
     where
         B: MessageBody + 'static,
     {
diff --git a/actix-http/src/body/message_body.rs b/actix-http/src/body/message_body.rs
index ca8a3b2c3..0c62d52c7 100644
--- a/actix-http/src/body/message_body.rs
+++ b/actix-http/src/body/message_body.rs
@@ -82,11 +82,12 @@ pub trait MessageBody {
     ///
     /// Implementation shouldn't call `BoxBody::new` on `self` as it would cause infinite
     /// recursion.
+    #[inline]
     fn boxed(self) -> BoxBody
     where
         Self: Sized + 'static,
     {
-        BoxBody::new_raw(self)
+        BoxBody::new(self)
     }
 }
 
diff --git a/actix-http/src/response.rs b/actix-http/src/response.rs
index 9f799f669..aee9e80b4 100644
--- a/actix-http/src/response.rs
+++ b/actix-http/src/response.rs
@@ -194,7 +194,7 @@ impl<B> Response<B> {
     where
         B: MessageBody + 'static,
     {
-        self.map_body(|_, body| BoxBody::new(body))
+        self.map_body(|_, body| body.boxed())
     }
 
     /// Returns body, consuming this response.
diff --git a/src/response/response.rs b/src/response/response.rs
index 1900dd845..4fb4b44b6 100644
--- a/src/response/response.rs
+++ b/src/response/response.rs
@@ -244,8 +244,7 @@ impl<B> HttpResponse<B> {
     where
         B: MessageBody + 'static,
     {
-        // TODO: avoid double boxing with down-casting, if it improves perf
-        self.map_body(|_, body| BoxBody::new(body))
+        self.map_body(|_, body| body.boxed())
     }
 
     /// Extract response body
diff --git a/src/service.rs b/src/service.rs
index 36b3858e6..9ccf5274d 100644
--- a/src/service.rs
+++ b/src/service.rs
@@ -451,7 +451,7 @@ impl<B> ServiceResponse<B> {
     where
         B: MessageBody + 'static,
     {
-        self.map_body(|_, body| BoxBody::new(body))
+        self.map_body(|_, body| body.boxed())
     }
 }
 

```

aliemjay:
This is exactly what I proposed in the comment above. But I'm not sure which way is better. Feel free to change it as  you like.

robjtede:
> This is exactly what I proposed in the comment above.

Right. I was wondering if there was a reason why there was a new method; maybe easier to trip into infinite loops or smth.

aliemjay:
I just felt it somewhat not right that `body.boxed()` and `BoxBody::new(body)` have different perf characteristics and behavior.

robjtede:
Assessed all usages of `BoxBody::new` in the code base, replaced the ones that needed `.boxed()` instead, and documented `new` with a link to `.boxed`. Should be good enough imo; I tend to prefer reduced exposed API.

robjtede (approved):

PR #2522, opened 2021-12-17
simplify MessageBody::complete_body interface

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

This tries to make MessageBody interface for take_complete_body less strict, more fail-safe and hence usable directly.

The idea is to store Bytes in boxed body so that `take_body` no longer needs to be object-safe.

@robjtede I'd appreciate your input at this point before finishing it.

### DIFF
--- a/actix-http/src/body/boxed.rs
+++ b/actix-http/src/body/boxed.rs
@@ -8,12 +8,13 @@ use std::{
 use bytes::Bytes;
 
 use super::{BodySize, MessageBody, MessageBodyMapErr};
+use crate::body;
 
 /// A boxed message body with boxed errors.
 pub struct BoxBody(BoxBodyInner);
 
 enum BoxBodyInner {
-    None,
+    None(body::None),
     Bytes(Bytes),
     Stream(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>),
 }
@@ -29,7 +30,7 @@ impl BoxBody {
         B: MessageBody + 'static,
     {
         match body.size() {
-            BodySize::None => Self(BoxBodyInner::None),
+            BodySize::None => Self(BoxBodyInner::None(body::None)),
             _ => match body.try_into_bytes() {
                 Ok(bytes) => Self(BoxBodyInner::Bytes(bytes)),
                 Err(body) => {
@@ -60,8 +61,8 @@ impl MessageBody for BoxBody {
     #[inline]
     fn size(&self) -> BodySize {
         match &self.0 {
-            BoxBodyInner::None => BodySize::None,
-            BoxBodyInner::Bytes(bytes) => BodySize::Sized(bytes.len() as u64),
+            BoxBodyInner::None(none) => none.size(),
+            BoxBodyInner::Bytes(bytes) => bytes.size(),
             BoxBodyInner::Stream(stream) => stream.size(),
         }
     }
@@ -72,7 +73,7 @@ impl MessageBody for BoxBody {
         cx: &mut Context<'_>,
     ) -> Poll<Option<Result<Bytes, Self::Error>>> {
         match &mut self.0 {
-            BoxBodyInner::None => Poll::Ready(None),
+            BoxBodyInner::None(_) => Poll::Ready(None),
             BoxBodyInner::Bytes(bytes) => Pin::new(bytes).poll_next(cx).map_err(Into::into),
             BoxBodyInner::Stream(stream) => Pin::new(stream).poll_next(cx),
         }
@@ -81,7 +82,8 @@ impl MessageBody for BoxBody {
     #[inline]
     fn try_into_bytes(self) -> Result<Bytes, Self> {
         match self.0 {
-            BoxBodyInner::Bytes(bytes) => Ok(bytes),
+            BoxBodyInner::None(none) => Ok(none.try_into_bytes().unwrap()),
+            BoxBodyInner::Bytes(bytes) => Ok(bytes.try_into_bytes().unwrap()),
             _ => Err(self),
         }
     }
--- a/actix-http/src/body/boxed.rs
+++ b/actix-http/src/body/boxed.rs
@@ -1,6 +1,6 @@
 use std::{
     error::Error as StdError,
-    fmt,
+    fmt, mem,
     pin::Pin,
     task::{Context, Poll},
 };
@@ -11,7 +11,13 @@ use super::{BodySize, MessageBody, MessageBodyMapErr};
 use crate::Error;
 
 /// A boxed message body with boxed errors.
-pub struct BoxBody(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>);
+pub struct BoxBody(BoxBodyInner);
+
+enum BoxBodyInner {
+    None,
+    Bytes(Bytes),
+    Stream(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>),
+}
 
 impl BoxBody {
     /// Same as `MessageBody::boxed`.
@@ -23,19 +29,29 @@ impl BoxBody {
     where
         B: MessageBody + 'static,
     {
-        let body = MessageBodyMapErr::new(body, Into::into);
-        Self(Box::pin(body))
+        match body.size() {
+            BodySize::None => Self(BoxBodyInner::None),
+            _ => match body.try_into_bytes() {
+                Ok(bytes) => Self(BoxBodyInner::Bytes(bytes)),
+                Err(body) => {
+                    let body = MessageBodyMapErr::new(body, Into::into);
+                    Self(BoxBodyInner::Stream(Box::pin(body)))
+                }
+            },
+        }
     }
 
     /// Returns a mutable pinned reference to the inner message body type.
     #[inline]
-    pub fn as_pin_mut(&mut self) -> Pin<&mut (dyn MessageBody<Error = Box<dyn StdError>>)> {
-        self.0.as_mut()
+    //pub fn as_pin_mut(&mut self) -> Pin<&mut (dyn MessageBody<Error = Box<dyn StdError>>)> {
+    pub fn as_pin_mut(&mut self) -> Pin<&mut Self> {
+        Pin::new(self)
     }
 }
 
 impl fmt::Debug for BoxBody {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // TODO show BoxBodyInner
         f.write_str("BoxBody(dyn MessageBody)")
     }
 }
@@ -45,7 +61,11 @@ impl MessageBody for BoxBody {
 
     #[inline]
     fn size(&self) -> BodySize {
-        self.0.size()
+        match &self.0 {
+            BoxBodyInner::None => BodySize::None,
+            BoxBodyInner::Bytes(bytes) => BodySize::Sized(bytes.len() as u64),
+            BoxBodyInner::Stream(stream) => stream.size(),
+        }
     }
 
     #[inline]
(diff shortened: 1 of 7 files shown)


robjtede:
This looks promising... can't believe I didn't think of the `Err(self)` pattern before.

Some initial **thoughts**; not reasons to reject this PR:
1. It's kinda a shame to lose the `try_into_bytes` impl on `[Pin]Box` due to not being able to get `self` for their wrapped types.
    - The return of `AnyBody`™️  as an implementation detail (`BoxBodyInner`) is a decent work-around though.
1. Implementations are now a little inconsistent when calling `try_into_bytes` multiple times. `&'static [u8]` would return Ok every with full bytes time but `Bytes` would return Ok with full bytes once and then Ok with empty Bytes every other time.
    - ~This might not matter given how we expect these fns to be used, but it's certainly a documentable point.~
    - Of course... receiving `self` prevents you calling it twice anyway. I wonder if it still matters for `&'static` types...
1. Spliting up the methods felt a bit like `poll_ready` vs `call` in the service trait. Eg. there may be some use case where you _want_ to make a choice based on it's result before destroying the original type.
    - Splitting up the methods was also the only way to guarantee the current `Pin<B>` impl is valid. Playing around with it again reminded me why I didn't go down that path before.
1. Are there situations where the message type contains more information that you'd want to keep around after extracting the Bytes?
    - Put another way, could it be `fn try_into_bytes(&mut self) -> Option<Bytes>`?
    - This signature would also address point 2 by receiving `&mut &'static [u8]` again.
    - This signature would give us back an impl on `Box<B>` (but not `Pin<B>` as mentioned in 3.)

Also note that I still haven't got bench numbers for how much this optimization matters or difference from current impl to this one.

robjtede:
Turns out you were 70% done already.

```diff
diff --git a/actix-http/src/body/boxed.rs b/actix-http/src/body/boxed.rs
index 2603aa762..d7f772c5a 100644
--- a/actix-http/src/body/boxed.rs
+++ b/actix-http/src/body/boxed.rs
@@ -8,7 +8,6 @@ use std::{
 use bytes::Bytes;
 
 use super::{BodySize, MessageBody, MessageBodyMapErr};
-use crate::Error;
 
 /// A boxed message body with boxed errors.
 pub struct BoxBody(BoxBodyInner);
@@ -57,7 +56,7 @@ impl fmt::Debug for BoxBody {
 }
 
 impl MessageBody for BoxBody {
-    type Error = Error;
+    type Error = Box<dyn StdError>;
 
     #[inline]
     fn size(&self) -> BodySize {
@@ -76,7 +75,7 @@ impl MessageBody for BoxBody {
         match &mut self.0 {
             BoxBodyInner::None => Poll::Ready(None),
             BoxBodyInner::Bytes(bytes) => Poll::Ready(Some(Ok(mem::take(bytes)))),
-            BoxBodyInner::Stream(stream) => Pin::new(stream).poll_next(cx).map_err(|err| Error::new_body().with_cause(err)),
+            BoxBodyInner::Stream(stream) => Pin::new(stream).poll_next(cx),
         }
     }
 
diff --git a/actix-http/src/body/message_body.rs b/actix-http/src/body/message_body.rs
index 7826ef54f..4b834c025 100644
--- a/actix-http/src/body/message_body.rs
+++ b/actix-http/src/body/message_body.rs
@@ -480,7 +480,7 @@ mod tests {
         let body = EitherBody::<(), _>::right(body);
         let body = Box::new(body);
         let body = Box::pin(body);
-        let mut body = body;
+        let body = body;
 
         assert_eq!(body.try_into_bytes().unwrap(), b"test".as_ref());
     }
diff --git a/actix-http/src/encoding/encoder.rs b/actix-http/src/encoding/encoder.rs
index fa294ab0d..55ddc8d8c 100644
--- a/actix-http/src/encoding/encoder.rs
+++ b/actix-http/src/encoding/encoder.rs
@@ -53,7 +53,7 @@ impl<B: MessageBody> Encoder<B> {
         }
     }
 
-    pub fn response(encoding: ContentEncoding, head: &mut ResponseHead, mut body: B) -> Self {
+    pub fn response(encoding: ContentEncoding, head: &mut ResponseHead, body: B) -> Self {
         let can_encode = !(head.headers().contains_key(&CONTENT_ENCODING)
             || head.status == StatusCode::SWITCHING_PROTOCOLS
             || head.status == StatusCode::NO_CONTENT
@@ -65,11 +65,9 @@ impl<B: MessageBody> Encoder<B> {
             return Self::none();
         }
 
-        let body = if body.is_complete_body() {
-            let body = body.take_complete_body();
-            EncoderBody::Full { body }
-        } else {
-            EncoderBody::Stream { body }
+        let body = match body.try_into_bytes() {
+            Ok(body) => EncoderBody::Full { body },
+            Err(body) => EncoderBody::Stream { body },
         };
 
         if can_encode {
@@ -133,21 +131,13 @@ where
         }
     }
 
-    fn is_complete_body(&self) -> bool {
+    fn try_into_bytes(self) -> Result<Bytes, Self>
+    where
+        Self: Sized,
+    {
         match self {
-            EncoderBody::None => true,
-            EncoderBody::Full { .. } => true,
-            EncoderBody::Stream { .. } => false,
-        }
-    }
-
-    fn take_complete_body(&mut self) -> Bytes {
-        match self {
-            EncoderBody::None => Bytes::new(),
-            EncoderBody::Full { body } => body.take_complete_body(),
-            EncoderBody::Stream { .. } => {
-                panic!("EncoderBody::Stream variant cannot be taken")
-            }
+            EncoderBody::Full { body } => Ok(body),
+            _ => Err(self),
         }
     }
 }
@@ -234,19 +224,20 @@ where
         }
     }
 
-    fn is_complete_body(&self) -> bool {
+    fn try_into_bytes(mut self) -> Result<Bytes, Self>
+    where
+        Self: Sized,
+    {
         if self.encoder.is_some() {
-            false
+            Err(self)
         } else {
-            self.body.is_complete_body()
-        }
-    }
-
-    fn take_complete_body(&mut self) -> Bytes {
-        if self.encoder.is_some() {
-            panic!("compressed body stream cannot be taken")
-        } else {
-            self.body.take_complete_body()
+            match self.body.try_into_bytes() {
+                Ok(body) => Ok(body),
+                Err(body) => {
+                    self.body = body;
+                    Err(self)
+                }
+            }
         }
     }
 }
diff --git a/actix-http/src/h1/dispatcher.rs b/actix-http/src/h1/dispatcher.rs
index 64bf83e03..ea0c2faa8 100644
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -22,7 +22,7 @@ use crate::{
     config::ServiceConfig,
     error::{DispatchError, ParseError, PayloadError},
     service::HttpFlow,
-    Extensions, OnConnectData, Request, Response, StatusCode,
+    Error, Extensions, OnConnectData, Request, Response, StatusCode,
 };
 
 use super::{
@@ -458,7 +458,9 @@ where
                             }
 
                             Poll::Ready(Some(Err(err))) => {
-                                return Err(DispatchError::Service(err.into()))
+                                return Err(DispatchError::Service(
+                                    Error::new_body().with_cause(err).into(),
+                                ))
                             }
 
                             Poll::Pending => return Ok(PollResponse::DoNothing),
diff --git a/src/dev.rs b/src/dev.rs
index d4a64985c..edcc158f8 100644
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -139,4 +139,12 @@ impl crate::body::MessageBody for AnyBody {
             AnyBody::Boxed { body } => body.as_pin_mut().poll_next(cx),
         }
     }
+
+    fn try_into_bytes(self) -> Result<crate::web::Bytes, Self> {
+        match self {
+            AnyBody::None => Ok(crate::web::Bytes::new()),
+            AnyBody::Full { body } => Ok(body),
+            _ => Err(self),
+        }
+    }
 }

```

aliemjay:
1. I think it is reasonable to drop support for `[Pin]Box<dyn MessageBody>`. It is not a law that a trait author should implement it on`Pin<Box>` and these impls are not used anyway since they were totally broken #2519.  If the user want type erasure, they can only use `BoxBody` and we can then paramterize it for error type (`BoxBody<anyhow::Error>` instead of `Pin<Box<dyn MessageBody<Error=anyhow:Error>>>`.

2. As you said, I don't see how is this a problem!

3. This should  be partially remedied by point 1.

4. If the type is more than `MessageBody` then it can provide borrowed access to the buffer through a different interface (like inherent impl). Otherwise `Bytes` is also a `MessageBody`, so this should not destroy type information.

Points 3 and 4 are sill legit concerns but they should be weighted against:
1. Regardless of whether we accept this design or not, `BoxBodyInner` is still nice to have since it enables zero-cost boxing for the most common body types, `None` and `Bytes`. Though this needs some benches.
2. Greatly simplify the MessageBody protocol for both the user and the implementer.

If we want to keep supporting the impl on `Pin<Box>`, then I agree that this is not the optimal interface. So that's  a decision to make.

robjtede:
Okay, glad we agree. I'd like to cut a release today and I think it's worth getting this in it. Can you get this updated soon or shall I apply the patch?

aliemjay:
I'll work on it

src/dev.rs
```diff
@@ -139,4 +139,12 @@ impl crate::body::MessageBody for AnyBody {
             AnyBody::Boxed { body } => body.as_pin_mut().poll_next(cx),
         }
     }
+
+    fn try_into_bytes(self) -> Result<crate::web::Bytes, Self> {
+        match self {
+            AnyBody::None => Ok(crate::web::Bytes::new()),
```
aliemjay, line 145: I don't think this s right . My argument is `AnyBody::None.size() != Bytes::new().size()`. This also needs to be consistent with `body::None`.

actix-http/src/body/none.rs
```diff
@@ -40,14 +40,4 @@ impl MessageBody for None {
...
-
-    #[inline]
-    fn is_complete_body(&self) -> bool {
-        true
-    }
-
-    #[inline]
-    fn take_complete_body(&mut self) -> Bytes {
-        Bytes::new()
-    }
```
robjtede, line 52: still think we should keep a impl for none

actix-http/src/body/message_body.rs
```diff
@@ -31,51 +31,11 @@ pub trait MessageBody {
...
-            of `take_complete_body`",
-            std::any::type_name::<Self>()
-        );
-
-        unimplemented!(
-            "type ({}) does not allow taking complete body; caller should make sure to \
-            check `is_complete_body` first",
-            std::any::type_name::<Self>()
-        );
+    fn try_into_bytes(self) -> Result<Bytes, Self>
```
robjtede, line 37: document

actix-http/src/body/boxed.rs
```diff
@@ -23,50 +28,62 @@ impl BoxBody {
...
 impl MessageBody for BoxBody {
-    type Error = Error;
+    type Error = Box<dyn StdError>;
 
     #[inline]
     fn size(&self) -> BodySize {
-        self.0.size()
+        match &self.0 {
+            BoxBodyInner::None => BodySize::None,
+            BoxBodyInner::Bytes(bytes) => BodySize::Sized(bytes.len() as u64),
```
robjtede: delegate `bytes.size()`

actix-http/src/body/boxed.rs
```diff
@@ -8,10 +8,15 @@ use std::{
...
 
 use super::{BodySize, MessageBody, MessageBodyMapErr};
-use crate::Error;
 
 /// A boxed message body with boxed errors.
-pub struct BoxBody(Pin<Box<dyn MessageBody<Error = Box<dyn StdError>>>>);
+pub struct BoxBody(BoxBodyInner);
+
+enum BoxBodyInner {
+    None,
```
robjtede: i kinda think this should wrap `body::None` and delegate so that the behavior is guaranteed to be the same

robjtede (changes requested):

src/dev.rs
robjtede: yeah, whatever the behavior the None variants should align with `body::None` I suggest handling this by wrapping that struct so that any future tweaks affect all places

actix-http/src/encoding/encoder.rs
```diff
@@ -133,21 +131,14 @@ where
...
-    }
-
-    fn take_complete_body(&mut self) -> Bytes {
-        match self {
-            EncoderBody::None => Bytes::new(),
-            EncoderBody::Full { body } => body.take_complete_body(),
-            EncoderBody::Stream { .. } => {
-                panic!("EncoderBody::Stream variant cannot be taken")
-            }
+            EncoderBody::None => Ok(Bytes::new()),
```
aliemjay, line 139: I'll try to dedupe EncoderBody and BoxBodyInner in a followup PR.

robjtede (approved):

PR #2526, opened 2021-12-17
allow any body type in Resource

robjtede:
## PR Type

Feature

## PR Checklist

## Overview
Similar to #2523

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -1,6 +1,6 @@
-use std::{cell::RefCell, fmt, future::Future, rc::Rc};
+use std::{cell::RefCell, fmt, future::Future, marker::PhantomData, rc::Rc};
 
-use actix_http::Extensions;
+use actix_http::{body::BoxBody, Extensions};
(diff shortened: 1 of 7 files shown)


src/resource.rs
```diff
@@ -534,11 +546,11 @@ mod tests {
...
-                // .wrap_fn(|req, srv| {
-                //     let fut = srv.call(req);
-                //     async { Ok(fut.await?.map_into_right_body::<()>()) }
-                // })
-                .wrap(Compat::new(DefaultHeaders::new()))
+                .wrap_fn(|req, srv| {
+                    let fut = srv.call(req);
+                    async { Ok(fut.await?.map_into_right_body::<()>()) }
+                })
+                .wrap(Compat::noop())
```
aliemjay, line 553: I think we should encourage returning `impl HttpServiceFactory` for `Scope` and `Resource`.

robjtede: I agree, will add it in. I included this test to show that the changes do not interfere with these possible-in-the-wild signatures.

PR #2531, opened 2021-12-20
Make guard before io creation when requiring new connection poll

### LINKED ISSUE #2249: Awc client didn't release when io error occured
On release web-v3.2.0, when using a client repeatly in a loop to do request and get io error, the client will run out of connection pool's limit, and any other request by the client would be blocked and get timeout error.

## Expected Behavior
Always print `error, Failed to connect to host: Connection refused (os error 111)`
## Current Behavior
Print "error, Failed to connect to host: Connection refused (os error 111)" for "limit(10)" times
And then always print `error, Timeout while waiting for response`

## Possible Solution
In actix-http-2.2.0/src/client/pool.rs line 130 to 134, the guard(which I thought is used to realse when io error occured) should be made before `connector.call(req).await?;`

```rust
use actix_web::client::{Client, Connector};
use actix_web::rt;
use std::time::Duration;
#[actix_web::main]
    let conector = Connector::new().limit(10).finish();
    let client = Client::builder().connector(conector).finish();
    loop {
        // unopened port 2333
        let uri = format!("http://127.0.0.1:2333");
        match client.get(uri).send().await {
            Ok(mut resp) => {
                let _payload = resp.body().await;
            }
            Err(e) => {
                println!("error, {}", e);
            }
        };
        rt::time::delay_for(Duration::from_millis(100)).await;
    }
}
```

## Context
I use client to fetch data from many different backends periodly, now I have to renew the client instance to work around this bug.

## Your Environment

* Rust Version (I.e, output of `rustc -V`):rustc 1.52.1 (9bc8c42bb 2021-05-09)
* Actix Web Version:web-v3.2.0

hikaricai:
## PR Type

Bug Fix

## PR Checklist

## Overview

fixes #2249

### DIFF
--- a/actix-http/src/client/pool.rs
+++ b/actix-http/src/client/pool.rs
@@ -126,12 +126,11 @@ where
                     ))
                 }
                 Acquire::Available => {
+                    let guard = OpenGuard::new(key, inner);
                     // open tcp connection
                     let (io, proto) = connector.call(req).await?;
 
                     let config = inner.borrow().config.clone();
-
-                    let guard = OpenGuard::new(key, inner);
 
                     if proto == Protocol::Http1 {
                         Ok(IoConnection::new(


robjtede:
just realized this is targetting 3.0 branch, unfortunately we're not accepting patches for v3

hikaricai:
Wating for 4.0

PR #2549, opened 2021-12-26
rework Handler trait

aliemjay:
## PR Type

Refactor

## PR Checklist

## Overview

It is now parameterized over Args only. The output type is now an associated type, which is more sematically correct and is also consistent with `std::ops::Fn` and `Service`.

This also removes the `Responder` bound from the trait. This is in order to have more clear errors for handlers with invalid output: `Responder not implmented` vs `Handler not implemented`

We can also move the `Future` bound from the trait to `XX::to()`  for the same reason as above, but It is a less common mistake to miss the `async` keyword from handlers, so I'm not sure about this!

At this point I think it is better renamed to `AsyncFn`, which I think describes it better since it more general now.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -17,7 +17,7 @@ use crate::{
 ///   [`HttpResponse`] (i.e., it implements the [`Responder`] trait).
 ///
 /// # Compiler Errors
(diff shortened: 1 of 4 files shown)


robjtede (approved):
Nice, I like the semantics improvement.

PR #2554, opened 2021-12-28
never return port in `realip_remote_addr`

### LINKED ISSUE #2528: `ConnectionInfo::realip_remote_addr()` may or may not contain a port number which makes it difficult to handle robustly
## Expected Behavior
The expected behavior is for `ConnectionInfo::realip_remote_addr()` to return a consistently formatted address, either with or without a port number. Preferably without as appending one when the `X-Forwarded-For` header is supplied would be incorrect (as the port number of the peer will not be the same port as the true peer if the request is being forwarded via a proxy, so copying the port number over is misleading).

## Current Behavior
If there is no `Forwarded` or `X-Forwarded-For` header, `.realip_remote_addr()` returns the same value as `.remote_addr()` which is a socket address containing both an IP address and a port number.

However, if one of those headers _is_ present then `realip_remote_addr()` will return just a bare IP address.

This makes it very difficult to parse and consume this value because `IpAddr::from_str()` will reject it in the former case and `SocketAddr::from_str()` will reject it in the latter case, so one is forced to attempt both.

Of course, since this value is set by the client and potentially untrustworthy (as the docs make very clear), then it may not be any kind of address at all. However, the issue here is that even _when_ it's valid, it's still unnecessarily difficult to work with.

## Possible Solution
Instead of directly forwarding to `.remote_addr()`, since the latter is set using `SocketAddr::to_string()` anyway, `realip_remote_addr` should default to `socket_addr.ip().to_string()`.

This shouldn't break any existing usages that are _correctly_ implemented as any attempt to parse this string as either a socket address or IP address without attempting both would have been error-prone to begin with.

This does have the downside of an extra, semi-redundant string allocation. As an optimization, the value stored internally could be an enum that's either a separately allocated `String` or a `Range` that can be used to slice the `remote_addr()` string to just the IP address segment. I'm not certain this would be necessary, though.

1. Write a request handler that attempts to parse the value of `ConnectionInfo::realip_remote_addr()` as a `SocketAddr`:
```rust
#[get("/")]
async fn log_realip_remote_addr(req: HttpRequest) -> HttpResponse {
    let realip_remote_addr: SocketAddr = req.connection_info().realip_remote_addr().unwrap().parse().unwrap();

    log::info!("Received connection from {}", realip_remote_addr);

    HttpResponse::Ok().finish()
}
```
2. Make a request to that handler and see that it completes normally.
3. Manually add `X-Forwarded-For: <any IP address>` to the request.
4. Observe the handler panic.

## Context
When using the API of Circle, a payment provider, mutating requests that touch real money require passing a `metadata` object with information that can be used to find the user that originated the request in case of an audit: https://developers.circle.com/reference#payments-payments-create

This includes the user's IP address.

When the Actix-web application is deployed as part of a Kubernetes cluster in Google Cloud, and configured with an external load balancer, the `.peer_addr()` of the request will be that of the load balancer itself in Gcloud, so this cannot be used. Gcloud will append the IP address that originated to the request to the `X-Forwarded-For` header, creating it if it does not exist: https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header

Thus, the `.realip_remote_addr()` will be a socket address when testing the application locally, but a bare IP address when deployed. Ideally it would just always be an IP address for consistency and ease of handling.

## Your Environment

* Rust Version (I.e, output of `rustc -V`): 1.56.0
* Actix Web Version: `4.0.0-beta.8` (looking at the code, the behavior is identical on `-beta.15`)

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
Normalizes `realip_remote_addr` so that a port is never included.

Closes #2528

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -128,7 +128,7 @@ impl AppConfig {
 
     /// Server host name.
     ///
-    /// Host name is used by application router as a hostname for url generation.
+    /// Host name is used by application router as a hostname for URL generation.
     /// Check [ConnectionInfo](super::dev::ConnectionInfo::host())
     /// documentation for more information.
     ///
@@ -137,7 +137,7 @@ impl AppConfig {
         &self.host
     }
 
-    /// Returns true if connection is secure(https)
+    /// Returns true if connection is secure (ie. running over `https:`).
     pub fn secure(&self) -> bool {
         self.secure
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -137,7 +137,7 @@ impl AppConfig {
         &self.host
     }
 
-    /// Returns true if connection is secure (ie. running over `https:`).
+    /// Returns true if connection is secure (i.e., running over `https:`).
     pub fn secure(&self) -> bool {
         self.secure
     }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 - The `Not` guard is now generic over the type of guard it wraps. [#2552]
 
 ### Fixed
+- Rename `ConnectionInfo::{remote_addr => peer_addr}`, deprecating the old name. [#2554]
+- `ConnectionInfo::peer_addr` will not return the port number. [#2554]
 - `ConnectionInfo::realip_remote_addr` will not return the port number if sourcing the IP from the peer's socket address. [#2554]
 
 [#2552]: https://github.com/actix/actix-web/pull/2552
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -137,7 +137,7 @@ impl AppConfig {
         &self.host
     }
 
-    /// Returns true if connection is secure(https)
+    /// Returns true if connection is secure (ie. running over `https:`).
```
popzxc:
```suggestion
    /// Returns true if connection is secure (i.e. running over `https:`).
```

popzxc (approved):

PR #2566, opened 2022-01-04
fully percent decode path segments when capturing

### LINKED ISSUE #2248: [RFC] percent-encoding in URL path paramters
# The problem
When extracting URL path parameters from `Path` extractor, the user is expected to be aware of percent-encoding because he would get some encoded characters. To date, all chars are decoded except for `/`, `+`, `%`.

This behavior is quite surprising for the user because of the lack of documentation and is also quite unusual for a web framework - percent-encoding is a different level of abstraction that the user should not be concerned about (`Query` extractor deals with this correctly). 

The whole reason for this behavior is that the router stores URL as a single string with slash separated segments and matches a single regex against it. This is why `/` need to be escaped, to not change semantics.

Related:  #182 #209 #556 #1157

# The proposal
What i think is a proper implementation is to store the url path and the matching pattern (`ResourceDef`)  as array of segments with each pattern segment matches a single url path segment. This has the benefit of simplifying code by offloading uand even making `regex` an optional dependency, but it also would not allow patterns, such as `"/user/u{id:.*}"`, that span multiple segments. Path tail patterns (e.g. `"/files/{path:.*}"`) can be implemented as a special case but they should then be de-serialized into `Vec<_>` not `String`.

An alternative easier implementation is to percent decode the matched url parameters when de-serializing `Path`, but then multi-segment matches should also be prohibited.

robjtede:
## PR Type

Fix

## PR Checklist

## Overview

Should close #2248

### DIFF
--- a/actix-router/src/de.rs
+++ b/actix-router/src/de.rs
@@ -2,7 +2,11 @@ use serde::de::{self, Deserializer, Error as DeError, Visitor};
 use serde::forward_to_deserialize_any;
 
 use crate::path::{Path, PathIter};
-use crate::ResourcePath;
+use crate::{Quoter, ResourcePath};
+
+thread_local! {
+    static FULL_QUOTER: Quoter = Quoter::new(b"+/%", b"");
+}
 
 macro_rules! unsupported_type {
     ($trait_fn:ident, $name:expr) => {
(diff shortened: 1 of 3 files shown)


actix-router/src/de.rs
```diff
@@ -185,7 +188,11 @@ impl<'de, T: ResourcePath + 'de> Deserializer<'de> for PathDeserializer<'de, T>
                 .as_str(),
             ))
         } else {
-            visitor.visit_str(&self.path[0])
+            let decoded = FULL_QUOTER
+                .with(|q| q.requote(self.path[0].as_bytes()))
+                .unwrap_or_else(|| self.path[0].to_owned());
+
+            visitor.visit_str(&decoded)
```
aliemjay:
```suggestion
            visitor.visit_string(decoded)
```
to avoid an allocation.

PR #2579, opened 2022-01-06
Lower MSRV to 1.51.0

david-mcgillicuddy-moixa:
Replace use of `str::split_once`, `str::rsplit_once`, and `concat!` inside doc comments, to enable 1.51 compat

## PR Type

Other

## PR Checklist

## Overview

Hi, as previously discussed in https://github.com/actix/actix-net/pull/434 we here at Moixa are stuck on `1.51.0` and currently stuck on old versions of the various actix beta crates. This fork (and PR) reverts the MSRV to 1.51.0, and not before that due to use of cargo's resolver feature.

I haven't quite finished the supporting changes such as the changelog, CI, and label changes since this PR is more substantial than the `actix-net` one and impacts the crate more significantly (but still invisibly from the outside) with the doc comments being more cumbersome to write and edit so I wanted to see what you thought about upstreaming it here first before doing that stuff.

I have generated the documentation and couldn't see any changes at all as I'd hope, and the new functions are already covered by the existing tests.

### DIFF
--- a/actix-http/src/header/shared/quality_item.rs
+++ b/actix-http/src/header/shared/quality_item.rs
@@ -99,8 +99,7 @@ impl<T: str::FromStr> str::FromStr for QualityItem<T> {
         let mut raw_item = q_item_str;
         let mut quality = Quality::MAX;
 
-        let parts = q_item_str
-            .rsplit_once(';')
+        let parts = str_rsplit_once(q_item_str, ';')
             .map(|(item, q_attr)| (item.trim(), q_attr.trim()));
 
         if let Some((val, q_attr)) = parts {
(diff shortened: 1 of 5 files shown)


robjtede:
I've realised that this might not be possible due our dependency on `cookie` [which is MSRV 1.53](https://github.com/SergioBenitez/cookie-rs/blob/master/CHANGELOG.md#breaking-changes) and `rustls` [which is MSRV 1.52.1](https://github.com/rustls/rustls/blob/main/.github/workflows/build.yml#L19).

To be honest, this is starting to look like more of a pain for maintenance that I would like.

david-mcgillicuddy-moixa:
Yes I agree on reflection, that doc comment change turns some awkward code into something more unpleasant. Okay forget this actix-web one then, although as said I'll continue to maintain the fork (and possibly fork those libraries too) at least while Moixa hasn't upgraded its build just to let anybody know who stumbles across this.

The actix-net one seems much more reasonable to me though so I would like to continue discussing that one if you're amenable.

Issue #2460, opened 2021-11-24
Improve naming in the cookie API


LukeMathWalker:
`HttpResponse` currently provides two state-changing methods for cookies:

- `add_cookie`;
- `del_cookie`.

Based on the naming, I'd expect `add_cookie` to add a `Set-Cookie` header that matches the cookie I passed as argument, while `del_cookie` would add a `Set-Cookie` header with `max-age` equal to 0 to unset the cookie.

That's not what happens though: our intuition is right for `add_cookie`, but it fails us on `del_cookie`, which is instead designed to remove a cookie that was previously added via `add_cookie`.

Possible ways forward:

- (backward-compatible) provide an `unset_cookie` method, while documenting the difference between `add_cookie` and `del_cookie`;
- (backward-incompatible) reshape the cookie API to expose a mutable container, therefore removing the need for `del_cookie`, while exposing an `unset_cookie` method to set removal cookies.

robjtede:
I've noticed that `HttpResponseBuilder::del_cookie`, which does operate on a `CookieJar` will produce removal cookies. These same-name methods should have the same behavior.

PR #2586, opened 2022-01-16
add `HttpResponse::add_removal_cookie`

robjtede:
## PR Type

Feature

## PR Checklist

## Overview

Addresses #2460

### DIFF
--- a/src/response/response.rs
+++ b/src/response/response.rs
@@ -27,7 +27,7 @@ use crate::{error::Error, HttpRequest, HttpResponseBuilder, Responder};
 /// An outgoing response.
 pub struct HttpResponse<B = BoxBody> {
     res: Response<B>,
-    pub(crate) error: Option<Error>,
+    error: Option<Error>,
 }
 
 impl HttpResponse<BoxBody> {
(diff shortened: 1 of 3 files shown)


src/response/response.rs
```diff
@@ -116,18 +116,48 @@ impl<B> HttpResponse<B> {
...
+    /// - an empty value;
+    /// - a max-age of `0`;
+    /// - an expiration date far in the past.
+    ///
+    /// # Errors
+    /// Returns an error if the given name results in a malformed `Set-Cookie` header.
+    #[cfg(feature = "cookies")]
+    pub fn add_removal_cookie(&mut self, name: &str) -> Result<(), HttpError> {
+        let mut removal_cookie = Cookie::new(name, "");
+        removal_cookie.make_removal();
```
LukeMathWalker, line 150: TIL this method exists.

robjtede:
Actually, it's a little wrong to add this in its current form. The app might have set a cookie with domain or path attributes and they need to match in the removal cookie in order to be effective.

LukeMathWalker:
Should we ask for a `Cookie` as input rather than a `name` and then call `make_removal` in the method?

PR #2590, opened 2022-01-19
deprecate `Path::path`

### LINKED ISSUE #2588: `actix_router::Path::{path, unprocessed}` are equivalent!
They both have exactly the same behavior.

Exisiting services may rely on this behavior (`Files` for example), so simply changing the behavior of `path` may introduce subtle bugs. I think renaming and the deprecating the old `path` is a better choice.

cc @robjtede. I think it is worth fixing before stabilizing `actix-router`.

## Your Environment

- Actix Web Version: v4.0-beta.20

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
Fixes #2588


aliemjay (approved):
Nice. Thanks!

PR #2601, opened 2022-01-24
remove impl Future for HttpResponse

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
This impl is useless in practice.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+### Removed
+- `impl Future for HttpResponse`. [#2601]
+
+[#2601]: https://github.com/actix/actix-web/pull/2601
 
 
 ## 4.0.0-beta.21 - 2022-01-21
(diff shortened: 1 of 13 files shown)


vbrandl:
Just a heads up:

I had the following handler function:

```rust
fn favicon32() -> HttpResponse {
    // FAVICON is a &[u8]
    HttpResponse::Ok().content_type("image/png").body(FAVICON)
}
```

And routed it using `.service(web::resource("/favicon.ico").route(web::get().to(favicon32)))`. This fails now with the following error:

```
error[E0277]: the trait bound `fn() -> HttpResponse {favicon32}: Handler<_>` is not satisfied
   --> src/lib.rs:488:72
    |
488 |             .service(web::resource("/favicon.ico").route(web::get().to(favicon32)))
    |                                                                     -- ^^^^^^^^^ the trait `Handler<_>` is not implemented for `fn() -> HttpResponse {favicon32}`
    |                                                                     |
    |                                                                     required by a bound introduced by this call
    |
```

I guess it was this change which caused the problem. I had to change the function signature to `async fn favicon32() -> HttpResponse` to make it work again.

If this is intended, that's OK. Just "This impl is useless in practice" might not be completely true...

PR #2615, opened 2022-01-28
Support serving pre-compressed files for static sites

### LINKED ISSUE #2415: static compression feature for actix-files

antonok-edm:
## PR Type

Feature

## PR Checklist

## Overview

This PR adds the `Files::try_compressed()` builder method. If a static files service is constructed with that method, each incoming request will automatically check for a pre-compressed `.gz`, `.br`, or `.zst` file on disk, in order of preference according to the `Accept-Encoding` header, before falling back to the normal uncompressed file. When a compressed file is selected, the `Content-Encoding` header is populated with the respective compression type, while the `Content-Type` and `Content-Disposition` fields retain the original file's values.

There should be no breaking changes here, ~although I've had to expose some fields in ways that may not be ideal (I'm looking at the `SUPPORTED_ENCODINGS` list in particular, and perhaps the new features I've copied straight from `actix-web` - let me know if you have a better idea of how to structure any of that!).~ Resolved, see https://github.com/actix/actix-web/pull/2615#issuecomment-1024539295

Closes https://github.com/actix/actix-web/issues/2415


antonok-edm:
After some reflection, I've realized there's no reason to tie the supported precompression formats here to the supported runtime compression formats in `actix-web`, since these ones don't actually depend on any additional dependencies. That means there's no need to reference the `SUPPORTED_ENCODINGS` from `actix-web`, and no need to have additional features exposed, and it makes things overall a bit neater.

I've pushed those changes.

lilnasy:
@robjtede miniserve depends on actix-files and would really benefit from this PR merging. Is there something blocking its review?

robjtede:
Just time allocation. Reviews from interested parties will help massively.

JohnTitor (approved):
Thank you for the PR and sorry for the delay!
LGTM with some tweaks.

PR #2620, opened 2022-02-01
Use `actix_rt::block_on` in examples

ibraheemdev:
## PR Type

Feature

## PR Checklist

## Overview

*Blocked on* https://github.com/actix/actix-net/pull/441.


robjtede:
Following on from the -net PR close...

Basically, I want it to be clear in these examples that a `System` is being created, which is still important to know in the case of using `awc` and WebSockets.

PR #2625, opened 2022-02-04
RC refinements

robjtede:
## PR Type

Refactor

## PR Checklist

## Overview
Stuff I've noticed while migrating our TFB code.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -2,11 +2,11 @@
 
 ## Unreleased - 2021-xx-xx
 ### Added
-- Implement `From<Vec<u8>>` for `Response<Vec<u8>>`. [#????]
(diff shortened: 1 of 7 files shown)


robjtede:
Actually `write_date_header` (previously `set_date`) [was doc(hidden) in v3](https://docs.rs/actix-http/2.2.2/src/actix_http/config.rs.html#179). It's used in TFB code though. Should we keep it hidden? What do you think @aliemjay?

aliemjay:
I've never been involved in this area before, but from a quick glance I see it like an alien api for ServiceConfig is being exposed. Making it doc(hidden) maybe better given it would make it exempt from our semver guarantees. Still not sure though!

robjtede:
> alien api for ServiceConfig

Yeah glad you brought that up, fresh pair of eyes are valuable. I agree; gonna hide it again.

PR #2631, opened 2022-02-05
add `middleware::logger::custom_response_replace`

### LINKED ISSUE #2627: feature: add ```middleware::logger::custom_response_replace```
## Expected Behavior
It should be possible to customize the logger with access to responses, in the same way we can customize it with access to requests.

## Current Behavior
Currently, it's only possible to customize with request access by using this api:
```rust
pub fn custom_request_replace(
    self,
    label: &str,
    f: impl Fn(&ServiceRequest) -> String + 'static
) -> Self
```

## Possible Solution
Have a more or less the same api for responses, which could look something like this:
```rust
pub fn custom_response_replace(
    self,
    label: &str,
    f: impl Fn(&ServiceResponse) -> String + 'static  // or maybe &HttpResponse?
) -> Self
```
However, it is important to note that after some code examinations, making this api for responses by simply following the existing api for requests turned out to be harder than I thought. ```HttpResponse<B>``` and ```ServiceResponse<B>``` both have a template parameter ```<B>``` which can be either dragged to a lot of other places (but then dictate a lot of uncomfortable and probably unnecassary changes) or dropped, but dropping it is also not trivial, since drop_body() takes ownership of self (and we only get the response by ref in that context).

Nevertheless, in my use case I only need to access the status code of the response. 
So a simplified approach to the matter can be to expose only the StatusCode. Since the status code can be easily obtained by both ```HttpResponse<B>``` and ```ServiceResponse<B>```, this prevents said problems and makes the implementation pretty straight forward. It could look something like:
```rust
pub fn custom_response_replace(
    self,
    label: &str,
    f: impl Fn(&StatusCode) -> String + 'static //just StatusCode instead of ServiceResponse or HttpResponse
) -> Self
```
But of course this solution is not as robust as the first one.. so I'm not entirely sure what is the best way to go here.

## Context
An example for a use case (what I personally needed for my application) is conditionally logging an error message with a reason depending on the response status. e.g. on 4xx/5xx log the error and the reason, and on 2xx/3xx log something else (or don't log anything at all).

[Discord thread](https://discord.com/channels/771444961383153695/939249525699256372)

benny-n:
## PR Type
Feature

## PR Checklist

## Overview
It is now possible to pass a closure to the logger, which recevies a ~`StatusCode`~ `ServiceResponse` and returns a `String` for use in the log line

Closes #2627


robjtede:
I figured out a way to execute the original plan of custom _response_-based replacements.

@benny-n let me know what you think of the solution.

robjtede (approved):

benny-n:
Hi @robjtede,
This is really great!
With these changes we can access a `ServiceResponse` in the custom function which is what I wanted in the original plan.
So if I understood correctly, the solution was to reconstruct the response into a temporary `ServiceResponse` without the actual body?

robjtede:
@benny-n Yeah, you got it. It works out okay because response bodies are not accessible with only shared refs so it doesn't matter what is in the body slot for our closure here. This way keeps the API free of generics.

actix-web/src/middleware/logger.rs
```diff
@@ -543,6 +633,16 @@ impl FormatText {
                 };
                 *self = FormatText::Str(s.to_string())
             }
+
+            FormatText::CustomResponse(_, status_fn) => {
+                let text = match status_fn {
```
benny-n: As we have `CustomResponse` instead of `CustomStatus` now, we should probably also change `status_fn` to `response_fn`

PR #2643, opened 2022-02-10
fix(docs): TestRequest example fixed

arpandaze:
## PR Type

Example Fixed

## PR Checklist

## Overview

Method signature for `insert_header` on `TestRequest` was changed from `insert_header(key: K, value: V)` to `insert_header(header: H)` on #1869 but example wasn't updated. This PR fixes the example.

### DIFF
--- a/actix-web/src/test/test_request.rs
+++ b/actix-web/src/test/test_request.rs
@@ -24,10 +24,10 @@ use crate::cookie::{Cookie, CookieJar};
 ///
 /// For unit testing, actix provides a request builder type and a simple handler runner. TestRequest implements a builder-like pattern.
 /// You can generate various types of request via TestRequest's methods:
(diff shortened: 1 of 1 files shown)


aliemjay (approved):
Thanks.

I wonder why it wasn't catched by doc tests! Maybe because of #[actix_web::test] attribute? We need to address the root cause.

arpandaze:
> Thanks.
> 
> I wonder why it wasn't catched by doc tests! Maybe because of #[actix_web::test] attribute? We need to address the root cause.

You were right about that. I tested by removing it and finally it showed errors. It also discovered two unnecessary `.unwrap()` few lines below. I have fixed and pushed them.

Regarding `#[actix_web::test]` line, I think I shouldn't remove it. How should that be fixed?

robjtede:
Bit of a hack but it works in this sorta special case. Thanks for spotting it @arpandaze.

PR #2658, opened 2022-02-22
prepare actix-router release 0.5.0

robjtede:
## PR Type

Release

## PR Checklist

## Overview
It's quite possible that actix-router will get folded into actix-web itself soon but worth making a stable release of 0.5.0 anyway.

Thoughts on the changelog pre-release rollup? Was thinking about doing this for other crates too. (See the rich diff / view file to see effect.)

### DIFF
--- a/actix-router/CHANGES.md
+++ b/actix-router/CHANGES.md
@@ -3,6 +3,77 @@
 ## Unreleased - 2021-xx-xx
 
 
+## 0.5.0 - 2022-02-22
+### Added
+- `Resource` is now implemented for `&mut Path<_>` and `RefMut<Path<_>>`. [#2568]
+- Add `Path::as_str`. [#2590]
+- Add `ResourceDef::set_name`. [#373][net#373]
+- Add `RouterBuilder::push`. [#2612]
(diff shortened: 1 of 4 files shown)


aliemjay (approved):
> Thoughts on the changelog pre-release rollup?

Just great :+1: . It would be nice if it also merges multiple iterations on the same api in a single entry,
like renaming capture_match_info and changing its signature. I can see it'd be a daunting task for actix-web though.

iirc clippy was complaining about adding must_use to ResourceDef::join. Might be a good idea to add it now.

robjtede:
> would be nice if it also merges multiple iterations on the same api in a single entry

Agree, think i spotted a couple of them here but will try a bit harder for AW; i can think of some obvious ones that should not be included.

> adding must_use to ResourceDef::join

Not sure what case you mean, and it would be non-breaking right?

aliemjay:
> Not sure what case you mean, and it would be non-breaking right?

Nevermind. I remembered having a clippy lint against ResourceDef::join not using must_use but It is clear now. maybe a clippy update!

PR #2690, opened 2022-03-10
[awc] change `client::Connect` to be public

joelwurtz:
## PR Type
Feature 

## PR Checklist

## Overview

The `client::Connector` is not usable ATM in a public version and we are using this to tunnel http request in case of an upgrade in a proxy middleware for actix-web.

However this may not be the wanted behavior (and maybe client:Connector should only be public to the crate then)

### DIFF
--- a/awc/CHANGES.md
+++ b/awc/CHANGES.md
@@ -1,6 +1,8 @@
 # Changes
 
-## Unreleased - 2022-xx-xx
+## Unreleased - 2023-xx-xx
+### Changed
+- `client::Connect` is now public to allow tunneling connection with `client::Connector`.
 
 
 ## 3.1.0 - 2023-01-21
--- a/awc/src/lib.rs
+++ b/awc/src/lib.rs
@@ -139,7 +139,7 @@ pub mod http {
 }
 
 pub use self::builder::ClientBuilder;
-pub use self::client::{Client, Connector};
+pub use self::client::{Client, Connect, Connector};
 pub use self::connect::{BoxConnectorService, BoxedSocket, ConnectRequest, ConnectResponse};
 pub use self::frozen::{FrozenClientRequest, FrozenSendBuilder};
 pub use self::request::ClientRequest;


joelwurtz:
Do you need more information or anything to review this PR ?

Having this in public allow us to create a reverse proxy with "Connection: Upgrade" support for websocket connection (or other protocols) otherwise it's not possible

popzxc (approved):

JohnTitor (approved):

JohnTitor:
@robjtede Any objections to this?

robjtede (approved):

robjtede:
Seems inconsequential but useful to someone so... sure :)

joelwurtz:
As an example here is how we use it, we write a reverse proxy and when we got an upgrade request (for websocket or any other protocol that use upgrade connection header) we do the following : 

```rust
let mut connector = Connector::new();
connector = connector.max_http_version(Version::HTTP_11);

let connect = Connect {
    uri: head.uri.clone(),
    addr: endpoint.map(|e| e.addr),
};

let conn_fut = match connector.finish().call(connect).await {
    Err(e) => {
        log::error!("cannot connect: {:?}", e);

        return (
            HttpResponse::build(StatusCode::SERVICE_UNAVAILABLE)
                .content_type("text/html charset=utf-8")
                .body(NOT_AVAILABLE),
            None,
        );
    }
    Ok(c) => c,
};

let (res_head, framed) = match conn_fut.open_tunnel(head).await {
    Err(e) => {
        log::error!("cannot send request: {:?}", e);

        return (
            HttpResponse::build(StatusCode::SERVICE_UNAVAILABLE)
                .content_type("text/html charset=utf-8")
                .body(NOT_AVAILABLE),
            None,
        );
    }
    Ok(r) => r,
};

let mut response_builder = HttpResponse::build(res_head.status);
self.map_headers(&mut response_builder, res_head.version, res_head.status, res_head.headers());

let (backend_stream_read, backend_stream_write) = tokio::io::split(framed.into_parts().io);

let client_stream_read = PayloadIoStream {
    payload,
    state: PayloadIoStreamState::PendingChunk,
};

let response_stream = into_bytes_stream(backend_stream_read);
let copy_future = CopyFuture {
    reader: client_stream_read,
    writer: backend_stream_write,
    read_done: false,
    pos: 0,
    cap: 0,
    buf: vec![0; 8096].into_boxed_slice(),
};

let handle = actix_rt::spawn(copy_future);

if let Some(backend) = endpoint {
    self.pool.release(backend);
}

return (response_builder.streaming(response_stream), Some(handle))
```

PR #2699, opened 2022-03-17
fix(docs): update `TestRequest` example in `actix-http`

orhun:
## PR Type

Fix (example)

## PR Checklist

- [x] ~~A changelog entry has been made for the appropriate packages.~~

## Overview

Request builder example in [`actix_http::test::TestRequest`](https://docs.rs/actix-http/3.0.4/actix_http/test/struct.TestRequest.html) is outdated and also marked as `ignore` so it wasn't checked during doctests.

This PR updates this example and removes the `ignore` attribute.

### DIFF
--- a/actix-http/src/test.rs
+++ b/actix-http/src/test.rs
@@ -21,26 +21,31 @@ use crate::{
 
 /// Test `Request` builder
 ///
-/// ```ignore
-/// # use http::{header, StatusCode};
-/// # use actix_web::*;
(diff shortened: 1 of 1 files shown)


robjtede:
I've chosen to just remove this doc test for now. It shouldn't be using `actix-web` parts any more so until I have `actix-test` finalized this is just wrong and there's nothing suitable to replace it yet. Thanks for flagging this and the PR 👍🏻

PR #2709, opened 2022-03-22
rework actix_router::Quoter

aliemjay:
## PR Type

Refactor

## PR Checklist

- [x] ~~A changelog entry has been made for the appropriate packages.~~

## Overview

Refactor Quoter to simplify logic and remove useless safe table (see the first commit).

Performance also improved to match that of `percent_encoding` crate.

### DIFF
--- a/actix-router/src/quoter.rs
+++ b/actix-router/src/quoter.rs
@@ -241,4 +241,16 @@ mod tests {
         let q = Quoter::new(b"", b"");
         assert_eq!(q.requote(b"/abc/../efg"), None);
     }
+
+    #[test]
+    fn pointless_safe_table() {
+        for safe_ch in 0..=127 {
+            let q = Quoter::new(&[safe_ch], b"");
+
+            for i in 0..=255 {
+                let test = format!("/%{:02X}/", i);
+                assert_eq!(q.requote(test.as_bytes()).unwrap(), &[b'/', i, b'/']);
+            }
+        }
+    }
 }
--- a/actix-router/src/quoter.rs
+++ b/actix-router/src/quoter.rs
@@ -1,4 +1,16 @@
-/// Partial Percent-decoding.
+/// Partial percent-decoding.
+///
+/// Performs percent-decoding on a slice but can selectively skip decoding certain sequences.
+///
+/// # Examples
+/// ```
+/// # use actix_router::Quoter;
+/// // + is set as a protected character and will not be decoded...
+/// let q = Quoter::new(&[], b"+");
+///
+/// // ...but the other encoded characters (like the hyphen below) will.
+/// assert_eq!(q.requote(b"/a%2Db%2Bc").unwrap(), b"/a-b%2Bc");
+/// ```
 pub struct Quoter {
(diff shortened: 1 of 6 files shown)


actix-router/src/quoter.rs
```diff
@@ -1,132 +1,87 @@
...
-            }
+        let (prev, ch) = self.decode_next(&mut remaining)?;
+        let mut buf = Vec::<u8>::with_capacity(val.len());
+        buf.extend_from_slice(prev);
+        buf.push(ch);
 
-            idx += 1;
+        while let Some((prev, ch)) = self.decode_next(&mut remaining) {
+            // this ugly conditional achieves +50% perf in cases where this is a tight loop.
+            if prev.len() != 0 {
```
robjtede: to make clippy happy: `!prev.is_empty()`

actix-router/src/quoter.rs
```diff
@@ -1,132 +1,87 @@
...
-                if cloned.is_none() {
-                    let mut c = Vec::with_capacity(len);
-                    c.extend_from_slice(&val[..idx]);
-                    cloned = Some(c);
-                }
-            } else if let Some(ref mut cloned) = cloned {
-                cloned.push(ch)
-            }
+        let (prev, ch) = self.decode_next(&mut remaining)?;
+        let mut buf = Vec::<u8>::with_capacity(val.len());
```
robjtede, line 71: At this line, we know there are replacements right? So in the end there will be more added to the vec than `val.len()`. Shall we just add a magic number of extra capacity to account for it?

robjtede:
For reference, ran the benchmark on master and then on this PR:

```
quoter_unquoted:
    time:   [690.11 ns 692.04 ns 694.04 ns]
    change: [-65.345% -65.194% -65.060%] (p = 0.00 < 0.05)
    Performance has improved.

percent_encode_unquoted:
    time:   [658.60 ns 660.16 ns 662.08 ns]

quoter_quoted:
    time:   [3.2017 us 3.2133 us 3.2251 us]
    change: [-42.618% -42.415% -42.200%] (p = 0.00 < 0.05)
    Performance has improved.

percent_encode_quoted:
    time:   [2.9895 us 2.9967 us 3.0047 us]
```

So yeah, nice improvement and simpler code 👍🏻

actix-router/src/quoter.rs
```diff
@@ -135,89 +92,59 @@ impl Quoter {
...
-            assert_eq!(from_ascii_hex(hex[i]).unwrap(), expected[i]);
-        }
-    }
-
     #[test]
     fn custom_quoter() {
         let q = Quoter::new(b"", b"+");
         assert_eq!(q.requote(b"/a%25c").unwrap(), b"/a%c");
-        assert_eq!(q.requote(b"/a%2Bc").unwrap(), b"/a%2Bc");
+        assert_eq!(q.requote(b"/a%2Bc"), None);
```
robjtede, line 142: its definitely surprising that this returned something before, havent dug into why but definitely feels more like a fix than a break

robjtede (approved):

PR #2714, opened 2022-03-28
propagate handler visibility through route macros

### LINKED ISSUE #2696: not using handlers declared with routing macro should be detectable
Hello,

I've been using actix-web for a while and it's working very well! There is one little thing that's been bothering me, though.
When I write services with actix-web I really like to use the macros for specifying method and path, as in the example below:
```rust
use actix_web::{get, web, App, HttpServer, Responder};

#[get("/hello/{name}")]
async fn greet(name: web::Path<String>) -> impl Responder {
    format!("Hello {name}!")
}

#[actix_web::main] // or #[tokio::main]
    HttpServer::new(|| {
        App::new().service(greet)
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```

I think it's great to have the path and method next to the implementation of the endpoint. However, nearly every time I add a new endpoint I forget to add the `service` to the app in when setting up the server, leading to a confused developer and a 404 when testing it out :)

I am not experienced enough with Rust's macro system to know if it's actually possible to do, but it would really be great if just using the macro could somehow add the service.

Chaostheorie:
## PR Type
Fix

## PR Checklist

## Overview

The visibility of the instrumented function is inherited to the generated struct by employing syn's `ItemFn.vis` attribute.

Closes #2696

### DIFF
--- a/actix-web-codegen/CHANGES.md
+++ b/actix-web-codegen/CHANGES.md
@@ -1,7 +1,7 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-
+- Preserve function visibility for routing macros
 
 ## 4.0.0 - 2022-02-24
 - Version aligned with `actix-web` and will remain in sync going forward.
--- a/actix-web-codegen/CHANGES.md
+++ b/actix-web-codegen/CHANGES.md
@@ -1,7 +1,10 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
-- Preserve function visibility for routing macros
+- Preserve function visibility for routing macros. [#2714]
+
+[#2714]: https://github.com/actix/actix-web/pull/2714
+
 
 ## 4.0.0 - 2022-02-24
 - Version aligned with `actix-web` and will remain in sync going forward.
--- a/actix-web-codegen/src/lib.rs
+++ b/actix-web-codegen/src/lib.rs
@@ -19,7 +19,8 @@
 //!
 //! # Single Method Handler
 //! There is a macro to set up a handler for each of the most common HTTP methods that also define
-//! additional guards and route-specific middleware.
+//! additional guards and route-specific middleware. This macros will inherit the visibility
+//! of the underlying handler.
 //!
 //! See docs for: [GET], [POST], [PATCH], [PUT], [DELETE], [HEAD], [CONNECT], [OPTIONS], [TRACE]
 //!
--- a/actix-web-codegen/src/route.rs
+++ b/actix-web-codegen/src/route.rs
@@ -291,6 +291,7 @@ impl ToTokens for Route {
             resource_type,
             doc_attributes,
         } = self;
+        let visibility = &ast.vis;
         let resource_name = resource_name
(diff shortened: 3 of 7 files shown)


robjtede:
I suspect this could be tested in a trybuild test and asserting the compiler output contains the desired warning.

Chaostheorie:
It doesn't seem like trybuild supports checking of compiler warnings. It would be possible to instead create a test that tests the visibility with modules instead, is this sufficient @robjtede ?

For example, the following should fail …

```rust
mod test {
#[get("/test")]
async fn test() ->  … { … }
}

fn main() ->  … {
    …
    App::new()
            .service(test::test)
    …    
}
```

… while on the other hand, this should work …

```rust
mod test {
#[get("/test")]
pub async fn test() ->  … { … }
}

fn main() ->  … {
    …
    App::new()
            .service(test::test)
    …    
}
```

robjtede:
Good idea. That would work too, yeah.

Chaostheorie:
The idea with trybuild tests worked out. Though, this most likely is a breaking change, right?

robjtede:
Only the question about breakage is keeping this from being merged. I've asked the other contributors for input.

My view is this could reasonably be called a fix and resulting compile errors "relying on a bug".

Also toying with the idea of issuing a warning at compile time or otherwise trying to figure out how to release this in a non-technically-breaking way but nothing promising has come up yet.

Chaostheorie:
Well, the docs AFAIK never actually mention the 'public' visibility of the routing macros, but since they've been public for at least v3 and now v4.0.1. Further, all examples of the routing macros in the actix-codegen docs are 'private' so they most likely need to be adjusted.

A warning for private visibility is also a problem since, e.g., the following pattern would be valid even after the change:

 <details>
  <summary>ServiceConfig Pattern</summary>

```rust
mod example {
    #[route("/test", method = "GET", method = "HEAD")]
    async fn example() -> HttpResponse {
        HttpResponse::Ok().finish()
    }

    pub fn configure_routes(cfg: &mut ServiceConfig) {
        cfg.service(example);
    }
}

#[actix_web::main]
pub fn main() -> … {
    HttpServer::new(move || {
        App::new()
            .service(web::scope("/example").configure(example::configure_routes))
            .build()
    });
}
```

</details> 

The first thought coming to my mind is making the 'default' public visibility, a crate feature that is part of the default feature set. This would limit the problem to projects employing `default-features = false`, but the 'migration' would at least be easy/ cheap for those affected. In the long term, the feature could be made optional and/ or be removed.

robjtede:
Makes me sad but we can't really justify this before the next breaking change and I don't see a way to do it without breaks.

Chaostheorie:
Well, that's to be expected for such a change. Please ping me before the next breaking change, and I'll rebase and, if required, update the patch.

porkbrain:
Looking forward to this change. It will help us keep our app with 100s of endpoints cleaner.

Chaostheorie:
Closing this as it is outdated

PR #2718, opened 2022-03-30
`routes` macro allowing multiple paths per handler

### LINKED ISSUE #2713: Attribute macro multiple paths
I posted this originally as a comment in a different issue, but I think it deserves it's own issue for discussion.

Even tho I'm a bit late to the party with the `route` macro already existing, but I would propose something along the lines of 
```rs
#[routes]
#[delete("/{id}")]
#[get("/{id}/delete")]
async fn delete(id: Path<String>) -> GetResponse { 
//...
}
```

If this is something you would consider, I could also implement it. This would allow defining not only multiple methods but also routes, e.g. to allow the deprecated path to still function after refactoring the API schema.

This could also be implemented in a way that keeps the current functionality in the `#[route]` macro.

Theoretically, the `#[route]` could even be dropped. But that would mean that the first macro would implicitly do the `route` macros job, and all lower macros are just helpers that do not actually support specification via path etc.

_Originally posted by @ModProg in https://github.com/actix/actix-web/issues/1360#issuecomment-1079458839_

ModProg:
## PR Type

Feature

## PR Checklist

## Overview

Adds support for `routes` macro, allowing multiple paths and methods to be specified.

Closes #2713


ModProg:
This is now working and documented (already using it my self in https://github.com/ModProg/paste).

Would like some feedback on the code to get this in a merge able state ASAP, so I don't need to run on a git version forever.

actix-web-codegen/tests/trybuild.rs
```diff
@@ -12,6 +12,8 @@ fn compile_macros() {
     t.compile_fail("tests/trybuild/route-unexpected-method-fail.rs");
     t.compile_fail("tests/trybuild/route-malformed-path-fail.rs");
 
+    t.pass("tests/trybuild/routes-ok.rs");
```
popzxc, line 15: Maybe add some negative tests? E.g. with `routes` macro without any actual routes supplied?

popzxc (approved):
LGTM

actix-web-codegen/tests/trybuild.rs
ModProg: I think in that case it would just do nothing, but an error would make sense.

actix-web-codegen/src/lib.rs
```diff
@@ -104,6 +115,39 @@ pub fn route(args: TokenStream, input: TokenStream) -> TokenStream {
...
+/// # Syntax
+/// ```plain
+/// #[routes]
+/// #[<method>("path", ...)]
+/// #[<method>("path", ...)]
+/// ...
+/// ```
+///
+/// # Attributes
+/// The `routes` macro it self has no parameters, but allows specifying the attribute macros for
```
popzxc:
```suggestion
/// The `routes` macro itself has no parameters, but allows specifying the attribute macros for
```

actix-web-codegen/src/lib.rs
```diff
@@ -104,6 +115,39 @@ pub fn route(args: TokenStream, input: TokenStream) -> TokenStream {
...
+/// the different methods, e.g. [`GET`](macro@get) or [`POST`](macro@post).
```
popzxc:
```suggestion
/// the multiple paths and/or methods, e.g. [`GET`](macro@get) and [`POST`](macro@post).
```

popzxc (approved):
Nice!

ModProg:
Is there anything left for me to do to make this PR mergeable?

robjtede:
I haven't reviewed yet but I'm targetting this for v4.2, not the upcoming v4.1 so it will be merged after that release.

robjtede (approved):

robjtede:
CI issue was just missing a feature flag on `syn`. Cleaned up a bit too. Looking good now.

Thanks for the contribution.

ModProg:
Thanks, really looking forward to 4.2 now :)

PR #2734, opened 2022-04-15
Improve documentation for FromRequest::Future

mattfbacon:
## PR Type

Other: Documentation

## PR Checklist

- [N/A] Tests for the changes have been added / updated.
- [N/A] A changelog entry has been made for the appropriate packages.
- [N/A] (Team) Label with affected crates and semver status.

## Overview

Improve documentation for the `Future` associated type of the `FromRequest` trait. I was confused by this and made [a Rust forum post](https://users.rust-lang.org/t/actix-web-fromrequest-where-future-ready/74280/3) where I got some help so I thought it would make sense to add this to the documentation.

No breaking changes.

### DIFF
--- a/actix-web/src/extract.rs
+++ b/actix-web/src/extract.rs
@@ -67,6 +67,10 @@ pub trait FromRequest: Sized {
     type Error: Into<Error>;
 
     /// Future that resolves to a Self.
+    ///
+    /// To refer to the type of an async function or block here, you have two options.
+    /// The first is to use a boxed future such as `futures::future::BoxFuture`. This works on stable and nightly.
+    /// The second is to use `impl Future`, which requires `#![feature(type_alias_impl_trait)]` so only works on nightly.
     type Future: Future<Output = Result<Self, Self::Error>>;
 
     /// Create a Self from request parts asynchronously.
--- a/actix-web/src/extract.rs
+++ b/actix-web/src/extract.rs
@@ -68,9 +68,13 @@ pub trait FromRequest: Sized {
 
     /// Future that resolves to a Self.
     ///
-    /// To refer to the type of an async function or block here, you have two options.
-    /// The first is to use a boxed future such as `futures::future::BoxFuture`. This works on stable and nightly.
-    /// The second is to use `impl Future`, which requires `#![feature(type_alias_impl_trait)]` so only works on nightly.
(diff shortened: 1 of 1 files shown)


robjtede:
The best practice for Actix Web is to use `LocalBoxFuture` instead. I'd also prefer not to mention nightly features in our docs for now.

mattfbacon:
> The best practice for Actix Web is to use `LocalBoxFuture` instead. I'd also prefer not to mention nightly features in our docs for now.

OK

fakeshadow:
You can also make a new type and implement `Future` trait and `LocalBoxFuture` is just an alias for `Pin<Box<dyn Future>>`  .The doc of actix-web should not try to offer a hand holding tutorial on basic language feature. It can get outdated easily(async Rust is still in development) and it can't do a good job on explaining it thoroughly.

mattfbacon:
I tried to find resources on this information and couldn't find any. And even when I asked on the forum, a Rust (specifically async; Alice is a primary maintainer of Tokio) expert said something different than what you all are saying. So it's not true that it's a "basic language feature" (it's not a language feature at all, but that's beside the point).

Anyway, I think it would be helpful. There are no examples that I could find in the docs that used it.

mattfbacon:
Sorry, part of my previous comment was incorrect. I found one example using LocalBoxFuture in one place. :smiley:

fakeshadow:
I'm not saying the language feature have a good discovery. I'm saying said discovery does not belong to actix-web library. Making related question more easy to get answer of from other source is better than burden some library that not maintained by rust-lang domain.
With major improvement of async Rust language feature someone have to keep the information updated and who would do that for actix-web?

fakeshadow:
BTW `async-trait`'s readme have a detailed explain for this issue https://github.com/dtolnay/async-trait

mattfbacon:
>  I'm saying said discovery does not belong to actix-web library

The existence of boxed futures, no, but the Actix convention to use LocolBoxFuture yes.

> With major improvement of async Rust language feature someone have to keep the information updated

If the convention changes from LocalBoxFuture to something else (maybe `impl Future` once that stabilizes), that should be documented.

>  who would do that for actix-web?

Whoever created and/or will change the convention.

robjtede (approved):

PR #2745, opened 2022-04-25
fix erroneous 500 responses when connection is unclean

### LINKED ISSUE #2695: Actix Multipart Panics & Closes Connection on Early Response
## Expected Behavior
In a Multipart request, as parts are received a server may validate it, check the size for maximums, and more. If an error occurs and a user need to know about it, the server can issue an "early" response, where the other parts are ignored.

## Current Behavior
Actix-web returns an internal server error, closes connection, ends up with no content even though content was specified, and the worker panics with the message:
``thread 'actix-rt|system:0|arbiter:1' panicked at 'Multipart polled after finish'``

## Possible Solution
https://github.com/actix/actix-web/blob/03456b8a33a4550b94785a7612e16d715755cd00/actix-http/src/h1/dispatcher.rs#L739

This line of code on the actix web server should change. Requests should not return an error if part of the payload is dropped.

https://github.com/Geometrically/actix-web-bug-poc
Run the actix server above, and request using CURL.
```
curl -i -X POST http://127.0.0.1:8080 \ 
   --header 'Content-Type: multipart/form-data; boundary=---------BOUNDARY' \
   --data-binary @file.txt
```
The expected output is: `this does not show up :(`
However, no content is returned, and a panic is printed in the console.
`thread 'actix-rt|system:0|arbiter:1' panicked at 'Multipart polled after finish'`

## Context
My website, https://modrinth.com uses actix-web to power our open-source API (https://github.com/modrinth/labrinth). In our website, users can upload mods which have many files. We use multipart requests to do this. However, files are validated, so if an error occurs in validation it is returned early. Instead of returning the error, actix-web panics, causing the connection to close, so no error shows up to the user. Because of the panic, we cannot rollback steps in the project creation process.

## Your Environment
See minimum reproducible example, barebones actix environment with actix-multipart `0.4.0`

- Rust Version: 1.59.0
- Actix Web Version: 4.0.1/4

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
The fix for #2624 tipped the scale too far the other way and caused 500 error responses to get queued. This moves the unclean connection detection to after the responses are queued and adds `Connection: close` headers.

Expect some minor perf drop (untested) but this is an important fix.

fixes #2695

### DIFF
--- a/actix-http/src/h1/dispatcher_tests.rs
+++ b/actix-http/src/h1/dispatcher_tests.rs
@@ -861,56 +861,26 @@ async fn handler_drop_payload() {
             ",
         ));
 
-        assert!(h1.as_mut().poll(cx).is_pending());
-
-        // polls: manual => manual
-        assert_eq!(h1.poll_count, 2);
-
-        let mut res = BytesMut::from(buf.take_write_buf().as_ref());
-        stabilize_date_header(&mut res);
-        let res = &res[..];
-
-        // expect response immediately even though request side has not finished reading payload
-        let exp = http_msg(
-            r"
-            HTTP/1.1 200 OK
-            content-length: 15
-            date: Thu, 01 Jan 1970 12:34:56 UTC
-
-            payload dropped
-            ",
-        );
-
-        assert_eq!(
-            res,
-            exp,
-            "\nexpected response not in write buffer:\n\
-               response: {:?}\n\
-               expected: {:?}",
-            String::from_utf8_lossy(res),
-            String::from_utf8_lossy(&exp)
-        );
-    })
-    .await;
-
-    lazy(|cx| {
         assert!(h1.as_mut().poll(cx).is_ready());
 
-        // polls: manual => manual => manual
+        // polls: manual => manual => shutdown
         assert_eq!(h1.poll_count, 3);
 
         let mut res = BytesMut::from(buf.take_write_buf().as_ref());
         stabilize_date_header(&mut res);
         let res = &res[..];
 
-        // expect that unrequested error response is sent back since connection could not be cleaned
+        // expect response immediately even though request side has not finished reading payload
+        // since write buffer was "too short" we should expect a closed connection hint
         let exp = http_msg(
+            // connection: close
             r"
-            HTTP/1.1 500 Internal Server Error
-            content-length: 0
+            HTTP/1.1 200 OK
+            content-length: 15
             connection: close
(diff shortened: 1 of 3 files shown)


actix-http/src/h1/dispatcher_tests.rs
```diff
@@ -858,56 +858,26 @@ async fn handler_drop_payload() {
...
-
-        // expect that unrequested error response is sent back since connection could not be cleaned
-        let exp = http_msg(
-            r"
-            HTTP/1.1 500 Internal Server Error
-            content-length: 0
             connection: close
             date: Thu, 01 Jan 1970 12:34:56 UTC
 
+            payload dropped
```
Geometrically, line 883: Does this code retain responses a server gives when a payload is dropped? That is the intended behavior IMO, and is what actix did before 4.0.

Geometrically (changes requested):

Geometrically:
Regarding my previous comment, I don't think this is the case. We have deployed this fix (from this PR) in production at Modrinth, and users still are getting payload drops on their multipart requests.

This is a really big issue as when the payload drops and actix closes the connection, CORS headers aren't set so we can't access the response. This PR hasn't fixed the bug in my issue.

I feel like the previous approach was a lot better. The previous "fix" was for a very small use case. This issue is affecting big projects like MeiliSearch, and I don't think it fits with the actix architecture.

IMO, if I return an error in my response to a user, they should get that error. The connection shouldn't be instantly closed.

(Sorry if this wording was harsh- it's just a pretty big issue my team and I have been dealing with for the past couple of months)

robjtede:
I understand the severity of the issue (we're hitting this at my work as well) and also that this PR is an incomplete fix, hence why it is not merged and released.

I'm going to revert the change that created this problem, which will give us back the other bug that existing in v3 (citation needed) which was relatively less problematic. This will give me and anyone else who wants to try some more time to debug and fix both problems finally.

Expect v4.1 to be out by EOD Thursday.

Geometrically:
Thanks a lot for your work on actix :smile: 

Glad this issue is being fixed and love the work the team is doing. Besides this, actix has been an amazing platform for us to build our API on, and for that I am very thankful :heart: 

Can't wait for 4.1!

Geometrically:
Hey, I noticed this was moved to v4.2. Is there a fix for this issue being planned? Or at least a fork we can use which resolves it to the prior system?

PR #2752, opened 2022-05-05
Expose option for setting TLS handshake timeout

lulf:
## PR Type

Feature

## PR Checklist

## Overview

Apologies for just raising this PR without any prior discussion.  At present the TLS handshake timeout is hardcoded at 3 seconds internally in actix-tls. Having the ability to up this timeout is important for servers handling embedded clients that may take a while to do a handshake.

I'm attempting to expose the TLS handshake timeout as available on the Acceptor type in the actix-tls crate. The way I've done this is to add another option on the HttpServer if built with openssl or rustls. I'm very unfamiliar with the code, and I'm not particularily happy with the modification to the openssl/rustls-specific bits in actix-http (which is a bit incomplete missing the h1/h2 specific bits), which somewhat breaks the API, so I thought I'd just raise a draft PR as is to get the discussion going and get a better understanding on how you'd like this to be exposed.


robjtede:
This breaking API change would certainly not be ok to merge. It would need a new method adding.

lulf:
Thanks for the feedback @robjtede. I've made the following changes:

* Create the type TlsAcceptorConfig which holds the desired acceptor configuration, default being to use whatever actix-tls chooses.
* Create a rustls_with_config/openssl_with_config which takes the TlsAcceptorConfig parameter

Longer term, it feels like this could use some refactoring.

lulf:
Would anyone be able to review this again?

JohnTitor (commented):
Haven't reviewed deeply but adding timeout config makes sense to me (if no one reviews for a few weeks, I'm going to do a more detailed review).

robjtede (approved):
I reduced the API surface a bit and added relevant documentation but this approach is good to be added.

lulf:
Thanks!

PR #2769, opened 2022-05-30
Implement `ResponseError` for `Infallible`

SabrinaJewson:
## PR Type
Feature

## PR Checklist

    - N/A? I can’t really add tests that wouldn’t just be repeating what the code says
    - Also N/A IMO, I don’t think it really needs docs

## Overview

I saw someone on Discord encounter a problem because it wasn’t present earlier today. It seems reasonable, so I added it. The use case was in the `Result` returned from [`actix_form_data::Field::file`](https://docs.rs/actix-form-data/latest/actix_form_data/enum.Field.html#method.file).


SabrinaJewson:
Ah…CI is failing but it doesn’t seem to be related to this PR, just that hashbrown no longer compiles on the MSRV.

robjtede (approved):
Neat change. Thank you 👍 

We'll sort through CI later.

PR #2772, opened 2022-06-02
Read remaining bytes when cleaning dropped payload #2764

### LINKED ISSUE #2764: Failure to flush Payload when it is dropped before handler consumes all bytes
In cases where a handler exits early without reading a supplied `Payload`, there is some new code in actix-web (see PR #2624) that consumes the rest of the bytes in the request.

It seems not to work with large request bodies (approaching 512k).

## Expected Behavior
If a `Payload` is dropped before all bytes are consumed, the rest of the client request should still be consumed.

## Current Behavior
If a `Payload` is dropped before all bytes are consumed, the rest of the request is consumed if the request is small, but larger requests stop reading and drop the connection with a message `"handler did not read whole payload and dispatcher could not drain read buf; return 500 and close connection"`.

## Possible Solution
Haven't quite found the root cause yet, but it appears that the `read_buf` in the `Dispatcher` is empty. When `PayloadDecoder::decode()` is called with `read_buf` as `src`, it reports zero bytes available, but `self.kind` contains a nonzero `Kind::Length` of remaining bytes. If they were both correctly zero, it would return `PayloadItem::Eof`, and the flush would terminate correctly. If they were both correctly nonzero, it would successfully return `PayloadItem::Chunk(buf)`, which would continue the flush loop until exhausted.

I'm not sure which one is wrong here, but it only seems to happen if there are a certain number of reads required to flush the request.

Minimal server:
```rust
use actix_web::{post, web::Payload, App, HttpServer};

#[post("/hello")]
async fn greet(_: Payload) -> &'static str {
    "Ok\n"
}

#[actix_web::main]
    env_logger::init();
    HttpServer::new(|| App::new().service(greet))
        .bind(("127.0.0.1", 5000))?
        .run()
        .await
}
```
Client-side - I was able to duplicate this by sending a 511k file (succeeded), and a 512k file (failed), but managed to narrow down to exact number of bytes by using `curl`'s `-C` to skip initial bytes in the file. Byte ranges may vary across OSs, not sure.
```bash
# create a 512k file of random data
dd if=/dev/random of=512k bs=1024 count=512
# this succeeds - continue sending data from byte 1753
curl -C 1753 -H 'Content-Type: application/octet-stream' -d @512k -v http://127.0.0.1:5000/hello
# this fails - continue sending data from byte 1752
curl -C 1752 -H 'Content-Type: application/octet-stream' -d @512k -v http://127.0.0.1:5000/hello
```

## Context
I'd like to be able to return from a handler without having to worry about whether I've finished reading the `Payload` completely. For example, streaming a file to disk, then having a write fail, it's much easier to handle with a simple `?` and a handler returning a `Result`, than having to wrap the whole thing in a handler that cleans up the `Payload` before returning an error.

## Your Environment
- MacOS 12.4 x86_64
- actix-web 4.0.1
- actix-http 3.0.0
- rustc 1.61.0 (fe5b13d68 2022-05-18)

squidpickles:
## PR Type

Bug Fix

## PR Checklist

## Overview

Existing behavior did not completely read client data if a `Payload` was dropped without consuming all of its data. This calls `read_available()` before each `Payload` chunk is discarded. This does add a `read_available_projected()` so it can be called when only a pinned reference to the dispatcher is available.

Closes #2764

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -1,6 +1,8 @@
 # Changes
 
-## Unreleased - 2021-xx-xx
+## Unreleased - 2022-xx-xx
+### Fixed
+- Consume bytes from read buffer when `Payload` is dropped early. [#2764]
 
 
 ## 3.0.4 - 2022-03-09
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -706,6 +706,9 @@ where
                     debug!("handler dropped payload early; attempt to clean connection");
                     // ...in which case poll request payload a few times
                     loop {
+                        if this.read_buf.is_empty() {
+                            Self::read_available_projected(&mut this, cx)?;
+                        }
                         match this.codec.decode(this.read_buf)? {
                             Some(msg) => {
                                 match msg {
@@ -1010,8 +1013,22 @@ where
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Result<bool, DispatchError> {
-        let this = self.project();
+        let mut this = self.project();
+        Self::read_available_projected(&mut this, cx)
+    }
 
+    /// Returns true when I/O stream can be disconnected after write to it.
+    /// Meant to be called when there is already access to a projected
+    /// `InnerDispatcher` available.
+    ///
+    /// It covers these conditions:
+    /// - `std::io::ErrorKind::ConnectionReset` after partial read;
+    /// - all data read done.
+    #[inline(always)] // TODO: bench this inline
+    fn read_available_projected(
+        this: &mut InnerDispatcherProj<'_, T, S, B, X, U>,
+        cx: &mut Context<'_>,
+    ) -> Result<bool, DispatchError> {
         if this.flags.contains(Flags::READ_DISCONNECT) {
             return Ok(false);
         };
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -22,7 +22,7 @@ use crate::{
     config::ServiceConfig,
     error::{DispatchError, ParseError, PayloadError},
     service::HttpFlow,
-    ConnectionType, Error, Extensions, OnConnectData, Request, Response, StatusCode,
+    Error, Extensions, OnConnectData, Request, Response, StatusCode,
 };
 
 use super::{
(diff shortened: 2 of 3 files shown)


squidpickles:
I noticed curl stops sending bytes when the request is over a certain size, and the server returns an error response. (curl message `HTTP error before end of send, stop sending`) 

I can see that we're reading zero bytes in the added read here https://github.com/squidpickles/actix-web/blob/30004d5b9f7fb7f03ed98f5a6f0f26ee67c17186/actix-http/src/h1/dispatcher.rs#L710 but it's not clear whether the `true` return (called `should_disconnect` elsewhere) is correct in all cases. Investigating... (also not clear how to write a unit test that mimics this client behavior)

squidpickles:
Still working on this, partly because I realized I'm not totally clear what the desired behavior should be here.

Right now (in the released code), if the Payload is dropped before it has been read completely, the Dispatcher will attempt to "clean" the connection by consuming any bytes in the read buffer. The trouble is that if there are more than `decoder::MAX_BUFFER_SIZE` bytes sent by the client, the cleaning process will fail (because `PayloadDecoder::decode()` will be called with bytes remaining and an empty read buffer), and it will jump to the 500 error and connection close.

I added an additional read into the cleaning loop if the read buffer is empty, so now the bytes will be consumed even if there are more than a single read buffer's worth.

However, the situation is complicated because a handler that does not consume the whole Payload will return a response to the client before all bytes are read from the wire. Some clients, such as cURL, will stop sending data if they get an early error response.

I think the main thing I would change here is the 500 error response, and the error logging. It seems best not to log a server error when it's a relic of client behavior. In this case, the 500 is triggered by a client either disconnecting or stopping sending bytes before the cleaning process completes. The handler has already sent a response if it was ever going to, so nothing further really needs to be sent to the client. The only case where it should be logging an error is if there's some reason the server can't read, but the client is still sending bytes. I'm not sure that's something we'd catch here and not elsewhere, so probably not worth treating as a potential error in this situation.

So my proposal is to disconnect when client send terminates early (fewer bytes are sent than the content-length header suggests) as the server does now, but not to send an additional 500 response, and not to log an error. All while preserving the extra read added in this PR, so larger requests still get cleaned properly.

squidpickles:
Ok, well, it looks like that extra read hangs the worker. I haven't worked out the polling mechanism well enough to understand why. Going to leave this PR open and maybe someone with better understanding can comment, but for now, I'll work around it and explicitly consume the entire Payload before returning from my handlers.

PR #2777, opened 2022-06-06
update msrv to 1.56

### LINKED ISSUE #2747: consider making firestorm dependency of actix-router optional
I've noticed `firestorm` dependency in Cargo.lock of my application. This dependency comes from `actix-router`. My understanding is that this is  a dev-tool used for internal profiling of actix-router. 

I understand that it is a no-op crate by default, but I still would prefer not to have this entry in my Cargo.lock to avoid worrying about supply chain attacks and spending time on figuring out why this relatively unknown crate is there.

robjtede:
## PR Type

MSRV

## PR Checklist

## Overview
Bump MSRV to 1.56, following `hashbrown`.

Generally, it's fine to do this in a minor release.


JohnTitor:
CI failed:
```
error: package `dashmap v5.3.4` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.56.0
```

We have to bump up to 1.59, it seems. It also may be better to describe that this is due to our dependency on changelog, not to confuse users.

JohnTitor (approved):
LGTM with updating changelog (I mean something like "... is now 1.56. This is because one of our dependencies updated MSRV without a major release.". Some of users should be aware of it already but just in case someone wonders why actix breaks MSRV in a minor version.)

PR #2784, opened 2022-06-15
Add ability to set default error handlers to the `ErrorHandler` middleware

### LINKED ISSUE #2667: Provide/document a way to set a blanket/default handler for 4xx/5xx errors
## Expected Behavior
There should be a documented way to set a default handler for 4xx and 5xx errors.

## Current Behavior
```rust
 .wrap(ErrorHandlers::new()
    // XXX: Check for new `StatusCode` variants on each actix-web upgrade
    .handler(StatusCode::UNAUTHORIZED, render_error)  // Visible on 'Cancel' click
    .handler(StatusCode::FORBIDDEN, render_error)
    .handler(StatusCode::NOT_FOUND, render_error)
    .handler(StatusCode::METHOD_NOT_ALLOWED, render_error)
    .handler(StatusCode::GONE, render_error)
    .handler(StatusCode::IM_A_TEAPOT, render_error)
    // ...
    .handler(StatusCode::INTERNAL_SERVER_ERROR, render_error)
    .handler(StatusCode::NOT_IMPLEMENTED, render_error))
    // ...
```

...which is an opportunity for bugs if the developer accidentally adds something to the list that shouldn't be there.

## Possible Solution
If this isn't already concisely possible, maybe add a `.default_handler` function to the `ErrorHandlers` middleware to specify a handler that will be invoked if none of the more specific handlers match?

If it is concisely possible, Google wasn't helping and I didn't see anything relevant mentioned in the parts of the rustdocs I'm already aware of, so maybe add a link from the `ErrorHandlers` documentation if that's the case.

The main issue I foresee is ensuring that documentation is clear about how things are intended to work in situations like "most routes just use the default response, but one sets a custom body that should be passed through".

## Context
I'm trying to ensure that all error messages that the browser might present to the user will use a template that is stylistically consistent with the 200 OK base template.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): rustc 1.59.0 (9d1b2106e 2022-02-23)
- Actix Web Version: 4.0.1

junbl:
## PR Type

Feature

## PR Checklist

## Overview

Currently, each error handler has to be set separately for each status code. It's cumbersome to have to add a line for each status code, and ripe for errors if some codes are missed.

This PR allows the user to set a default handler either for all errors, or separate defaults for client and server errors, while still allowing other handlers to be tied to specific status codes. It defines the following new methods on `middleware::ErrorHandlers`:
 - `default_handler`: sets a default handler that will be used if the given status code doesn't have another handler specified
 - `default_handler_client`: sets a default handler only for client errors (400-499)
 - `default_handler_server`: sets a default handler only for server errors (500-599)

This PR contains no breaking changes; other than the additions above, the only changes are to private fields/types.

Closes #2667.


actix-web/src/middleware/err_handlers.rs
```diff
@@ -53,7 +55,38 @@ type ErrorHandler<B> = dyn Fn(ServiceResponse<B>) -> Result<ErrorHandlerResponse
 ///     .wrap(ErrorHandlers::new().handler(StatusCode::INTERNAL_SERVER_ERROR, add_error_header))
 ///     .service(web::resource("/").route(web::get().to(HttpResponse::InternalServerError)));
 /// ```
+/// ## Registering default handler
```
robjtede, line 70: Explain the precedence of overlapping handlers.

junbl: hope the new docs are clearer. lmk if anything still doesn't make sense!

popzxc (approved):

junbl:
Check failed due to <https://github.com/rust-lang/rust/issues/99261> which looks to have been fixed on latest nightly

robjtede (approved):

robjtede:
Thanks for the contribution 👍🏻

PR #2785, opened 2022-06-17
Fixed code example in README

deviatorslegacy:
A call to the `format!` macro was implying the use of a named argument
called `name` in the format string, but the argument wasn't specified. I fixed
this by using the simple {} notation in the format string and passing
the `name` variable as a second argument.

Would it be appropriate to log this in a CHANGES.md file?

This is my first ever pull request, so let me know if I'm doing anything wrong! :)

## PR Type

Documentation

## PR Checklist

## Overview

The change will not affect any in-project code, rather the example in the README.
The example has an error on line 5 due to the format! macro expecting a named argument.
Anyone using the example should be able to figure out how to fix the code, but I figured that fixing it in the README
would help people out :)

### DIFF
--- a/actix-web/README.md
+++ b/actix-web/README.md
@@ -58,7 +58,7 @@ use actix_web::{get, web, App, HttpServer, Responder};
 
 #[get("/hello/{name}")]
 async fn greet(name: web::Path<String>) -> impl Responder {
-    format!("Hello {name}!")
+    format!("Hello {}!", name)
 }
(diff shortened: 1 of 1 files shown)


deviatorslegacy:
My apologies, I just found out that a new rust version (1.58.0) uses these, and I'm on 1.57.0. I'll close the pull request here, since the code works fine with an updated version :)

Issue #2175, opened 2021-04-19
Allow non-destructive access to a request/payload of a ServiceRequest


andy128k:
## Expected Behavior
Expose references (mutable references) to a request and a payload of a ServiceRequest.

## Current Behavior
`ServiceRequest` allows to get a request and a payload by invoking `into_parts` method. But this method consumes a `ServiceRequest`.

## Possible Solution
1. Add getters (mutable getters).
2. Make fields public.

## Context
A middleware may have a need to access a request (or a payload) by a borrowed reference to A `ServiceRequest`.
Related issues: https://github.com/actix/actix-web/pull/2168 https://github.com/actix/actix-web/pull/1821

ilyvion:
I really don't understand why it was decided to only add a `parts_mut` and not also a `parts`. I'm in a situation where I have a `&ServiceRequest` and I can't find any (non-UB) way of getting a hold of the inner `&HttpRequest`.

aliemjay:
> I really don't understand why it was decided to only add a `parts_mut` and not also a `parts`. I'm in a situation where I have a `&ServiceRequest` and I can't find any (non-UB) way of getting a hold of the inner `&HttpRequest`.

This quite reasonable. Please consider opening another issue that covers `ServiceResponse` as well.

PR #2786, opened 2022-06-17
Add getters for `&ServiceRequest`

junbl:
## PR Type

Feature

## PR Checklist

- [x] ~~Tests for the changes have been added / updated.~~*

*no tests necessary; none for `parts_mut/into_parts` or other getters

## Overview

Adds getters for cases when you only have a `&ServiceRequest` and need access to the inner parts.

 - `parts`: returns immutable refs to the inner request and payload. Non-owning counterpart to `into_parts` and `parts_mut`.
 - `request`: just gives access to the `HttpRequest` since there's not a lot you can do with a `&Payload` currently.

This allows access to parts of the `HttpRequest` not already covered by the existing getters. Additionally, this allows `FromRequest`ing types that don't require access to the `Payload`, like `Identity`. (`ServiceRequest::extract`, while covering a similar use case, requires a `&mut ServiceRequest` and thus isn't usable in this context.)

No breaking changes.

Use case mentioned in #2175, but it doesn't seem like the associated issue was ever opened.


robjtede (approved):
Thanks

PR #2788, opened 2022-06-19
Improve documentation for `actix-web-actors`

Nerixyz:
## PR Type

Documentation update

## PR Checklist

## Overview

The documentation for `actix-web-actors` didn't provide any examples on how to use the library. The examples are either on the actix.rs website or in the examples repository. While the examples are good, the documentation should still provide at least basic examples on how to use major structs/modules.

I've added:

* An example ([Websocket example on actix.rx](https://actix.rs/docs/websockets/)) and links (like for actix-web) to documentation and other resources.
* An example (same as on the crate-level) to the `ws` module.
* An example to the `WsResponseBuilder` struct (previously it only included a minimal example with no context).
* An example to `HttpContext`.

I'm not sure about the duplicate example in the `ws` module and on crate-level. I feel like there sould be an example in both locations, maybe even the example from `HttpContext` but I'd like to avoid duplicate examples.

### DIFF
--- a/actix-web-actors/CHANGES.md
+++ b/actix-web-actors/CHANGES.md
@@ -2,9 +2,6 @@
 
 ## Unreleased - 2022-xx-xx
 - Minimum supported Rust version (MSRV) is now 1.57 due to transitive `time` dependency.
-- Improve crate documentation. [#2788]
-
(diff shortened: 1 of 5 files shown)


robjtede (approved):
looks good, thanks

PR #2790, opened 2022-06-22
fix unrecoverable Err(Overflow) in websocket frame parser

oatoam:
## PR Type

Bug Fix

## PR Checklist

- [ ] ~~Documentation comments have been added / updated.~~

### Current
When received a frame with length larger than `max_size`, the websocket frame parser will goes into a `Err(ProtocolError::Overflow)` state, in which the buffer index cannot be advanced so the parser keep parsing the oversized frame again and again, result in endless `Err(ProtocolError::Overflow)`.

### new
Addvance the buffer index before throw a `Err(ProtocolError::Overflow)`, so the incoming frame data will be parsed properly.


actix-http/src/ws/frame.rs
```diff
@@ -339,6 +335,25 @@ mod tests {
...
+    fn test_parse_frame_max_size_recoverability() {
+        let mut buf =
+            BytesMut::from(&[0b0000_0001u8, 0b0000_0010u8, 0b0000_0000u8, 0b0000_0000u8][..]);
+        buf.extend(&[0b0000_0010u8, 0b0000_0010u8, 0xffu8, 0xffu8]);
+
+        assert_eq!(buf.len(), 8);
+        if let Err(ProtocolError::Overflow) = Parser::parse(&mut buf, false, 1) {
+        } else {
+            unreachable!("error");
+        }
```
robjtede: make this an `assert!(matches!(...))`

actix-http/src/ws/frame.rs
```diff
@@ -339,6 +335,25 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_parse_frame_max_size_recoverability() {
+        let mut buf =
+            BytesMut::from(&[0b0000_0001u8, 0b0000_0010u8, 0b0000_0000u8, 0b0000_0000u8][..]);
+        buf.extend(&[0b0000_0010u8, 0b0000_0010u8, 0xffu8, 0xffu8]);
```
robjtede: describe the set up

robjtede (approved):
Great. Thanks 👍

PR #2793, opened 2022-06-25
Fix common grammar mistakes and add small documentation for AppConfig's Default implementation

PeterPierinakos:
## PR Type

Grammar improvement and documentation.

## Overview

This PR fixes several grammar errors regarding the incorrect use of the "it's" instead of "its". Additionally, I have added some small documentation for the `AppConfig'`s `Default`. The reason I added the documentation is because I have wasted an unreasonable amount of time in the past trying to figure out why I couldn't access my application outside of my local network when using the special `127.0.0.1` loopback address, and I didn't want other people to have to go through that common troubleshooting too. Hope the note will help future readers.


robjtede (approved):
Thanks

PR #2817, opened 2022-07-23
fix partial ranges for video content

### LINKED ISSUE #2815: Streaming & Seeking in large video served using NamedFile doesn't work
## Expected Behavior
Serving a large *(in my case ~2GB)* static video file *(mp4, h264)* from actix on a given route should allow the html5 `video` player to start playback immediately, acting instantly on the arriving video chunks. The player should also be able to seek forward, which should cause a new get request with an offset range.

This is the behavior that can be observed when using nodejs + `express`, or `warp` to do the same thing.

## Current Behavior
For the large 2GB video file: The video is loading with no playback until the full file is downloaded. Only after the get request is completed, the video can start playing.

For smaller video files: The videos do start playing immediately (as would be expected), but seeking beyond the buffered range doesn't work. Instead of stopping the current buffering get request and starting a new one at the seeked range, it doesn't do anything when seeking beyond the buffer. The seek cursor just jumps back.

- Obtain a video file that is playable in browser (mp4, h264), for example [this random test video](http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/Sintel.mp4)
- If the video file is smaller than ~2GB (like the test video), either use another one or make it longer (for the example video: `ffmpeg -stream_loop 10 -i Sintel.mp4 -c copy Sintel_long.mp4`)
- Use the following `actix-web` rust program to serve the video (adjust the filename if needed)
```rust
use actix_files::NamedFile;
use actix_web::{get, App, HttpRequest, HttpResponse, HttpServer};

#[get("/video.mp4")]
pub async fn serve_video(req: HttpRequest) -> HttpResponse {
    let istream = NamedFile::open_async("Sintel_long.mp4").await.unwrap();
    istream.into_response(&req)
}

#[actix_web::main]
    HttpServer::new(|| App::new().service(serve_video))
        .bind(("127.0.0.1", 8080))?
        .run()
        .await
}
```

- Open the following html file in the browser while the `actix-web` server is running
```html
<!DOCTYPE html>
<html lang="en">
<head></head>
<body>
    <video controls preload="none">
        <source src="http://127.0.0.1:8080/video.mp4" type="video/mp4">
    </video>
</body>
</html>
```
- Press play in the browser (current behavior: doesn't play until the full get request is completed (2GB download))
- With a smaller video file: Try to seek forward beyond the buffered range (current behavior: current position cursor jumps back instead of seeking forward)

- Try the same thing with the following `nodejs` / `express` code (should work as described in *expected*)
```javascript
const express = require("express")
const app = express()
app.use("/video.mp4", express.static("Sintel_long.mp4"))
app.listen(8080)
```

- Try the same thing with the following `warp` code (should work as described in *expected*)
```rust
use warp::Filter;

#[tokio::main]
pub async fn main() {
    let route_file = warp::path("video.mp4").and(warp::fs::file("Sintel_long.mp4"));
    warp::serve(route_file).run(([127, 0, 0, 1], 8080)).await;
}
```

## Context
I'm trying to serve static video files using `actix-web` and play them using the html5 video player. This should have the *good user experience* of being able to play the video without waiting for multiple minutes and also seeking forward.

## Your Environment
- Rust Version: `rustc 1.62.0 (a8314ef7d 2022-06-27)`
- Actix Web Version: `4.1.0`
- Actix Files Version: `0.6.1`
- Target: `x86_64-pc-windows-msvc`
- Browser: `Firefox` & `Chromium`

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
The comment next to the code explains well.

Fixes #2815

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -1,7 +1,10 @@
 # Changes
 
 ## Unreleased - 2022-xx-xx
+- Allow partial range responses for video content to start streaming sooner. [#2817]
 - Minimum supported Rust version (MSRV) is now 1.57 due to transitive `time` dependency.
+
+[#2817]: https://github.com/actix/actix-web/pull/2817
 
 
 ## 0.6.1 - 2022-06-11
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -528,11 +528,26 @@ impl NamedFile {
                     length = ranges[0].length;
                     offset = ranges[0].start;
 
-                    // don't allow compression middleware to modify partial content
-                    res.insert_header((
-                        header::CONTENT_ENCODING,
-                        HeaderValue::from_static("identity"),
-                    ));
+                    // When a Content-Encoding header is present in a 206 partial content response
+                    // for video content, it prevents browser video players from starting playback
+                    // before loading the whole video and also prevents seeking.
+                    //
+                    // See: https://github.com/actix/actix-web/issues/2815
+                    //
+                    // The assumption of this fix is that the video player knows to not send an
+                    // Accept-Encoding header for this request and that downstream middleware will
+                    // not attempt compression for requests without it.
+                    //
+                    // TODO: Solve question around what to do if self.encoding is set and partial
+                    // range is requested. Reject request? Ignoring self.encoding seems wrong, too.
+                    // In practice, it should not come up.
+                    if req.headers().contains_key(&header::ACCEPT_ENCODING) {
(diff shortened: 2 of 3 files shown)


robjtede:
> Note that the coding named "identity" is reserved for its special role in [Accept-Encoding](https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding) and thus SHOULD NOT be included.

- <https://www.rfc-editor.org/rfc/rfc9110#section-8.4>

Further work required it seems.

PR #2823, opened 2022-07-26
fix pending stream drop

### LINKED ISSUE #1313: A Stream object in a pending state won't be dropped after the client disconnect

simorex80:
## PR Type

But Fix

## PR Checklist

## Overview

Fix the problem described in bug:
A Stream object in a pending state won't be dropped after the client disconnect #1313
When the disconnection events is triggered, no state check is needed.
The streaming process, due no access to TcpStream, don't have the necessary elements to change the state and this caused the stuck described in the bug.

Closes #1313


simorex80:
Moved to https://github.com/actix/actix-web/pull/2830

PR #2833, opened 2022-08-10
Expose inner value of Path

sebastianv89:
## PR Type

Feature

## PR Checklist

## Overview
The inner value is already exposed via the `into_inner` function, so we may as well expose it directly to improve ergonomics. I have updated the doc example to illustrate this.

Note that this aligns the implementation with that of Query, Json, Header, and Form.


robjtede:
This will be included with the next breaking release. Actix-web-lab has a version with public T you can use in the meantime.

https://docs.rs/actix-web-lab/latest/actix_web_lab/extract/struct.Path.html

PR #2858, opened 2022-08-31
Add ConditionOption middleware

KoltesDigital:
## PR Type

Feature

## PR Checklist

## Overview

I'm interested in conditionally applying a middleware. But with `Condition`, I find strange that the middleware has to be created even if the condition is false.

One simple, quite idiomatic improvement would have been to use a lambda to lazily generate the middleware.

But there's a more idiomatic way to do, using `Option`!

As `Transform` is defined in another crate, a newtype is required, which I called `ConditionOption`. Sadly it has to be mentioned for `wrap` to work.

### DIFF
--- a/actix-web/src/middleware/condition_option.rs
+++ b/actix-web/src/middleware/condition_option.rs
@@ -1,0 +1,60 @@
+//! For middleware documentation, see [`ConditionOption`].
+
+use futures_core::future::LocalBoxFuture;
+use futures_util::future::FutureExt as _;
+
+use crate::{
+    body::EitherBody,
+    dev::{Service, ServiceResponse, Transform},
+    middleware::condition::ConditionMiddleware,
+};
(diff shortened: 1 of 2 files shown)


fakeshadow:
This should be added to actix-service where Transform can impl Option<T>

KoltesDigital:
That's what I thought at the beginning, then I saw that the latest documentation for Transform mentions impls for Arc and Rc but they've disappeared from the master branch, so I thought that its impls have to be removed.

In the other hand, that would move `ConditionMiddleware` as well. I don't master your architecture, I can't make such a bold move.

fakeshadow:
You can start by making an issue in [actix-net](https://github.com/actix/actix-net) repo. We want to get it right since we already know what shoud be done. Rather than just add more duplicated (purpose) code.

PR #2859, opened 2022-09-04
Update TechEmpower benchmark link to latest

YoniFei:
## PR Type
Other

## Overview
Just a README update


fakeshadow:
https://github.com/TechEmpower/FrameworkBenchmarks/issues/7475#issuecomment-1190855363

There is an issue in said benchmark that does not calculate the score correctly. To avoid confusion/misleading this PR should be accepted only after said issue resolved.

PR #2881, opened 2022-09-18
PathDeserializer: use `deserialize_str` for `deserialize_any`

### LINKED ISSUE #318: Path extractor not working for enums
It looks like the Path extractor doesn't currently work with enums. To illustrate, I’m getting

```
 WARN 2018-06-14T23:46:13Z: actix_web::pipeline: Error occured during request handling: unsupported type: 'any'
 INFO 2018-06-14T23:46:13Z: actix_web::middleware::logger: 127.0.0.1:63742 [14/Jun/2018:16:46:13 -0700] "GET /my_resource/52635231 HTTP/1.1" 404 0 "-" "curl/7.54.0” 0.000458
```

whenever I try to do the following:

```rust
.resource(“/my_resource/{thing}", |r| {
    r.method(Method::GET).with(handle_mypath)
})
```

```rust
#[derive(Clone, Debug, Eq, Hash, PartialEq, Deserialize, Serialize)]
#[serde(untagged)]
pub enum MyThing {
    String(String),
    Int(u32),
}

pub fn handle_mypath<A>((state, path): (State<Arc<A>>, Path<MyThing>)) -> HttpResponse {
    …
}
```

It appears to never reach the handler, despite compiling fine. As soon as I convert it to a bare type or a struct it seems to work fine. Interestingly, Query extractors seem to work fine with this.

nitn3lav:
## PR Type

probably bug fix?

## PR Checklist

## Overview

There are cases in which one might want to use types using [`deserialize_any`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_any) in a [`web::Path`](https://actix.rs/actix-web/actix_web/web/struct.Path.html). This is especially useful for `enum`s with fields or when the same type is also used for other things, e. g. persisting to a database.
Since URLs are usually represented as Strings, I believe trying to deserialize a String if `deserialize_any` is called to be a reasonable default.

### Examples

In this example it is unclear to the `Deserialize` `impl` whether to deserialize a str or u32 and the decision is therefore left to the Deserializer.

```rs
use actix_web::{get, web::Path, App, HttpResponse, HttpServer};
use serde::{de::Visitor, Deserialize};
use std::io;

#[derive(Clone, Debug, Eq, Hash, PartialEq)]
pub enum MyThing {
    String(String),
    Int(u32),
    Other,
}

impl<'de> Deserialize<'de> for MyThing {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        struct Vis;
        impl<'de> Visitor<'de> for Vis {
            type Value = MyThing;

            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                write!(formatter, "my thing")
            }

            fn visit_u32<E>(self, v: u32) -> Result<Self::Value, E>
            where
                E: serde::de::Error,
            {
                Ok(MyThing::Int(v))
            }

            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
            where
                E: serde::de::Error,
            {
                Ok(match v.parse() {
                    Ok(v) => MyThing::Int(v),
                    Err(e) => match v {
                        "other" => MyThing::Other,
                        _ => MyThing::String(format!("'{v}' is not an int because {e}")),
                    },
                })
            }
        }

        deserializer.deserialize_any(Vis)
    }
}

#[actix_web::main]
    HttpServer::new(move || App::new().service(root))
        .bind("127.0.0.1:8080")?
        .run()
        .await
}

#[get("/{my_thing}")]
async fn root(my_thing: Path<MyThing>) -> HttpResponse {
    HttpResponse::Ok().body(format!("it works: {my_thing:?}"))
}
```

In the current version of `actix-web` this fails because `deserialize_any` is not supported:
```console
# curl "http://localhost:8080/123"
unsupported type: any%
```

Because this PR uses `deserialize_str` for `deserialize_any`, this now works as expected:
```console
# curl "http://localhost:8080/123"
it works: Path(Int(123))%

# curl "http://localhost:8080/12334435543"
it works: Path(String("'12334435543' is not an int because number too large to fit in target type"))%

# curl "http://localhost:8080/123abc"
it works: Path(String("'123abc' is not an int because invalid digit found in string"))%

# curl "http://localhost:8080/other"
it works: Path(Other)%
```

Another example for a commonly used type like this is [`bson::uuid::Uuid`](https://docs.rs/bson/latest/bson/uuid/index.html). MongoDB stores UUIDs in a map containing the binary data along with a binary subtype. However deserializing Strings or raw binary data is supported, too. Therefore [`deserialize_any`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_any) is used to indicate that many formats are supported.

This PR makes it possible to use types like `MyThing` and [`bson::uuid::Uuid`](https://docs.rs/bson/latest/bson/uuid/index.html) in a [`web::Path`](https://actix.rs/actix-web/actix_web/web/struct.Path.html).

Closes #318


robjtede:
Marking as semver-major for now until I dig into this change properly.

In the meantime, I'd definitely like to see a test case added that would fail on master and pass on this PR.

nitn3lav:
This PR adds the option to deserialize types whose `Deserialize` implementation uses `deserialize_any` which previously failed with `unsupported type: any`. Every other deserialization that worked before will continue to work as it did before.
The error always arises when a route tries to deserialize a type that uses `deserialize_any` and is therefore independent of any HTTP request sent by a client. I believe it to be very unlikely that an application using `actix-web` relies on this error. I therefore think that this can be safely added in a non-breaking minor change.

actix-router/src/de.rs
```diff
@@ -713,6 +722,104 @@ mod tests {
...
+        let rdef = ResourceDef::new("/{key}/{value}");
+
+        let mut path = Path::new("/other/123");
+        rdef.capture_match_info(&mut path);
+        let de = PathDeserializer::new(&path);
+        let segment: (AnyEnumCustom, AnyEnumDerive) =
+            serde::Deserialize::deserialize(de).unwrap();
+        assert_eq!(segment.0, AnyEnumCustom::Other);
+        // Deserializes to `String` instead of `Int` because deserializing defaults to `str` and serde doesn't automatically implement parsing numbers from `&str`s
+        assert_eq!(segment.1, AnyEnumDerive::String("123".to_string()));
```
robjtede: This is my current hold-up. Ideally, if the `AnyEnumDerive` enum had `Int` variant before `String` then it should be able to deserialize to `Int` here. It's good that we can support it with a custom deserialize but I feel it's more distruptive to change this behavior later if it can be done at all.

robjtede:
There's another interesting case who's possibility gets shut off by this change:

```
#[derive(Debug, serde::Deserialize)]
struct PathParams {
    #[serde(flatten)]
    unknown: BTreeMap<String, String>,
}
```

If you use this in `Path<_>` you get `Error("unsupported type: any")`.

actix-router/src/de.rs
JohnTitor: Addressed in https://github.com/actix/actix-web/pull/2881/changes/9c453c07860233335d9a34bcf73119e8fff492e4.
We now does best-effort™ support for numeric values.

JohnTitor (approved):
Thank you for the PR!

PR #2883, opened 2022-09-20
actix-multipart: Feature: Add typed multipart form extractor

### LINKED ISSUE #2849: Support for multipart form extractors
If you have a look on crates.io there are a proliferation of crates for doing multipart form uploads with files in actix. 

Including:

- [actix-easy-multipart](https://github.com/jacob-pro/actix-easy-multipart) by me
- [actix-multipart-extract](https://github.com/JSH32/actix-multipart-extract) by @JSH32
- [actix-multipart-derive](https://github.com/robjtede/actix-multipart-derive) by @robjtede 
- [actix-web-multipart-file](https://github.com/KeenS/actix-web-multipart-file) by @KeenS
- [actix_extract_multipart](https://github.com/KosekiDev/ActixExtractMultipart) by @KosekiDev
- [awmp](https://github.com/kardeiz/awmp) by @kardeiz
- [actix-form-data](https://crates.io/crates/actix-form-data) by @asonix

I think each of these have some unique improvements and features that the others do not. It seems silly to have so many unofficial crates, and is not very friendly to people who just want to quickly get started with actix.

I am wondering if we should have an effort to consolidate all of these crates into one official one in the actix project?

jacob-pro:
## PR Type
Feature

## PR Checklist

## Overview

This introduces a new multipart form/data extractor for recieving a multipart upload into a struct. 

I believe this should be enough to cover all the various features of the existing similar crates as per the discussion here: https://github.com/actix/actix-web/issues/2849

An example form looks like:

```rust
#[derive(MultipartForm)]
struct Upload {
    description: Option<Text<String>>,
    timestamp: Text<i64>,
    #[multipart(rename="image_set[]")
    image_set: Vec<Tempfile>,
}

async fn route(form: MultipartForm<Upload>) -> impl Responder {
    ...
}
```

The key feature is that each field just needs to implement the `FieldReader` trait. This trait allows the user to provide their own abitarary async handler for processing a field, for example they may want to stream the data to S3.

I look forward to your feedback! @robjtede @asonix @JSH32 @e-rhodes

List of features:
- [x] Optional fields
- [x] Lists/Vec fields ([see RFC](https://www.rfc-editor.org/rfc/rfc7578#section-4.3))
- [x] Field renaming 
- [x] User customisable field handlers 
- [x] Global and field level data limits
- [x] Configurable action on duplicate fields
- [x] Allow denying unknown fields  
- [x] User customisable error messages
- [x] Stream fields to temporary files
- [x] Deserialize integers, floats, enums from plain text fields
- [x] Deserialize complex structs from JSON fields   

Update: If anyone wants to use these features right now I have backported them to my existing libary [actix-easy-multipart v3.0.0](https://github.com/jacob-pro/actix-easy-multipart)


JSH32:
Looks good! What is the purpose of using `TextField` instead of `String` directly? Can you show an example of how someone might implement their own field or make their own field compatible with `TextField`? (just for making it easy to reference later on)

jacob-pro:
@JSH32 yep that is a good question.

We want to allow reading into not just String itself, but also things like integers, floats, enums, i.e. `Text<T: DeserializeOwned>`. This leads to two problems:

#### Trait Conflicts

But we can't simultaneously implement `FieldReader` for any `T: DeserializeOwned`, whilst also allowing the user to use native `Vec` and `Option`, giving us two choices:

Choice 1 (what I have implemented)

```rust
#[derive(MultipartForm)]
struct Upload {
    numbers: Vec<Text<i64>>,
}
```

Choice 2 (we would have to use Vec and Option wrapper types)

```rust
#[derive(MultipartForm)]
struct Upload {
    numbers: VecWrapper<i64>,
}
```

This is because of conflicting trait implementations (although one day this might be solved by [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html))

```
impl<'t, T: DeserializeOwned> FieldReader<'t> for T 

error[E0119]: conflicting implementations of trait `form::FieldGroupReader<'_>` for type `std::option::Option<_>`
   --> actix-multipart\src\form\mod.rs:233:1
    |
158 | / impl<'t, T> FieldGroupReader<'t> for Option<T>
159 | | where
160 | |     T: FieldReader<'t>,
161 | | {
...   |
194 | |     }
195 | | }
    | |_- first implementation here
...
233 | / impl<'t, T> FieldGroupReader<'t> for T
234 | | where
235 | |     T: FieldReader<'t>,
236 | | {
...   |
272 | |     }
273 | | }
    | |_^ conflicting implementation for `std::option::Option<_>`

For more information about this error, try `rustc --explain E0119`.
```

(Technically we could implement directly for String itself (rather than generic T), but it seems pointless since we would still need `Text<T: DeserializeOwned>` to work also)

#### Deserialization Ambiguity

The multipart standard doesn't define any serialization format for the data within the fields themselves. So even though we may want to use `serde` to automatically deserialize the contents of a field, there is no correct answer to which serde backend to use.

Instead by using the `Text` type it allows the user to specically opt-in to using `serde_plain`. Alternatively they can use the `Json` type, and the text would be deserialized using `serde_json` instead.

For example:

```rust
#[derive(MultipartForm)]
struct DeserializationMethods {
    json: Json<HashMap<String, String>>,
    plain: Text<String>,
}

async fn send() {
        let mut form = multipart::Form::default();
        // We can send the exact same input, but it is up to the server to choose how to deserialize it
        form.add_text("json", "{\"key1\": \"value1\", \"key2\": \"value2\"}");
        form.add_text("plain", "{\"key1\": \"value1\", \"key2\": \"value2\"}");
        ...
}
```

#### Compatibility

> Can you show an example of how someone might implement their own field or make their own field compatible with TextField

I'm not 100% sure what you mean by this - but the idea is that `Text` works with any types that implement `Deserialize`  (provided that they are compatible with [serde_plain](https://docs.rs/serde_plain/latest/serde_plain/))

If you wanted to use a more complex type e.g. arbitrary structs & complex enums you could use `Json` instead.

If you wanted to implement your own field type, then you just need to `impl<'t> FieldReader<'t> for YourField` - have a look at `Bytes`, `Json`, `Text`, or `Tempfile` as examples

actix-multipart/src/form/mod.rs
```diff
@@ -0,0 +1,832 @@
...
+        // Exceeds the 20 byte overall limit
+        let mut form = multipart::Form::default();
+        form.add_text("field", "this string is 28 bytes long");
+        let response = send_form(&srv, form, "/file").await;
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+    }
+
+    #[derive(MultipartForm)]
+    struct TestFieldLevelLimits {
+        #[multipart_form(limit = "30B")]
```
asonix: are limits like this runtime-configurable?

jacob-pro: Field level limits are currently only configured at compile time by the macro.

There are also global limits that are configured via `MultipartFormConfig` in the same way that [JsonConfig](https://docs.rs/actix-web/latest/actix_web/web/struct.JsonConfig.html#method.limit) already works in actix-web.

Did you have a more advanced use-case for how these limits should work? I guess right now if you wanted truly dynamic runtime limits, you would have to write your own `FieldReader` implementation (which has access to the `HttpRequest` context), and that allows you to do whatever you want including ignoring the global limits if you wanted to.

asonix: i think my use-case is covered by the global limits

jacob-pro: cool 👍 
Yeah there is definitely some scope for more allowing more advanced configuration future - I was thinking about maybe changing the `Limits` struct into a generic trait - but wasn't planning on doing it for now until I have a better idea how people would actually use it.

actix-multipart/src/form/json.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// Sets whether or not the field must have a valid `Content-Type` header to be parsed.
+    pub fn content_type_required(mut self, content_type_required: bool) -> Self {
+        self.content_type_required = content_type_required;
+        self
+    }
+}
+
+impl Default for JsonFieldConfig {
+    fn default() -> Self {
+        DEFAULT_CONFIG.clone()
```
junbl: These `clone`s on the *Config Default impls seem unnecessary, is there a reason for them? (on `MultipartFormConfig`, `TempfileFieldConfig`, `TextFieldConfig`)

actix-multipart/src/form/json.rs
```diff
@@ -0,0 +1,183 @@
...
+impl ResponseError for JsonFieldError {
+    fn status_code(&self) -> StatusCode {
+        StatusCode::BAD_REQUEST
+    }
+}
+
+/// Configuration for the [`JsonField`] reader.
+#[derive(Clone)]
+pub struct JsonFieldConfig {
+    err_handler: Option<Arc<dyn Fn(JsonFieldError, &HttpRequest) -> Error + Send + Sync>>,
```
junbl: all the configs have a very similar (complex) type for the error handler-- what do you think about
```rust
type FormErrorHandler<T> = Option<Arc<dyn Fn(T, &HttpRequest) -> Error + Send + Sync>>;
```

actix-multipart/src/form/bytes.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::{Field, MultipartError};
+use actix_web::HttpRequest;
+use bytes::{Bytes, BytesMut};
+use futures_core::future::LocalBoxFuture;
+use futures_util::TryStreamExt;
+use mime::Mime;
+
+/// Read the field into memory.
+#[derive(Debug)]
+pub struct BytesField {
```
junbl: ik we talked about this earlier, but maybe we can come to a compromise lol--what do you think about removing the `Field` suffix on all these (`BytesField`, `JsonField`, etc)? there's some naming collisions, which is annoying on the library side for sure. maybe split these into a module `field` for clarity so users will have `field::Bytes` as an option instead, and the library can deal with reexporting?

just feels bad as a user to type the same thing over and over, and the purpose of the wrappers is pretty clear in a user's code when you see them in a `Multipart` struct, so I don't think specifying `Field` gives that much more information

junbl (commented):
mostly nitpicks form me. great work, I'm excited to use this! ❤️

actix-multipart-derive/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            fn handle_field<'t>(
+                req: &'t ::actix_web::HttpRequest,
+                field: ::actix_multipart::Field,
+                limits: &'t mut ::actix_multipart::form::Limits,
+                state: &'t mut ::actix_multipart::form::State,
+            ) -> ::futures_core::future::LocalBoxFuture<'t, Result<(), ::actix_multipart::MultipartError>> {
+                use ::actix_multipart::form::FieldGroupReader;
+                use ::futures_util::{TryFutureExt, FutureExt};
```
jacob-pro: This is a bug I need to fix - the macro implementation requires some dependencies from futures_util and futures_core - so if the end user's crate didn't declare them as dependencies it would currently fail to compile when using the macro.

Should I re-export these (maybe as doc-hidden?) from actix-multipart so that they may then be reliably used from within the macro?

junbl: reexporting seems best, that's what I've seen other libraries do in this case

it also doesn't look like you need `TryFutureExt` from what I can see, and the others (`LocalBoxFuture` and `boxed_local`) are basically one liners. if you don't anticipate using more deps from futures, it'd be trivial to reimplement

actix-multipart/src/form/bytes.rs
jacob-pro: Yeah that sounds like a good idea - and I'm definitely happy to reconsider the names of basically every type in this PR 😆 

One thing I am worried about is a repeat of this issue: https://github.com/jacob-pro/actix-easy-multipart/issues/2#issuecomment-1225384360

Which could easily happen since we would have both a multipart `JsonConfig` and a regular actix-web `JsonConfig` structs - which might get very confusing...

actix-multipart/src/form/json.rs
jacob-pro: I tried to replicate the same pattern from here for consistency: https://github.com/actix/actix-web/blob/894effb8569e53227e98480fd72becc04ea3131d/actix-web/src/types/json.rs#L281-L295

I think the idea is that if none is provided then the shared reference to `DEFAULT_CONFIG` is used, but when the user wants to create their own version the default is cloned, allowing it to then be modified:

```rust
.app_data(JsonFieldConfig::default().content_type_required(false))
```

actix-multipart/src/form/bytes.rs
junbl: there's definitely room for confusion, but I think it's mostly solved by more detailed documentation. plus the situation here is mitigated by the fact that the name conflicts are in different crates-- much easier to spot the difference between `actix_web::web::Json` and `actix_web_multipart::field::Json` (or whatever the path might be) than two types in `actix_web_multipart`

would also be nice to shorten the attribute macro `multipart_form` -> `multipart`-- matches the crate name, which is at least what I would expect.

actix-multipart/src/form/json.rs
junbl: that's interesting. the setup with the const separate from the `Default` impl makes sense to get a reference, but the clone shouldn't be necessary there either.

tbh I think that should be changed too in `actix_web`. it looks like `FormConfig` uses the same pattern, but without the clone: 

https://github.com/actix/actix-web/blob/bcdde1d4ea2f4bf6c01ccf8b5b6118f7c9a3b2bf/actix-web/src/types/form.rs#L257-L267

actix-multipart-derive/src/lib.rs
jacob-pro: Thanks, yep you are right, these are very simple to just write directly

actix-multipart/src/form/bytes.rs
jacob-pro: Makes sense - I have renamed them all as described

actix-multipart/src/form/json.rs
jacob-pro: Oh I see - yep you are completely right - I was thinking this was a reference to a static (because of the comment above)

actix-multipart/src/form/json.rs
```diff
@@ -11,35 +11,35 @@ use std::sync::Arc;
...
     fn read_field(req: &'t HttpRequest, field: Field, limits: &'t mut Limits) -> Self::Future {
         Box::pin(async move {
-            let config = JsonFieldConfig::from_req(req);
+            let config = JsonConfig::from_req(req);
             let field_name = field.name().to_owned();
 
             let valid_content_type = field.content_type().subtype() == mime::JSON
                 || field.content_type().suffix() == Some(mime::JSON);
 
-            if !valid_content_type && config.content_type_required {
```
junbl: probably a negligible performance improvement but this could be:
```rust
if config.validate_content_type {
    let valid_content_type = field.content_type().subtype() == mime::JSON
        || field.content_type().suffix() == Some(mime::JSON);
    if !valid_content_type {
```

actix-multipart/src/form/bytes.rs
jacob-pro: Regarding the macro name (and also extractor / trait name) - I was trying to be careful to keep the names separate from the existing low-level implementation.

People may still want to use the existing `Multipart` extractor which just provides access to stream itself, or they can use the higher-level `MultipartForm` extractor which allows automatic extraction of typed fields.

junbl: makes sense, I was just referring to the attribute macro `#[multipart_form]`, by analogy with `serde`-- i.e. using `#[serde]` as opposed to `#[deserialize]`, and so `#[multipart]` instead of `#[multipart_form]`. I don't think the lower level `Multipart` has a proc macro/attributes, but if it does it certainly makes sense to keep those separate.

jacob-pro: oh I see what you mean - yep that is a good idea - done 👍

actix-multipart/src/form/json.rs
jacob-pro: fixed

jacob-pro:
@robjtede when you have time, please can you review this PR - and let me know what you think?

actix-multipart/src/form/mod.rs
```diff
@@ -0,0 +1,832 @@
...
+            let mut field_limits = HashMap::<String, Option<usize>>::new();
+
+            while let Some(field) = payload.try_next().await? {
+                // Retrieve the limit for this field
+                let entry = field_limits
+                    .entry(field.name().to_owned())
+                    .or_insert_with(|| T::limit(field.name()));
+                limits.field_limit_remaining = entry.to_owned();
+
+                T::handle_field(&req, field, &mut limits, &mut state).await?;
```
asonix: I realize this might be difficult with the &mut limits and &mut state in here, but would there be any way to race `handle_field` against `payload.try_next()`

In my crate, actix-form-data, I have a FuturesUnordered that I push the field handler futures into. This allows field handlers to drop their Field stream upon completion and make the remaining work concurrent with the next field handlers

jacob-pro: hmmm... that is interesting I never even thought about that. I think to do that we would have to synchronise on the limits & state map, which is do-able.

I do wonder though if it actually brings any noticable performance improvement - since the multipart as a whole is being received as a single stream, presumably the next part can't be received until all of the previous part has been buffered - I guess it would depend on how much buffering actix-multipart does internally?

asonix: if a handler does something like
```rust
let bytes = Vec::new();
while let Some(res) = field.next().await {
    bytes.extend_from_slice(&res?);
}
drop(field);
let result =  send_these_bytes_somewhere(bytes).await;
```
then the `send_these_bytes_somewhere` could be run concurrently with the next field handler, since the field was completely read and dropped before that is executed.
 
I use this in my application pict-rs to better allow concurrently uploading files to object storage: https://git.asonix.dog/asonix/pict-rs/src/branch/main/src/store/object_store.rs#L246
This allows making additional requests to object storage without waiting for the existing in-flight requests to complete

jacob-pro: oh I see - you're effectively adding your own internal buffer to the handler

jacob-pro: does this mean your server memory usage is only bounded by the overall form limit, e.g. if `send_these_bytes_somewhere` / your object storage backend is running quite slowly, whilst the multipart upload is very fast, then nearly all of the contents could be be read into memory before getting uploaded?

asonix: Yes, that is true. actix-form-data has per-field size limits and total field count limits, so the maximum possible memory usage is `max_field_size * max_field_count`, and in theory that limit could be hit by slow object storage.

asonix: I've done some thinking on this and I think I can get away with using `spawn` to make this concurrent without changing the design here.

robjtede:
Sorry for the mega delay; finally getting around to reviewing this. Excited by what I've gone through so far :)

jacob-pro:
Thanks for looking at this @robjtede - I'm glad you figured out some improvements to the trait name 😆 !

vbocan:
What is the status of this PR? This is excellent work and I'd really love to have an "official" way to handle forms that have text fields as well as binary files. I am currently evaluating [actix-easy-multipart v3.0.0](https://github.com/jacob-pro/actix-easy-multipart) but I'd rather use actix-multipart with the same features. Thanks @jacob-pro for the hard work!

actix-multipart/src/error.rs
```diff
@@ -46,12 +49,41 @@ pub enum MultipartError {
...
+
+    /// Unknown field
+    #[display(fmt = "Unsupported field `{}`", _0)]
+    #[from(ignore)]
+    UnsupportedField(#[error(not(source))] String),
 }
 
 /// Return `BadRequest` for `MultipartError`
 impl ResponseError for MultipartError {
     fn status_code(&self) -> StatusCode {
```
jacob-pro, line 82: @robjtede you may also want to add an override to the `error_response()` function as well

DuckyBlender:
Any status update?

robjtede:
Thanks for the nudge @DuckyBlender.

Added some trybuild tests. Once CI passes we'll get this merged and released 🎉

robjtede (approved):

PR #2885, opened 2022-09-21
Fix actix-multipart field content_type() to return an Option

### LINKED ISSUE #2880: actix_multipart: Incorrect field content type defaults
## Current Behavior

The [Field::content_type()](https://docs.rs/actix-multipart/latest/actix_multipart/struct.Field.html#method.content_type) method returns a default of application/octet-stream in the case that it is not present

https://github.com/actix/actix-web/blob/bd5c0af0a6bf9a720cf0685b3334151de1d43c32/actix-multipart/src/server.rs#L364-L368

## Expected Behavior

According to the [multipart specification](https://www.rfc-editor.org/rfc/rfc7578#section-4.4):

>    Each part MAY have an (optional) "Content-Type" header field, which
   defaults to "text/plain".  If the contents of a file are to be sent,
   the file data SHOULD be labeled with an appropriate media type, if
   known, or "application/octet-stream".

So in the case of a text part, then we should actually get a default of "text/plain" and not "application/octet-stream".

## Possible Solution

Unfortunately there is no simple way to provide a default, since from the server's perspective it is impossible to tell if the field is a "file" or "text part", this is because the filename parameter is optional even for files:

> For form data that represents the content of a file, a name for the
   file SHOULD be supplied as well, by using a "filename" parameter of
   the Content-Disposition header field.  The file name isn't mandatory
   for cases where the file name isn't available or is meaningless or
   private; this might result, for example, when selection or drag-and-
   drop is used or when the form data content is streamed directly from
   a device.

Therefore if we want to provide reliable information to library users, the best thing we can do is make `Field::content_type()` return `Option<&Mime>` instead, to reflect the true content type that was set by the uploader.

jacob-pro:
## PR Type
Bug Fix

## PR Checklist

## Overview

The [Field::content_type()](https://docs.rs/actix-multipart/latest/actix_multipart/struct.Field.html#method.content_type) is changed to return Option rather than implicitly using a default value.

This is necessary because the multipart spec does not define a universal default content type - the default content type depends on whether the field is "text" or a "file" (which is impossible to tell).

Closes https://github.com/actix/actix-web/issues/2880

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased - 2022-xx-xx
 - Minimum supported Rust version (MSRV) is now 1.59 due to transitive `time` dependency.
+- `Field::content_type()` now returns `Option<&mime::Mime>` [#2880]
 
 
 ## 0.4.0 - 2022-02-25
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -4,6 +4,8 @@
 - Minimum supported Rust version (MSRV) is now 1.59 due to transitive `time` dependency.
 - `Field::content_type()` now returns `Option<&mime::Mime>` [#2880]
 
+[#2880]: https://github.com/actix/actix-web/pull/2880
+
 
 ## 0.4.0 - 2022-02-25
 - No significant changes since `0.4.0-beta.13`.
--- a/actix-multipart/src/server.rs
+++ b/actix-multipart/src/server.rs
@@ -361,17 +361,18 @@ impl InnerMultipart {
                 return Poll::Ready(Some(Err(MultipartError::NoContentDisposition)));
             };
 
-            let ct: mime::Mime = headers
+            let ct: Option<mime::Mime> = headers
                 .get(&header::CONTENT_TYPE)
                 .and_then(|ct| ct.to_str().ok())
-                .and_then(|ct| ct.parse().ok())
-                .unwrap_or(mime::APPLICATION_OCTET_STREAM);
+                .and_then(|ct| ct.parse().ok());
 
             self.state = InnerState::Boundary;
 
             // nested multipart stream is not supported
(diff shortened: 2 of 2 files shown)


jacob-pro:
@robjtede is this good to merge?

robjtede:
Almost. Can you add some more documentation on the method about the reasons for it being optional, linking to the spec if appropriate.

jacob-pro:
Sure. 

However having re-read the spec, there is the question instead of whether we should just change the default to `text/plain` instead?

> Each part MAY have an (optional) "Content-Type" header field, which
defaults to "text/plain". If the contents of a file are to be sent,
the file data SHOULD be labeled with an appropriate media type, if
known, or "application/octet-stream".

All I am certain of is that the current default "application/octet-stream" is incorrect...

What do you think @robjtede ? Either one is a breaking change - it is just a question of which would have the biggest impact

robjtede:
I think the `-> Option<&Mime>` route is better (with expanded docs) and other crates seem to agree:
- [`multer`](https://docs.rs/multer/latest/multer/struct.Field.html#method.content_type) : optional
- [`multipart`](https://docs.rs/multipart/latest/multipart/server/struct.FieldHeaders.html#structfield.content_type) : optional

jacob-pro:
👍 I have updated the comment on `content_type()` @robjtede

robjtede (approved):
I may tweak the wording a bit but this is good, thanks 👍🏻

jacob-pro:
thanks @robjtede :)

jacob-pro:
any idea on when this will be released?

robjtede:
Thanks for the bump. Since it's breaking I was gonna try and combine it with changes from #2883.

jacob-pro:
thanks @robjtede  - do you know if yourself or any of the other actix maintainers are able to review #2883 soon?

jacob-pro:
Otherwise if #2883 is unlikely to be merged soon, then I would appreciate having this fix released on its own - so that I can update my own multipart library ?

PR #2903, opened 2022-10-07
fix xhtml pages' `content-disposition`

kkocdko:
## PR Type

Bug Fix

## PR Checklist

- [x] ~(Is this needed?) Tests for the changes have been added / updated.~
- [x] ~(Is this needed?) Documentation comments have been added / updated.~

## Overview

Currently a `index.xhtml` will starts download dialog in browser due to the `content-disposition: attachment`. This PR change it to `inline`.

It's a tiny fix, but still a breaking change? Do we promised this behaviour (the crate `mime_guess`)?

### DIFF
--- a/actix-files/src/named.rs
+++ b/actix-files/src/named.rs
@@ -133,6 +133,7 @@ impl NamedFile {
                 mime::APPLICATION => match ct.subtype() {
                     mime::JAVASCRIPT | mime::JSON => DispositionType::Inline,
                     name if name == "wasm" => DispositionType::Inline,
+                    name if name == "xhtml" => DispositionType::Inline,
                     _ => DispositionType::Attachment,
                 },
                 _ => DispositionType::Attachment,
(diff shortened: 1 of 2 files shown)


actix-files/src/named.rs
```diff
@@ -133,6 +133,7 @@ impl NamedFile {
                 mime::APPLICATION => match ct.subtype() {
                     mime::JAVASCRIPT | mime::JSON => DispositionType::Inline,
                     name if name == "wasm" => DispositionType::Inline,
+                    name if name == "xhtml" => DispositionType::Inline,
```
JohnTitor: We could shorten like:

```suggestion
                    name if name == "wasm" || name == "xhtml" => DispositionType::Inline,
```

JohnTitor:
> It's a tiny fix, but still a breaking change? Do we promised this behaviour (the crate `mime_guess`)?

It's indeed a behavioral change, but for correctness. Therefore I'd call it a fix, not a breaking change.

robjtede (approved):
thanks 👍🏻

PR #2948, opened 2022-12-15
add setup to compress middleware

### LINKED ISSUE #2928: Allow setting compression levels
## Expected Behavior

Although the default compression levels are sane, some users might want to adjust it to better fit their needs. Being able to set the compression levels gives users more control and flexibility.

## Current Behavior

```rust
.wrap(Compress::default())
```

That's super easy to add, but there are no ways to fine-tune the compression algos.

## Possible Solution

I like the defaults, although the available compression algos aren't that transparent, in a sense that it's tricky to figure out which ones are going to be supported from looking at the init code alone. I think that keeping it that way is fine, but it would be nice to have an alternative way to configure compression, here is a pseudo-code:

```rust
.wrap(Compress::setup(
    vec![
        Gzip::Level_5,
        Brotli::Level_7,
    ]
))
```

I'm relatively new to Rust, so take it with a grain of salt, I'm sure there are some more convenient/idiomatic patterns which can be used to setup the compression levels.

## Context

I'm fine with the current defaults but I have some spare CPU capacity so it wouldn't hurt to keep it busy. Also, I'm planning to add Tor support to one of the APIs I'm working on, so optimizing for bandwidth would certainly help to speed things up a bit.

topenkoff:
## PR Type

Feature

## PR Checklist

## Overview

Add level setup for `middleware::Compress`

Closes #2928


robjtede:
It seems to me that setting one level to apply to all compression formats is not a good enough solution. We'd need a compression level setting per encoding.

robjtede:
i still think the idea of #2928 is sensible, but my previous comment stands

therefore, closing as stale

if you or someone else is interested in this work, please propose an API in the issue before spending time on a PR 🙏🏻

PR #2949, opened 2022-12-21
add Allow header to resource's default 405 handler

### LINKED ISSUE #2359: `allow` response header missing for `405 Method Not Allowed` generated by router
## Expected Behavior

I am expecting to see an `allow` header for `405 Method Not Allowed` responses generated by the router.

As per [RFC7231#6.5.5](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5):

> […] The origin server MUST generate an
> Allow header field in a 405 response containing a list of the target
> resource's currently supported methods.

## Current Behavior

There is no `allow` header to be seen.

## Possible Solution

When the router is sending a 405 in response to a request to an existing route, but with an unknown HTTP method it should add a `allow` header with the known methods.

main.rs
```rust
use actix_web::{web, App, HttpServer, Responder};

async fn index() -> impl Responder {
    "Index"
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(web::resource("/").route(web::get().to(index)))
    })
    .bind("[::]:8080")?
    .run()
    .await
}
```

Cargo.toml
```toml
[package]
name = "actix_test"
version = "0.1.0"
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
actix-web = "3.3.2"
```

Allowed method:
```http
$ http GET 'localhost:8080/'
HTTP/1.1 200 OK
content-length: 5
content-type: text/plain; charset=utf-8
date: Mon, 09 Aug 2021 19:15:05 GMT

Index
```

Disallowed method:
```http
$ http PUT 'localhost:8080/'
HTTP/1.1 405 Method Not Allowed
content-length: 0
date: Mon, 09 Aug 2021 19:15:00 GMT
```

## Context

This bug report was created purely from a standard correctness perspective with actix violating a MUST.

This was previously reported in #885, but I believe the remark about RFC correctness was missed in the issue.

## Your Environment

* Actix Web Version: 3.3.2

robjtede:
## PR Type

Fix

## PR Checklist

## Overview
[Spec](https://httpwg.org/specs/rfc9110.html#rfc.section.15.5.6) says MUST so...

closes #2359

### DIFF
--- a/actix-web/src/guard/mod.rs
+++ b/actix-web/src/guard/mod.rs
@@ -298,7 +298,6 @@ impl Guard for MethodGuard {
         let registered = ctx.req_data_mut().remove::<RegisteredMethods>();
 
         if let Some(mut methods) = registered {
-            eprintln!("here one");
             methods.0.push(self.0.clone());
             ctx.req_data_mut().insert(methods);
         } else {
--- a/actix-web/src/guard/mod.rs
+++ b/actix-web/src/guard/mod.rs
@@ -286,11 +286,26 @@ pub fn Method(method: HttpMethod) -> impl Guard {
     MethodGuard(method)
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct RegisteredMethods(pub(crate) Vec<HttpMethod>);
+
 /// HTTP method guard.
-struct MethodGuard(HttpMethod);
+#[derive(Debug)]
+pub(crate) struct MethodGuard(HttpMethod);
 
 impl Guard for MethodGuard {
     fn check(&self, ctx: &GuardContext<'_>) -> bool {
+        let registered = ctx.req_data_mut().remove::<RegisteredMethods>();
+
+        if let Some(mut methods) = registered {
+            eprintln!("here one");
+            methods.0.push(self.0.clone());
+            ctx.req_data_mut().insert(methods);
+        } else {
+            ctx.req_data_mut()
+                .insert(RegisteredMethods(vec![self.0.clone()]));
+        }
+
         ctx.head().method == self.0
     }
 }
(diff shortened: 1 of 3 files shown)


actix-web/src/guard/mod.rs
```diff
@@ -286,11 +286,26 @@ pub fn Method(method: HttpMethod) -> impl Guard {
...
-struct MethodGuard(HttpMethod);
+#[derive(Debug)]
+pub(crate) struct MethodGuard(HttpMethod);
 
 impl Guard for MethodGuard {
     fn check(&self, ctx: &GuardContext<'_>) -> bool {
+        let registered = ctx.req_data_mut().remove::<RegisteredMethods>();
+
+        if let Some(mut methods) = registered {
+            eprintln!("here one");
```
JohnTitor: Looks like a left-over on debugging?

actix-web/src/resource.rs
```diff
@@ -40,8 +41,8 @@ use crate::{
 ///         .route(web::get().to(|| HttpResponse::Ok())));
 /// ```
 ///
-/// If no matching route could be found, *405* response code get returned. Default behavior could be
-/// overridden with `default_resource()` method.
+/// If no matching route is found, a 405 response is returned with an appropriate Allow header.
```
JohnTitor: This is my favor so feel free to ignore, but it'd be helpful to add the spec link just in case someone notices this change and asks why?

JohnTitor (approved):
> [Spec](https://httpwg.org/specs/rfc9110.html#rfc.section.15.5.6) says MUST so...

TIL, thank you for the fix!

PR #2953, opened 2022-12-29
Hide authorization header in httprequest debug output

ctrlaltf24:
## PR Type
Bug fix

## PR Checklist

## Overview
Hides Authorization header from HTTPRequest Debug output. In the case of HTTP Basic Authentication, this header would have the username and password of the authenticated user. Preventing this header from showing up in debug would prevent this password leakage into the logs. In other schemes the value may not be as sensitive, but still in most cases would allow impersonation of the user in some form.

Whether or not you want this conditional or not is of course up to you. Willing to consider other locations for this check if you want a more general approach to prevent all debug outputs of sensitive headers. 

Thanks for a great library, keep it up!


JohnTitor (approved):
Makes sense to me!

robjtede:
I, personally, don't think this should be the case. What if one wants to debug that header, for example?

JohnTitor:
So, should we guard it with a log level?

LukeMathWalker:
I think the approach is sensible, but I'd suggest printing "<header name>: ***" or `redacted` rather than behaving as if the header wasn't there.

LukeMathWalker:
I think the approach is sensible, but I'd suggest printing `<header name>: ***` or `redacted` rather than behaving as if the header wasn't there.

If they want to debug that header, they can always print it on its own. It's still possible if that's what you _want_ to do, but it makes it more difficult to do it accidentally.
The issue with this approach is that Debug isn't configurable so it can't accomodate other user-defined sensitive headers.

robjtede:
In terms of "what counts as a sensitive header", we have some precedent already: https://github.com/actix/actix-web/blob/master/awc/src/middleware/redirect.rs#L294. I think if we go ahead with this change, all these should be included.

ctrlaltf24:
> In terms of "what counts as a sensitive header", we have some precedent already: https://github.com/actix/actix-web/blob/master/awc/src/middleware/redirect.rs#L294. I think if we go ahead with this change, all these should be included.

Fixed

robjtede:
I've come around to this change; can't hurt.

robjtede (approved):

Issue #2893, opened 2022-09-26
Allow custom HTTP methods in route macro


SeriousBug:
Your issue may already be reported!
Please search on the [Actix Web issue tracker](https://github.com/actix/actix-web/issues) before creating one.

## Expected Behavior

You should be able to use custom HTTP methods in the `route` macro:

```rs
#[route("/my/path", method = "CREATE")]
async fn create() -> HttpResponse {
  // ...
```

This is already possible with the `web::method` function, like this:

```rs
App::new()
  .route("/", web::method(Method::from_str("CREATE").unwrap()).to(|| async { "Hello CREATE" }))
```

## Current Behavior

The code above results in the error "Unexpected HTTP method: \`CREATE\`."

## Possible Solution

The parser in `actix-web-codegen` should be made more permissive so it will allow custom methods. The parsing code is already in place in `actix-web`, so that could likely be reused.

## Context

Trying to extend my server with additional HTTP methods. Not a dealbreaker and already possible, just a little more verbosely.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): rustc 1.64.0 (a55dd71d5 2022-09-19)
- Actix Web Version: 4.2.1

andrewpollack:
Looking into this now -- is there a way I can assign myself to this issue?

andrewpollack:
My desired code change spot is going to be at the `try_from` logic at line 154 in the `MethodType::try_from` portion https://github.com/actix/actix-web/blob/6f0a6bd1bb7ffdd98fa5ce825b24a73c4d71d9a7/actix-web-codegen/src/route.rs#L153-L157

One hangup I have is that by allowing for essentially arbitrary strings to define a Method, we nearly avoid any logic that defines the `method_type!` enum at https://github.com/actix/actix-web/blob/6f0a6bd1bb7ffdd98fa5ce825b24a73c4d71d9a7/actix-web-codegen/src/route.rs#L44-L54

How should I go about this? Should the `method_type!` logic be re-configured to just deal in arbitrary strings, or should I look to change behavior elsewhere?

alexandrebouthinon:
Hi @andrewpollack, are you still working on it or can I try to add this feature too?

andrewpollack:
I'm not sure I'll get a response on my request for feedback above, so I'm probably not the best owner. 

Happy to pass this to another interested party

robjtede:
[...]

PR #2963, opened 2023-01-11
Add support for all the IANA registered HTTP methods

alexandrebouthinon:
## PR Type

Feature

## PR Checklist

## Overview

This PR add support for all the IANA RFCs described HTTP verbs except `MOVE` and `*` because they're rust keywords (if you know a trick to get around the problem and thus complete the list, I am interested 🙂).
See the complete list on the [IANA website](https://www.iana.org/assignments/http-methods/http-methods.xhtml#methods)


ibraheemdev:
Can we not just support any string literal? I don't see the point of restricting it if we're already accepting non-standard methods.. other than not catching typos.

robjtede:
related to #2893

alexandrebouthinon:
> Can we not just support any string literal? I don't see the point of restricting it if we're already accepting non-standard methods.. other than not catching typos.

If you want I can close this one and make a new one to manage a custom string. But I don't know if Andrew is still working on it

alexandrebouthinon:
I've been thinking about this and my opinion is that the request of custom literals support for HTTP methods in a framework is almost always motivated by the need to use one or more of the additional methods described by IANA. 

  Moreover, I have looked a bit and I think that adding this support will increase drastically the complexity and the difficulty of maintaining the project (especially without wanting to release a new major version and therefore making design concessions during implementation) 😅 

What do you think?

JohnTitor:
Because we use the http crate for HTTP methods in some places, I think it makes some sense to keep consistency with [their methods list](https://docs.rs/http/latest/http/method/struct.Method.html) to simplify handling, and accepting other methods as user-defined sounds better to me.
Also, modifying only `method_type!` inputs adds partial support, not full support, right? Then, uhm, I guess it'd also add a certain complexity.

alexandrebouthinon:
> Because we use the http crate for HTTP methods in some places, I think it makes some sense to keep consistency with [their methods list](https://docs.rs/http/latest/http/method/struct.Method.html) to simplify handling, and accepting other methods as user-defined sounds better to me. Also, modifying only `method_type!` inputs adds partial support, not full support, right? Then, uhm, I guess it'd also add a certain complexity.

That's right, so I'm going to close this PR and work on the string literals support. Thank you for taking time to follow up my questions 🙂

PR #2979, opened 2023-02-13
add methods for mapping error responses

robjtede:
## PR Type

Feature

## PR Checklist

## Overview

Give ability for users, middleware in particular, to alter the response generated from `Error`s.


Samielakkad (commented):
The idea of response mappers is interesting, but the public API as written has an extra generic type parameter that callers cannot infer:

```rust
pub fn add_mapper<F, B>(&mut self, mapper: F)
where
    F: Fn(HttpResponse) -> HttpResponse + 'static,
```

`B` is not used in the argument list, return type, or bounds, so a normal call like:

```rust
err.add_mapper(|res| res)
```

will require the caller to specify a meaningless generic parameter or fail type inference. This should probably just be `pub fn add_mapper<F>(...)` unless `B` was intended to be part of the mapper signature.

It would also be good to add a small compile/pass test or unit test that actually calls `add_mapper` through the public API, so this kind of inference issue is caught.

robjtede:
@JohnTitor basically an attempt to work around https://github.com/actix/actix-extras/issues/680 in a nonbreaking way

robjtede:
@JohnTitor while it feels a bit hacky, i do think this is a reasonable enough workaround to a real problem while staying non-breaking, would appreciate you taking a look at it when you've got some time please 🙏🏻

PR #3000, opened 2023-03-12
add `body::to_bytes_limited` (#3000

robjtede:
## PR Type

Feat

## PR Checklist

## Overview

Provides easy way to read a response body in a middleware up to a certain size to protect against accidental memory exhaustion.

Open questions:
- Good function name?
- Return type should be nested Result or generic error type?

### DIFF
--- a/actix-http/src/body/utils.rs
+++ b/actix-http/src/body/utils.rs
@@ -11,7 +11,7 @@ use super::{BodySize, MessageBody};
 ///
 /// Any errors produced by the body stream are returned immediately.
 ///
-/// Consider using [`to_bytes_limit`] instead to protect against memory exhaustion.
+/// Consider using [`to_bytes_limited`] instead to protect against memory exhaustion.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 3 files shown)


actix-http/src/body/utils.rs
```diff
@@ -7,71 +7,184 @@ use futures_core::ready;
...
+///
+/// let body = Bytes::from_static(b"123");
+/// let bytes = to_bytes_limit(body, 10).await.unwrap().unwrap();
+/// assert_eq!(bytes, "123");
+///
+/// let body = Bytes::from_static(b"123");
+/// assert!(to_bytes_limit(body, 2).await.is_err());
+/// # });
+/// ```
+pub async fn to_bytes_limit<B: MessageBody>(
```
LukeMathWalker: `to_bytes_with_limit`? Pure bikeshedding.

actix-http/src/body/utils.rs
```diff
@@ -7,71 +7,184 @@ use futures_core::ready;
 
 use super::{BodySize, MessageBody};
 
-/// Collects the body produced by a `MessageBody` implementation into `Bytes`.
+/// Collects all the bytes produced by `body`.
 ///
 /// Any errors produced by the body stream are returned immediately.
 ///
+/// Consider using [`to_bytes_limit`] instead to protect against memory exhaustion.
```
LukeMathWalker: I wonder if marking it as `deprecated` would be a stronger/better way to deliver the message, especially to those who are _today_ using `to_bytes` and exposed to memory exhaustion.

robjtede: Hmm, maybe in a later release we could deprecate it?

actix-http/src/body/utils.rs
robjtede: `to_bytes_limited` ?

robjtede: `up_to_bytes` 😂  ?

popzxc (approved):

PR #3015, opened 2023-04-14
Improvements on streaming()

### LINKED ISSUE #2306: [awc] HTTP client ignores Content-Length if Body is stream.
## Expected Behavior

HTTP client should respect Content-Length even body of request is stream.

## Current Behavior

Content-Length header is removed by client itself.

## Possible Solution

respect Content-Length header if present.
also an enhancement, if first item of body stream is known as `None` it is better set Content-Length to zero. not inserting chunked encoding header.

Send request with empty stream. 

## Context

Cloudfront will reject you.

## Your Environment

* Actix Web Version: awc 3.0.0-beta.7

dynos01:
## PR Type

Bug Fix

## PR Checklist

## Overview

This PR changes the behavior of `actix_web::response::builder::HttpResponseBuilder::streaming()`. 

Before this PR, if `Content-Type` is not set by the user, then there is no such header in the final response. This PR adds a default `Content-Type` of `application/octet-stream`, just like `actix_web::response::builder::HttpResponseBuilder::json()` defaults the `Content-Type` to `application/json`.

Before this PR, if the user does not call `actix_web::response::builder::HttpResponseBuilder::no_chunking()` but sets `Content-Length`, then the header is silently dropped, as described in #2306. This PR respects the user's input: if the user sets `Content-Length`, then `actix_web::response::builder::HttpResponseBuilder::no_chunking()` is automatically called with the corresponding stream length.

A small illustration program for the change: 
```rust
use std::{
    pin::Pin,
    task::{Context, Poll},
};

use actix_web::{get, web, App, HttpRequest, HttpResponse, HttpServer, Responder, http::header};
use futures::stream::Stream;

struct ZeroStream {
    n: u64,
}

impl Stream for ZeroStream {
    type Item = Result<web::Bytes, actix_web::Error>;

    fn poll_next(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        let resp = vec![0u8; self.n as usize];
        Poll::Ready(Some(Ok(web::Bytes::copy_from_slice(&resp))))
    }
}

#[get("/{n}")]
async fn zero(req: HttpRequest) -> impl Responder {
    let n: &str = match req.match_info().get("n") {
        Some(n) => n,
        None => return HttpResponse::BadRequest().finish(),
    };
    let n: u64 = match n.parse() {
        Ok(n) => n,
        Err(_) => return HttpResponse::BadRequest().finish(),
    };

    let stream = ZeroStream {
        n,
    };

    HttpResponse::Ok()
        .append_header((header::CONTENT_LENGTH, format!("{n}")))
        .streaming(stream)
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(zero)
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```
Request:
```
GET /100 HTTP/1.1
```
The response is now:
```
< HTTP/1.1 200 OK
< content-length: 100
< content-type: application/octet-stream
< date: Fri, 14 Apr 2023 15:49:33 GMT
```
instead of:
```
< HTTP/1.1 200 OK
< transfer-encoding: chunked
< date: Fri, 14 Apr 2023 16:19:26 GMT
// Downloading endlessly
```

Closes #2306


dinmukhamedm:
@robjtede is this planned for any of the future releases? I think I am still getting this with `actix-web 4.11.0`.

Here's a snippet of how I return from a route registered with a `get` macro

```rust
let get_response: std::pin::Pin<Box<dyn Stream<Item = web::Bytes> + Send>> = storage
    .get_stream(&parquet_path)
    .await?;

let filename = parquet_path.split('/').last().unwrap_or(&name);

let mut response = HttpResponse::Ok();

response
    .content_type("application/octet-stream")
    .insert_header((
        "Content-Disposition",
        format!("attachment; filename=\"{}\"", filename),
    ))
    .insert_header(("Cache-Control", "no-cache"))
    // content_length is a non-zero u64
    .insert_header(("Content-Length", content_length.to_string()));

Ok(response.streaming(get_response.map(|chunk| Ok::<_, anyhow::Error>(chunk))))
```

And here's what I get in the output of `wget`

```
---response begin---
HTTP/1.1 200 OK
transfer-encoding: chunked
content-disposition: attachment; filename="0002.parquet"
content-type: application/octet-stream
cache-control: no-cache
date: Thu, 07 Aug 2025 09:40:29 GMT

---response end---
200 OK
Registered socket 6 for persistent reuse.
Length: unspecified [application/octet-stream]
Saving to: ‘0002.parquet’
```

JohnTitor (approved):
Makes sense.
Opened https://github.com/actix/actix-web/pull/3737 to tweak some code and add some tests (I cannot push directly to your branch)

PR #3019, opened 2023-04-20
set up dependabot

oriontvv:
## PR Type
Other

## PR Checklist

## Overview
This tech PR adds dependabot config for managing dependencies. Not sure if checklist should be processed.


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "docker"
```
JohnTitor: I think we don't need this?

JohnTitor (commented):
It'd be great if we also could have the "open-pull-request-limit" to reduce the noise.

JohnTitor:
@robjtede Any objections on this? I'm worried this could be noisy for you.

.github/dependabot.yml
oriontvv: Sure, fixed.

oriontvv:
@robjtede Hello, sorry for mentioning - is this okay? I think it would be helpful for auto fixing security issues like #3022

robjtede:
I'm not 100% sure that adding dependabot is the right call for crates. From what I've seen, it tends to restrict version requirements unnecessarily.

Also fixing security issues are done at the application level and can never be auto fixed like this.

I'll have to delve into documentation of the options to figure out if there's a config that works for us.

robjtede:
My feeling is I'll be disabling (parts of) this almost immediately, but willing to try it out to get a sense of what it's trying to do and documenting the config file with it. I think at least, the GHA updater could be useful.

Issue #2287, opened 2021-06-24
[awc] HTTP Client does not respect standard when no content-length or transfer encoding is present


joelwurtz:
## Expected Behavior

A server return a http response without the "Content-Length" or "Transfer-Encoding: chunked" header, but with a body and a "Connection: close" header

HTTP Client is expected to have a the body sent by the server according to: https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.3 (last choice in the list)

## Current Behavior

Body is empty

## Steps to Reproduce (for bugs)

This can be tested with the embed php server (https://www.php.net/manual/en/features.commandline.webserver.php) and a 'echo "foo";' as a php file, (by default php close the connection without those headers.

## Your Environment

* Rust Version 1.53.0
* Actix Web Version:
    * awc "3.0.0-beta.5"

ArtBlnd:
I got same issue here. 
![image](https://user-images.githubusercontent.com/2209400/124597255-32001f80-de9e-11eb-8e57-cb2e40207005.png)

Server sends payload but client send me empty response and close the socket.
easily reproducible just by setting HTTP version to HTTP 1.0 so that streamed server disables chunked response.

NOTE: I am using awc 3.0.0-beta.6

ArtBlnd:
Seems this is H1 client bug. 
https://github.com/actix/actix-web/blob/2504c2ecb0906d261688cd61d93444d4f0537cde/actix-http/src/h1/decoder.rs#L164-L166

This should emit `PayloadLength::None` instead of `PayloadLength::Payload(PayloadType::Stream(PayloadDecoder::eof()))` but seems fixing this makes other issue like time out.

ArtBlnd:
Anyone help me to fix this out? Our whole system is working with actix ecosystem, so this is critical for us.

zesty:
Can either @ArtBlnd  or @joelwurtz post some code to reproduce? I'm not getting a clear picture of what the precise content/sequence of request/response, actual/expected behavior is.

Is there an applicable section in the spec for HTTP 1.0? OP refers to 1.1, but the additional repro was with 1.0.

aliemjay:
> Can either @ArtBlnd or @joelwurtz post some code to reproduce?

server:
`$ echo -ne "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\nhello\n" | nc -lp 8080`

client:
Response body should be "hello\n" instead of `None`, same as with curl `$ curl http://localhost:8080/`

[...]

PR #3038, opened 2023-05-20
[awc] Fix response body when no content-length header

armandmgt:
## PR Type

Bug Fix for awc

## PR Checklist

## Overview

As noted in #2287 response body should be read until connection is closed when neither Transfer-Encoding nor Content-Length are specified in HTTP/1.1 (item 7 in https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3).
In order not to break any current tests I added a check for the presence of the `Connection: close` header (which is not required by the spec).
For reference this is how go's standard library is doing it: https://github.com/golang/go/blob/aeb0644bd33e67f24e2411a651ac9ff72ddc96b4/src/net/http/transfer.go#L548


armandmgt:
Not sure this is really relevant / up-to-date
Feel free to re-open if needed !

PR #3068, opened 2023-07-10
fix(http): set error to Payload before dispatcher disconnect

### LINKED ISSUE #3067: lack validation of whether request payload match content-length
### code
```rust
use actix_web::{post, web, App, HttpServer};
use serde::Deserialize;

use log::warn;

#[derive(Deserialize)]
struct Info {
    name: String,
}

#[post("/")]
async fn index(web::Form(form): web::Form<Info>) -> String {
    warn!("name length: {}", form.name.len());
    format!("Welcome {}!", form.name)
}

#[actix_web::main]
    std::env::set_var("RUST_LOG", "debug");
    env_logger::init();

    HttpServer::new(|| {
        let form_config = web::FormConfig::default()
            .limit(4096)
            .error_handler(|err, _| {
                warn!("Form error: {}", err);
                actix_web::error::InternalError::from_response(
                    err,
                    actix_web::HttpResponse::Conflict().finish(),
                )
                .into()
            });
        App::new()
            .app_data(form_config)
            .service(index)
            .wrap(actix_web::middleware::Logger::default())
    })
    .bind(("127.0.0.1", 1234))?
    .run()
    .await
}
```

```toml
[dependencies]
actix-web = "4.3.1"
env_logger = "0.10.0"
log = "0.4.19"
serde = { version = "1.0", features = ["derive"] }
```
### steps
1. `cargo run`
2. send a request with content-length larger thn body
```bash
curl -v --location 'http://127.0.0.1:1234' --header 'Content-Length: 1000' --header 'Content-Type: application/x-www-form-urlencoded' --data-urlencode 'name=haha'
```
3. `ctrl-c` stop the `curl`

## Current Behavior
```bash
❯ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/rust_study`
[2023-07-08T01:59:18Z INFO  actix_server::builder] starting 8 workers
[2023-07-08T01:59:18Z INFO  actix_server::server] Actix runtime found; starting in Actix runtime
[2023-07-08T01:59:26Z WARN  rust_study] name length: 4
[2023-07-08T01:59:26Z INFO  actix_web::middleware::logger] 127.0.0.1 "POST / HTTP/1.1" 200 13 "-" "curl/8.1.2" 3.468221
```

## Expected Behavior
1. log something like `Form error: incomplete request`
2. http-status code 400 not 200

## Context

> [rfc-2616](https://www.rfc-editor.org/rfc/rfc2616#page-33)
> When a Content-Length is given in a message where a message-body is allowed, its field value MUST exactly match the number of OCTETs in the message-body. HTTP/1.1 user agents MUST notify the user when an invalid length is received and detected.

the incomplete request should not be handled as normal. because:
1. not match http sepc
3. the handler does not have the `Content-Length`, it can not valid this by itself.

## Possible Solution

take form as an example, check payload length after got
https://github.com/actix/actix-web/blob/ce3af777a05e6be2673290417b25f714e6e25830/actix-web/src/types/form.rs#L383-L396

ihciah:
## PR Type

Fix

## PR Checklist

## Overview

Now actix-web decode body with Content-Length like that:

1. PayloadDeocder: Use Framed in an unconventional way which maintains data length left and returns `Ok(Some(Bytes))` even when data is not sufficient. Returns `PayloadItem::Chunk` or `PayloadItem::Eof`.
2. Codec: Convert `PayloadItem` to `Message::Chunk(Some(Bytes))` or `Message::Chunk(None)`.
3. poll_request: Handling Message, for `Message::Chunk(Some(..))`, call `feed_data`; for `Message::Chunk(None)`, call `feed_eof`.
4. poll: call poll_request.

However, when connection closed, we have to tell Payload. So here( https://github.com/actix/actix-web/blob/ce3af777a0/actix-http/src/h1/dispatcher.rs#L1139 ) actix-web call `feed_eof`. But since the content length infomation is only inside the decoder, and the outside can only sense it by waiting the Stream ends. And here the stream ends in advance when connection closed.

If the payload sender exists, it means there's still data to receive, so we can just set error here before waking the receiver.

Close #3067


robjtede:
I'm apprehensive to accept PRs like this without a test case that fails before the patch.

ihciah:
Here is the case I think: https://github.com/actix/actix-web/issues/3067

This bug also caused an accident in our product environment.

ihciah:
I added a test case to prove the fix. Before the patch it fails.
Also the changelog is updated.

Note: This test might be a bit verbose as I'm not familiar with some of the actix built-in utils.

ihciah:
This issue may cause server writing wrong data to db in the case of unstable network. Do you think we should publish a new version for it? Theoretically all users could be affected by this.

ihciah:
ping @robjtede Can you review it?

JohnTitor (approved):
Makes sense, thanks!

PR #3086, opened 2023-07-22
Add feature for limit request to log in middleware logger using status code

tglman:
Hi,
This is yet a preliminary change to just collect feedback:

So this PR add a new feature on the logger middleware, to limit the logging to a range of status code, so that are logged only some specific status in the range, a simple example that will log only the 400+ request/responses  is:

```
        Logger::default().statuses(StatusCode::BAD_REQUEST..);
```

Do you like this API is the naming right or you prefer a different one ?
Do you like this? shall I continue  :smile:  ?

## PR Checklist

       Could not really find a simple way to test this without an application, I didn't see any similar tests for other similar features where could I find/plug a test case for this case ?
   work in progress
   will do


tglman:
Hi,

I did add a simple dev-dependency to capture logs and check them in the tests, so I could add some tests that check the full flow, I updated also some existing tests to check that the output is what is expected.

Now this PR should be complete of all the requirements.

Regards

robjtede:
The only problem I see with this is the issue of disjoint ranges. I can see it being reasonable to want everything except redirects being logged; in which case it's either:
- `100..=299 ∪ 400..=999`
- `100..=999 ∖ 300..=399`

It's tempting to go with the set subtraction technique since more common patterns can be expressed with just one builder method. Despite being terse, it might be confusing, too.

So, the API could be `log_enabled: Vec<Range<StatusCode>>`.

PR #3087, opened 2023-07-22
Added bincode extractor and responder

JesusGuzmanJr:
## PR Type

Feature

## PR Checklist

    - I used `cargo +nightly fmt` to format because `cargo fmt` gave me a warning that unstable features are only available in nightly channel.

## Overview

This PR introduces the `web::Bincode<T>` extractor and responder.  [Bincode](https://docs.rs/bincode/latest/bincode/) is a popular Rust crate for encoding and decoding using a tiny binary serialization strategy. By using a binary representation in the payload of http requests and responses, one can avoid the unnecessary (de)serialization of field names like is done in JSON. Fundamentally this is due to the nature of JSON; as a self-describing data format, JSON must encode the shape of the data as well as the data itself and that is, what I consider, unnecessary overhead. With a binary format, like Bincode, the shape of the data is already known by server and client, so only the data need be transmitted. This lowers network traffic and aligns with Rust's Zero Cost Abstractions. 

## Implementation Details

The new code is more-or-less a copy-paste from the `web::Json<T>` extractor. Similar to the JSON extractor, the `BincodeBody<T>` handles the overflow checking to prevent DDOS attacks. 

No breaking change since this is a new feature.

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -9,6 +9,7 @@
 - Add `web::Payload::to_bytes[_limited]()` helper methods.
 - Add missing constructors on `HttpResponse` for several status codes.
 - Add `http::header::ContentLength` typed header.
+- Add `web::Bincode<T>` extractor and responder behind new `bincode` crate feature.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


robjtede:
This extractor is not appropriate for inclusion in Actix Web itself. It's too specific of a format for general use and not worth a feature flag here.

I can see https://crates.io/crates/actix-bincode so maybe this is useful to you and/or you can contribute to it. If not, I'd accept a submission to `actix-web-lab`.

JesusGuzmanJr:
Oh I didn't know about that crate. Thanks for pointing it out!

PR #3093, opened 2023-08-01
Test: WebSocket close behavior.

mstyura:
## PR Type

Test/Question/Other

## PR Checklist

## Overview

This PR emphasises my question regarding proper behavior of `WebSocket` close procedure. 
It seems like neither receiving close and sending it back to client nor initiating close and receiving it back from client ([close handshake](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.2)) does not cause web socket stream to finish and TCP connection released.

I'm not sure if it's my expectation from API is wrong or there is a bug. But anyway the question is what is the proper way to terminate WebSocket (except `ctx.stop()` which seems to terminate connection ungracefully, possibly loosing pending messages)?

I've inspected code and it seems like there are some parts which are intended to do some special handling of web socket closure, but this code seems to be never executed. E.g. there is [`Message::Close`](https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-http/src/ws/dispatcher.rs#L163-L164) with single usage across `actix-web` repo, so it's not created anywhere:
https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-http/src/ws/dispatcher.rs#L357-L360

There is also some code within [`WebsocketContextFut`](https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-web-actors/src/ws.rs#L671) to mark `WebSocket` as closed, which marks WS as closed when `None` is read from `messages` queue:
https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-web-actors/src/ws.rs#L703-L710
while there is no possibly to write `None` to outgoing messages queue:
https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-web-actors/src/ws.rs#L605-L638 

The only part of close which seems to work as expected is when underlying connection is closed by client app and handled here:
https://github.com/actix/actix-web/blob/ff8fd2f7b54dce405ede595fae568561a8465d48/actix-web-actors/src/ws.rs#L769-L784

So maybe maintainers could kindly provide answers to questions I've raised here: 
1. Is provided modifications to tests should be passed or not? (currently not passed)
2. If no, what is the proper way to gracefully close WebSocket connection by server side?


centromere:
Will this get investigated? If I run `f = stream.next().await` and the remote peer sends a TCP FIN, `f` will not resolve.

JohnTitor:
Closing in favor of https://github.com/actix/actix-extras/pull/476

PR #3096, opened 2023-08-02
web: add `HttpRequest::full_url()`

### LINKED ISSUE #2895: Request.uri() returns the Path instead of the full URL
## Expected Behavior

I need the full URL in a response handler.
`Request.uri()` should print the full URL (e.g. `https://example.com/hello-world`).

## Current Behavior

Prints the path (`/hello-world`), which is _not_ a [valid URI](https://www.rfc-editor.org/rfc/rfc3986).

## Possible Solution

Suggestion:

- The [`URI` struct](https://docs.rs/actix-web/latest/actix_web/http/struct.Uri.html) should always be a valid URI, which means it must have a schema + host + origin.
- The current behavior of `req.uri()` returns the `path` of the URI. I think a `.path()` method is a better name for this. Add a `full_uri()` method that returns the full URI.

```rust
pub async fn handle_get_resource(
    req: actix_web::HttpRequest,
) -> AtomicServerResult<HttpResponse> {
    println!("URL: {}", req.uri());
```

## Your Environment

- Rust Version (I.e, output of `rustc -V`): rustc 1.63.0 (4b91a6ea7 2022-08-08)
- Actix Web Version: 4.1.0

TimoCak:
## PR Checklist

## Overview

Hi Guys, 

i implemented a new method called `full_uri()` for the `HttpRequest` Type. To get the full uri of a request since the already existing method `uri()` does not have this functionality.
It seemed to me like an easy solution to get informations which will be provided from `HttpRequest` methods.
So i used the `connection_info()` method to extract the scheme and the host from the Request.
The origin can be provided by the `uri()` method.
I used the `Uri::builder()` to form the return value of the method.

I tested it with a simple handler which returns the uri:

grafik
grafik

Thats basically it. If there is any information missing, please let me know! 

Cheers,

Timo

Closes #2895


robjtede (approved):
Made some tweaks to name and return type.

Thanks for getting this started, 👍🏻

joepio:
Thank you for this! :D

PR #3107, opened 2023-08-20
feat: change log level based on response status code

fawni:
## PR Type

Feature

## PR Checklist

## Overview

Changes the log level to `Warn` on client errors and redirection messages, and to `Error` on server errors.

The current behavior is to log everything on an `Info` level.

This helps logging crates customize the output. Here's an example using the [femme](https://crates.io/crates/femme) crate:

Before:

20_08_2023-17_16_WindowsTerminal_0vZ7vIB

After:

20_08_2023-17_17_WindowsTerminal_dYRJ1Pq

It would also be colored red on a 5xx status code.

I do not know how to test or document those changes so I apologize if I should have done that.

### DIFF
--- a/actix-web/src/middleware/logger.rs
+++ b/actix-web/src/middleware/logger.rs
@@ -16,7 +16,7 @@ use actix_service::{Service, Transform};
 use actix_utils::future::{ready, Ready};
 use bytes::Bytes;
 use futures_core::ready;
-use log::{debug, warn};
+use log::{debug, warn, Level};
 use pin_project_lite::pin_project;
 use regex::{Regex, RegexSet};
 use time::{format_description::well_known::Rfc3339, OffsetDateTime};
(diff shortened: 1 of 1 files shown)


tglman:
Hi @fawni,

I was thinking to do a similar thing, but instead of directly set a log level for a response type (which still do make sense), I was thinking to set the log level as parameter, and than filter the range of status to log, for the range filtering I did already a PR here : https://github.com/actix/actix-web/pull/3086 

so let the user choose a what level to log request, with multiple loggers, something like:

```rust
 Logger::new("....").statuses(StatusCode::OK..StatusCode::BAD_REQUEST).level(Level::Info)
 Logger::new("....").statuses(StatusCode::BAD_REQUEST..).level(Level::Warn)
```

WDYT ?

robjtede:
In middleware, any response (regardless of status code) that is sent is considered a success. Logging as INFO level is the policy of _this_ textual logger. If you want to separately track errors, I encourage you to write a middleware to add in your stack.

Also, we're gradually transitioning to the `tracing` ecosystem and the whole Logger will need a re-think. Until then I'd like to keep this one relatively simple with obvious behavior.

Thank you for your ideas and contribution.

PR #3136, opened 2023-09-14
Scope macro

### LINKED ISSUE #1709: `#[scope]` macro
an idea from a fork that seem cool:

https://docs.rs/actix-web-cute-codegen/0.1.0/actix_web_cute_codegen/attr.scope.html

pmd3d:
## PR Type
Start of adding proc macro scope feature based on issue #1709.

## PR Checklist

## Overview

Like the project the issue referenced, I added scope proc macro that takes a path. I couldn't figure out how the rest of the features would work int he referenced project so did not implement them. 

This is my first major pull request. I decided to remove non path arguments to the scope proc macro to reduce the size of this pull request assuming there may be other issues given my inexperience. Also the issue references a handler macro that I did not implement because I wasn't sure exactly if perhaps that was covered by a get macro or should be something more.

Issue #1709

For example, the scope macro together with a prefix path would be declarated before module like below.
Then, web methods in the module would get the prefix path concatenated to the path argument.

```rust
use actix_web_codegen::{scope};
#[scope("/test")]
mod scope_module {
  use actix_web::{get, HttpResponse, Responder};
  #[get("/test")]
  pub async fn test() -> impl Responder {
    // this has path /test/test
    HttpResponse::Ok().finish()
  }
}
```


actix-web-codegen/src/lib.rs
```diff
@@ -240,3 +240,92 @@ pub fn test(_: TokenStream, item: TokenStream) -> TokenStream {
...
+/// pub fn mod_test2() -> impl actix_web::Responder {
+///    actix_web::HttpResponse::Ok().finish()
+/// }
+///};
+/// ```
+///
+/// # Note
+///
+/// Internally the macro generate struct with name of scope (e.g. `mod_inner`)
+/// And create public module as `<name>_scope`
```
ghost:
```suggestion
/// Internally the macro generates struct with name of scope (e.g. `mod_inner`).
/// And creates public module as `<name>_scope`.
```

pmd3d:
Converted to draft. Rust 1.72 was mentioned in changes markdown. It seems like Rust 1.72 supports adding the proc macro directly to a module so I'm going to rework this pull request to remove the const function workaround and try having the macro directly on a module.

robjtede:
Done a bit of re-organization and docs tweaking. Think it's in a good state to merge.

One more thing I want to add is to disallow trailing slashes to align with this piece of documentation and it's reasoning: https://docs.rs/actix-web/latest/actix_web/struct.Scope.html#avoid-trailing-slashes

robjtede (approved):
great work 👍🏻  thanks for being patient for the review 🙃

PR #3157, opened 2023-10-03
feat: add ability to specify log level for access logger

tglman:
Hi,

This add a new feature to the `Logger` middleware to set the level of the log.

# Example 
` Logger::default().level(Level::Debug);`

This would ideally go together with https://github.com/actix/actix-web/pull/3086 

## PR Checklist

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add `Logger::level` to change the log level.
+
 ### Changed
 
 - Updated `zstd` dependency to `0.13`.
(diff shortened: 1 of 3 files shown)


tglman:
Hi,

This is complete also,  it may conflict with the pr: https://github.com/actix/actix-web/pull/3086  because it add the same deps for test the logs, I may rebase the one of the two pr based on the one that is merged before.

Regards

robjtede:
Thanks for the work you've put into this.

I'm planning on migrating all the actix/* codebases over to `tracing` gradually, but dynamic logging levels is not something that's well supported by that ecosystem. The benefits definitely outweigh that but it means this PR is probably never going to land as-is.

We can revisit this idea after the tracing migration.

PR #3168, opened 2023-10-16
fix: content type required flag

### LINKED ISSUE #3162: Setting `JsonConfig::content_type_required(false)` results in weird behavior
When spinning up the following server

```rs
#[derive(Debug, Serialize, Deserialize)]
struct MyObj {
    name: String,
    number: i32,
}

/// This handler uses json extractor
async fn index(item: web::Json<MyObj>) -> HttpResponse {
    println!("model: {:?}", &item);
    HttpResponse::Ok().json(item.0) // <- send response
}

#[actix_web::main]
    env_logger::init_from_env(env_logger::Env::new().default_filter_or("trace"));
    info!("starting HTTP server at http://localhost:8080");

    HttpServer::new(|| {
        App::new().wrap(middleware::Logger::default()).service(
            web::resource("/")
                .route(web::post().to(index))
                .app_data(web::JsonConfig::default().content_type_required(false)),
        )
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```

We get the following outputs when running requests against the server 

```bash
$ curl -d '{"name": "biz", "number": 2}' localhost:8080
Content type error # fails

$ curl -H 'Content-Type: muhahah' -d '{"name": "biz", "number": 2}' localhost:8080
{"name":"biz","number":2} # works

$ curl -H 'Content-Type: muhah/ah' -d '{"name": "biz", "number": 2}' localhost:8080
Content type error # fails
```

It seems like the server still tries to parse the content-type header even though the API seems to be telling it to ignore it

MarcioOrdonez:
## PR Type

Bug Fix

## PR Checklist

## Overview

Change the order of content type validation in the `JsonBody::new()` method. 

This PR fixes #3162

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fix content type required validation when set to false. [#3162]
+
 ## 4.4.0
 
 ### Added
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fix validation of `Json` extractor when `JsonConfig::validate_content_type()` is set to false. [#3162]
+- Fix validation of `Json` extractor when `JsonConfig::validate_content_type()` is set to false.
 
 ## 4.4.0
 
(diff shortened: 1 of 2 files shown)


robjtede (approved):
thanks for picking this up 👍🏻 😄

PR #3192, opened 2023-11-17
feat: use RPITIT for FromRequest

robjtede:


mkulke:
Will this work enable impl async traits in `State`?

I referenced this branch, but I'll see the same error as on `master`:

image

code [here](https://github.com/mkulke/async-trait-playground/blob/actix-web/src/main.rs)

PR #3199, opened 2023-11-22
Do not encode zero-sized response bodies

phooijenga:
## PR Type

Bug Fix

## PR Checklist

## Overview

While debugging [unleash-edge#341](https://github.com/Unleash/unleash-edge/issues/341), I noticed a that when a client requests compression, the Compress middleware uses `actix_http::encoding::Encoder` to encode the response body. If the response body is empty (as a result of using `()` or `Bytes::new()` or something like `HttpResponse::NotModified().finish()`), the resulting "compressed" response is is no longer empty.

This does not happen if `actix_web::body::None` is used, or the response code is 204 No Content.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Updated `zstd` dependency to `0.13`.
 
+### Fixed
+
+- Do not encode zero-sized response bodies
+
 ## 3.4.0
 
 ### Added
--- a/actix-http/src/encoding/encoder.rs
+++ b/actix-http/src/encoding/encoder.rs
@@ -52,7 +52,7 @@ impl<B: MessageBody> Encoder<B> {
 
     pub fn response(encoding: ContentEncoding, head: &mut ResponseHead, body: B) -> Self {
         // no need to compress an empty body
-        if matches!(body.size(), BodySize::None) {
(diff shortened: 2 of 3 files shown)


robjtede:
This will definitely need a test added that would fail on master. I think it's possible that some codecs rightly encode 0-byte inputs as non-zero.

phooijenga:
Fair enough, but unfortunately I'm not really familiar enough with this codebase to know where to add such tests. 
I have now added a test to the compression middleware that demonstrates my original issue.

robjtede:
> I think it's possible that some codecs rightly encode 0-byte inputs as non-zero.

It was too early in the morning when I wrote this.

Since this process is a negotiation, it would be silly to sent anything more than "identity" encoding if the codec would make it bigger. This question extends even to a minimum size body that gets compressed, say 512 bytes, but we'll leave that for later.

For now, can I just ask that the test you added (good place for it) also checks a brolti request and asserts that the content-encoding header is set to "identity" (or missing, I forget which happens).

phooijenga:
I've added a copy of the test which uses `Accept-Encoding: br` instead. I added a second test mod for this. I tried moving the `#[cfg(feature = "compress-gzip")]` on the module to more specific tests, but came to the conclusion that most of the tests don't really make sense without compression enabled (although interestingly, many of them pass anyway).

robjtede (approved):

Issue #780, opened 2019-04-15
AppEntry should be exported


otavio:
When splitting applications, it is expected to have an `app` or similar function which constructs the routes. 

For example:

```rust
pub fn app(addr: Addr<actor::Machine>) -> App<AppEntry, Body> {
    App::new()
        .data(API::new(addr))
        .route("/info", web::get().to(API::info))
        .route("/log", web::get().to(API::probe))
        .route("/probe", web::post().to(API::probe))
        .route(
            "/update/download/abort",
            web::post().to(API::download_abort),
        )
}
```

is expected to work. However, `ApiEntry` is not exposed.

It is used at:

https://github.com/actix/actix-web/blob/7a28b32f6d12b51d26a51e93a182e3e83504eb68/src/app.rs#L41-L56

So it'd make sense to have it exported.

fafhrd91:
`App` is not designed for this kind of usage. Would `App::configure()` help?

otavio:
Possibly. I will try to adapt our code for it.

A mention of this would be good, on the migration document.

otavio:
Yes, I went with `App::configure` to define the routes and like. It worked well.

Bellow is the code change:

```rust
pub fn configure(cfg: &mut web::RouterConfig, addr: Addr<actor::Machine>) {
    cfg.data(API::new(addr))
        .route("/info", web::get().to(API::info))
        .route("/log", web::get().to(API::log))
        .route("/probe", web::post().to(API::probe))
        .route(
            "/update/download/abort",
            web::post().to(API::download_abort),
        );
}
```

LukeMathWalker:
I just ran in this exact same issue - it should probably be highlighted in `App`'s documentation.

winston0410:
Exact same issue.

robjtede:
> App is a complex type and not intended to be returned from a function directly. The App::configure method provides a way to move routing setup to a shared function. In rare cases, a simple macro can be used for App setup. References: #780 #1005 #1156 #2039 #2073 #2082 #2301

- https://github.com/actix/actix-web/wiki/FAQ#how-can-i-return-app-from-a-function--why-is-appentry-private

PR #3240, opened 2024-01-08
add pub use crate::app_service::AppEntry;

singersbalm:
## PR Type

Other

## Overview

Just makes AppEntry public for usage in functions. example: ` fn register(&self, app: App<AppEntry>) -> App<AppEntry>;`

### DIFF
--- a/actix-web/src/lib.rs
+++ b/actix-web/src/lib.rs
@@ -121,6 +121,8 @@ pub use crate::{
     server::HttpServer,
     types::Either,
 };
+pub use crate::app_service::AppEntry;
+
 
 macro_rules! codegen_reexport {
     ($name:ident) => {


robjtede:
see https://github.com/actix/actix-web/issues/780#issuecomment-894704144

singersbalm:
@robjtede and how can i use it with a trait?

```
cannot be made into an object Note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically

**Comment by robjtede** [comment]
can you post full errors please

**Comment by singersbalm** [comment]
```
error[E0038]: the trait `B` cannot be made into an object
  --> src/lib.rs:18:45
   |
18 |     pub data: Arc<Mutex<dyn B + Send + 'static>>,
   |                                             ^^^^^^^^^^^^^^^^^ `B` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
  --> src/lib.rs:44:8
   |
42 | pub trait B {
   |           ------- this trait cannot be made into an object...
44 |     fn register<T: ServiceFactory<
   |        ^^^^^^^^ ...because method `register` has generic type parameters
   = help: consider moving `register` to another trait

```
Code
```rs
pub struct A {
    pub data: Arc<Mutex<dyn B + Send + 'static>>,
}

pub trait  B {
    fn register<T: ServiceFactory<
        ServiceRequest,
        Response=ServiceResponse<impl MessageBody>,
        Config=(),
        InitError=(),
        Error=Error,
    >>(&self, app: App<T>) -> App<T>;
}
```

robjtede:
Would it be okay for you to use:
```rust
    fn register<
        T: ...
    >(&self, app: App<T>) -> App<T>
    where
        Self: Sized;
```

singersbalm:
i tried it with sized, but it seems like this wont work. 
```rs
error: the `register` method cannot be invoked on a trait object
  --> src/main.rs:35:37
   |
35 |             app = v.lock().unwrap().register(app);
   |                                     ^^^^^^^^
   |
  ::: src/lib.rs:51:50
   |
51 |     >>(&self, app: App<T>) -> App<T> where Self: Sized {
   |                                                  ----- this has a `Sized` requirement


```

robjtede:
Ahh. I'd hoped you wouldn't need that method behind a dyn reference.

singersbalm:
any ideas what could solve this?

PR #3291, opened 2024-02-15
Add `unicode` feature to switch between `regex` and `regex-lite` crates as a trade-off between full unicode support and binary size

yujincheng08:
## PR Type

Feature

## PR Checklist

## Overview

Address #3290. It aims at removing a large dependency `regex-automa` for binary size.

Before `cargo bloat`:
```
226.2KiB regex_automata
```
After `cargo bloat`:
```
29.6KiB regex_lite
```

### DIFF
--- a/actix-web/Cargo.toml
+++ b/actix-web/Cargo.toml
@@ -40,7 +40,7 @@ name = "actix_web"
 path = "src/lib.rs"
 
 [features]
-default = ["macros", "compress-brotli", "compress-gzip", "compress-zstd", "cookies", "http2"]
+default = ["macros", "compress-brotli", "compress-gzip", "compress-zstd", "cookies", "http2", "unicode"]
 
 # Brotli algorithm content-encoding support
 compress-brotli = ["actix-http/compress-brotli", "__compress"]
@@ -72,6 +72,9 @@ rustls-0_21 = ["http2", "actix-http/rustls-0_21", "actix-tls/accept", "actix-tls
 # TLS via Rustls v0.22
 rustls-0_22 = ["http2", "actix-http/rustls-0_22", "actix-tls/accept", "actix-tls/rustls-0_22"]
 
+# Full unicode support
+unicode = ["dep:regex", "actix-router/unicode"]
+
 # Internal (PRIVATE!) features used to aid testing and checking feature status.
 # Don't rely on these whatsoever. They may disappear at anytime.
 __compress = []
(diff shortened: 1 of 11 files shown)


robjtede:
So the benefit of `regex` over `regex-lite` is it's unicode support so I think what I'd prefer is a feature flag called `unicode` (bikeshedding tbd), enabled by default, where disabling it falls back to `regex-lite`. Basically flip this around to make it _additive_.

yujincheng08:
I think there's also performance gain when using `regex` over `regex-lite`.

> The principal difference between the `regex` and `regex-lite` crates is that the latter prioritizes smaller binary sizes and shorter Rust compile times over **performance** and functionality.

yujincheng08:
@robjtede modified as requested.

robjtede:
> prioritizes smaller binary sizes and shorter Rust compile times over performance

Another good reason to have this be an on-by-default feature since this will be sort of "infectious" as a transitive dep.

Thankfully, most third party libs use default-features = false already.

yujincheng08:
Hi @robjtede, I rebased this PR again. Is it ready to be merged?

robjtede:
I've made some changes to isolate the conditional Regex logic (for actix-router at least) in a separate module which also allows us to keep using `regex::RegexSet` for better perf.

robjtede (approved):

actix-web/Cargo.toml
```diff
@@ -72,6 +72,9 @@ rustls-0_21 = ["http2", "actix-http/rustls-0_21", "actix-tls/accept", "actix-tls
 # TLS via Rustls v0.22
 rustls-0_22 = ["http2", "actix-http/rustls-0_22", "actix-tls/accept", "actix-tls/rustls-0_22"]
 
+# Full unicode support
+unicode = ["dep:regex", "actix-router/unicode"]
```
SteveLauC, line 76: It seems that this broke repos that are using actix-web from git? The `unicode` feature is not available in actix-router 0.5, but it is required here:

```
error: failed to select a version for `actix-router`.
    ... required by package `actix-web v4.5.1 (workspace/actix-web/actix-web)`
    ... which satisfies path dependency `actix-web` of package `xxx v0.x.x (xxx)`
versions that meet the requirements `^0.5` are: 0.5.2, 0.5.1, 0.5.0

the package `actix-web` depends on `actix-router`, with features: `unicode` but `actix-router` does not have these features.



failed to select a version for `actix-router` which could resolve this conflict
```

yujincheng08: @SteveLauC

```toml
[dependencies]
actix-web = { git = "https://github.com/actix/actix-web.git", rev = "c10f05a8672e53ecada8ec2f15bf9d927efade57", default-features = false, features = [
    "macros",
] }
[patch.crates-io]
actix-router = { git = "https://github.com/actix/actix-web.git", rev = "c10f05a8672e53ecada8ec2f15bf9d927efade57" }
```

SteveLauC: Yeah, that would make things work!

PR #3310, opened 2024-03-16
`#[any]` macro

### LINKED ISSUE #3263: Add macro for route handler accepting all HTTP methods
As discussed in <https://github.com/actix/actix-web/discussions/3259>, implement macros like `#[any("...")]` and `#[route("...", method = any)]` to create route handlers accepting all HTTP methods (which is currently only possible with something like `app.route("...", web::route().to(handler))`)

sabinonweb:
## PR Type

 It's a draft PR for progress checking
 DRAFT PR

PR_TYPE

## PR Checklist


actix-web-codegen/src/lib.rs
```diff
@@ -196,6 +196,7 @@ method_macro!(Connect, connect);
 method_macro!(Options, options);
 method_macro!(Trace, trace);
 method_macro!(Patch, patch);
+method_macro!(All, all);
```
robjtede, line 200: This will almost certainly be easier with a new macro creator rather than trying to retro-fit `method_macro!`

actix-web-codegen/tests/trybuild/route-any-method.rs
```diff
@@ -0,0 +1,25 @@
+use std::str::FromStr;
+
+use actix_web::http::Method;
+use actix_web_codegen::route;
+
+#[route("/single", method = "ALL")]
+async fn index() -> String {
+    "Hello Single!".to_owned()
+}
```
robjtede, line 9: I think we were trying to aim more so for the dedicated macro style:

```rust
#[any("/path")]
async handler() {
  // ...
```

actix-web-codegen/src/lib.rs
sabinonweb: Is making a custom http server and defining any a way or is there a better way? @robjtede

robjtede: What it would need to generate would be similar to the other macros (an HttpServiceFactory implementation) except without adding any method guards to the route.

sabinonweb: I am sorry but I am not sure if I understand @robjtede

sabinonweb: Screenshot 2024-03-17 123942
Do I need to define it here?
Screenshot 2024-03-17 124134
Or in service.rs in actix-http. 
Pardon me for asking so many questions btw

sabinonweb:
Would adding methods in with_any_methods, similar to with_methods work @robjtede ?

PR #3334, opened 2024-04-15
Return 415 rather than 400 on Urlencoded Content-Type mismatch

mpalmer:
## PR Type

Bug Fix

## PR Checklist

## Overview

When the `Content-Type` of a request that was parsed by `UrlEncoded` wasn't appropriate, a generic 400 response was returned.  Given that there is a dedicated response code for this (415), I thought it better to return that instead, so now it does.


robjtede (approved):

robjtede:
LGTM. I think this is more correct so worth the potential issues in downstream tests.

PR #3336, opened 2024-04-17
Update cookie dependency to 0.18

ffreddow:
## PR Type

Update dependency

PR_TYPE

## PR Checklist

## Overview

Update the cookie dependency to 0.18 since increasingly more browsers require the "Partitioned" attribute to be able to send cross-site-requests containing cookies. It was added to the cookie crate with version 0.18.1. 

The changes in the cookie crate introduce breaking changes in the creation of cookies.

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - Minimum supported Rust version (MSRV) is now 1.72.
+- Updated `cookie` dependency to `0.18`.
 
 ## 4.5.1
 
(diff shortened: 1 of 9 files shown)


robjtede:
It should be possible to use `cookie` v0.18 manually if you desperately need the experimental Partitioned flag.

I want to point out that `cookie` v0.18 is just one option for the v5.0 upgrade path. Migrating to `biscotti` is also something I'd like to try out before making any decisions.

ffreddow:
Yea for us this is not urgent. Currently we workaround by adding the partitioned flag manually which works just perfectly fine.

PR #3343, opened 2024-04-26
Strip non-address characters from Forwarded for=

mpalmer:
## PR Type

Bug Fix

## PR Checklist

## Overview

This is something of a followup to #2528, which asked for port information to not be included in  when it was taken from the local socket.

The  header's  element may optionally contain port information (https://datatracker.ietf.org/doc/html/rfc7239#section-6). However, as I understand it,  is *supposed* to only contain an IP address, without port (per #2528).

This PR corrects that discrepancy, making it easier to parse the result of this method in application code.

There should not be any compatibility concerns, as anyone parsing the output of  would already need to handle both port and portless cases anyway.


robjtede (approved):

robjtede:
LGTM, the prevalence (or lack thereof) of IPv6 makes this change low risk.

PR #3348, opened 2024-05-03
Fix type confusion in some scenarios

asonix:
## PR Type

Bug Fix

## PR Checklist

## Overview

When the feature for rustls 0.22 is enabled, and rustls 0.23 is also present in a project, there suddently exist multiple paths for errors when building middleware chains due to the use of two consecutive `?` operators without specifying the intermediate error type.

This commit addresses the issue by removing the first `?`, so that the first error type will always be known, and the second `?` always has a well defined implementation.

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - Minimum supported Rust version (MSRV) is now 1.72.
+- Avoid type confusion in rare circumstances
 
 ## 4.5.1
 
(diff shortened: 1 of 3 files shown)


asonix:
I originally found this issue while working on pict-rs. Here's a minimum reproduction:

```toml
[package]
name = "type-confusion"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
actix-web = { version = "4.5.1", features = ["rustls-0_22"] }
rustls = "0.23.5"
```

```rust
#[actix_web::main]
    actix_web::HttpServer::new(|| actix_web::App::new().route("/", actix_web::web::get().to(index)))
        .bind("127.0.0.1:9090")?
        .run()
        .await
}

async fn index() -> &'static str {
    "Hewwo Mr Obama"
}
```

asonix:
Looking into updating actix-tls in actix-http for rustls 0.23 support and running into these same compile errors :pensive:

robjtede (approved):

PR #3350, opened 2024-05-04
awc: gate TlsConnectorService behind any feature that uses it

asonix:
## PR Type

Bug Fix

## PR Checklist

## Overview
Gate this struct behind the features that use it, preventing a compiler warning introduced in 1.78 for unused code detection


asonix:
This still fails the public-api-diff because the prior version has the issue :p

PR #3351, opened 2024-05-05
actix-test: allow the configuration of the TestServer address

mpalmer:
## PR Type

Feature

## PR Checklist

## Overview

This is useful if you're running (say) Selenium tests against a running TestServer, and the Selenium workers are Docker containers elsewhere in the network. Not a *particularly* common use case, perhaps, but one that I can attest happens every now and then.


mpalmer:
CI failures appear to be the `TlsConnectorService` issues that are being addressed in #3350.

robjtede (approved):

robjtede:
thanks for the addition 👍🏻 

looks like it needs a quick look at the new test failure

mpalmer:
Those test failures are weird -- there's no immediately-obvious relationship to my changes.  However, since I can't see any recent changes on `master` that could be to blame, I'll roll up my sleeves and dig in, I guess.

mpalmer:
After some rummaging around, I'm fairly confident my changes *are* at fault, if only because the three failing tests are the ones that are calling `actix_test::start`, while the other tests are spawning a testing server using other mechanisms.  Complicating the problem is that I can't repro the test failures locally, but I'll figure that out one way or another.

mpalmer:
Test suite passes now.  The problem appears to be that, for some mysterious reason, an IPv6 address in the `actix-test`-constructed URL is being passed in the resolver library, which proceeds to lose its mind.  But only in CI -- it works fine for me locally, even when I hard-code an IPv6 listen address.  :shrug:

robjtede:
The CI runners, on their VMs, probably have some crazy IPv6 things going on.

I think the quick fix is to use 127.0.0.1 as the default instead of localhost.

robjtede:
Ahh I see that's actually what you did. Sorry I was on mobile earlier and didn't check the newest commit.

PR #3361, opened 2024-05-12
actix-http: Add rustls 0.23

asonix:
## PR Type

Feature

## PR Checklist

- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt.
- [x] (Team) Label with affected crates and semver status.

## Overview

Adds support for rustls 0.23 to actix-http in the same way rustls 0.22 support was exposed

PR #3362, opened 2024-05-12
awc: Add rustls 0.23

asonix:
## PR Type

Feature

## PR Checklist

## Overview
Followup to https://github.com/actix/actix-web/pull/3361

- Add rustls 0.23 in awc via three features, `rustls-0_23` for custom-configuration only, and the webpki-roots and native-roots features like in rustls 0.22
- Fix root store lookup when rustls-0_22-native-roots is enabled
- Add documentation to ClientBuilder::new, and Connector::new
- Add rustls-0.23 to actix-test


asonix:
This one was a bit tricky since awc with webpki or native roots now implicitly relies on the rustls default provider, and creating the ClientConfig will panic if there is no global default and the number of crypto backends enabled is not exactly 1

I did what I could with documentation & another base rustls 0.23 feature to try to mitigate this issue

robjtede:
seems everything was included in #3363

Issue #3367, opened 2024-05-14
High memory usage / large websocket buffer per client remains in memory after message is sent


RandomGHUser:
Issue: the websocket buffer per client will remain in memory and potentially continue expanding until stopping/closing the connection.  This becames a major memory usage issue when sending a large amount of data to client(s).  

Related to https://github.com/actix/actix-web/issues/3198 and https://github.com/actix/actix-web/issues/1967

I would expect the in-memory buffer to flush after a message is sent to a client (please correct me if I'm mistaken with the expectation).

Example below creates a websocket connection, connect to it and send a message "[x]M" (i.e. 5M or 10M) to generate a client message of x million random characters.

Memory starts around 4MB in the example video.  Then the memory jumps to expand the in-memory buffer per client to accomodate the message size, but doesn't flush after being sent.

```rs
use std::{env, thread};
use ::actix::{Actor, StreamHandler};
use rand::distributions::{Alphanumeric, DistString};
use actix_web_actors::ws::{self};
use actix_web::{web, App, HttpResponse, HttpServer, HttpRequest, Error, get};

pub fn get_address() -> String {
    match env::var("LISTEN_ADDRESS") {
        Ok(e) =>e.to_owned(),
        Err(_e) => "127.0.0.1:8080".to_string()
    }
}
pub fn assert_address() {
    let addr = get_address();
    std::net::TcpListener::bind(&addr).expect(&format!("Could not open port on {addr}"));
}
fn main() {
    assert_address();
    let actix = thread::spawn(|| {
        start().unwrap();
        std::process::exit(0);
    });
    actix.join().unwrap();
}

// actix

#[get("/ws")]
async fn websocket(req: HttpRequest, stream: web::Payload) -> Result<HttpResponse, Error> {
    let resp = ws::start(MyWs{}, &req, stream);
    resp
}
#[actix_web::main]
pub async fn start() -> std::io::Result<()> {
    let address = get_address();
    println!("Serving on {}", address);
    println!("websocat ws://{address}/ws -S -B 1000000000");
    HttpServer::new(|| {
        App::new()
        .service(websocket)
    })
    .bind(address)?
    .run()
    .await
}

// websocket

pub struct MyWs{}
impl Actor for MyWs {
    type Context = ws::WebsocketContext<Self>;
}
impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for MyWs {
[...]

PR #3369, opened 2024-05-18
actix-web-actors: take the internal buffer when yielding

asonix:
## PR Type

Bug Fix i think

## PR Checklist

## Overview
This just takes the internal buffer when yielding in the websocket context future rather than splitting it. The result is that the internal buffer is not re-used between yields, but it also no longer grows perpetually

relates to https://github.com/actix/actix-web/issues/3367


robjtede (approved):
makes sense 👍🏻

PR #3372, opened 2024-05-19
Do not require actix-router default features from actix-web-codegen

dbanty:
## PR Type

Bug Fix

## PR Checklist

## Overview

This should allow you to omit the `regex` dependency when using `actix-web` 4.6.0 with the `macros` feature enabled (and no `unicode`).


robjtede (approved):

robjtede:
Turns out we actually have a concession here in that the errors are strictly worse with the `unicode` feature disables. Only those who explicitly disable it will see them though, so I'm okay with it.

```diff
 13 | #[get("/{}")]
    | ^^^^^^^^^^^^^
    |
-   = help: message: Wrong path pattern: "/{}" regex parse error:
-               ((?s-m)^/(?P<>[^/]+))$
-                            ^
-           error: empty capture group name
+   = help: message: Wrong path pattern: "/{}" empty capture group names are not allowed
```

PR #3388, opened 2024-06-06
From Boxed ResponseError impl added

nyzd:
## PR Type

Feature (kind of)

## PR Checklist

## Overview

`Error` struct has a private cause property with type `Box<dyn ResponseError>` so why not implement `From<Box<dyn ResponseError>>` for it? It is basically a `new` method for the `Error` struct


robjtede (approved):
don't see any reason why not 👍🏻

PR #3393, opened 2024-06-09
Re-export `http::status::InvalidStatusCode`

SamuelMarks:
## PR Type

Feature

## PR Checklist

## Overview
Adding a custom error type as per your guide and want to percolate up the errors rather than `unwrap` in any child functions:
```rs
fn status_code(&self) -> Result<actix_web::http::StatusCode, actix_web::http::status::InvalidStatusCode> {
```

But I can't do that because you don't export [`InvalidStatusCode`](https://docs.rs/http/latest/http/status/struct.InvalidStatusCode.html). Happy to export the whole `status` module or just this one `struct`.

### DIFF
--- a/actix-http/src/lib.rs
+++ b/actix-http/src/lib.rs
@@ -31,7 +31,7 @@
 #![doc(html_favicon_url = "https://actix.rs/favicon.ico")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-pub use http::{uri, uri::Uri, Method, StatusCode, Version, status::InvalidStatusCode};
+pub use http::{uri, uri::Uri, Method, StatusCode, Version};
 
 pub mod body;
 mod builder;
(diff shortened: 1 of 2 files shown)


actix-http/src/lib.rs
```diff
@@ -31,7 +31,7 @@
 #![doc(html_favicon_url = "https://actix.rs/favicon.ico")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-pub use http::{uri, uri::Uri, Method, StatusCode, Version};
+pub use http::{uri, uri::Uri, Method, StatusCode, Version, status::InvalidStatusCode};
```
robjtede: I'm fine re-exporting this, but I think it belongs in the `error` module.

SamuelMarks: Great. Ok moved.

robjtede (approved):

PR #3401, opened 2024-06-14
[actix-web/src/helpers.rs] Make `MutWriter` `pub`lic

SamuelMarks:
## PR Type

Feature

## PR Checklist

## Overview

Due to insufficient document I usually run through your codebase to find the best way of doing things. For example, to `impl actix_web::ResponseError for` my error type, I see:
https://github.com/actix/actix-web/blob/5e29726/actix-web/src/error/response_error.rs#L43

Which uses your `pub(crate) MutWriter`; which I cannot use. This PR makes it usable. Can you consider either merging this PR or changing the body of `actix-web/src/error/response_error.rs` to only use `pub`lic types / functions / symbols?

🙏

### DIFF
--- a/actix-web/src/helpers.rs
+++ b/actix-web/src/helpers.rs
@@ -8,7 +8,7 @@ use bytes::BufMut;
 ///
 /// This is slightly faster (~10%) than `bytes::buf::Writer` in such cases because it does not
 /// perform a remaining length check before writing.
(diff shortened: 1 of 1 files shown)


robjtede:
If it's important for your app to have the minor optimization then I'd encourage you to copy this small snippet to your code. I don't want to make any public API guarantees about this type, though.

Thanks for the contribution, all the same.

PR #3408, opened 2024-06-19
Fix Rustls 0.22 & 0.23 are limited to 256 handshakes per second.

### LINKED ISSUE #3407: TLS handshake rate limit adjustment does not work for rustls 0.22 & 0.23
Your issue may already be reported! Please search on the [Actix Web issue tracker](https://github.com/actix/actix-web/issues) before creating one.

## Expected Behavior

[HttpServer::method.max_connection_rate](https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.max_connection_rate) should allow to adjust TLS handshakes per second as suggested by documentation when used with any version of rustls or openssl.

## Current Behavior

When used with `rustls 0.22` or `0.23` due to conditional compilation part of code to adjust handshake per seconds is ingored.

## Possible Solution

Will be provided as pull request.

1. Receive `> 256` new TLS connections per second per worker.

## Context

In case of incoming TLS handshakes exceed default limit of `256` per second service to start accepting connection slower with part of the connections reset at TCP level.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): `rustc 1.79.0 (129f3b996 2024-06-10)`
- Actix Web Version: `actix-web 4.7.0`

mstyura:
## PR Type

Bug Fix

## PR Checklist

## Overview

Currently due to possibly accidentally missed features with `rustls 0.22` and `0.23` above configuration statement the server does not allow configuration of TLS handshakes limit per seconds.
The proposed approach is to define internal feature `__tls` and rely on it when writing feature conditions where it is expected to work on any tls implementation. Similar approach is already used by `__compress` feature. Such approach will make addition of new version of rustls less error prone.

The breaking change proposed is to make [HttpServer::method.max_connection_rate](https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.max_connection_rate) conditionally defined based on `__tls` feature. This will make it obvious that this method take effect only for TLS connections rate.
This breaking change is completely avoidable just by moving feature condition inside method as it was before.

Closes #3407

### DIFF
--- a/actix-web/src/server.rs
+++ b/actix-web/src/server.rs
@@ -7,13 +7,7 @@ use std::{
     time::Duration,
 };
 
-#[cfg(any(
-    feature = "openssl",
-    feature = "rustls-0_20",
-    feature = "rustls-0_21",
-    feature = "rustls-0_22",
-    feature = "rustls-0_23",
-))]
+#[cfg(feature = "__tls")]
 use actix_http::TlsAcceptorConfig;
 use actix_http::{body::MessageBody, Extensions, HttpService, KeepAlive, Request, Response};
 use actix_server::{Server, ServerBuilder};
@@ -190,7 +184,7 @@ where
     /// By default max connections is set to a 256.
     #[allow(unused_variables)]
     pub fn max_connection_rate(self, num: usize) -> Self {
-        #[cfg(any(feature = "rustls-0_20", feature = "rustls-0_21", feature = "openssl"))]
+        #[cfg(feature = "__tls")]
         actix_tls::accept::max_concurrent_tls_connect(num);
         self
     }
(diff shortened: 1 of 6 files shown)


actix-web/src/server.rs
```diff
@@ -189,8 +183,8 @@ where
     ///
     /// By default max connections is set to a 256.
     #[allow(unused_variables)]
+    #[cfg(feature = "__tls")]
     pub fn max_connection_rate(self, num: usize) -> Self {
-        #[cfg(any(feature = "rustls-0_20", feature = "rustls-0_21", feature = "openssl"))]
```
robjtede: moving the flag like this is a breaking change, please move it back and we'll be able to get this merged

mstyura:
Not sure if first two items in checklist are relevant. So will mark as ready for review and wait for a feedback.

actix-web/src/server.rs
mstyura: Moved back to statement level, thanks for review!

robjtede (approved):

PR #3432, opened 2024-07-18
Support deserializing paths to sequences for multi-component (eg. tail) matches

maartendeprez:
## PR Type

Feature

## PR Checklist

## Overview

When extracting path parameters with a custom regex matching multiple path components (eg. "/path/to/{tail:.*}"), there is currently no way to percent-decode the components individually. This pull request changes that by adding support for sequences in the path deserializer. When deserializing the path into a Vec or a tuple, the path segment is first split on slashes, then individual path components are percent-decoded and offered to the deserializing type. Empty path components are ignored.

Since the parameter deserializer did not previously support deserializing to a sequence, I expect this to be fully backward compatible.


actix-web/src/types/path.rs
```diff
@@ -53,6 +53,26 @@ use crate::{
...
+/// use actix_web::{get, web};
+/// use serde::Deserialize;
+///
+/// #[derive(Deserialize)]
+/// struct Tail {
+///     tail: Vec<String>,
+/// }
+///
+/// // extract `Tail` from a path using serde
+/// #[get("/path/to/{tail:.*}")]
```
robjtede: please can we use the suggested syntax for tail matches in the tests and docs
https://docs.rs/actix-router/latest/actix_router/struct.ResourceDef.html#tail-segments

```suggestion
/// #[get("/path/to/{tail}*")]
```

actix-router/CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ## 0.5.3
 
 - Add `unicode` crate feature (on-by-default) to switch between `regex` and `regex-lite` as a trade-off between full unicode support and binary size.
+- Add support for extracting multi-component path params into a sequence (Vec, tuple, ...)
```
robjtede: move to unreleased section

maartendeprez:
Updated!

JohnTitor (approved):
Thank you and sorry for taking long time to be merged.

PR #3454, opened 2024-08-16
Consume complete message body even on error

### LINKED ISSUE #3182: Connections hang in CLOSE-WAIT state
If a post request **with a large body** gets error (413 or 404 or any other) as a response, the connection doesn't close, but hangs in `CLOSE-WAIT` state

## Expected Behavior

After getting error connection is closed successfully

## Current Behavior

Connection hangs in `CLOSE-WAIT` state with some bytes in `Recv-Q`

## Possible Solution

As it only happens to requests with a large body, I guess there is something to do with the bytes that weren't read from a socket at the time the error happens. So we need to make sure these bytes are being read before closing the socket.

1. Creating simple server
```rust
use actix_web::{App, HttpServer};

#[actix_web::main]
pub async fn main() {
    HttpServer::new(move || {
        App::new()
    })
        .bind("127.0.0.1:5556").unwrap()
        .run().await.unwrap();
}
```
2. Making request with large body, that will end up with an error (404 in this case)
```python
import requests
requests.post('http://127.0.0.1:5556/data', json={"data": "a"*1000000})
```

3. Checking sockets - we can see the connections to our server in `CLOSE-WAIT` state with a lot of bytes in `Recv-Q`
```bash
$ ss -n4t | grep 5556
FIN-WAIT-2  0       0            127.0.0.1:59240        127.0.0.1:5556          
CLOSE-WAIT  787381  0            127.0.0.1:5556         127.0.0.1:59240
```

## Context

We have an application that uses actix-web as http-api server. We've discovered that at some point it starts leaking the memory and connections. We've also found that another application is sending huge files to our app and keeps retrying as it gets 413 error. Future investigation showed that these things are connected and every 413 error leaves a connection in `CLOSE-WAIT` state. Later we managed to reproduce this problem with other errors (like 404 in the minimal example)

## Your Environment

- Rust Version: 1.70.0
- Actix Web Version: 4.4.0

gschulze:
## PR Type

Bug Fix

## PR Checklist

## Overview

Currently, if there is an error during an HTTP request, the message body is not completely consumed in `actix-web/src/types/payload.rs`. In cases where HTTP keep-alive is enabled, this causes the connection to hang in the `CLOSE-WAIT` state, with the TCP receive queue still containing some bytes that need to be consumed. This PR modifies the `Future` implementation for `HttpMessageBody` to always consume all available data and return with success or error at the end. Besides issue #3182, there might be similar issues that are resolved by this PR.

Closes #3182

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -23,6 +23,7 @@
 - Always remove port from return value of `ConnectionInfo::realip_remote_addr()` when handling IPv6 addresses. from the `Forwarded` header.
 - The `UrlencodedError::ContentType` variant (relevant to the `Form` extractor) now uses the 415 (Media Type Unsupported) status code in it's `ResponseError` implementation.
 - Apply `HttpServer::max_connection_rate()` setting when using rustls v0.22 or v0.23.
+- Always consume complete HTTP message body, even on error
 
 ## 4.7.0
 
--- a/actix-web/src/types/payload.rs
+++ b/actix-web/src/types/payload.rs
@@ -417,24 +417,23 @@ impl Future for HttpMessageBody {
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.get_mut();
 
-        if let Some(err) = this.err.take() {
-            return Poll::Ready(Err(err));
-        }
+        while let Some(chunk) = ready!(Pin::new(&mut this.stream).poll_next(cx)) {
+            if this.err.is_some() {
+                continue;
+            }
 
-        loop {
-            let res = ready!(Pin::new(&mut this.stream).poll_next(cx));
-            match res {
-                Some(chunk) => {
-                    let chunk = chunk?;
-                    if this.buf.len() + chunk.len() > this.limit {
-                        return Poll::Ready(Err(PayloadError::Overflow));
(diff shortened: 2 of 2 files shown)


gschulze:
Hey @robjtede, what are your thoughts on this?

JohnTitor:
I've tried your patch on my local, but the issue remains. Please check again that the fix is actually valid, and add a regression test for this case somehow (I know it's challenging, though).

gschulze:
@JohnTitor I remember having a thorough test setup one year ago when I investigated the issue. I need to check whether I can find the code from back then.

robjtede:
i think the issue is related to #3715 where we've discussed and are testing a fix at the actix-http level, i think that supersedes this PR

JohnTitor:
> i think the issue is related to #3715 where we've discussed and are testing a fix at the actix-http level, i think that supersedes this PR

Good to know! It'd be worth to confirming once https://github.com/actix/actix-web/pull/3721 is merged (I'll take a look later).

JohnTitor:
@gschulze Hey, another (possible) fix for the issue has been shipped, could you confirm it really fixes your issue by updating actix-http to v3.11.2?

JohnTitor:
Closing as fixed, let me know if the issue still remains. Thanks for the PR anyway!

PR #3455, opened 2024-08-17
Add uri typed headers

keithamus:
## PR Type

Feature

## PR Checklist

## Overview

This adds the `Location`, `Content-Location` and `Referer` typed headers. It also adds the `TryIntoHeaderValue` trait for `Uri` which allows `Uri` to be a typed value for these headers.

While `Location` _technically_ is a different type than `Content-Location` and `Referer` (`Location` allows for protocol relative URIs) I think it's not a large enough distinction to warrant a separate type, as using protocol relative urls is quite rare. However we could use `url::Url` instead of `Uri` for `Location` if we wanted to allow for that distinction.

The motivation for this PR is to allow for using `Referer` as an Extractor, and the other two were easy enough to also add, and complete the trifecta of URI related headers (to the best of my knowledge).


actix-http/src/header/into_value.rs
```diff
@@ -129,3 +129,12 @@ impl TryIntoHeaderValue for Mime {
...
 }
+
+impl TryIntoHeaderValue for Uri {
+    type Error = InvalidHeaderValue;
+
+    #[inline]
+    fn try_into_value(self) -> Result<HeaderValue, Self::Error> {
+        HeaderValue::from_str(&self.to_string())
+    }
+}
```
robjtede, line 140: deserves a changelog entry

actix-web/src/http/header/content_location.rs
```diff
@@ -0,0 +1,36 @@
+use super::{Uri, CONTENT_LOCATION};
+
+crate::http::header::common_header! {
+    /// `Content-Location` header, defined
+    /// in [RFC 7231 §3.1.4.2](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.4.2)
```
robjtede: please update references and text according to new spec RFC 9110

actix-web/src/http/header/mod.rs
```diff
@@ -14,6 +14,7 @@ use std::fmt;
 // - the few typed headers from actix-http
 // - header parsing utils
 pub use actix_http::header::*;
+pub use actix_http::Uri;
```
robjtede: `Uri` is already re-exported in the `http` module, i don't think there should be another one here

robjtede (changes requested):

keithamus:
@robjtede those changes have been made. Let me know if you have any more feedback.

keithamus:
I'm still interested in this patch, if you think it's worth merging in.

robjtede:
Just come back to this and I still like it, but I'm going to make the `Location` (and possibly `ContentLocation` after some more research) use `url::Url` instead of `http::Uri` so that fragments (`/path?query#fragment`) are supported.

I'll cut a release with what's been merged already and get to this one afterwards. Should be easy to fixup and get out this weekend.

robjtede:
Run into a hurdle with that idea, that `url::Url` doesn't support relative references, e.g.: `/new-path`. So both types seem incapable of expressing what we actually want from Location/Content-Location.

PR #3463, opened 2024-08-29
Ensure incompatible versions of time are not allowed

kornelski:
## PR Type

Other

## Overview

Rust 1.80 [broke compatibility with many versions of the `time 0.3` crate](https://github.com/rust-lang/rust/issues/127343). `actix-web` is one of the most popular crates using it, so bumping it here may help users avoid ever getting a broken `time` version, even if they have an older lockfile.

### DIFF
--- a/actix-web/Cargo.toml
+++ b/actix-web/Cargo.toml
@@ -164,7 +164,7 @@ serde_json = "1.0"
 serde_urlencoded = "0.7"
 smallvec = "1.6.1"
 socket2 = "0.5"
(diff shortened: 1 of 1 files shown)


robjtede:
Isn't it just as easy to `cargo update -p=time` as it would be to `cargo update -p=actix-web` as would be needed to get this patch if released?

kornelski:
Yes, although it helps for `cargo add`.

robjtede:
I don't feel that the upside justifies an over-specified dependency.

PR #3477, opened 2024-10-04
Improve worker_max_blocking_threads documentation

luca-iachini:
## PR Type

Documentation

## PR Checklist

## Overview
The current documentation for `HttpServer::worker_max_blocking_threads` says that the default number of blocking threads is "512 divided by the number of workers." However, this could lead someone to believe that the default value depends on the number of workers set in `HttpServer`, when it actually depends on the system's `available_parallelism`.


robjtede (approved):
Nice catch, thanks 👍🏻

PR #3492, opened 2024-11-01
Wake Future when EOF is received

centromere:
## PR Type

Bug fix

## PR Checklist

## Overview

When a TCP FIN is received, the Future needs to be awakened. Without this, it is not possible to detect a connection closure in end-user applications.

Relates to: https://github.com/actix/actix-web/pull/3093


actix-http/CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - Add `header::CLEAR_SITE_DATA` constant.
+- Fix bug which prevents end-users from discovering when the remote peer has closed its side of the TCP connection.
```
asonix: This should probably be in a new `### Fixed` section rather than in "Added"

actix-http/src/h1/payload.rs
```diff
@@ -205,6 +205,7 @@ impl Inner {
     #[inline]
     fn feed_eof(&mut self) {
         self.eof = true;
+        self.wake();
```
asonix, line 205: I think this makes sense. I can imagine a scenario where Payload is polled before EOF comes in, and then EOF comes in and Payload is not woken.

asonix (approved):
Makes sense to me

centromere:
@robjtede @asonix, may this get merged?

JohnTitor:
Seems superseded by https://github.com/actix/actix-web/pull/3749 and I'm going to close. Thanks for the PR anyway!

PR #3496, opened 2024-11-06
Add support for `permessage-deflate` WebSocket extension

segfault87:
## PR Type

Feature

## PR Checklist

## Overview

This PR implements DEFLATE compression (specifically `permessage-deflate` extension which defined at [RFC 7692](https://datatracker.ietf.org/doc/html/rfc7692)) support for WebSocket.
The whole feature is flagged behind `compress-ws-deflate` feature.
Currently, flate2 requires zlib for [certain feature](https://docs.rs/flate2/latest/flate2/struct.Compress.html#method.new_with_window_bits) required by `permessage-deflate` extension. Therefore using this feature will introduce extra zlib dependency.

Also, `Codec` has been modified to split into `Encoder` and `Decoder` as there are some cases when encoding and decoding take place in different producer/consumers. Still its interface is unmodified in order to prevent API compatibility breakage.

A PR counterpart on [actix-ws](https://github.com/actix/actix-extras/) will follow.


segfault87:
On the idea of implementing the extension outside of actix-http, I elaborated a bit at this comment: https://github.com/actix/actix-extras/issues/479#issuecomment-2466053016

asonix (approved):
There's a lot here but I think overall this looks good :) It's unfortunate that this is a breaking change to the public API in a few places. I'm not sure what the plan is for a new major version of actix-http/actix-web although I assume there's no immediate plans for such a release.

segfault87:
@robjtede Could you have a look at this?

PR #3515, opened 2024-12-06
feat(awc): allow to configure connector per request

joelwurtz:
## PR Type

Feature

## PR Checklist

## Overview

Goal of this PR is to allow configure some connector options per request instead of needed different clients 

It's not finished but i'm wondering how far this should go.

Since it's based on #3522 it allow to have a different pool on different conn life time or other config (like h2 windows size) which will be better as it ensure proper connection option.


joelwurtz:
EDIT: This pr has been rebased on #3522 so we can benefit from the pool key being determined by the connect object (instead of authority)

PR #3522, opened 2024-12-09
feat(awc): allow to set a specific sni host on the request

joelwurtz:
## PR Type

Feature

## PR Checklist

## Overview

This PR allow client to set a specific sni host for a request, this can be useful if you want to connect to a specific hostname, but do SSL validation on another one

In order to do that some thing have been changed : 

 * Uri is no longer cloned during connection, instead a specific object is created with only wanted information, it may be better as we don't need all uri information in the connection (only hostname, port and if it should use tls most of the time), so this should also reduce memory consumption on certain case.
 * A new object is used for tcp / tls connection : HostnameWithSni (can be changed to another name), basically it's the host with the optional sni host and a state (to know which host to use given the connector: we use hostname during tcp, for dns resolution if needed, we use sni_host if available during tls connection for server name)
 * The pool now use the connect object for its key of pooled connection, i believe the old way was not adequate (did not do test, since it was only relying on the authority of the uri, if you pass a specific local address by example, after a first request it would not have use the existing pool, and i believe it should not in this case ?)
 
There may be better implementation on some point, as i have done what it's the most pragmatic way as a first step, will be happy to fix those if you have some inputs.

### DIFF
--- a/awc/src/connect.rs
+++ b/awc/src/connect.rs
@@ -13,7 +13,10 @@ use futures_core::{future::LocalBoxFuture, ready};
 
 use crate::{
     any_body::AnyBody,
-    client::{Connect as ClientConnect, ConnectError, Connection, ConnectionIo, SendRequestError},
+    client::{
+        Connect as ClientConnect, ConnectError, Connection, ConnectionIo, SendRequestError,
+        ServerName,
+    },
     ClientResponse,
 };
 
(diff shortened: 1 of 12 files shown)


awc/src/connect.rs
```diff
@@ -103,17 +107,41 @@ where
...
+            ConnectRequest::Client(ref head, .., addr, ref sni_host) => {
+                (head.as_ref(), addr, sni_host.clone())
+            }
+            ConnectRequest::Tunnel(ref head, addr, ref sni_host) => (head, addr, sni_host.clone()),
+        };
+
+        let authority = if let Some(authority) = head.uri.authority() {
+            authority
+        } else {
+            return ConnectRequestFuture::Error {
```
joelwurtz, line 124: there may be a better way, but a lot of code need to be updated and was not sure about the correct / standard thing to do here, this is a pragmatic way of doing it, but if you have an idea on how to avoid this will be glad to hear.

awc/src/frozen.rs
```diff
@@ -26,6 +26,8 @@ pub struct FrozenClientRequest {
     pub(crate) response_decompress: bool,
     pub(crate) timeout: Option<Duration>,
     pub(crate) config: ClientConfig,
+    // @TODO: find a way to avoid cloning
+    pub(crate) sni_host: Option<String>,
```
joelwurtz: I'm not sure here if we keep cloning this, or maybe have an `Option<Rc<String>>` ? (or maybe another borrowed / enum like it's done with ServerName for rustls ?)

joelwurtz: Fixed : i used a specific servername object which can be owned or borrowed

PR #3524, opened 2024-12-10
feat(web): add trusted proxies feature

joelwurtz:
## PR Type

Feature

## PR Checklist

## Overview

This PR aims to add the possibility for end user to correctly trust information from the `ConnectInfo` object : host, scheme and real ip of end user.

Actually everything is trusted which can be a security risk if some endpoint need to check the ip of end user as someone may forge a request with a `Forwarded` or `X-Forwarded-*` headers and then expose sensitive data to this person.

Even if the doc clearly state that it can be a security risk trusting this, you have to implement this logic by yourself, i do believe proposing a correct implementation for trusting ip and other headers has its place under actix also this enforce better security by default.

Current default will trust
  * IpV4 loopback address
  * IpV4 private networks
  * IpV6 loopback address
  * IpV6 private networks
  * `Forwarded` header
  
Which means that if this is merged other headers will not be trusted anymore, i'm wondering if this should be considered as a BC break then (since user may not have the same information as before given their network / proxy configuration in front of their actix web server)

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -5,6 +5,7 @@
 - On Windows, an error is now returned from `HttpServer::bind()` (or TLS variants) when binding to a socket that's already in use.
 - Update `brotli` dependency to `7`.
 - Minimum supported Rust version (MSRV) is now 1.75.
+- Add trusted proxies features to allow for customizing the list of trusted proxies and headers for determining host, scheme and ip of the `ConnectionInfo` object.
 
 ## 4.9.0
 
--- a/actix-web/Cargo.toml
+++ b/actix-web/Cargo.toml
@@ -152,6 +152,7 @@ futures-core = { version = "0.3.17", default-features = false }
 futures-util = { version = "0.3.17", default-features = false }
 itoa = "1"
 impl-more = "0.1.4"
+ipnet = "2.10.1"
 language-tags = "0.3"
 log = "0.4"
 mime = "0.3"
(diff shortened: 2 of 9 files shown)


robjtede:
I think this is slightly too opinionated to add to Actix Web itself. We could do with some experimentation. I've done similar work specific to the Cloudflare headers here: https://github.com/robjtede/actix-web-lab/tree/main/actix-client-ip-cloudflare

What do you think about moving this to the -lab crate initially so we can freely mess with the API?

joelwurtz:
I'm fine with that, what would be the API for this then, only a helper that given a trusted proxies gives you all field ?

However, I dont think it's too much opinionated i have based the current implementation given the current RFC recommendation and also existing web framework (mostly [symfony](https://github.com/symfony/http-foundation/blob/7.2/Request.php#L1996), which does that from a long time and it works fine for most users)

Also we could have a default implementation that accepts everything by default so there is no behavior change.

EDIT : I have just saw the ip cloudflare crate on lab, and yeah it may be better to do that here. What do you think of having a crate that use this code and add a feature for cloudflare which allow to fetch their ip to automatically populate the trusted proxies ?

joelwurtz:
I have extracted the implementation into its own crate https://github.com/redirectionio/trusted-proxies, it rely on http 1.2.0 at the moment so it's not directly compatible, but it's just a trait to implement to this should be fairly easy if needed.

Will close this then

PR #3525, opened 2024-12-10
feat: do not use host header on http2 for guard

joelwurtz:
## PR Type

Fix

## PR Checklist

## Overview

Host header should not be used when the request comes from HTTP2.

Also i'm wondering if this Host should be checked with the connect info object instead so it can use forwarded header ? (But need #3524 to be safe)


robjtede (approved):
great fix, thanks 👍🏻

PR #3526, opened 2024-12-10
Fix continuous integration

joelwurtz:
## PR Type

Fix

## PR Checklist

## Overview

Fix CI:

 * install nightly for public api check
 * fix divan to version 0.1.15 not needing 1.80 msrv (otherwise it requires some new stable api)
 * fix parse size to 1.0.0 (otherwise requires msrv 1.81)
 
Not sure if it's the right fix, otherwise msrv need to be updated to 1.81

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,6 +12,7 @@ fmt:
 downgrade-for-msrv:
     cargo update -p=parse-size --precise=1.0.0
     cargo update -p=clap --precise=4.4.18
+    cargo update -p=divan --precise=0.1.15
 
 msrv := ```
     cargo metadata --format-version=1 \


.github/workflows/lint.yml
```diff
@@ -105,6 +105,9 @@ jobs:
         with:
           toolchain: ${{ vars.RUST_VERSION_API_DIFF }}
 
+      - name: Add nightly toolchain for cargo-public-api
+        run: rustup install nightly --profile minimal
+
```
robjtede: installing nightly rust is done above

if the version needs bumping i can do that in the repo variables settings

robjtede:
for msrv compat points, this is the technique which allows testing with newer versions on stable and older versions on msrv https://github.com/actix/actix-web/blob/002c1b5a196ba38ccf59a2d76927d8f4d8a05263/justfile#L10-L14

.github/workflows/lint.yml
joelwurtz: It seems that cargo public api requires `nightly-2024-10-18`, i removed this then since it's only a setting update

joelwurtz:
I added divan in the downgrade section of the justfile and remove other commits should be good (only need the update of the nightly version in settings)

.github/workflows/lint.yml
joelwurtz: Hum, it's strange, it seems to install stable version on every PR since https://github.com/actix/actix-web/pull/3501 var seems to be empty, maybe as user we don't have access to this variable so it's empty by default for us (but not for you ?).

robjtede:
>  Anyone with collaborator access to the repositories with access to a secret or variable can use it for Actions. They are not passed to workflows that are triggered by a pull request from a fork. 

urgh you gotta be kidding

robjtede:
Screenshot 2024-12-10 at 16 04 16

thought it would allow variables to pass to PRs :(

robjtede (approved):
thanks 👍🏻

joelwurtz:
Yeah, i think they do that because some people put sensitive information even in variable (not secret) there is a thread on this subject here https://github.com/orgs/community/discussions/44322 but there is no work around (only work around is to not use variables for this)

PR #3528, opened 2024-12-10
feat(awc): replace trust dns by hickory dns

### LINKED ISSUE #3558: Upgrade trust-dns-resolver dependency in awc
cargo audit is reporting a security issues in idna  see:
```
Crate:     idna
Version:   0.4.0
Title:     `idna` accepts Punycode labels that do not produce any non-ASCII when decoded
Date:      2024-12-09
ID:        RUSTSEC-2024-0421
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0421
Solution:  Upgrade to >=1.0.0
Dependency tree:
idna 0.4.0
└── trust-dns-proto 0.23.2
    └── trust-dns-resolver 0.23.2
        └── awc 3.5.1
            └── nucleus_beelzebot 0.2.1

error: 1 vulnerability found!
```
Please update the dependency, right now our pipelines have to explicitly ignore this CVE, as it's surprisingly difficult to ignore this one.
 
Chers, -dennis

joelwurtz:
## PR Type

Fix / Feature (not sure)

## PR Checklist

## Overview

See https://bluejekyll.github.io/blog/posts/announcing-hickory-dns/

Also current trust dns resolver force idna 0.4.0 to be used which have some CVE so it's kind of a security issue for people using the feature.

Closes https://github.com/actix/actix-web/issues/3558


DennisBerger1984:
upvote from me. It's the followup project from the same author, license is the same. Solves also my issue #3558

robjtede:
replacing the feature is a breaking change, can we adjust this PR so that it's additive instead?

joelwurtz:
> replacing the feature is a breaking change, can we adjust this PR so that it's additive instead?

I don't see why it's a bc break ? old feature is keep but is an alias to the new one it should not change anything for end user ?

JohnTitor (approved):

JohnTitor:
Agree with Joel, the main change here should be one related to trademarks and the old feature should work with the (almost) same functionality.

@robjtede Could you clarify a scenario that may cause a breaking change?

robjtede:
I think you're right. My reaction was too cautious. I haven't been able to contrive a breakage.

I'm going to push an update to 0.25 of hickory-resolver.

robjtede (approved):

PR #3531, opened 2024-12-13
feat(files): remove spawning a thread for each file handling, as it does more harm than actually blocking

joelwurtz:
## PR Type

Fix/Feat (not always sure with those kind of patch)

## PR Checklist

## Overview

I was looking at a recent comment that indicates than actix files was having bad perf, after some investigation it seems that using the sync api make it way better, than having to spawn a thread to make it async.

In my local environment i want from 11k req/s to 600k req/s (so roughly a x54 increase of performance)

It may be a problem if you only handle big files with poor fs i/io speed (so all actix web  threads will be blocked).

Maybe we could still spawn this thread if the size of the file is superior to some limit ?

### DIFF
--- a/actix-files/CHANGES.md
+++ b/actix-files/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Minimum supported Rust version (MSRV) is now 1.75.
+- Avoid spawning a thread for each file which does more harm than good
 
 ## 0.6.6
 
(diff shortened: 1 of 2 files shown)


robjtede:
Having some threshold on the spawn might be a good idea. I can see this current patch being quite risky for high concurrency situations.

joelwurtz:
Maybe, we can provide a default strategy that tell that half of worker can do sync file serving (using a sémaphore) and allow people to change this strategy if they want.

PR #3534, opened 2024-12-16
feat(web): use an extension for encoding, which allow a specific handler to disable compression if needed

joelwurtz:
## PR Type

Feature

## PR Checklist

## Overview

This PR aims to have better control of the compression middleware.

Actually the compression middleware will always encode the response once it used, however there is some case where we don't want compression, some middleware force the accept encoding to avoid this, but there is no possibility to prevent compression once it has been calculated.

With this PR, new middleware will have the ability to avoid compression just by inserting a Identity encoding into the `extensions_mut` of the request.

Also an user would be able to avoid compression a request by removing this data from the request extension (or setting it to identity)

Also this would allow other middleware that are **after** the compression middleware to still correctly handle the compression without end user needing to take care of the middleware order

Will finalize this PR if feature is accepted


JohnTitor:
I like the idea, it should be useful for some usecases.
@joelwurtz Are you still interested in finishing up this PR? Would love to see it.

PR #3560, opened 2025-01-31
implement Responder for Result<(), E: Error>

axos88:
## PR Type

Feature

## PR Checklist

## Overview

Implement Responder for ~Option<()> and~ Result<(), E: Error>.
See #1108 and #2391 on why it is not implemented / should not be implemented for (), but implementing for these types should not be an issue.


robjtede:
I still think implementing for `()` at all is too opinionated. I can forsee us getting issues opened about why `Some(())` is 204 and not 200 for example.

I feel that pushing users of Actix Web towards explicitly understandable return values is good thing.

axos88:
Why would you assume its a 200? 204 exactly means no content, which is what the unit type is - something with no content. For a 200 without a body an empty string could be used.

robjtede:
I fully agree.

After thinking about this, it's easy enough to just give the reasoning if it comes up. Pushing folks towards the correct response for empty content seems like a good change.

robjtede:
The Option impl is more questionable though due to the None variant. I'll think about it more.

axos88:
Yeah,  agreed Some(()) is indeed a weird value to return. Honestly i just need the result variant, i included the option for symmetry without enough thought it seems

axos88:
Although one could argue that None should be rendered as a 404, and i believe it is - I'm on my phone right now,  so in that case it makes sense

robjtede:
Yea I think I'd like to progress without the `Option<()>` impl for now. It does seem very weird to use as a `Responder`.

robjtede (approved):

axos88:
I just stumbled upon a case where Result<()> should be rendered as a 202, instead of 204.

A POST request that has no output (eg. delete this record) should be rendered as a 202 Accepted, and I believe I would model it same Result<(), Error> as a GET operation for an resource that exists, but has no data (like getting the preferences for the current user). 

So indeed, this might not be a very good idea after all, but it would be a good idea to guide users on the nuances of this.

PR #3566, opened 2025-02-05
Fix order of template arguments in Handler documentation example

justahero:
## PR Type

Documentation fix

## PR Checklist

## Overview

Before this change, the illustrative code example for the [`Handler` trait](https://github.com/actix/actix-web/blob/master/actix-web/src/handler.rs#L73-L84) did not match the order of template arguments in the [`Handler` implementation](https://github.com/actix/actix-web/blob/master/actix-web/src/handler.rs#L132).


robjtede (approved):
thanks 👍🏻

PR #3579, opened 2025-02-14
Added serde derive to close reason and close code

### LINKED ISSUE #3578: CloseReason does not optionally implement serde::Serialize and Deserialize
As per title, the `CloseReason` does not implement serde::Serialize and Deserialize. Why is this?

https://github.com/actix/actix-web/blob/9656383646463e95d33d17458cc4b02ba5930f2c/actix-http/src/ws/proto.rs#L197-L205

LucaCappelletti94:
## PR Type

Feature

## PR Checklist

## Overview

Added feature-gated support for `serde::Serialize` and `serde::Deserialize` to `CloseReason` and `CloseCode`

Closes  #3578

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Add `header::CLEAR_SITE_DATA` constant.
 - Add `Extensions::get_or_insert[_with]()` methods.
+- Add feature-gated `serde::Serialize` and `serde::Deserialize` implementations for `CloseCode` and `CloseReason`
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


JohnTitor:
@LucaCappelletti94 
Hey, thank you for the PR! As Rob mentioned in the issue, I'd like to know your use case of this feature to determine whether to accept or not.

LucaCappelletti94:
Thank you for reminding me of this PR, I have replied in the issue. TLDR, I wanted to use that struct within an error enum, which derives serde. If this struct does not support the serde derive, I would have to create a new struct (my current solution) which is functionally a clone, and I would rather not do that.

robjtede:
I don't think this should be merged.

Since there's no canonical serialisation of this in generic data formats, adding a serde derive is semantically wrong.

It's a public enum so all variants are matchable so you can write your own serialize_with function that fits your use case.

PR #3588, opened 2025-02-27
Make web::Path innards public

IriaSomobu:
## PR Type

Other?

## PR Checklist

- [x] ~Documentation comments have been added / updated~

## Overview

This change allows us to do something like this:
```rust
async fn delete_channel_message(
    web::Path((channel_id, message_id)): web::Path<(Snowflake, Snowflake)>,
) {
  // here channel_id and message_id are Snowflakes,
  // and not web::Path<Snowflake>
}
```

Dunno why other web types are public and this is not.

### DIFF
--- a/actix-web-codegen/tests/scopes.rs
+++ b/actix-web-codegen/tests/scopes.rs
@@ -24,7 +24,7 @@ mod scope_module {
 
     #[get("/twice-test/{value}")]
     pub async fn twice(value: web::Path<String>) -> impl actix_web::Responder {
-        let int_value: i32 = value.parse().unwrap_or(0);
+        let int_value: i32 = value.0.parse().unwrap_or(0);
         let doubled = int_value * 2;
         HttpResponse::Ok().body(format!("Twice value: {}", doubled))
     }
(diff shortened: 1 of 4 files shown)


robjtede:
This field was made private for v4.0 but later it was realized this was the wrong trade-off. Making it public again is certainly planned for v5.0.

This is a breaking change due to the Deref impl, so it should also be removed in this PR.

For now, you can use the [Path type from actix-web-lab](https://docs.rs/actix-web-lab/latest/actix_web_lab/extract/struct.Path.html) to get destructuring.

IriaSomobu:
> This is a breaking change due to the Deref impl, so it should also be removed in this PR.

Could you elaborate please?

robjtede:
The thing we were seeing in v3.x was users reporting confusion using this type without destructuring because the Deref impl allowed access to tuple elements like `path.1` but then `path.0.1` also worked so it was weird. Anyway, making this public will make the `path.0` case ambiguous to the compiler in certain cases, making this a breaking change.

IriaSomobu:
Okey, I get it. `path.0.1` really looks funny. Gimme 5min -- I'll remove that deref.

robjtede:
No rush. Like I say, this will need to wait for v5.0.

PR #3594, opened 2025-03-03
feat(web): initial support for route and HTTP method introspection

### LINKED ISSUE #1462: Route introspection
It would be nice to be able to get more information on what routes are configured and methods on Path for getting a url string without the parameters in them.

Use cases:

1. Swagger api generation by crawling the routes. It would be nice to be able to automatically serve up api for any routes that are configured.
2. Get url for matching a route without parameters. This would be nice for logging prometheus metrics. I don't want a label for every fully parameterized url, just the services/routes.

Thanks! Sorry if this is already possible. I could not figure it out :)

### LINKED ISSUE #2677: way to get a list of registered services
it doesn't currently look like there is a good way to get a list of registered services/endpoints.
such functionality would be useful for creating index sites that list all endpoints (my usecase).

dertin:
## PR Type

Feature

## PR Checklist

## Overview

Hello, this is my first contribution to the project. This PR introduces a new resource introspection feature for actix-web that allows retrieving all configured route paths along with their associated HTTP methods after the server has started. Currently, it supports only basic `Guards` that determine HTTP methods. The feature is activated using the `experimental-introspection` feature flag.

I developed this feature because it's extremely useful for auditing services with many route paths that change constantly. I hope it proves valuable for others as well, and I welcome any feedback or suggestions for improvement.

Closes #1462 
Closes #2677


actix-web/src/guard/mod.rs
```diff
@@ -146,7 +147,7 @@ impl Guard for Rc<dyn Guard> {
 /// ```
 pub fn fn_guard<F>(f: F) -> impl Guard
 where
-    F: Fn(&GuardContext<'_>) -> bool,
+    F: Fn(&GuardContext<'_>) -> bool + 'static,
```
dertin: This change introduces the requirement for F to be 'static in fn_guard. This could be a breaking modification for some existing use cases. The reason for this requirement is to allow the use of downcast_ref on Guard.

If there is an alternative that is less breaking, it is welcome.

dertin:
Hi, @robjtede 

Here's a limited-scope approach for listing routes and methods from the `Guards` it’s a starting point. 
Extending support to extract additional metadata might require more breaking changes 
(potentially involving proc-macro / `actix-web-codegen`). 

Please review and let me know your thoughts!

actix-web/Cargo.toml
```diff
@@ -128,6 +128,9 @@ compat = [
 # Opt-out forwards-compatibility for handler visibility inheritance fix.
 compat-routing-macros-force-pub = ["actix-web-codegen?/compat-routing-macros-force-pub"]
 
+# Enabling the retrieval of metadata for initialized resources, including path and HTTP method.
+resources-introspection = []
```
dertin: Can you suggest a better name for the feature flag? Another option could be `experimental-introspection` to better reflect its current state of development. Open to other naming suggestions that capture the intent of the feature.

actix-web/src/app_service.rs
```diff
@@ -105,6 +126,11 @@ where
         let rmap = Rc::new(rmap);
         ResourceMap::finish(&rmap);
 
+        #[cfg(feature = "resources-introspection")]
+        {
+            crate::introspection::process_introspection(Rc::clone(&rmap), rdef_methods);
```
dertin: Surely there is a better place to make the call to `introspection::process_introspection`. This observation applies to all locations where I am currently making this call, including `scope.rs`, `route.rs`, `resource.rs`, and `app_service.rs`. Given some time, I can improve this implementation, always aiming to minimize breaking changes. Any ideas?

actix-web/src/introspection.rs
```diff
@@ -0,0 +1,263 @@
...
+};
+
+use serde::Serialize;
+
+use crate::http::Method;
+
+static REGISTRY: OnceLock<Mutex<IntrospectionNode>> = OnceLock::new();
+static DETAIL_REGISTRY: OnceLock<Mutex<HashMap<String, RouteDetail>>> = OnceLock::new();
+static DESIGNATED_THREAD: OnceLock<thread::ThreadId> = OnceLock::new();
+static IS_INITIALIZED: AtomicBool = AtomicBool::new(false);
```
dertin: Using static global variables for shared state isn’t ideal, since all App instances in the same process will share the same data. I chose this approach for now to avoid bigger changes, but it should be improved later to support multiple independent App instances.

dertin:
Hi @robjtede 

I’ve finished the work on this PR. It adds the experimental introspection feature, JSON and plain-text reports, includes examples, unit and integration tests, and the corresponding CHANGELOG entry. If you notice anything missing, please let me know.

Could you take a look when you have a moment? If you’d like to make further changes or prefer a different approach for any part of the implementation, I’ll be happy to revise it.

Thanks!
Guillermo

JohnTitor (approved):
Thank you for the great implementation!
I've tweaked a bit but let's test this feature as experimental.
I'm going to declare the policy for experimental features in https://github.com/actix/actix-web/pull/3916.
If you need tweaks after releasing, a follow-up is appriciated!

PR #3595, opened 2025-03-03
Add from_bytes/u8_bytes to dev::Payload

### LINKED ISSUE #3589: Create a means of convinently creating `Payload` from `web::Bytes`
# Problem Statement
I needed to extract json data in my middleware for some validations and restore the `SeriveRequest` so that extractors can be used in the handler code. This seems a bit of a common pattern but i had to searched a lot only to find out that, this functionality is available in `actix-http` which is used in `actix-web` but isn't exposed. So I had to add the `actix-http` crate and add the `bytes_to_payload fn` to achieve something that ideally should be already doable.

# Suggestions
1. Add `dev::Payload::from_bytes` for creating payload from `web::Bytes`
2. We can add a req.set_payload_from_bytes that take a `web::Bytes` and does what `bytes_to_payload fn` do and then set the payload field
3. Add `bytes_to_payload fn` on SerivceRequest
4. Expose the `actix_http::h1/h2::Payload` because currently there doesn't seem to be a way of creating a new payload, you can only `take()` fn access payload with `into_parts/parts()` fn. So you have to add `actix-http` then use its `Payload` type to implement `bytes_to_paylod` but It seems a redundant since `actix-web` already uses `actix-http` internally

A quick search on [Source graph](https://sourcegraph.com/search?q=bytes_to_payload&patternType=keyword&sm=0&__cc=1&df=%5B%22lang%22%2C%22Rust%22%2C%22lang%3Arust%22%5D) shows a number of public `actix-web` repos are aready using this pattern

**I am open to implementing any of these suggestions if accepted**

bernardassan:
## PR Type

Feature

## PR Checklist

## Overview

Currently one needs to add the `actix-http` crate and write the fn `bytes_to_payload`
```rust
let bytes_to_payload =
    |buf: web::Bytes| -> dev::Payload {
        let (_, mut pl) =
            actix_http::h1::Payload::create(true);
        pl.unread_data(buf);
        dev::Payload::from(pl)
    };

req.set_payload(bytes_to_payload(
     web::Bytes::from(login_req_body)
));
```
to be able to create a `Payload` from `bytes::Bytes` or `Vec<u8>` bytes but with the added `from_bytes` and `from_u8_bytes` this becomes

```rust
req.set_payload(dev::Payload::from_u8_bytes(
    login_req_body
));
```

This allows convenient construction of Payload from bytes which is very useful in middlewares.

closes actix/actix-web#3589

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -6,7 +6,8 @@
 
 - Add `header::CLEAR_SITE_DATA` constant.
 - Add `Extensions::get_or_insert[_with]()` methods.
-- Add `dev::Payload::from_[bytes/u8_bytes]()` methods.
+- Implement `From<Bytes>` for `Payload`.
+- Implement `From<Vec<u8>>` for `Payload`.
 
 ### Changed
 
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Add `header::CLEAR_SITE_DATA` constant.
 - Add `Extensions::get_or_insert[_with]()` methods.
+- Add `dev::Payload::from_[bytes/u8_bytes]()` methods.
 
 ### Changed
 
--- a/actix-http/src/payload.rs
+++ b/actix-http/src/payload.rs
@@ -41,13 +41,31 @@ pin_project! {
 }
 
 impl<S> From<crate::h1::Payload> for Payload<S> {
+    #[inline]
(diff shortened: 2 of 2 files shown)


bernardassan:
@robjtede could you review and merge this change

robjtede (approved):

robjtede:
I changed the impl to use `From` for both. Makes the API a bit cleaner.

Thanks for the contibution 👍🏻

PR #3619, opened 2025-04-08
Fix HttpRequest::peer_addr documentation

silverpill:
## PR Type

Other

## PR Checklist

## Overview

In my tests, `HttpRequest::peer_addr` returns `None` when server listens on an unix socket. This was not documented.

### DIFF
--- a/actix-web/src/info.rs
+++ b/actix-web/src/info.rs
@@ -158,7 +158,7 @@ impl ConnectionInfo {
     /// The address is resolved through the following, in order:
     /// - `Forwarded` header
     /// - `X-Forwarded-For` header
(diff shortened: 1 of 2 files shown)


robjtede:
Thank you for this. I've cherry-picked bd8e3cf5fc14b2ff9b9e38f9049c2df89393cfd9 onto master but I'm leaving the other one unmerged.

`peer_addr` not being present on UDS is a bug, so I'll try to make the fix instead of updating this documentation.

robjtede:
after some more looking, the _peer_ address doesn't have a semi-sensible return value when using UDS, I was thinking about _local_ address (which we set to 127.0.0.1)

therefore, i'll take the second committed suggestion, too, thank you again! 👍🏻

PR #3649, opened 2025-05-16
Refactor and Optimization of Match Resource Definition

### LINKED ISSUE #3645: Refactor and Optimize Match Resource Definition
### **Current Behavior**

Currently, to check if a resource matches a definition using capture_match_info_fn, we always need to pass a function to perform additional checks.

This causes two problems:
*  The check function is called inside the capture_match_info_fn, so we cannot determine if fails on the check or matching part.

*  If the caller doesn't want to perform any extra checks, they still have to pass a dummy closure.

### **Possible Solution**

Instead of calling capture_match_info_fn directly and applying the guard check function (or another function passed as a parameter) inside the function, we can change the approach. Let the capture_match_info function only handle the task of resource matching using the service request and we can create a more general function wrapper where runs the matching and check.

Image

Also, what we could do in the routing layer is execute the check function only when capture_match_info returns true (meaning the resource definition matches) and only one time.
Then we stop the flow always but returning the resource definition only if check function returns success.

This avoid looping through all resource definitions always (if route not found or found but fails in check function), but we lose the flexibility of having multiple resource definitions with same routing configuration's but different guards (this flexibility is really needed/useful? Always routes can be defined using the join of the two defined route guards).

This change improves performance and reduces resource usage in the routing layer using early return.
It also improves abstraction and responsibility separation by avoiding the need to pass dummy closures when no check is needed.

Would be a pleasure to do my first contribution!
Kind Regards, 
Manuel

manuelgdlvh:
## PR Type

Feature / Refactor 

## PR Checklist

### Current Behavior

Currently, to check if a resource matches a definition using capture_match_info_fn, we always need to pass a function to perform additional checks.

This causes two problems:

* The check function is called inside the capture_match_info_fn, so we cannot determine if fails on the check or matching part.
* If the caller doesn't want to perform any extra checks, they still have to pass a dummy closure.

### New Behavior

The capture_match_info function was refactored into the following components:
* resolve_path() -> Processes the resource match information (Path Length and Dynamic Segments) using the resource definition.
* resolve_path_if_match() -> Checks if a resource matches resource definition and, if so, calls resolve_path() to extract perform path processing
* capture_match_info(): Now solely responsible for matching a single resource against a resource definition.

Additionally, the Router Builder has been enhanced with a new mechanism to track path conflicts:
When a resource definition is added, it increments a conflict counter if its path pattern matches one previously added (ignoring guards and other configurations).
At the end of the build process, it computes the maximum number of path conflicts.

This enables an early return optimization in recognize_fn()—if the number of matching paths reaches the recorded conflict maximum, further checks can be skipped.

### Old Implementation
Screenshot from 2025-05-16 22-02-40

### New Implementation
Screenshot from 2025-05-16 22-01-59

After multiple test suite executions, the following performance improvements were observed:
* With Guard Check Failure: Reduced from 27µs to 10µs
* Without Guard Check Failure: Reduced from 12µs to 10µs

The performance gain in the "no guard check failed" scenario is primarily due to the elimination of unnecessary data structure initializations (specifically, segment allocations) when performing static resource path matching.

Closes #3645


actix-router/src/router.rs
```diff
@@ -96,7 +121,18 @@ impl<T, U> RouterBuilder<T, U> {
         val: T,
         ctx: U,
     ) -> (&mut ResourceDef, &mut T, &mut U) {
+        if let Some((_, path_conflicts)) = self
+            .path_conflicts
+            .iter_mut()
+            .find(|(route_idx, _)| rdef.eq(&self.routes.get(*route_idx).unwrap().0))
```
JohnTitor, line 127: I think `rdef.eq` here doesn't work well for some code, e.g. when `/yay` and `/{val}` are registered at the same time. These have different patterns but should be routed both, i.e. treated as the same path when requesting `/yay` and we have to fall back into `/{val}` when `/yay` doesn't work.

https://github.com/actix/actix-web/blob/2c0be64b68b8466239c66c05fa7cf9f315fecf80/actix-router/src/resource.rs#L1086

JohnTitor (changes requested):
Thanks for the PR!
While we'd have some wins for specific cases, I think we need more tests to validate changes and confirm there's no behavior change.
Also I'm happy to see clippy warnings are addressed.

PR #3653, opened 2025-05-20
Resolved collisions between `missing_docs` clippy lints

LucaCappelletti94:
## PR Type

This PR is a bug fix.

## PR Checklist

- [x] Tests for the changes have been added / updated. - **does not really apply**
- [x] Documentation comments have been added / updated. - **does not really apply**
- [x] A changelog entry has been made for the appropriate packages. - **should this minimal change be added?**

## Overview

At the time of writing, when setting `missing_docs = "forbid"` in the workspace, all the generated macros like `get` produce the following error:

```
allow(missing_docs) incompatible with previous forbid
`forbid` lint level was set on command line
```

This error is caused by addressing the missing doc by silencing at a lower level by adding the `allow(missing_docs)` decorator. This PR removes this decorator, and adds a ```#[doc = concat!("Creates a new response builder with the status code `", stringify!($status), "`.")]``` for the `pub fn $name() -> HttpResponseBuilder` function, which satisfies the `missing_docs="forbid"` constraint.

This PR does not introduce any breaking change.


actix-web/src/response/http_codes.rs
```diff
@@ -6,7 +6,8 @@ use crate::{HttpResponse, HttpResponseBuilder};
 
 macro_rules! static_resp {
     ($name:ident, $status:expr) => {
-        #[allow(non_snake_case, missing_docs)]
+        #[allow(non_snake_case)]
+        #[doc(hidden)]
         pub fn $name() -> HttpResponseBuilder {
             HttpResponseBuilder::new($status)
         }
```
robjtede: this is undesireable as it removes all the documentation for the quick httpresponse builders listed here: https://docs.rs/actix-web/latest/actix_web/struct.HttpResponse.html#method.Ok

actix-web-codegen/src/route.rs
```diff
@@ -466,7 +467,8 @@ impl ToTokens for Route {
 
         let stream = quote! {
             #(#doc_attributes)*
-            #[allow(non_camel_case_types, missing_docs)]
+            #[doc(hidden)]
+            #[allow(non_camel_case_types)]
             #vis struct #name;
```
robjtede: this is undesireable as it hides any copied documentation from handler functions

using `forbid(missing_docs)` feels like a misuse of forbid to me; `deny` is plenty and allows these overrides to work as intended

LucaCappelletti94: > this is undesireable as it hides any copied documentation from handler functions
> 
> using `forbid(missing_docs)` feels like a misuse of forbid to me; `deny` is plenty and allows these overrides to work as intended

While this may be reasonable in some settings, in my project it has been decided that `forbid` (and the clear red messages) are needed, so I believe we should find a solution where both approaches work appropriately.

Would it be a valid alternative to replace `#[doc(hidden)]` with some auto-generated doc (to be better defined), in the general idea of `#[doc = "API entrypoint #name"]`? This would both continue to show the copied documentation, and satisfy clippy.

I had proposed to go for `#[doc(hidden)]` as I didn't know there was other documentation within the generated code.

actix-web/src/response/http_codes.rs
LucaCappelletti94: I have now replaced the `#[doc(hidden)]` here with ```#[doc = concat!("Creates a new response builder with the status code `", stringify!($status), "`.")]``` - happy to update the generated documentation comment with anything that you believe is more appropriate.

actix-web-codegen/src/route.rs
LucaCappelletti94: I have now replaced the `#[doc(hidden)]` here with (some code omitted for brevity):

```rust
let documentation =
    format!("Struct generated by the routing macro for the handler function `{name}`.",);

let stream = quote! {
    #[doc = #documentation]
    #[allow(non_camel_case_types)]
    #vis struct #name;
};
```

robjtede: I don't think the stated rationale is strong enough to make this proposed change. I see some solutions which fit your problem statement though:
- document your handlers so that the documentation is copied over to these generated structs, thus avoiding the lint errors;
- disable the `compat-routing-macros-force-pub` feature to match up visibility of your handlers, thus avoiding the lint errors which only apply to public items.

LucaCappelletti94: My handlers are documented, as far as I understand - there aren't any additional macro parameters, right?

In the two points that are interested by my PR, the documentation is missing, and therefore clippy is correct in ranting.

One entrypoint example that is currently raising the aforementioned clippy error in the current `actix-web`'s `master` branch is:

```rust
#[get("/healthchecker")]
/// Health checker endpoint to verify if the server is running.
async fn health_checker_handler() -> impl Responder {
    const MESSAGE: &str = "Build API with Rust, Diesel, Postgres,and Actix Web";

    HttpResponse::Ok().json(json!({"status": "success","message": MESSAGE}))
}
```

which raises:

```
error[E0453]: allow(missing_docs) incompatible with previous forbid
 --> web/backend/src/api/healthchecker.rs:5:1
  |
5 | #[get("/healthchecker")]
  | ^^^^^^^^^^^^^^^^^^^^^^^^ overruled by previous forbid
  |
  = note: `forbid` lint level was set on command line
  = note: this error originates in the attribute macro `get` (in Nightly builds, run with -Z macro-backtrace for more info)
```

robjtede: Okay, that's a bug then. As you can see from the line above your changes, the intention is to copy docs over.

I can merge the rest of this PR and look into it later if you'd like.

LucaCappelletti94: No, my bad - the issue in the `master` branch is solely related to the use of the `allow(missing_docs)` decorator for what entails the `#vis struct #name;`. Only the `pub fn $name() -> HttpResponseBuilder {` necessitates of a doc comment which is currently absent in the `master` branch.

robjtede (approved):

PR #3662, opened 2025-06-01
Added max_buffer_size as a configuration item on the httpservice

ngrondin:
## PR Type

Feature

## PR Checklist

## Overview

No breaking changes.
This new feature allows the user to configure the max_buffer_size used by the http service dispatcher, rather than rely on a const value.
This is important to control the chunk size in streaming scenarios. For example, when streaming images in a multipart/x-mixed-replace, if the images are smaller than the const MAX_BUFFER_SIZE, they will buffer and affect the end-user's refresh rate. With this feature, the user can select a lower max_buffer_size and stream arbitrarily small chunks.


aniruddhaadak80:
Exposing `max_buffer_size` protects the HTTP service from massive payload OOM (Out Of Memory) crashes. It's a great knob to have for high-throughput or file uploading endpoints!

themavik (commented):
`max_buffer_size` plumbed from `HttpServer` / `HttpServiceBuilder` into `h1::Dispatcher` replaces the fixed `MAX_BUFFER_SIZE` cap in both payload encode loops — wiring is consistent. nit: builder docs say "maximum buffer size" generically; spelling out that it is the h1 write-buffer coalescing threshold avoids confusion with read-side limits.

PR #3665, opened 2025-06-07
fix(actix-http): wake before return pending when read half disconnect and need to shutdown

### LINKED ISSUE #1313: A Stream object in a pending state won't be dropped after the client disconnect

bitcapybara:
## PR Type

Bug Fix

## PR Checklist

## Overview

Closes #1313

In our application, we use actix-web to forward websocket data to another service. During runtime, we discovered that when the browser closes the connection, actix-web cannot detect this event, causing tokio's io::copy call to block indefinitely. Through testing, we found that in the Dispatcher's poll method, there are the following two issues:

Because inner_p.state is not None, READ_DISCONNECT cannot be detected
After returning Poll::Pending, the Dispatcher will not be polled again

This PR resolves the above two issues, allowing the client disconnect event to be properly detected and the process to exit normally.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -1182,7 +1182,7 @@ where
                     let state_is_none = inner_p.state.is_none();
 
                     // read half is closed; we do not process any responses
-                    if inner_p.flags.contains(Flags::READ_DISCONNECT) && state_is_none {
+                    if inner_p.flags.contains(Flags::READ_DISCONNECT) {
                         trace!("read half closed; start shutdown");
                         inner_p.flags.insert(Flags::SHUTDOWN);
                     }
@@ -1216,6 +1216,9 @@ where
                         inner_p.shutdown_timer,
                     );
 
+                    if inner_p.flags.contains(Flags::SHUTDOWN) {
+                        cx.waker().wake_by_ref();
+                    }
                     Poll::Pending
                 };
 


JohnTitor (approved):
Makes sense to me, thank you!
Adding a test for this case is a hassle and I think it should be ok as-is.

thalesfragoso:
This doesn't seem like a proper fix. The client might have closed its writer side of the connection with a FIN, but still want to receive the response.

That's was why that `state_is_none` check was there. If the client really closed the whole connection, then you would be able to detect it when trying to write something to it, can you provide more information regarding your blocked `io::copy`?

bitcapybara:
Our program flow is divided into three parts:
 1. The browser initiates a websocket connection and closes the connection upon exit.
 2. Service A acts as a proxy for the websocket connection. actix-web receives web::Payload, upgrades it to a websocket connection using reqwest::Client, and forwards it to Service B. The request is relayed using tokio::io::copy(payload_upgrade_reader, client_upgrade_writer).
 3. Service B uses actix_ws::Session to receive websocket requests. Upon receiving a Message::Close or detecting a network interruption, it closes the connection using session.close().

service A (pseudocode)
```rust
async fn proxy(payload: web::Payload, client: web::Data<reqwest::Client>, target_url: &str) -> Result<HttpResponse, actix_web::Error> {
	let mut builder = client.get(target_url);
	let target_response = builder.send.await.map_err(...)?;
	let target_upgrade = target_response.upgrade().await.map_err(...)?;
	let (target_rx, mut target_tx) = tokio::io::split(target_upgrade);

	let mut client_reponse = HttpResponse::SwitchingProtocols();
	client_reponse.upgrade("websocket");

	let cnacel = CancellationToken::new();

	tokio::task::spawn_local({
		let cancel = cancel.clone();
        async move {
            let mut client_stream = payload.map_err(std::io::Error::other);
            let mut client_read = tokio_util::io::StreamReader::new(&mut client_stream);
            if let Some(Err(e)) = cancel
                .run_until_cancelled(tokio::io::copy(&mut client_read, &mut target_tx))
                .await
            {
                tracing::error!("Error proxying websocket client bytes to target: {e}");
            };

            tracing::info!("Websocket client closed");
        }
	});

	// =======StreamWithCancel will cancel the CancellationToken when stream get None=============
	let target_stream =
        StreamWithCancel::new(tokio_util::io::ReaderStream::new(target_rx), cancel);
    Ok(client_response.streaming(target_stream))
}
```

service B (pseudocode)

```rust
pub(crate) async fn send_message(
    req: HttpRequest,
    stream: Payload,
) {
	let (res, session, msg_stream) = actix_ws::handle(&req, stream)?;
    let cancel = CancellationToken::new();
    rt::spawn(send_message_ws(
        task_id,
        req,
        session.clone(),
        cancel.clone(),
    ));
    rt::spawn(echo_heartbeat_ws(session, msg_stream, cancel));
    Ok(res)
}
async fn echo_heartbeat_ws(
    mut session: actix_ws::Session,
    mut msg_stream: actix_ws::MessageStream,
    cancel: CancellationToken,
) {

	let _cancel_guard = cancel.drop_guard();

	let mut interval = interval(HEARTBEAT_INTERVAL);
	let reason = loop {
        let tick = interval.tick();
        pin!(tick);

        match future::select(msg_stream.next(), tick).await {
            Either::Left((Some(Ok(msg)), _)) => {
                tracing::trace!("msg: {msg:?}");

                match msg {
                    Message::Close(reason) => {
                        break reason;
                    }

                    msg_to_process => {...}
                };
            }

            Either::Left((Some(Err(err)), _)) => {
                tracing::error!("{}", err);
                break None;
            }

            // client WebSocket stream ended
            Either::Left((None, _)) => break None,

            // heartbeat interval ticked
            Either::Right((_inst, _)) => {
                ...
            }
        }
    };

    // attempt to close connection gracefully
    let _ = session.close(reason).await;
}
```

thalesfragoso:
@bitcapybara Your program was supposed to detect the EOF on the read stream indeed, but this is different from some of the complains in #1313, where people were referring to the writer side.

While investigating this problem, I noticed that when we feed an error or EOF through the `PayloadSender`, we don't wake up the receiving `web::Payload`, that should probably be the real culprit on your example.

This change seems to work for you because the dispatcher simply gives up after receiving the FIN from the client, but I don't think this is correct behavior. I will open an issue to fix the `PayloadSender` and then another one where we can discuss if we want to revert this change.

PR #3678, opened 2025-06-17
fix: close connection after receiving GO_AWAY from remote

mitja-kleider:
## PR Type

Bug Fix

PR_TYPE

## PR Checklist

## Overview

If the remote gracefully shuts down a HTTP/2 connection by sending GO_AWAY, the corresponding h2 error is returned.
Before this change, the connection remained in the pool and using it again results in the same h2 error.
With this change the connection is released from the pool as intended by the remote.

### DIFF
--- a/awc/src/client/h2proto.rs
+++ b/awc/src/client/h2proto.rs
@@ -121,7 +121,7 @@ where
             fut.await.map_err(SendRequestError::from)?
         }
         Err(err) => {
-            io.on_release(err.is_io());
+            io.on_release(err.is_io() || (err.is_go_away() && err.is_remote()));
             return Err(err.into());
         }
     };


mitja-kleider:
I found it difficult to write a suitable test, we would need to shut down the h2 connection gracefully after it has been established and before sending the response. Do you have an idea how to achieve that?

kevin-logan:
I've just found this PR, which is resolving an issue very similar to one I found: https://github.com/actix/actix-web/issues/3788

Unfortunately I don't expect your fix here would resolve my issue, but it's the same underlying problem: that all non-IO errors are ignored when releasing a connection, but some non-IO errors are permanent.

My proposed fix is similar to yours but doesn't bother to check if it's a remote GO_AWAY (as local GO_AWAY errors are precisely my issue) _and_ to ensure the same is done on error handling for both `io.poll_ready` and `io.send_request`.

Similarly I found it very difficult to write a test to show the issue. I've managed to write one that is flaky (to demonstrate with _some_ confidence there is indeed an issue) but nothing deterministic.

PR #3695, opened 2025-07-16
fix: include content-length with bytes test payload

imgurbot12:
## PR Type

Bug Fix

## PR Checklist

## Overview

I noticed when writing my own tests for an actix-web service that the "Content-Length" is not sent when using [`TestRequest::set_form`](https://docs.rs/actix-web/latest/actix_web/test/struct.TestRequest.html#method.set_form) and [`TestRequest::set_json`](https://docs.rs/actix-web/latest/actix_web/test/struct.TestRequest.html#method.set_json) methods.

Both call [`HttpTestRequest::set_payload`](https://docs.rs/actix-http/latest/actix_http/test/struct.TestRequest.html#method.set_payload) underneath so this should cover all cases at the root.

Hope this helps, thanks! :)

### DIFF
--- a/actix-http/src/test.rs
+++ b/actix-http/src/test.rs
@@ -98,6 +98,8 @@ impl TestRequest {
     }
 
     /// Set request payload.
+    ///
+    /// This sets the `Content-Length` header with the size of `data`.
     pub fn set_payload(&mut self, data: impl Into<Bytes>) -> &mut Self {
         let mut payload = crate::h1::Payload::empty();
         let bytes = data.into();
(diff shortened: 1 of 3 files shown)


actix-http/src/test.rs
```diff
@@ -100,7 +100,9 @@ impl TestRequest {
     /// Set request payload.
```
JohnTitor, line 100:
```suggestion
    /// Set request payload.
    ///
    /// This sets the `Content-Length` header with the size of `data`.

JohnTitor (approved):
Makes sense, thanks!

imgurbot12:
> Makes sense, thanks!

Anytime! thanks for the review :)

PR #3699, opened 2025-07-26
feat(actix-files): support multiple fallback directory indexes

imgurbot12:
## PR Type

Feature / Refactor

## PR Checklist

## Overview

This modification allows for the support of multiple index file fallbacks, similar to apache2/nginx.
Eg. `[index.actix.html, index.html, index.txt, ...]`

The implementation appends the indexes to a vector rather than the default option, and is configured
using the existing `Files::index_file` which can be called multiple times to append each entry.

This design should avoid any breaking changes as the default behavior of just calling it once results in the same functionality before the modification.

This PR supports a personal [project of mine](https://github.com/imgurbot12/bob) attempting to implement an caddy/apache-style reverse-proxy service. Supported by a [bunch of services](https://github.com/imgurbot12/actix-services) I've written recently.

Thank you :)


imgurbot12:
Is it possible to re-run the jobs on this one? It looks like some unrelated h2 unit-test timed out after getting stuck for almost an hour lol.

JohnTitor:
Triggered :+1: 

Notes:
I've looked the implementation and the best form of this change might require a breaking change.
I'll revisit this once we can go ahead for v5, but feel free to decline my assumption if you have different views!

imgurbot12:
Thanks! 

> I've looked the implementation and the best form of this change might require a breaking change.

Fair enough I guess, I'm using my fork of actix-web until all of my PRs end up merged to support [that project](https://github.com/imgurbot12/bob) I mentioned, though I'm not sure what you mean by "best form".

Samielakkad (commented):
The multiple-index API is useful, but the lookup path probably wants to avoid `Path::exists()` here:

```rust
let index = this
    .indexes
    .iter()
    .map(|i| path.join(i))
    .find(|p| p.exists());
```

Two reasons:

- `exists()` is a synchronous filesystem stat inside the async service path.
- More importantly, it changes fallback behavior. If the first configured index path exists but `NamedFile::open_async` fails because it is unreadable, a directory, races with deletion, etc., the service handles that error immediately and never tries the next configured index. With multiple fallbacks, I would expect it to attempt `index.html`, then `index.htm`, etc. until one actually opens.

A cleaner shape may be to iterate the configured names and call `NamedFile::open_async(path.join(index)).await` for each, returning the first successful open and only falling back to listing/default/error after all configured index files fail with not-found-style errors.

PR #3705, opened 2025-07-31
fix(awc): some request methods incorrectly send body & body-headers

imgurbot12:
## PR Type

Bug Fix

## PR Checklist

## Overview

The existing implementation for awc incorrectly sends a request-body for http methods that don't support one, such as `GET/HEAD/OPTIONS/TRACE`. This is especially apparent when passing a stream into the request for all methods since it uses chunked content-encoding.

I discovered this whilst implementing a reverse-proxy using actix-web and awc.

Example:

```
GET / HTTP/1.1
transfer-encoding: chunked
accept-encoding: br, gzip, deflate, zstd
host: localhost:8000
date: Thu, 31 Jul 2025 03:51:41 GMT

0
```

You can verify this yourself with a simple script (I'm using netcat to view the request `nc -lvp 8000`):

```rust
use std::{
    pin::Pin,
    task::{Context, Poll},
};

use awc::{error::PayloadError, http::Method};
use bytes::Bytes;
use futures_core::Stream;

struct ReqStream;

impl Stream for ReqStream {
    type Item = Result<Bytes, PayloadError>;
    fn poll_next(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        Poll::Ready(None)
    }
}

#[actix_rt::main]
async fn main() {
    let client = awc::Client::default();
    let res = client
        .request(Method::GET, "http://localhost:8000")
        .send_stream(ReqStream {})
        .await;
    println!("response: {res:?}");
}
```

The following changes ensure that the body and its associated headers are not included for the standard request methods, resulting in a correct request:

```
GET / HTTP/1.1
accept-encoding: br, gzip, deflate, zstd
host: localhost:8000
date: Thu, 31 Jul 2025 04:00:59 GMT
```

### DIFF
--- a/actix-http/tests/test_client.rs
+++ b/actix-http/tests/test_client.rs
@@ -139,7 +139,7 @@ async fn h1_expect() {
 
     // test expect would fail to continue
     let request = srv
-        .request(http::Method::GET, srv.url("/"))
+        .request(http::Method::POST, srv.url("/"))
         .insert_header(("Expect", "100-continue"));
 
     let response = request.send_body("expect body").await.unwrap();
@@ -147,7 +147,7 @@ async fn h1_expect() {
 
     // test expect would continue
     let request = srv
-        .request(http::Method::GET, srv.url("/"))
+        .request(http::Method::POST, srv.url("/"))
         .insert_header(("Expect", "100-continue"))
         .insert_header(("AUTH", "996"));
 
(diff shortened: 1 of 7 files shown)


themavik (commented):
`AnyBody::from_message_body` now strips payload for GET/HEAD/OPTIONS/TRACE before connect; flipping expect/h2 tests to POST matches real body semantics. nit: anyone relying on non-standard GET bodies through awc will see a behavior change — your CHANGES line flags it; CONNECT tunneling with a body would hit the same path.

PR #3718, opened 2025-08-05
fix: eof hanging bodies on streamed requests

### LINKED ISSUE #3715: Client Connection timeouts when running behind loadbalancers
## Current Behavior
When hosting Actix behind Traefik or Envoy, if a POST request is made with invalid data, or invalid headers, Actix does not close the channel from Traefik/Envoy's perspective, but stops polling it, so next request that comes in will never get an answer from Actix.

Using curl directly against actix with the same incorrect body, does not seem to reproduce the issue.

## Expected Behavior
I would not expect a previous request to cause new requests to experience timeouts, even if there's a loadbalancer in front of Actix

## Possible Solution

Our workaround is putting nginx in our docker container to interface with traefik/envoy and then forward requests to actix using proxy_pass, this has removed our 499s, but adds more cpu load to the nodes running the docker container.

https://github.com/chriswk/actix-web-lb-connection
To reproduce: 
Follow the readme, I've narrowed the actix application down to the absolute minimum here, no extra middlewares, no custom code. This is what makes me think this is deep in actix internals.

## Context

We first noticed this problem when deploying our actix app to an AWS App runner (Uses envoy for loadbalancing). All metrics on our side looked fine, but our customer experienced client timeouts. Later we saw the same issue when deploying the same app inside our k8s cluster which uses Traefik 2.11.

Using haproxy or nginx or connecting directly to the actix app does not seem to surface the same problem.

As an extra aside here, I could not reproduce it with the same minimum example, just using axum instead of actix-web, not matter which load-balancer I put it behind. But since the app we're actually running has grown rather large, I would really like to not have to migrate.

## Your Environment

- Rust version: 1.88 (though we've experienced this on 1.85 and 1.81 as well, in both arm and amd64 flavours)
- Actix Web Version: 4.11 (though reproducible in 4.8 and 4.9 as well)
- Traefik: 3.5 / 2.11 (both exhibit same behaviour)
- Amazon App Runner (uses envoy behind the scenes)

sighphyre:
Fixes #3715

Not convinced this is a perfect fix so open to suggestions and feedback here. The docs tests seem to be failing on my fork of master here as well so I don't _think_ it's anything I've done here.

There's no tests for this because I cannot for the life of me find a way to reproduce this with a test harness. But #3715 provides a functioning example of the bug.

### Description

This PR introduces a fix for a a timeout issue with streaming proxies like Traefik. 

The bug occurs when a client sends an invalid request to the server as a multipart TCP packet. Actix may be able to detect that the request will not succeed, for example due to an invalid header against a request that requires a body, the server will then respond immediately to the client while leaving the body bytes unread. When a second request is made, those body bytes are then read as the first part of the message, causing the server to block until it hits its own timeout threshold, since they can't be parsed as a valid message.

This fix forces a eof to be sent in cases where the request will not receive further processing but bytes remain unread from that request.

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -20,18 +20,18 @@ use tokio_util::codec::{Decoder as _, Encoder as _};
 use tracing::{error, trace};
 
 use super::{
-    Message, MessageType,
     codec::Codec,
     decoder::MAX_BUFFER_SIZE,
     payload::{Payload, PayloadSender, PayloadStatus},
     timer::TimerState,
+    Message, MessageType,
 };
 use crate::{
-    Error, Extensions, OnConnectData, Request, Response, StatusCode,
     body::{BodySize, BoxBody, MessageBody},
     config::ServiceConfig,
     error::{DispatchError, ParseError, PayloadError},
     service::HttpFlow,
+    Error, Extensions, OnConnectData, Request, Response, StatusCode,
 };
 
 const LW_BUFFER_SIZE: usize = 1024;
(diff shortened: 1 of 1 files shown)


ivarconr:
@robjtede this is critical for us to fix. How can we help get this issue resolved?

robjtede:
As it's critical for you, you can use cargo's patching system for your projects for now.

I don't currently have access to my development machine since it was stolen recently so PRs requiring scrutiny like this one are held up atm. Sorry about that.

ivarconr:
> As it's critical for you, you can use cargo's patching system for your projects for now.
> 
> I don't currently have access to my development machine since it was stolen recently so PRs requiring scrutiny like this one are held up atm. Sorry about that.

Thanks for your feedback, and I am sorry about your laptop.

robjtede:
prefering #3721 as the fix for #3715

PR #3721, opened 2025-08-07
actix-http: h1: stop pipelining when not reading full requests

### LINKED ISSUE #3715: Client Connection timeouts when running behind loadbalancers

asonix:
An alternate solution to https://github.com/actix/actix-web/issues/3715

The existing pipelining behavior of the h1 dispatcher can cause client timeouts if the entire request body isn't read. It puts the dispatcher into a state where it refuses to read more (payload dropped) but there are still bytes in the buffer from the request body.

This solution adds the SHUTDOWN flag in addition to the FINISHED flag when completing a response when both the following are true:

1. There are no messages in `this.messages`
2. There is still a payload in `this.payload`

This combination implies two things. First, that we have not parsed a pipelined request after the request we have just responded to. Second, that the current request payload has not been fed an EOF. Because there are no pipelined requests, we know that the current request payload belongs to the request we have just responded to, and because the request payload has not been fed an EOF, we know we never finished reading it.

When this occurs, adding the SHUTDOWN flag to the dispatcher triggers a `flush` and a `poll_shutdown` on the IO resource on the next poll.

## PR Type

Bug Fix

## PR Checklist


Sytten:
Considering the recent releases of new http desync techniques at defcon (https://portswigger.net/research/http1-must-die) we will want to double check that code a bit more. I tend to support any preventive early shutdown since the base of those attacks is leaving data in the buffer that is interpreted as the next release.

robjtede (approved):

asonix:
I've added a test that fails without the change and passes with the change. specifically the line that fails the test is the one that panics if the poll of h1 is not pending. With the change present, calling poll with an unfinished request shuts down the dispatcher.

asonix:
Is there anything else you want me to do on this PR? I see you approved it but I don't have permission to merge it :p

robjtede:
Nothing on top of mind, I just havent checked out the added test yet.

JohnTitor (approved):
LGTM! Thank you for the fix and detailed explanation!

lovasoa:
Possibly related: https://github.com/actix/actix-web/issues/3967

PR #3722, opened 2025-08-09
fix(http): preserve manual Content-Length headers for HEAD requests p…

tao-guo:
fix(http): preserve manual Content-Length headers for HEAD requests per RFC 7230

## PR Checklist

## Overview

Implements RFC 7230 [Section 3.3.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2) which allows servers to send Content-Length headers in HEAD responses. This enables support for APIs like AWS S3 HeadObject that require Content-Length headers even with empty response bodies.

Previously, manually set Content-Length headers on empty body responses were being overridden by automatic "content-length: 0" headers. Now the encoder checks for existing Content-Length headers before adding automatic ones.

### DIFF
--- a/actix-http/src/h1/encoder.rs
+++ b/actix-http/src/h1/encoder.rs
@@ -60,8 +60,13 @@ pub(crate) trait MessageType: Sized {
         config: &ServiceConfig,
     ) -> io::Result<()> {
         let chunked = self.chunked();
-        let mut skip_len = length != BodySize::Stream;
+        let mut skip_len = false;
         let camel_case = self.camel_case();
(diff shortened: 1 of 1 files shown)


robjtede:
in general allowing content-length to be set manually is a huge footgun

there is already a pattern to allow the S3 HEAD style requests as demonstrated in the examples: https://github.com/actix/examples/blob/0a076922ef8389d895c9af846f9e5bc7d339fe03/forms/multipart-s3/src/main.rs#L67

tao-guo:
Thanks for the information！I will close this pr.

PR #3737, opened 2025-08-23
fix(actix-web): improve streaming response behavior for Content-Length and Content-Type

### LINKED ISSUE #2306: [awc] HTTP client ignores Content-Length if Body is stream.

JohnTitor:
## PR Type

Bug Fix

## PR Checklist

## Overview

This is a rebase of https://github.com/actix/actix-web/pull/3015/ as I cannot push additional commits.

- Set default Content-Type to application/octet-stream for streaming responses
- Respect Content-Length header by automatically calling `no_chunking()` when set
- Resolves issue where Content-Length was ignored in streaming responses

Fixes #2306, closes #3015

### DIFF
--- a/actix-web/src/response/builder.rs
+++ b/actix-web/src/response/builder.rs
@@ -318,12 +318,33 @@ impl HttpResponseBuilder {
     /// Set a streaming body and build the `HttpResponse`.
     ///
     /// `HttpResponseBuilder` can not be used after this call.
+    ///
+    /// If `Content-Type` is not set, then it is automatically set to `application/octet-stream`.
+    ///
+    /// If `Content-Length` is set, then [`no_chunking()`](Self::no_chunking) is automatically called.
     #[inline]
     pub fn streaming<S, E>(&mut self, stream: S) -> HttpResponse
     where
         S: Stream<Item = Result<Bytes, E>> + 'static,
         E: Into<BoxError> + 'static,
     {
+        // Set mime type to application/octet-stream if it is not set
+        if let Some(parts) = self.inner() {
+            if !parts.headers.contains_key(header::CONTENT_TYPE) {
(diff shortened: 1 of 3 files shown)


actix-web/src/response/builder.rs
```diff
@@ -318,12 +318,33 @@ impl HttpResponseBuilder {
...
+
+        if let Some(parts) = self.inner() {
+            if let Some(length) = parts.headers.get(header::CONTENT_LENGTH) {
+                if let Ok(length) = length.to_str() {
+                    if let Ok(length) = length.parse::<u64>() {
+                        self.no_chunking(length);
+                    }
+                }
+            }
+        }
```
robjtede, line 346: this won't do anything due to https://github.com/actix/actix-web/blob/8996198f2c922e7c052a85d69022515440cac66d/actix-http/src/h1/encoder.rs#L147

we should encourage using SizedStream if the size is known, also plays better with body-altering middleware like compression

JohnTitor: Sorry for the late reply! (Actually I didn't notice you comment, sorry.)

I might be wrong but `skip_len` would be false when no_chunking?
https://github.com/actix/actix-web/blob/8996198f2c922e7c052a85d69022515440cac66d/actix-http/src/h1/encoder.rs#L94-L104

But in general I agree with using SizedStream, I'll submit a follow-up patch!

PR #3757, opened 2025-09-04
permanent redirect status code instead of found

### LINKED ISSUE #3543: Files::redirect_to_slash_directory() should use HTTP status code 308
## Expected Behavior

When serving files with `actix_files::Files::new().redirect_to_slash_directory()` the redirect response to the client should have status code 308 (permanent redirect) or 301 (moved permanently).

## Current Behavior

Currently the response has status code 302 (found).

The culprit is this `HttpResponse::Found()`
https://github.com/actix/actix-web/blob/5aeb0dd950214d1cf0f4ac2045b8a8116d89678b/actix-files/src/service.rs#L142-L153

## Context

302 is the wrong status in this case. [MDN says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) that search engines ignore temporary redirects. In this case the redirect is permanent, because we want all requests forever and always to use a trailing slash.

## Your Environment

- Rust Version 1.83.0
- Actix Web Version: 4.9.0
- Firefox Version 133

francesco-gaglione:
## PR Type

Close #3543 

## PR Checklist

### DIFF
--- a/actix-files/src/lib.rs
+++ b/actix-files/src/lib.rs
@@ -736,7 +736,7 @@ mod tests {
         .await;
         let req = TestRequest::with_uri("/tests").to_request();
         let resp = test::call_service(&srv, req).await;
-        assert_eq!(resp.status(), StatusCode::PERMANENT_REDIRECT);
+        assert_eq!(resp.status(), StatusCode::TEMPORARY_REDIRECT);
 
         // should redirect if files listing is enabled
         let srv = test::init_service(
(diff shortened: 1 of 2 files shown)


robjtede:
I agree Found is less appropriate but permanent redirects are a risky default for a framework due to their caching implications.

I'd accept a change to TemporaryRedirect or allow a customization.

francesco-gaglione:
@robjtede if customization is prefered I can try to work on it. I wa thinking to change `redirect_to_slash` bolean to an Option containint the redirect mode. What do you think?
I can work on it on a different PR if prefered leaving here the TemporaryRedirect

robjtede:
> change redirect_to_slash

a new method would be preferred to avoid breaking changes, but in spirit i agree it's a good path to take

francesco-gaglione:
@robjtede working on it. This merge request will be obsolete and will be included in the new one.

francesco-gaglione:
Opened #3767

PR #3767, opened 2025-09-11
Add support for permanent redirects in `Files` service

### LINKED ISSUE #3543: Files::redirect_to_slash_directory() should use HTTP status code 308

francesco-gaglione:
## PR Type

Feature

## PR Checklist

## Overview

Current behavior redirect only with `FOUND`.
This pull request:
- Change default redirect to temporary redirect
- Add support for permanent redirects in `Files` service

Close #3543

### DIFF
--- a/actix-files/src/files.rs
+++ b/actix-files/src/files.rs
@@ -150,7 +150,7 @@ impl Files {
     /// Redirect with permanent redirect status code
     ///
     /// By default redirect with temporary redirect status code
-    pub fn with_permanent_redirect_directory(mut self) -> Self {
+    pub fn with_permanent_redirect(mut self) -> Self {
         self.with_permanent_redirect = true;
         self
     }
--- a/actix-files/src/files.rs
+++ b/actix-files/src/files.rs
@@ -41,6 +41,7 @@ pub struct Files {
     index: Option<String>,
     show_index: bool,
     redirect_to_slash: bool,
+    with_permanent_redirect: bool,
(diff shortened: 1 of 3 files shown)


actix-files/src/files.rs
```diff
@@ -144,6 +147,14 @@ impl Files {
         self
     }
 
+    /// Redirect with permanent redirect status code
+    ///
+    /// By default redirect with temporary redirect status code
```
JohnTitor:
```suggestion
    /// Redirect with permanent redirect status code (308).
    ///
    /// By default redirect with temporary redirect status code (307).
```

JohnTitor (approved):

robjtede (approved):
looks pretty good, thanks 👍🏻

PR #3779, opened 2025-09-26
add function single generic type support

### LINKED ISSUE #2866: Cannot infer type for type parameter when defining handler with #[get("/")] and generics
## Expected Behavior
Pass type checking

## Current Behavior
```console
  --> examples/cannnot-compile.rs:28:10
   |
28 | async fn index<T: UserRepository>(client: web::Data<T>) -> impl Responder {
   |          ^^^^^ cannot infer type for type parameter `T` declared on the function `index`

Some errors have detailed explanations: E0107, E0282.
For more information about an error, try `rustc --explain E0107`.
error: could not compile `my-project` due to 2 previous errors
```

## Possible Solution
Use `.route("/", web::get().to(index::<T>)` instead of `service(index::<T>)` and `get` method macro, but is there any good way to keep using `get` macro?

```rust
use actix_web::{
    dev::Server,
    get,
    web::{self, Data},
    App, HttpServer, Responder,
};
use serde::Serialize;

#[derive(Debug, Serialize, Clone, Copy)]
pub struct User {
    id: u64,
}

pub trait UserRepository {
    fn get_user(&self) -> User;
}

#[derive(Clone)]
struct UserClient;

impl UserRepository for UserClient {
    fn get_user(&self) -> User {
        User { id: 0 }
    }
}

// when uncommenting following the line, the type checking is unaccepted
// because of cannot infer type parameter T
// #[get("/")]
async fn index<T: UserRepository>(client: web::Data<T>) -> impl Responder {
    let user = client.into_inner().get_user();
    web::Json(user)
}

pub fn create_server<T: UserRepository + Send + Sync + 'static + Clone>(
    search: T,
) -> Result<Server, std::io::Error> {
    let server = HttpServer::new(move || {
        App::new()
            .app_data(Data::new(search.clone()))
            .route("/", web::get().to(index::<T>))
        // .service(index::<T>)
    })
    .bind("127.0.0.1:8080")?
    .run();
    Ok(server)
}

#[actix_web::main]
    let user_client = UserClient;
    create_server(user_client).unwrap().await
}

```

## Your Environment

- Rust Version rustc 1.62.0 (a8314ef7d 2022-06-27)
- Actix Web Version: 4.0.1

uraneko:
## PR Type

Other(improvement) 

## PR Checklist

## Overview

Closes #2866


JohnTitor:
Thanks for the PR! I've tweaked to support more generics and have better test suite.
But the current implementation has _technically_ breaking change around Default impl although I think the current is one of good form to land.
If you have a good idea here, I'm happy to hear and see it :)

uraneko:
Thanks for the tweaks, I think the current pr state looks pretty good.

> But the current implementation has technically breaking change around Default impl

I'm not very familiar with the code base, so I may be missing something obvious, but, what's the technically breaking change? 

My understanding is that `app.route(...)` can still be used with the same syntax, while the method macro + `app.service(...)` pattern is now extended to support generics by registering `service_function::<Generics>::default()`. Or is the breaking change around something else?

Could you show me an example user code that would break due to this change?

By the way. should I add a function to the test file to clarify that lifetimes are also supported? Something like this:

```rust
// actix-web-codegen/tests/trybuild/route-generic-ok.rs

- trait UserRepository {                 
-     fn get_user(&self) -> u64;
- }                                      

- impl UserRepository for UserClient { 
-     fn get_user(&self) -> u64 {     
-         99
-     }
- }

+ trait UserRepository {
+    fn get_user(&self) -> &str;
+ }

+ impl UserRepository for UserClient { 
+     fn get_user(&self) -> &str {     
+        "99"                         
+     }                                
+ }

//...

+ #[get("/lifetime")]                                                      
+ async fn with_lifetime<'a, T>(client: web::Data<&'a T>) -> &'a str       
+ where                                                            
+    T: UserRepository + Send + Sync + 'static,                   
+ {                                                                
+    client.get_ref().get_user()                                  
+ }                                                              

//...

+ .app_data(web::Data::new(&UserClient))
+ .service(with_lifetime::<UserClient>::default())
```

PR #3811, opened 2025-10-27
test: disable docsrs test for now

JohnTitor:
## PR Type

PR_TYPE

## PR Checklist


JohnTitor:
Ok, turned out we have to wait for crypto crates to update their dependencies and ship a stable release to drop the `generic-array v0.14` dependency. The test does nothing anyway so disabling cfg would be a reasonable stopgap, I think.

PR #3826, opened 2025-11-10
chore: Add public export for `EitherExtractError`

### LINKED ISSUE #3541: Transform the Error of a FromRequest of an Either is now impossible
Creating a FromRequest for a type that uses a FromRequest of an Either under the hood is now impossible to the opaque error type of the EitherExtractFut.

Migrating the following code post 4.0.0 seems impossible:

```
pub enum Permission {... }
pub struct Permissions(HashSet<Permission>);
pub type AuthInfo = Either<Auth0AuthData, ApiKeyAuthData>;

impl FromRequest for Auth0AuthData { ... }
impl FromRequest for ApiKeyAuthData { ... }


impl FromRequest for Permissions {
  type Error = PermissionError;
  type Future = ResponseFuture<Result<Permissions, PermissionError>>;

  fn from_request(req: &HttpRequest, _payload: &mut Payload) -> Self::Future {
    let auth_info_fut = AuthInfo::extract(req);

    let f = async move {
      let auth_info = auth_info_fut.await.map_err(|e| match e {
        EitherExtractError::Bytes(e) => PermissionError::Other(e.to_string()),
        EitherExtractError::Extract(PermissionError::NoCredentials, r) => r,
        EitherExtractError::Extract(l, _) => l,
      )?;

      match auth_info {
        Either::Left(auth0) => auth0.permissions(&config_addr).await,
        Either::Right(ApiKey { permissions, .. }) => Ok(Permissions(permissions)),
      }
    };

    Box::pin(f)
  }
}

```

## Expected Behavior

EitherExtractError should be accessible

## Current Behavior

It's opaque

## Possible Solution

Revert the removal from the public api

## Your Environment

- Actix Web Version: 4.0.0-beta-10

Ruchir28:
## PR Type

Refactor

## PR Checklist

- ~~Tests for the changes have been added / updated.~~
- ~~Documentation comments have been added / updated.~~
## Overview

Exports `EitherExtractError` in the `error` module, making it accessible 
for users who need to handle errors when working with `Either<A, B>` extractors

Closes #3541

### DIFF
--- a/actix-web/src/lib.rs
+++ b/actix-web/src/lib.rs
@@ -121,7 +121,7 @@ pub use crate::{
     route::Route,
     scope::Scope,
     server::HttpServer,
-    types::{Either, EitherExtractError},
+    types::Either,
 };
 
 macro_rules! codegen_reexport {
--- a/actix-web/src/lib.rs
+++ b/actix-web/src/lib.rs
@@ -121,7 +121,7 @@ pub use crate::{
     route::Route,
     scope::Scope,
     server::HttpServer,
-    types::Either,
+    types::{Either, EitherExtractError},
 };
 
 macro_rules! codegen_reexport {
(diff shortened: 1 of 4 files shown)


actix-web/src/lib.rs
```diff
@@ -121,7 +121,7 @@ pub use crate::{
     route::Route,
     scope::Scope,
     server::HttpServer,
-    types::Either,
+    types::{Either, EitherExtractError},
```
robjtede: i don't want to export this as the top level, imo should either come under `dev` or `error`

JohnTitor: +1, I'd go with `error`.

Ruchir28: Updated

robjtede (approved):

PR #3936, opened 2026-02-17
fix(http): drain unread body if chunked and dropped payload

### LINKED ISSUE #2972: Sockets left open and client get stuck, when PUT returns error
## Expected Behavior
Obviously server should run smoothly, no sockets should be left in ESTABLISHED or CLOSE_WAIT and client should not get stuck.

## Current Behavior
Sockets are left in  ESTABLISHED or CLOSE_WAIT and clients get stuck.

## Possible Solution
No clue

### Server:
`Cargo.toml`:
```toml
[dependencies]
anyhow = "1"
actix-rt = "2.8"
actix-web = "4.3"
```
`main.rs`:
```rs
use actix_web::{error::ErrorConflict, web, App, HttpResponse, HttpServer};
use anyhow;

async fn upload() -> Result<HttpResponse, actix_web::Error> {
    Err(ErrorConflict("conflict"))
}

#[actix_rt::main]
async fn main() -> anyhow::Result<()> {
    HttpServer::new(move || {
        App::new().service(web::resource("/upload/{name}").route(web::put().to(upload)))
    })
    .bind("127.0.0.1:8000")?
    .run()
    .await?;

    Ok(())
}
```

### Client
`client.py`:
```py
import requests

def chunks():
    yield bytes(1024*1024*b'a')
    yield bytes(1024*1024*b'b')
    yield bytes(1024*1024*b'c')
    yield bytes(1024*1024*b'd')

resp = requests.put(f"http://localhost:8000/upload/test.txt", data=chunks())
print(repr(resp))
```

### Steps:
- Start the server
- Run `client.py` few times.

### Aftermath:
After running `client.py` few times lsof shows for example
```
> lsof -i -a -p 17671 -n -P 
COMMAND   PID USER   FD   TYPE  DEVICE SIZE/OFF NODE NAME
rs      17671 etam    9u  IPv4 2823931      0t0  TCP 127.0.0.1:8000 (LISTEN)
rs      17671 etam   21u  IPv4 2823936      0t0  TCP 127.0.0.1:8000->127.0.0.1:34484 (CLOSE_WAIT)
rs      17671 etam   23u  IPv4 2827508      0t0  TCP 127.0.0.1:8000->127.0.0.1:36270 (CLOSE_WAIT)
rs      17671 etam   24u  IPv4 2827545      0t0  TCP 127.0.0.1:8000->127.0.0.1:39514 (CLOSE_WAIT)
rs      17671 etam   25u  IPv4 2827585      0t0  TCP 127.0.0.1:8000->127.0.0.1:39516 (CLOSE_WAIT)
rs      17671 etam   26u  IPv4 2825173      0t0  TCP 127.0.0.1:8000->127.0.0.1:48344 (CLOSE_WAIT)
rs      17671 etam   27u  IPv4 2825215      0t0  TCP 127.0.0.1:8000->127.0.0.1:48358 (ESTABLISHED)
rs      17671 etam   28u  IPv4 2829316      0t0  TCP 127.0.0.1:8000->127.0.0.1:43466 (ESTABLISHED)
rs      17671 etam   29u  IPv4 2826652      0t0  TCP 127.0.0.1:8000->127.0.0.1:43474 (ESTABLISHED)
rs      17671 etam   30u  IPv4 2826663      0t0  TCP 127.0.0.1:8000->127.0.0.1:43482 (ESTABLISHED)
```

and `client.py` instead of printing `<Response [409]>` might get stuck, where it has to be killed with Ctrl+C:
```
Traceback (most recent call last):
  File "/home/etam/tmp/./upload.py", line 11, in <module>
    resp = requests.put(f"http://localhost:8000/upload/test.txt", data=chunks())
  File "/usr/lib/python3.10/site-packages/requests/api.py", line 130, in put
    return request("put", url, data=data, **kwargs)
  File "/usr/lib/python3.10/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python3.10/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3.10/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python3.10/site-packages/requests/adapters.py", line 526, in send
    low_conn.send(i)
  File "/usr/lib64/python3.10/http/client.py", line 998, in send
    self.sock.sendall(data)
KeyboardInterrupt
```

## Context
It's a simple service for uploading files. But sometimes we want to return `ErrorConflict`.

## Your Environment
- Rust Version: 1.67.0
- Actix Web Version: 4.3.0
- python3-requests: 2.28.2

JohnTitor:
## PR Type

PR_TYPE

## PR Checklist

## Overview

Fix #2972 

The problem here is that our dispatcher doesn't clean up socket when payload is dropped.
For chunked requests, we should be able to drain payloads safely if it's dropped.

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -8,11 +8,13 @@
 - Fix truncated body ending without error when connection closed abnormally. [#3067]
 - Add config/method for `TCP_NODELAY`. [#3918]
 - Do not compress 206 Partial Content responses. [#3191]
+- Fix lingering sockets and client stalls when responding early to dropped chunked request payloads. [#2972]
 
 [#3638]: https://github.com/actix/actix-web/issues/3638
 [#3067]: https://github.com/actix/actix-web/pull/3067
 [#3918]: https://github.com/actix/actix-web/pull/3918
 [#3191]: https://github.com/actix/actix-web/issues/3191
+[#2972]: https://github.com/actix/actix-web/issues/2972
 
 ## 3.11.2
 
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -31,7 +31,7 @@ use crate::{
     config::ServiceConfig,
     error::{DispatchError, ParseError, PayloadError},
     service::HttpFlow,
-    Error, Extensions, OnConnectData, Request, Response, StatusCode,
+    Error, Extensions, HttpMessage, OnConnectData, Request, Response, StatusCode,
 };
 
 const LW_BUFFER_SIZE: usize = 1024;
@@ -157,6 +157,8 @@ pin_project! {
         pub(super) state: State<S, B, X>,
         // when Some(_) dispatcher is in state of receiving request payload
         payload: Option<PayloadSender>,
+        // true when current request uses chunked transfer encoding (drainable when payload is dropped)
+        payload_drainable: bool,
         messages: VecDeque<DispatcherMessage>,
 
         head_timer: TimerState,
@@ -269,6 +271,7 @@ where
 
                     state: State::None,
                     payload: None,
+                    payload_drainable: false,
                     messages: VecDeque::new(),
 
                     head_timer: TimerState::new(config.client_request_deadline().is_some()),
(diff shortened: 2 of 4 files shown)


asonix:
Is this the correct solution to the problem? 
I would have expected that https://github.com/actix/actix-web/pull/3721 solved the initial problem of sockets not getting cleaned, and I'm not sure if reading and discarding payloads indefinitely is a good choice

JohnTitor:
@asonix 
IIUC #3721 fixed a case by shutting down when a response completes with unread payload, but shutdown without draining can still stall sender-side clients. For example, when the client keeps writing chunked data and doesn’t read the response yet, which should be the case of #2972.
I agree we should avoid unbounded draining, though. I need to refine that.

Issue #2993, opened 2023-03-07
Unexpected middleware chaining behaviour - needs explanation / documentation


brecht-derooms:
## Expected Behavior
The actix documentation currently mentions:
"Middleware is registered for each App, scope, or Resource and executed in the opposite order as registration"
When I read that phrase I would expect the following to work and call the 'bypass auth middleware' first for the authentication endpoints before the CookieAuthMiddleware
```
wrap(auth_middleware::middleware::CookieAuthMiddlewareFactory::new())
            .app_data(app_state.clone())
            .service(web::scope("api").configure(controller::api::conf))
            .service(
                web::scope("oauth2")
                    .wrap(auth_middleware::middleware::BypassAuthMiddlewareFactory::new())
                    .configure(auth_middleware::controller::oauth2::conf),
```

## Current Behavior
Currently, it seems that middleware defined within a scope is not called before the 'global' middleware although I consider that later in the 'order of registration'. Maybe the registration in a scope is delayed or something else is amiss. It probably is worthwhile to complete the documentation with more advanced usages of middleware and how multiple middleware would interact with each other in different configurations. 

## More context
The equivalent working version of what I intended above is: 
```
service(
                web::scope("api")
                    .wrap(auth_middleware::middleware::CookieAuthMiddlewareFactory::new())
                    .configure(controller::api::conf),
            )
            .service(
                web::scope("oauth2")
                    .wrap(auth_middleware::middleware::CookieAuthMiddlewareFactory::new())
                    .wrap(auth_middleware::middleware::BypassAuthMiddlewareFactory::new())
                    .configure(auth_middleware::controller::oauth2::conf),
```

However, what I really want to achieve here is to have all endpoints protected by default (secure by default) and allow people to override / bypass authentication only in certain cases since I personally have better experiences by securing everything by default and opening up things than having everything open by default and securing each endpoint or group of endpoints separately, it's easy to forget such a line with rather bad effects.

PR #3947, opened 2026-02-22
docs: clarify cross-level middleware ordering between App and Scope

ashishkaushik05:
## Summary

Fixes the documentation gap described in #2993.

The existing docs state:
> Middleware is registered for each App, Scope, or Resource and executed in opposite order as registration.

This is accurate _within_ a single level, but misleading when middleware is registered at both the `App` level and a `Scope` level. Users reasonably expect a scope-level middleware to run before an app-level one (since it appears "later in registration"), but that is not how it works.

## Root Cause

App-level middleware structurally wraps the entire request router. A `Scope` is a service registered _inside_ that router. This means:
- App MW is always the outermost layer
- Routing to a scope happens _inside_ App MW
- Scope MW cannot bypass or run before App MW

This is by design and correct — but it was undocumented.

## Changes

**`actix-web/src/middleware/mod.rs`**
- Qualify the opening "opposite order" statement to note it applies within a single level only
- Add a new **"Cross-Level Middleware Ordering"** section with:
  - Prose explanation of why App MW always runs first
  - ASCII diagram showing the full nesting: App MW → Router → Scope MW → Resource MW → Handler
  - Explicit note that scope MW cannot bypass App MW
  - Code example demonstrating the recommended per-scope pattern

**`actix-web/src/app.rs`**
- Add a **"# Middleware Ordering"** note to `App::wrap`'s doc comment pointing to the new section

## What this does NOT change

- No behavior change — documentation only
- Does not touch `Scope::wrap`, `Resource::wrap`, or `authors-guide.md` (keeping PR scope focused)

### DIFF
--- a/actix-web/src/middleware/mod.rs
+++ b/actix-web/src/middleware/mod.rs
@@ -15,7 +15,10 @@
 //! - Access external services (e.g., [sessions](https://docs.rs/actix-session), etc.)
 //!
 //! Middleware is registered for each [`App`], [`Scope`](crate::Scope), or
-//! [`Resource`](crate::Resource) and executed in opposite order as registration.
+//! [`Resource`](crate::Resource) and executed in opposite order as registration. Note that this
+//! reverse ordering applies _within_ a single level (`App`, `Scope`, or `Resource`). When
+//! middleware exists at multiple levels, App-level middleware always runs first. See
+//! [Cross-Level Middleware Ordering](#cross-level-middleware-ordering) for details.
 //!
 //! # Simple Middleware
 //!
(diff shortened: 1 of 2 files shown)


actix-web/src/middleware/mod.rs
```diff
@@ -120,6 +123,62 @@
...
+//!
+//! ```
+//! use actix_web::{web, middleware, App};
+//!
+//! // Instead of applying auth middleware at the App level and trying
+//! // to bypass it on certain scopes, apply it per-scope:
+//! let app = App::new()
+//!     .service(
+//!         web::scope("/api")
+//!             .wrap(middleware::DefaultHeaders::new()) // e.g., auth middleware
```
JohnTitor, line 172: This is not auth tho?

actix-web/src/middleware/mod.rs
```diff
@@ -120,6 +123,62 @@
 //! middleware are run in reverse order for [`App`] _and then_ in reverse order for the wrapped
 //! service.
 //!
+//! # Cross-Level Middleware Ordering
+//!
+//! When middleware is registered on _both_ an [`App`] and a [`Scope`](crate::Scope) (or
+//! [`Resource`](crate::Resource)), the App-level middleware **always** runs before the
```
JohnTitor, line 129: I think this isn't correct for outbounds.

JohnTitor:
Have you used coding agent or something else on this PR? If so, it's totally ok to use but I'd recommend reading and understanding diff before submitting a PR.

ashishkaushik05:
Thanks for pointing that out.

You're right — that statement only applies to the inbound request flow. For outbound responses the order is reversed (resource/scope before app). I overlooked that while reviewing the change.

I'll update the documentation to clarify the distinction between inbound and outbound middleware order.

JohnTitor:
Sigh, answering using GenAI is quite rude, we do not accept such a contribution.

PR #3953, opened 2026-02-23
awc: add camel_case headers api

### LINKED ISSUE #3954: WebSocket camel-case header are incorrect
Websocket http headers are incorrect in camel-case

## Expected Behavior

In some devices such as Embedded devices like `IP Camera`, they have an WebServer that only accept `camel-case` header format

for examples following header will reject immediately
```
sec-websocket-version
sec-websocket-key
```

## Current Behavior

I already patch support camel-case api for websocket in `awc` but in `write_camel_case` function the output will be 
```
Sec-Websocket-Version
Sec-Websocket-Key
```

which is incorrect an will reject from server

This is my [PR](https://github.com/actix/actix-web/pull/3944) for support camel-case in websocket

i already verify correct camel-case headers for websocket with python and postman it should be like this
```
Sec-WebSocket-Version
Sec-WebSocket-Key
```

## Possible Solution

Easiest solution is to add detect `web` word and make uppercase next letter after it
Second approach is to detect exactly websocket headers and write correct camel-case header

Connect with websocket to an WebServer that not support header case-insensitive for example an IP Camera (MC-
L12B)
## Your Environment

```
rustc 1.93.0 (254b59607 2026-01-19)
```
- awc Version: 3.8.2

Ali-Mirghasemi:
## PR Type

Feature

## PR Checklist

## Overview

I add `camel_case_headers` and `set_camel_case_headers` to `awc` websocket to provice support header camel case
also fix `Connection: Upgrade` camel case in `actix-http`

Close #3954

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Encode the HTTP/1 `Connection: Upgrade` header in Camel-Case when camel-case header formatting is enabled.[#3953]
+
+[#3953]: https://github.com/actix/actix-web/pull/3953
+
 ## 3.12.0
 
 - Minimum supported Rust version (MSRV) is now 1.88.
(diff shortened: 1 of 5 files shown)


JohnTitor (approved):
Thanks!

PR #3970, opened 2026-03-10
fix: yield lines correctly in readlines across split chunks

### LINKED ISSUE #3968: `Readlines` does not enforce per-line limits across payload chunks and can return split lines
`actix_web::dev::Readlines` mishandles a logical line when the newline arrives in a later payload chunk.

In `actix-web/src/types/readlines.rs`, the newline-found branch for freshly received bytes checks only `ind + 1 > this.limit` and then returns only `bytes.split_to(ind + 1)`. It does not include the already buffered prefix from `this.buf`.

That creates two user-visible problems:

1. The configured per-line limit is not enforced across chunk boundaries.
2. A single logical line can be split into two separate yielded lines.

When the first chunk contains a partial line and the second chunk contains the terminating `\n`, the current implementation treats only the second chunk as "the line" for both limit checking and output construction.

## Current Behavior

For example, with a limit of `10`:

- first chunk: `AAAAAAAAAA`
- second chunk: `A\n`

The logical line is `AAAAAAAAAAA\n` (12 bytes), so it should be rejected with `ReadlinesError::LimitOverflow`.

Instead, the current code accepts the second chunk because `ind + 1 == 2`, returns `"A\n"`, and leaves the original ten-byte prefix buffered for a later poll.

That means callers can observe corrupted line boundaries and oversized logical lines can bypass the configured line limit.

## Expected Behavior

When a newline is found in a newly received chunk, `Readlines` should:

- compute the effective line length as `this.buf.len() + ind + 1`
- return `LimitOverflow` if that combined length exceeds the configured limit
- otherwise return the full logical line assembled from the buffered prefix plus the newline-terminated suffix from the current chunk

## Possible Solution

In the newline-found branch for `bytes`:

- check `this.buf.len() + ind + 1 > this.limit`
- prepend `this.buf` to the returned line before decoding
- leave only the bytes after the newline buffered for future polls

Adding the following unit test to `actix-web/src/types/readlines.rs`:

```rust
#[test]
fn test_readlines_limit_across_chunks() {
    use std::{
        pin::Pin,
        task::{Context, Poll},
    };

    use actix_http::{h1, Request};
    use futures_util::task::noop_waker_ref;

    use crate::error::ReadlinesError;

    let (mut sender, payload) = h1::Payload::create(false);
    let payload: actix_http::Payload = payload.into();
    let mut req = Request::with_payload(payload);
    let mut stream = Readlines::new(&mut req).limit(10);
    let mut cx = Context::from_waker(noop_waker_ref());

    sender.feed_data(Bytes::from_static(b"AAAAAAAAAA"));
    assert!(matches!(
        Pin::new(&mut stream).poll_next(&mut cx),
        Poll::Pending
    ));

    sender.feed_data(Bytes::from_static(b"A\n"));
    assert!(matches!(
        Pin::new(&mut stream).poll_next(&mut cx),
        Poll::Ready(Some(Err(ReadlinesError::LimitOverflow)))
    ));
}
```

```text
cargo test -p actix-web --lib types::readlines::tests::test_readlines_limit_across_chunks -- --exact
```

Observed result:

```text
test types::readlines::tests::test_readlines_limit_across_chunks ... FAILED

thread 'types::readlines::tests::test_readlines_limit_across_chunks' panicked at
actix-web/src/types/readlines.rs:...:
assertion failed: matches!(Pin::new(&mut stream).poll_next(&mut cx),
    Poll::Ready(Some(Err(ReadlinesError::LimitOverflow))))
```

robjtede:
## PR Type

Fix

## PR Checklist

## Overview

Fixes #3968

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -4,6 +4,7 @@
 
 - Panic when calling `Route::to()` or `Route::service()` after `Route::wrap()` to prevent silently dropping route middleware. [#3944]
 - Fix `HttpRequest::{match_pattern,match_name}` reporting path-only matches when route guards disambiguate overlapping resources. [#3346]
+- Fix `Readlines` handling of lines split across payload chunks so combined line limits are enforced and complete lines are yielded.
 
 [#3944]: https://github.com/actix/actix-web/pull/3944
 [#3346]: https://github.com/actix/actix-web/issues/3346
--- a/actix-web/src/types/readlines.rs
+++ b/actix-web/src/types/readlines.rs
@@ -65,6 +65,23 @@ where
             err: Some(err),
         }
     }
+
+    /// Decodes one complete logical line using the request's configured encoding.
+    ///
+    /// Callers are expected to pass only the bytes that belong to the line being yielded,
+    /// whether they came from the internal buffer, the current payload chunk, or both.
+    fn decode(encoding: &'static Encoding, bytes: &[u8]) -> Result<String, ReadlinesError> {
+        if encoding == UTF_8 {
+            str::from_utf8(bytes)
+                .map_err(|_| ReadlinesError::EncodingError)
+                .map(str::to_owned)
+        } else {
+            encoding
+                .decode_without_bom_handling_and_without_replacement(bytes)
+                .map(Cow::into_owned)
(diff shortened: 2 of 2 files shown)


JohnTitor (approved):
LGTM!

PR #3971, opened 2026-03-10
fix: correct HeaderMap iterators' len and size_hint for multi-value

### LINKED ISSUE #3969: `HeaderMap` iterators violate `ExactSizeIterator` for multi-value headers and panic on `.len()`
`actix_http::header::HeaderMap` marks `Iter`, `IntoIter`, and `Drain` as `ExactSizeIterator`, but those iterators flatten multi-value headers while their `size_hint()` only reflects the number of unique header keys in the backing hash map.

This violates the `ExactSizeIterator` contract for any header name with multiple values, such as `Set-Cookie`.

In practice, this is worse than just returning the wrong length: calling `.len()` on these iterators panics in the standard library because `size_hint()` returns `(lower_bound, None)` even though the iterator claims to be exact.

## Why This Is a Bug

For a map containing:

- `Set-Cookie: a=1`
- `Set-Cookie: b=2`
- `Set-Cookie: c=3`

the iterators yield **3** `(name, value)` pairs.

However:

- `Iter::size_hint()` returns `(1, None)`
- `IntoIter::size_hint()` returns `(1, None)`
- `Drain::size_hint()` returns `(1, None)`

because they delegate to the inner hash map's notion of size, which counts unique keys, not flattened pairs.

Since all three iterators also implement `ExactSizeIterator`, `.len()` assumes an exact upper bound exists and panics.

## Expected Behavior

One of the following should be true:

1. The iterators return an exact flattened remaining item count and `size_hint()` returns `(n, Some(n))`, or
2. They do not implement `ExactSizeIterator`.

Current behavior satisfies neither.

## Possible Solution

Either:

- remove `ExactSizeIterator` from `Iter`, `IntoIter`, and `Drain`, or
- track the flattened remaining length across multi-value entries and return an exact size hint

Given the current implementation structure, removing `ExactSizeIterator` is likely the safer minimal fix.

Add the following unit tests to `actix-http/src/header/map.rs`:

```rust
#[test]
fn iter_len_counts_values() {
    let mut map = HeaderMap::new();
    map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));

    assert_eq!(map.iter().count(), 3);
    assert_eq!(map.iter().len(), 3);
}

#[test]
fn into_iter_len_counts_values() {
    let mut map = HeaderMap::new();
    map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));

    assert_eq!(map.clone().into_iter().count(), 3);
    assert_eq!(map.into_iter().len(), 3);
}

#[test]
fn drain_len_counts_values() {
    let mut map = HeaderMap::new();
    map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
    map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));

    let mut drained = map.clone();
    assert_eq!(map.drain().count(), 3);
    assert_eq!(drained.drain().len(), 3);
}
```

Commands run locally:

```text
cargo test -p actix-http --lib header::map::tests::iter_len_counts_values -- --exact
cargo test -p actix-http --lib header::map::tests::into_iter_len_counts_values -- --exact
cargo test -p actix-http --lib header::map::tests::drain_len_counts_values -- --exact
```

Observed result for all three:

```text
thread 'header::map::tests::...' panicked at .../core/src/iter/traits/exact_size.rs:122:9:
assertion `left == right` failed
  left: None
 right: Some(1)
```

robjtede:
## PR Type

Fix

## PR Checklist

## Overview

Fixes #3969

### DIFF
--- a/actix-http/src/header/map.rs
+++ b/actix-http/src/header/map.rs
@@ -1160,6 +1160,40 @@ mod tests {
         assert!(vals.next().is_none());
     }
 
+    #[test]
+    fn iter_len_counts_values() {
+        let mut map = HeaderMap::new();
+        map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));
+
+        assert_eq!(map.iter().count(), 3);
+        assert_eq!(map.iter().len(), 3);
+    }
+
+    #[test]
+    fn into_iter_len_counts_values() {
+        let mut map = HeaderMap::new();
+        map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));
+
+        assert_eq!(map.clone().into_iter().count(), 3);
+        assert_eq!(map.into_iter().len(), 3);
+    }
+
+    #[test]
+    fn drain_len_counts_values() {
+        let mut map = HeaderMap::new();
+        map.append(header::SET_COOKIE, HeaderValue::from_static("a=1"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("b=2"));
+        map.append(header::SET_COOKIE, HeaderValue::from_static("c=3"));
+
+        let mut drained = map.clone();
+        assert_eq!(map.drain().count(), 3);
+        assert_eq!(drained.drain().len(), 3);
+    }
+
     fn owned_pair<'a>((name, val): (&'a HeaderName, &'a HeaderValue)) -> (HeaderName, HeaderValue) {
         (name.clone(), val.clone())
     }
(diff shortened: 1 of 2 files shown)


actix-http/src/header/map.rs
```diff
@@ -537,7 +537,7 @@ impl HeaderMap {
     /// assert!(pairs.contains(&(&header::SET_COOKIE, &HeaderValue::from_static("two=2"))));
     /// ```
     pub fn iter(&self) -> Iter<'_> {
-        Iter::new(self.inner.iter())
+        Iter::new(self.inner.iter(), self.len())
```
JohnTitor, line 540: pre-existing but `len()` does a bit more work than usual: https://github.com/actix/actix-web/blob/c86f2309f1fc202a211d7f6df16f31fb659c6ce4/actix-http/src/header/map.rs#L169
I'm not sure how its full scan affects performance, but I guess we could inline the length inside map 🤔

JohnTitor (approved):
LGTM with or without addressing my comment.

actix-http/src/header/map.rs
robjtede: ye i was thinking about that too, good optimisation target for another PR

Issue #2357, opened 2021-08-07
Hang or protocol error when sending early response to a POST request with delayed body


Riateche:
## Expected Behavior

This is a minimal example of a client that reproduces the problem:
```
[dependencies]
tokio = { version = "1.0", features = ["full"] }
reqwest = { version = "0.11", features = ["stream"] }
futures = "0.3.16"
```
```
use futures::stream;
use std::time::Duration;
use std::{env, io};
use tokio::time::sleep;

#[tokio::main]
async fn main() {
    let client = reqwest::Client::new();
    loop {
        let stream = stream::unfold(0u32, |state| async move {
            if state == 0 {
                sleep(Duration::from_secs(1)).await;
                println!("yield data");
                Some((io::Result::Ok(&b"abc"[..]), state + 1))
            } else {
                println!("yield none");
                None
            }
        });
        let response = client
            .post(env::args().nth(1).unwrap())
            .body(reqwest::Body::wrap_stream(stream))
            .send()
            .await
            .unwrap();
        println!("status: {:?}", response.status());
        let bytes = response.bytes().await.unwrap();
        println!("body: {:?}", bytes);
        println!("========");
        sleep(Duration::from_secs(3)).await;
    }
}
```
The client sends a POST request with a body, but the data for the body is not available right away. If the server does not respond with an error right away, the client will send the body as it becomes available, then the server will return the response:
```
cargo run -- https://httpbin.org/post 
yield data
yield none
status: 200
body: b"{\n  \"args\": {}, \n  \"data\": \"abc\", \n  \"files\": {}, \n  \"form\": {}, \n  \"headers\": {\n    \"Accept\": \"*/*\", \n    \"Host\": \"httpbin.org\", \n    \"Transfer-Encoding\": \"chunked\", \n    \"X-Amzn-Trace-Id\": \"Root=1-610f0968-7e9f910541c7bc430ad6928e\"\n  }, \n  \"json\": null, \n  \"origin\": \"94.72.10.210\", \n  \"url\": \"https://httpbin.org/post\"\n}\n"
======== (repeats forever)
``` 
The server can also return an error right away, in which case the client doesn't poll the body stream at all:
```
cargo run -- https://httpbin.org/404 
status: 404
[...]

PR #3972, opened 2026-03-10
implement h1 early-response linger

### LINKED ISSUE #3967: early response during upload breaks http semantics and leads to nginx 502s
## Expected Behavior

When an Actix server sends a final response before consuming the entire request body, it should either continue reading and discarding the rest of the body, or close the connection gracefully after sending the response, so that proxies like nginx can still forward the application response reliably instead of returning `502 Bad Gateway`.

## Current Behavior

With nginx request buffering enabled, a large `POST` to an Actix-based upstream that responds before reading the full body can fail with:

- nginx response: `502 Bad Gateway`
- nginx error log: `writev() failed (32: Broken pipe) while sending request to upstream`

This means the application-level response is lost and replaced by a proxy error.

Sending a final response before consuming the request body is allowed by HTTP/1.1, but if the server does not continue reading the body then it needs to close the connection only after sending the response in a way that does not cause the response to be lost in transit.

[RFC 9112, §9.3](https://www.rfc-editor.org/rfc/rfc9112#section-9.3) says:

> A server MUST read the entire request message body or close the connection **after** sending its response

- related nginx discussion (2013): https://mailman.nginx.org/pipermail/nginx/2013-September/040472.html

## Possible Solution

Avoid an immediate close while the sender is still uploading the request body. Either keep reading and discarding the unread request body, or perform a graceful/lingering close after sending the final response.

## Steps to Reproduce

```rs
use std::{convert::Infallible, io};

use actix_http::{HttpService, Request};
use actix_server::Server;

async fn handler(_req: Request) -> Result<&'static str, Infallible> {
    Ok("ok")
}

#[actix_rt::main]
    let f = || HttpService::build().finish(handler).tcp();
    Server::build().bind("r", ("0.0.0.0", 8080), f)?.run().await
}
```

```nginx
worker_processes 1;
events { worker_connections 1024; }
http {
    resolver 127.0.0.11;

    proxy_request_buffering on;
    client_body_buffer_size 10m;
    client_max_body_size 5G;

    server {
        listen 80;
        location = /111 { set $u http://v3111:8080; proxy_http_version 1.1; proxy_pass $u; }
    }
}
```

A minimal Linux/docker repro gist: https://gist.github.com/lovasoa/74de94a81829c255d5f4c6a81b7af4c2

1. Start the repro:
```sh
docker compose -f repro/actix-http-early-303/docker-compose.yml up --build
```
2. Send a large `POST` through nginx:
```sh
dd if=/dev/zero count=1M | curl -v --data-binary @- localhost:8090/111
```
3. Observe that nginx returns `502 Bad Gateway` (timing sensitive, not fully deterministic, may need to retry multiple times)
4. Check nginx logs and observe:
```text
writev() failed (32: Broken pipe) while sending request to upstream
```

<details>
<summary>reproduction</summary>

```
❯ dd if=/dev/zero count=1M | curl -v --data-binary @- localhost:8090/111
1048576+0 records in
1048576+0 records out
536870912 bytes (537 MB, 512 MiB) copied, 0.335004 s, 1.6 GB/s
* Host localhost:8090 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8090...
* Connected to localhost (::1) port 8090
* using HTTP/1.x
> POST /111 HTTP/1.1
> Host: localhost:8090
> User-Agent: curl/8.15.0
> Accept: */*
> Content-Length: 536870912
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
>
< HTTP/1.1 100 Continue
<
* upload completely sent off: 536870912 bytes
< HTTP/1.1 502 Bad Gateway
< Server: nginx/1.27.5
< Date: Mon, 09 Mar 2026 22:46:03 GMT
< Content-Type: text/html
< Content-Length: 157
< Connection: keep-alive
<
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.27.5</center>
</body>
</html>
* Connection #0 to host localhost left intact
```

</details>

## Context

This bug caused a production issue on my [sqlpage](https://sql-page.com/) server today.
File uploads from unauthenticated users started failing with `502 Bad Gateway` instead of redirecting to the login page.

This matters when Actix applications are deployed behind nginx and may return an early final response for file uploads, such as auth-related rejections.

In that setup, a valid application response can be replaced by a `502`.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): `rustc 1.91.1`
- Actix Web Version: reproduced with a minimal `actix-http 3.11.2` server; no `actix-web` app layer required

lovasoa:
## Summary

Fixes #3967 : HTTP/1 early responses to unread fixed-length request bodies now linger for 1s instead of immediately closing the connection.

Related issues:

- https://github.com/actix/actix-web/issues/3967
- https://github.com/actix/actix-web/issues/3715
- https://github.com/actix/actix-web/issues/2357

**Disclaimer**: This code was written by AI under heavy human supervision, and tested by both AI and human.

## Testing

- Focused dispatcher tests:
  - buffered fixed-length upload ignored by handler does not shut down immediately
  - idle linger times out and closes cleanly
- Direct no-nginx socket tests against a test server:
  - slow POST body: got `200 OK`, then linger timed out after about 1s, then later writes failed as expected
  - normal follow-up direct requests still succeeded after both timeout cases
- Linux/docker + nginx repro:
  - reproduced the original `502` / `writev() failed (32: Broken pipe)` before the fix
  - after the fix, repeated buffered uploads returned `200`
- nginx stress matrix after the fix:
  - concurrency up to `1000`
  - payload sizes from `64 KiB` to `1 GiB`
  - nginx buffer settings varied across `client_body_buffer_size`, `proxy_buffer_size`, `proxy_buffers`, and `proxy_busy_buffers_size`
  - no reproduced `502`/`504` for the original bug after the fix; large buffered uploads up to `1 GiB` returned `200`

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -32,8 +32,7 @@ use crate::{
     config::ServiceConfig,
     error::{DispatchError, ParseError, PayloadError},
     service::HttpFlow,
-    ConnectionType, Error, Extensions, HttpMessage, OnConnectData, Request, Response,
-    StatusCode,
+    ConnectionType, Error, Extensions, HttpMessage, OnConnectData, Request, Response, StatusCode,
 };
 
 const LW_BUFFER_SIZE: usize = 1024;
@@ -383,7 +382,10 @@ where
         }
     }
 
-    fn poll_linger(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Result<Poll<()>, DispatchError> {
+    fn poll_linger(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Result<Poll<()>, DispatchError> {
         if self.as_mut().poll_flush(cx)?.is_pending() {
             return Ok(Poll::Pending);
         }
@@ -459,7 +461,11 @@ where
 
                 self.as_mut().project().state.set(State::None);
             }
-            _ => self.as_mut().project().state.set(State::SendPayload { body }),
+            _ => self
+                .as_mut()
+                .project()
+                .state
+                .set(State::SendPayload { body }),
         }
 
         Ok(())
@@ -598,10 +604,8 @@ where
                                 // this.payload was the payload for the request we just finished
                                 // responding to. We can check to see if we finished reading it
                                 // yet, and if not, shutdown the connection.
-                                let linger = should_linger(
-                                    this.payload.as_ref(),
-                                    *this.payload_drainable,
-                                );
+                                let linger =
+                                    should_linger(this.payload.as_ref(), *this.payload_drainable);
                                 let not_pipelined = this.messages.is_empty();
 
                                 // payload stream finished.
@@ -652,10 +656,8 @@ where
                                 // this.payload was the payload for the request we just finished
                                 // responding to. We can check to see if we finished reading it
                                 // yet, and if not, shutdown the connection.
-                                let linger = should_linger(
-                                    this.payload.as_ref(),
-                                    *this.payload_drainable,
-                                );
+                                let linger =
+                                    should_linger(this.payload.as_ref(), *this.payload_drainable);
                                 let not_pipelined = this.messages.is_empty();
 
                                 // payload stream finished.
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -6,6 +6,7 @@ use std::{
     pin::Pin,
     rc::Rc,
     task::{Context, Poll},
+    time::{Duration, Instant},
 };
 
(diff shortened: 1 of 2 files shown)


JohnTitor:
Thanks! The idea here looks great but I'd like to make some tweaks, so I've opened another version as #3985 not to mess up this PR. We have to choose which way has a better fit so give us some moments to do so.

lovasoa:
up :) can we merge either of the pr?

JohnTitor:
Sorry, I've been waiting for reviews but it's time to go ahead maybe. I'll get back after finishing some neccessary stuff.

PR #3988, opened 2026-03-27
fix(windows): enable dual-stack IPv6 sockets by default

rofferom:
## PR Type

Bug-Fix

## PR Checklist

- [x ] A changelog entry has been made for the appropriate packages.

## Overview

The goal is to have uniform dual-stack behavior across all operating systems. On Linux and macOS, binding to `[::]` already accepts both IPv4 and IPv6 connections by default. Windows is the outlier.
   
The change is targeted to Windows only (#[cfg(windows)]) to limit the impact and avoid altering behavior on platforms where dual-stack already works as expected.

### DIFF
--- a/actix-web/CHANGES.md
+++ b/actix-web/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):
Thank you! I think it makes sense.

PR #4025, opened 2026-04-17
fix(multipart): set cap for parser buffering

JohnTitor:
## PR Type

PR_TYPE

## PR Checklist

## Overview

Make `poll_stream` defensive not to consume malformed requests. This adds some configurable limits around it. Pooling logic has rewritten largely but it now only consumes following limit instead of inf-looping.

### DIFF
--- a/actix-multipart/src/multipart.rs
+++ b/actix-multipart/src/multipart.rs
@@ -11,7 +11,7 @@ use actix_web::{
     dev,
     error::{ParseError, PayloadError},
     http::header::{self, ContentDisposition, HeaderMap, HeaderName, HeaderValue},
-    web::Bytes,
+    web::{self, Bytes},
     HttpRequest,
 };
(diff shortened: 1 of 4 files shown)


actix-multipart/src/multipart.rs
```diff
@@ -44,6 +44,46 @@ enum Flow {
     Error(Option<Error>),
 }
 
+/// [`Multipart`] extractor configuration.
+///
+/// Add to your app data to have it picked up by [`Multipart`] extractors.
+#[derive(Clone, Copy, Debug)]
```
robjtede, line 50: risky to add Copy to structs like this in case we want to add some sort of common thing like boxed customizer function for something

JohnTitor: Fair enough, addressed in https://github.com/actix/actix-web/pull/4047

PR #4028, opened 2026-04-17
test(multipart,derive): move trybuild tests

JohnTitor:
## PR Type

PR_TYPE

## PR Checklist

## Overview

To avoid cyclic dependency tree.


JohnTitor:
@robjtede By the way, could you add me as owner of multipart-derive?

robjtede:
@JohnTitor 👍🏻  added the core team as owners

robjtede:
i dont love this solution, disabling doctests etc, and i think there's a better way of avoiding it during publishes, testing now...

JohnTitor:
@robjtede 
In practice, I don't think it makes a big difference though. If the derive crate is meant to be used along with the base crate, that doc should be live there, I think. How about moving docs to the main crate and putting a link on the derive side?

JohnTitor:
@robjtede How about above?

JohnTitor:
@robjtede I'm going to merge this to unblock the release. I think it's not that hard to change the way later as it's not a functional but doc-related change.

PR #4055, opened 2026-05-03
fix(encoding): flush codec after each chunk for streaming responses 

### LINKED ISSUE #3410: middleware::Compress seems to be buffering content indefinitely
Initially reported https://github.com/lovasoa/SQLpage/issues/435

When making streaming responses and activating `middleware::Compress`, the compression seems to buffer content indefinitely, without ever flushing the streaming data when it's small enough.

Here is an example demonstrating the problem

```rust
use actix_web::{
    get, middleware,
    web::{self, Bytes},
    App, HttpResponse, HttpServer,
};
use futures::stream::{self, StreamExt};
use std::time::Duration;
use tokio::time::sleep;

#[get("/")]
async fn index() -> HttpResponse {
    let stream = stream::once(async {
        let initial_content = r#"
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Streaming HTML Example</title>
</head>
<body>
    <div>This content appears immediately</div>
"#;
        Ok::<_, std::io::Error>(Bytes::from(initial_content))
    })
    .chain(stream::once(async {
        sleep(Duration::from_secs(5)).await;
        let delayed_content = r#"
    <div>This content appears after 5 seconds</div>
</body>
</html>
"#;
        Ok::<_, std::io::Error>(Bytes::from(delayed_content))
    }));

    HttpResponse::Ok()
        .content_type("text/html; charset=utf-8")
        .streaming(stream)
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(
                web::scope("/compressed")
                    .service(index),
            )
            .service(index)
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

## Expected Behavior

 - When visiting I should see `This content appears immediately` immediateley, and then after 5 seconds `This content appears after 5 seconds`
 - When visiting http://localhost:8080/compressed/ the behavior should be the same, or as close as possible. 

## Current Behavior

 - When visiting http://localhost:8080/compressed/ the page stays blank for five seconds, then displays `This content appears immediately` and `This content appears after 5 seconds` at the same time. 

## Possible Solution

The compression middleware should flush the writer object after a few milliseconds without new input data, to avoid indefinitely retaining data in memory on the server when it could actually already be rendered on the client.  

It's possible to retain data in memory for a few milliseconds to ensure good compression ratios, but actix shouldn't keep streaming data in memory for multiple seconds.

## Context

We are trying to integrate easy native loading spinners without javascript in [SQLPage](https://sql.ophir.dev)

## Your Environment

- Actix Web Version: 4.8.0

Rachit2323:
Fixes #3410                                            

  ## Problem                                                                                                                                          
   
  When using `middleware::Compress` with a streaming response, chunks don't                                                                           
  arrive at the client progressively — instead everything is buffered and
  delivered all at once when the stream ends.
                                                                                                                                                      
  For example, a response that sends one chunk immediately and another after
  5 seconds will show nothing for 5 seconds, then both chunks appear at the                                                                           
  same time.                                                                                                                                          
                                                                                                                                                      
  ## Why it happens                                                                                                                                   
                                                                                                                                                      
  The gzip/deflate/zstd codecs buffer incoming data internally waiting for
  more bytes to compress efficiently. `write()` feeds data in but the codec                                                                           
  decides when to emit output. Without an explicit flush, it holds everything
  until `finish()` is called at stream EOF.                                                                                                           
                                          
  ## Fix                                                                                                                                              
                                                                                                                                                      
  Added `ContentEncoder::flush()` which calls sync flush on the underlying                                                                            
  codec, and call it after every `write()` in `poll_next`. This forces the                                                                            
  codec to release each chunk immediately rather than waiting for the stream
  to end.                                                                                                                                             
                                          
  Small trade-off: slightly lower compression ratio per chunk, but this is                                                                            
  the same behaviour as nginx/Apache and the right call for streaming.

### DIFF
--- a/actix-http/src/encoding/encoder.rs
+++ b/actix-http/src/encoding/encoder.rs
@@ -208,6 +208,7 @@ where
                     if let Some(mut encoder) = this.encoder.take() {
                         if chunk.len() < MAX_CHUNK_SIZE_ENCODE_IN_PLACE {
                             encoder.write(&chunk).map_err(EncoderError::Io)?;
+                            encoder.flush().map_err(EncoderError::Io)?;
                             let chunk = encoder.take();
                             *this.encoder = Some(encoder);
 
@@ -217,6 +218,7 @@ where
                         } else {
                             *this.fut = Some(spawn_blocking(move || {
                                 encoder.write(&chunk)?;
+                                encoder.flush()?;
                                 Ok(encoder)
                             }));
                         }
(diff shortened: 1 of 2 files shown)


JohnTitor:
Please do not paste AI output directly, we don't welcome slops.

PR #4070, opened 2026-05-14
fix: duplicated words in actix-multipart and actix-http h2 comments

oab24413gmai:
Two one-line typo fixes for duplicated words in source comments:
- `actix-multipart/src/form/mod.rs` — "// Exceeds the the 30 byte limit" → "// Exceeds the 30 byte limit"
- `actix-http/src/h2/dispatcher.rs` — "// When there is an in-flight ping-pong, poll pong and and keep-alive" → "...poll pong and keep-alive"

No code/behavior change.


robjtede (approved):
👍🏻

PR #4140, opened 2026-07-08
Add support for the HTTP QUERY method (RFC 10008)

mo-bruno:
## PR Type

Feature

## PR Checklist

## Overview

Adds first-class support for the HTTP `QUERY` method, wired exactly like the existing
`PATCH` method throughout routing, guards, the `web::` helpers, the routing macros, the
`awc` client, and the test servers.

[`QUERY`](https://www.rfc-editor.org/rfc/rfc10008.html) was standardized in **RFC 10008
(June 2026)** and registered in the [IANA HTTP Method Registry]. It is **safe** and
**idempotent** like `GET`, but carries a **request body** like `POST` — the first new HTTP
method since `PATCH` (RFC 5789, 2010). Because actix-web's extractors are method-agnostic,
body extraction (`web::Json`, `web::Bytes`, …) works for `QUERY` handlers with no special
handling.

This is a purely **additive** change (new public items only) → **semver-minor**; no breaking
changes.

[IANA HTTP Method Registry]: https://www.iana.org/assignments/http-methods/http-methods.xhtml

### What's included

`actix-web`
- `guard::Query()` — method guard, mirrors `guard::Patch()`
- `web::query()` — pre-configured `Route`, mirrors `web::patch()`
- `Resource::query()` — route shortcut, mirrors `Resource::patch()`
- `TestRequest::query()` — test helper, mirrors `TestRequest::patch()`
- re-export of the `#[query]` routing macro

`actix-web-codegen`
- `#[query]` attribute macro, mirrors `#[patch]`

`awc`
- `Client::query()` — mirrors `Client::patch()`

`actix-test` / `actix-http-test`
- `TestServer::query()` (and `actix-http-test`'s HTTPS `TestServer::squery()`) — mirror `patch`/`spatch`

Every method-convenience surface that has a `patch` now has a matching `query`, so `QUERY`
is a first-class method throughout.

### Notes for reviewers

- **`http` crate version.** actix-web currently pins `http` `0.2`, which has no
  `Method::QUERY` associated constant (that constant only exists in `http` `1.x`). This PR
  therefore constructs the method with `Method::from_bytes(b"QUERY").unwrap()` and leaves a
  `TODO` to switch to `Method::QUERY` once the dependency is bumped. This mirrors the
  existing `Method::from_bytes(...).unwrap()` pattern already used in `actix-web-codegen`
  for custom methods. This PR **does not** bump `http` — that is tracked separately in
  #3384. No new dependencies are added.
- **`actix-http`.** No change needed: `actix-http` re-exports `http::Method` and has no
  local method enum/constant list, so there is nothing to mirror there.
- **Scope.** Minimal and mechanical — every change mirrors an existing `PATCH` site. No
  unrelated refactoring.

### Tests

- `guard::Query()` matches a `QUERY` request and rejects a `GET`.
- A `web::query()` route returns `200` for `QUERY` and `405` (`Allow: QUERY`) for other methods.
- A `QUERY` handler reads and echoes a JSON body (proves body-carrying works).
- `#[query]` macro integration test mirrors the existing `#[patch]` test (`test_body`).
- `awc::Client::query()` builds a request whose method is `QUERY`.
- Round-trip tests for each test-server helper against a live server: `actix_test::TestServer::query()`
  (body echo through a `web::query()` route), `actix_http_test::TestServer::query()` (plaintext HTTP/1),
  and `actix_http_test::TestServer::squery()` (rustls; asserts the request arrives as `QUERY` over HTTP/2).

### Verification

Local, matching the project's CI gates:

- `cargo +nightly fmt --all -- --check` — clean.
- `cargo clippy --workspace --all-targets ... -D clippy::todo -D clippy::dbg_macro` — 0 warnings.
- `cargo +nightly doc --no-deps ... -D warnings` — my `[QUERY]` intra-doc link resolves; no new warnings.
- `just check-min` / `just check-default` (`cargo hack --workspace check [--no-default-features]`) — pass.
- `cargo +1.88` (MSRV) check — pass.
- Tests pass across `actix-web`, `actix-web-codegen`, `awc`, `actix-test`, `actix-http-test`
  (incl. doctests and awc integration/TLS tests), 0 failed.

### DIFF
--- a/actix-http-test/CHANGES.md
+++ b/actix-http-test/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `TestServer::{query, squery}()` for issuing HTTP `QUERY` requests (RFC 10008).
 - Minimum supported Rust version (MSRV) is now 1.88.
 
 ## 3.2.0
--- a/actix-http-test/CHANGES.md
+++ b/actix-http-test/CHANGES.md
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
-- Add `TestServer::{query, squery}()` for issuing HTTP `QUERY` requests (RFC 10008).
+- Add `TestServer::{query, squery}()` for issuing HTTP `QUERY` requests (RFC 10008). [#4140]
 - Minimum supported Rust version (MSRV) is now 1.88.
+
+[#4140]: https://github.com/actix/actix-web/pull/4140
 
 ## 3.2.0
 
--- a/actix-http-test/src/lib.rs
+++ b/actix-http-test/src/lib.rs
@@ -216,6 +216,16 @@ impl TestServer {
         self.client.patch(self.surl(path.as_ref()).as_str())
     }
 
+    /// Create `QUERY` request
+    pub fn query<S: AsRef<str>>(&self, path: S) -> ClientRequest {
+        self.client.query(self.url(path.as_ref()).as_str())
+    }
+
+    /// Create HTTPS `QUERY` request
+    pub fn squery<S: AsRef<str>>(&self, path: S) -> ClientRequest {
+        self.client.query(self.surl(path.as_ref()).as_str())
+    }
+
     /// Create `DELETE` request
     pub fn delete<S: AsRef<str>>(&self, path: S) -> ClientRequest {
         self.client.delete(self.url(path.as_ref()).as_str())
--- a/actix-http/tests/test_rustls.rs
+++ b/actix-http/tests/test_rustls.rs
@@ -193,6 +193,25 @@ async fn h2_1() -> io::Result<()> {
 }
 
 #[actix_rt::test]
+async fn h2_squery() -> io::Result<()> {
+    let mut srv = test_server(move || {
+        HttpService::build()
+            .h2(|req: Request| {
+                assert_eq!(req.method().as_str(), "QUERY");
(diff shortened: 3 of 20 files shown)


Samielakkad (commented):
The core QUERY routing and `awc::Client::query()` paths have coverage, but the new test-server helpers look untested:

- `actix_test::TestServer::query()`
- `actix_http_test::TestServer::query()`
- `actix_http_test::TestServer::squery()`

Those are small wrappers, but they are public convenience APIs and they can regress independently from `web::query()` / `TestRequest::query()` because they build full client requests from server URLs. It would be good to add a minimal test that starts a server with a `web::query()` route and reaches it through each new helper, at least for the non-TLS helpers if TLS coverage is awkward in this crate.

That would also prove the method survives the client/server round trip, not just the in-process service tests.

mo-bruno:
@Samielakkad added in fcbf37f, one per helper. `test_query_method` in actix-web's test_server.rs does the full round trip through a `web::query()` route, request body included. `h1_query` and `h2_squery` in the actix-http test files are the minimal version: the handler asserts the method comes through as QUERY and the client checks the response. TLS wasn't awkward after all since test_rustls.rs already uses `sget()`, and that path negotiates h2, so it also covers QUERY over HTTP/2, which nothing else did. Ready for another look.

PR #4142, opened 2026-07-09
fix: correct "Implementors" to "Implementers" in docs

maxtaran2010:
## Summary

Fixes two occurrences of the misspelling "Implementors" → "Implementers" in doc comments and migration guide.

- `actix-router/src/resource_path.rs` — doc comment for the `Resource` trait
- `actix-web/MIGRATION-4.0.md` — Guards API section

Found with `codespell`.


robjtede (approved):
i believe "implementor" is a legitimate way of spelling it but this is the more american english way, so fair enough

PR #4160, opened 2026-07-22
docs(web): clarify scope route matching

### LINKED ISSUE #2904: Overlapping scopes work differently than flat overlapping routes
## Expected Behavior

The docs seem to suggest that the `web::scope(...)` is a helper that groups different routes under the same prefix, but that otherwise the behaviour would be the same as with flattened routes. At least, I didn't see any mention of differences.

So, when I refactored my code from flat routes like in this minimal example:

```rust
use actix_web::{get, put, web, App, HttpServer, Responder};

#[get("/camera/dimensions")]
async fn camera_dimensions() -> impl Responder {
    "800x600"
}

#[get("/camera/owner")]
async fn camera_owner() -> impl Responder {
    "John Doe"
}

#[get("/focuser/distance")]
async fn focuser_focus() -> impl Responder {
    "42"
}

#[put("/focuser/distance")]
async fn focuser_focus_set(value: web::Path<f64>) -> impl Responder {
    format!("Set focus to {}", value)
}

#[get("/{device_type}/connected")]
async fn device_connected(device_type: web::Path<String>) -> impl Responder {
    format!("{} is connected", device_type)
}

#[put("/{device_type}/connect")]
async fn device_connect(device_type: web::Path<String>) -> impl Responder {
    format!("Connecting {}", device_type)
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(camera_dimensions)
            .service(camera_owner)
            .service(focuser_focus)
            .service(focuser_focus_set)
            .service(device_connected)
            .service(device_connect)
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```

to scoped groups like in this one:

```rust
use actix_web::{get, put, web, App, HttpServer, Responder};

#[get("/dimensions")]
async fn camera_dimensions() -> impl Responder {
    "800x600"
}

#[get("/owner")]
async fn camera_owner() -> impl Responder {
    "John Doe"
}

#[get("/distance")]
async fn focuser_focus() -> impl Responder {
    "42"
}

#[put("/distance")]
async fn focuser_focus_set(value: web::Path<f64>) -> impl Responder {
    format!("Set focus to {}", value)
}

#[get("/connected")]
async fn device_connected(device_type: web::Path<String>) -> impl Responder {
    format!("{} is connected", device_type)
}

#[put("/connect")]
async fn device_connect(device_type: web::Path<String>) -> impl Responder {
    format!("Connecting {}", device_type)
}

#[actix_web::main]
    HttpServer::new(|| {
        App::new()
            .service(
                web::scope("/camera")
                    .service(camera_dimensions)
                    .service(camera_owner),
            )
            .service(
                web::scope("/focuser")
                    .service(focuser_focus)
                    .service(focuser_focus_set),
            )
            .service(
                web::scope("/{device_type}")
                    .service(device_connected)
                    .service(device_connect),
            )
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```

I expected all routes to continue working like before.

## Current Behavior

Unfortunately, it seems that scopes are only matched against each other, and, once a scope is matched, no other scopes are tried even if it doesn't have the required method.

So, routes like `/camera/connected` and `/focuser/connected` worked before, but stop working after refactoring.

I thought reordering might help, putting the generic `/{device_type}` scope first, but that only seems to change which scope catches everything - now, routes like `/*/connected` work but specific device routes like `/camera/dimensions` don't.

Finally, I thought maybe I should use `/{device_type}` as `app.default_service()`, but that doesn't work either as `default_service()` seems to only accept a single handler and not a scope as its argument.

## Possible Solution

Flatten all routes at the app level when scopes are registered, so that there is no observable change in behaviour after grouping.

Run the minimal examples provided above with suggested requests like `http://localhost:8080/camera/connected` and observe the working response "camera is connected" in the first implementation and 404 error in the 2nd one.

## Context

I'm trying to refactor my complex list of routes into grouped scopes in separate functions to make code a bit more maintainable.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): rustc 1.63.0 (4b91a6ea7 2022-08-08)
- Actix Web Version: 4.2.1

AraMinjibir:
## PR Type

Documentation

## PR Checklist

## Overview

The documentation for `web::scope` now clarifies how scopes are matched during routing.

Once a request matches a scope's path prefix, routing continues only within that scope. If no route inside the matched scope can handle the request, Actix Web does not continue searching sibling scopes or services.

This clarifies that overlapping scopes do not behave the same as an equivalent flat route structure and helps explain why route organization using scopes can affect routing behavior.

Closes #2904

### DIFF
--- a/actix-web/src/scope.rs
+++ b/actix-web/src/scope.rs
@@ -30,6 +30,19 @@ type Guards = Vec<Box<dyn Guard>>;
 /// requests using the [`Path`](crate::web::Path) extractor or
 /// with [`HttpRequest::match_info()`](crate::HttpRequest::match_info).
 ///
+/// # Route Matching
+///
+/// Scopes are matched independently during routing. Once a request matches a
+/// scope's path prefix, routing continues only within that scope. If no route
+/// inside the matched scope can handle the request, Actix Web does not continue
+/// searching sibling scopes or services.
+///
+/// Consequently, overlapping scopes do not behave the same as an equivalent
+/// flat route structure.
+///
+/// If your goal is to organize routes into separate modules without changing
+/// routing behavior, consider using `App::configure` or `Scope::configure`.
+///
 /// # Avoid Trailing Slashes
 /// Avoid using trailing slashes in the scope prefix (e.g., `web::scope("/scope/")`). It will almost
 /// certainly not have the expected behavior. See the [documentation on resource definitions][pat]


actix-web/src/scope.rs
```diff
@@ -30,6 +30,19 @@ type Guards = Vec<Box<dyn Guard>>;
...
+/// Scopes are matched independently during routing. Once a request matches a
+/// scope's path prefix, routing continues only within that scope. If no route
+/// inside the matched scope can handle the request, Actix Web does not continue
+/// searching sibling scopes or services.
+///
+/// Consequently, overlapping scopes do not behave the same as an equivalent
+/// flat route structure.
+///
+/// If your goal is to organize routes into separate modules without changing
+/// routing behavior, consider using `App::configure` or `Scope::configure`.
```
JohnTitor, line 44: The below should be clearer:

```suggestion
/// Use [`App::configure`](crate::App::configure) for app-level services,
/// or [`Scope::configure`](crate::Scope::configure) for services already inside the same scope.
```

actix-web/src/scope.rs
```diff
@@ -30,6 +30,19 @@ type Guards = Vec<Box<dyn Guard>>;
 /// requests using the [`Path`](crate::web::Path) extractor or
 /// with [`HttpRequest::match_info()`](crate::HttpRequest::match_info).
 ///
+/// # Route Matching
+///
+/// Scopes are matched independently during routing. Once a request matches a
+/// scope's path prefix, routing continues only within that scope. If no route
+/// inside the matched scope can handle the request, Actix Web does not continue
+/// searching sibling scopes or services.
```
JohnTitor, line 38:
```suggestion
/// At each routing level, services are checked in registration order. A scope
/// is selected when its path prefix and all of its guards match the request.
/// Once selected, routing continues only within that scope. If none of its
/// nested services match, its default service is called; Actix Web does not
/// backtrack to sibling services registered after it.
```

Could we make the scope selection condition explicit here?
This would also capture the ordering behavior discussed in #2904.

PR #4169, opened 2026-08-03
Drain HTTP/1 connections during graceful shutdown

### LINKED ISSUE #3169: Graceful shutdown for persistent connections
When an `actix-web` server receives a SIGTERM signal, it will stop accepting new connections and wait for all active connections to close. After the shutdown timeout configured by `HttpServer::shutdown_timeout`, the server will force close all active connections.

In our case, we have several servers behind a layer 4 load balancer. When we want to deploy a new version of our service, we will perform a rolling update. We will stop one old server, deploy the new version of the service, and start the new server. Before the old server is stopped, the load balancer will stop making new connections to the old server. However, the old server will still have some persistent connections. If some request is received before the old server is stopped, but not finished before the shutdown timeout, the client will receive a `connection reset by peer` error.

After some research, we found that other web frameworks have a more graceful shutdown strategy. For HTTP/1.1, the server will add a `Connection: close` header to all subsequent responses. For HTTP/2, the server will send a `GOAWAY` frame to the client. This will allow the client to close the connection gracefully.

For example, in the `hyper` crate, the above strategy is implemented in the [code](https://github.com/hyperium/hyper/blob/v0.14.27/src/server/conn.rs#L744-L761). All `hyper`-based web frameworks, such as `axum`, can benefit from this graceful shutdown strategy.

We have tried to implement this graceful shutdown strategy in `actix-web` using middleware. However, some difficulties are encountered. For example, the middleware cannot access the shutdown signal, but luckily signals are the only shutdown condition in our case, and we can directly watch them. Another difficulty is that the middleware can only solve the problem in HTTP/1.1, but not in HTTP/2, as the middleware cannot access the underlying connection.

We hope that `actix-web` can implement this graceful shutdown strategy. This will make `actix-web` more suitable for production environments.

robjtede:
## Summary

- propagate the server's graceful shutdown notification into every HTTP service factory
- close idle HTTP/1 keep-alive connections as soon as graceful shutdown starts
- let the current HTTP/1 request finish, then close the connection
- discard queued pipelined requests without starting them
- cover both `ServerHandle::stop(true)` and `HttpServer::shutdown_signal()`

Requires actix-server 2.7, which contains the signal API from actix/actix-net#930.

## Root cause

Actix-server detaches each connection future and waits for its active-connection counter during graceful shutdown. The HTTP/1 dispatcher had no notification that shutdown had started, so an idle connection stayed open and a connection that completed its active request returned to keep-alive.

## Behavior

When graceful shutdown starts, an idle connection closes immediately. An active connection finishes only its current request, flushes the final response, and closes. Requests already queued behind the active request do not start.

## Validation

- `cargo test -p actix-web --test test_httpserver shutdown -- --nocapture`
- `cargo test -p actix-http --lib h1::dispatcher -- --nocapture`
- `just clippy`
- `cargo package --locked -p actix-web --allow-dirty --no-verify`

### DIFF
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -569,7 +569,6 @@ where
         'res: loop {
             let mut this = self.as_mut().project();
             match this.state.as_mut().project() {
-                // no future is in InnerDispatcher state; pop next message
                 StateProj::None if this.flags.contains(Flags::DRAINING) => {
                     this.messages.clear();
                     this.flags.remove(Flags::KEEP_ALIVE);
@@ -581,6 +580,7 @@ where
                     return Ok(PollResponse::DoNothing);
                 }
 
+                // no future is in InnerDispatcher state; pop next message
                 StateProj::None => match this.messages.pop_front() {
                     // handle request message
                     Some(DispatcherMessage::Item(req)) => {
--- a/actix-http/src/h1/dispatcher.rs
+++ b/actix-http/src/h1/dispatcher.rs
@@ -63,6 +63,11 @@ bitflags! {
         const LINGER           = 0b0100_0000;
 
         /// Set when the server is draining this connection during graceful shutdown.
+        ///
+        /// Unlike [`SHUTDOWN`](Self::SHUTDOWN), this state continues polling the current request
+        /// and response. It prevents queued requests from starting and transitions to `SHUTDOWN`
+        /// after the current response finishes. [`LINGER`](Self::LINGER) remains separate and is
+        /// used only to close cleanly after a response to an unread request payload.
         const DRAINING         = 0b1000_0000;
     }
 }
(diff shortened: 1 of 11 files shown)


robjtede:
note that the new `graceful_shutdown_closes_idle_http1_connection_after_in_flight_request` test fails on master

robjtede:
done a self review and looks good, want another pair of eyes though because it touches the dispatcher @JohnTitor 🙏🏻 TIA

actix-http/src/h1/dispatcher.rs
JohnTitor, line 1322: I guess this `else` branch won't be blocked and `poll_request` at L1341 can be called when the cond is met.
It'd be great if we could skip calling that so that we don't enter unwanted states (i.e. issuing a new req) while draining.

robjtede: Good catch. I don't love that we need to consider pipelined requests but added a test which exercised this conditional then it was a simple fix in poll_request.

JohnTitor (approved):
Thanks, looks great!

PR #4178, opened 2026-08-07
fix(http): terminate ByteString message bodies

### LINKED ISSUE #4177: `ByteString`'s `MessageBody` implementation never terminates when polled
## Expected Behavior

After yielding its bytes once, the `MessageBody` implementation for `bytestring::ByteString` should return `Poll::Ready(None)` to signal that the body is complete, like the other one-shot body implementations in the same module.

## Current Behavior

`ByteString::poll_next` unconditionally takes the current value and returns `Some(Ok(...))`. The first poll yields the content, but every later poll yields another empty chunk and the body never reaches end-of-stream.
<https://github.com/actix/actix-web/blob/937960ca67f20e14ffe2a075bf6d4593502be12c/actix-http/src/body/message_body.rs#L448-L468>

## Possible Solution

Check `self.is_empty()` before taking the value and return `Poll::Ready(None)` when it is empty. This matches the implementations for `Bytes`, `BytesMut`, `Vec<u8>`, `String`.

```rust
use std::{future::poll_fn, pin::pin};

use actix_http::body::MessageBody;
use bytes::Bytes;
use bytestring::ByteString;

#[actix_rt::test]
async fn bytestring_body_terminates() {
    let body = ByteString::from_static("hello");
    let mut body = pin!(body);

    let first = poll_fn(|cx| body.as_mut().poll_next(cx)).await;
    assert_eq!(first.unwrap().unwrap(), Bytes::from_static(b"hello"));

    let second = poll_fn(|cx| body.as_mut().poll_next(cx)).await;
    assert!(second.is_none());
}
```

The final assertion currently fails because `second` is `Some(Ok(Bytes::new()))`. Repeated polls continue returning the same value.

## Context

`MessageBody::poll_next` documents `Poll::Ready(None)` as the completion signal.

## Your Environment

- Rust Version (I.e, output of `rustc -V`): rustc 1.96.0 (ac68faa20 2026-05-25) 
- Actix Web Version: current main

WaterWhisperer:
## PR Type

Bug Fix

## PR Checklist

## Overview

The `MessageBody` implementation for `ByteString` kept producing empty chunks after yielding its contents, preventing polling consumers from observing completion.

Closes #4177

### DIFF
--- a/actix-http/CHANGES.md
+++ b/actix-http/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Fix `MessageBody` implementation for `ByteString` to signal completion after yielding its contents. [#4177]
+
+[#4177]: https://github.com/actix/actix-web/issues/4177
+
 ## 3.13.1
 
 - Fix HTTP/1 WebSocket upgrade responses being overwritten with `Connection: close` when the upgraded request payload remains open. [#4115]
--- a/actix-http/src/body/message_body.rs
+++ b/actix-http/src/body/message_body.rs
@@ -458,8 +458,12 @@ mod foreign_impls {
             self: Pin<&mut Self>,
             _cx: &mut Context<'_>,
         ) -> Poll<Option<Result<Bytes, Self::Error>>> {
-            let string = mem::take(self.get_mut());
-            Poll::Ready(Some(Ok(string.into_bytes())))
+            if self.is_empty() {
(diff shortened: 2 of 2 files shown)


robjtede (approved):
nice catch, fix is sound

robjtede:
this has been released in v3.13.2, cargo update to get it

PR #4179, opened 2026-08-08
Fix Connection and Expect header parsing

### LINKED ISSUE #2692: define behavior when multiple Connection header values are provided
see dicussion in comments of : https://github.com/actix/actix-web/discussions/2691#discussioncomment-2340435

in particular:
- what happens when receiving `close, upgrade`
- what happens when receiving `keep-alive, upgrade`
- other tricky combinations? TODO

WaterWhisperer:
## PR Type

Bug Fix

## PR Checklist

## Overview

- Parse comma-separated `Connection` options with `close` taking precedence.
- Parse `Expect` headers and only recognize `100-continue`.
- Correct misleading content-encoder trace messages and the `zstd` typo.

See [RFC 9110 §7.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.6.1) and [RFC 9110 §10.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1).

Closes #2692

### DIFF
--- a/actix-http/src/h1/decoder.rs
+++ b/actix-http/src/h1/decoder.rs
@@ -969,23 +969,25 @@ mod tests {
 
     #[test]
     fn test_conn_multi_value() {
-        let req = parse_ready!(&mut BytesMut::from(
-            "GET /test HTTP/1.1\r\n\
-             connection: keep-alive, Upgrade\r\n\r\n",
-        ));
-        assert_eq!(req.head().connection_type(), ConnectionType::Upgrade);
-
-        let req = parse_ready!(&mut BytesMut::from(
-            "GET /test HTTP/1.1\r\n\
-             connection: close, upgrade\r\n\r\n",
-        ));
-        assert_eq!(req.head().connection_type(), ConnectionType::Close);
-
-        let req = parse_ready!(&mut BytesMut::from(
-            "GET /test HTTP/1.1\r\n\
-             connection: upgrade, close\r\n\r\n",
-        ));
-        assert_eq!(req.head().connection_type(), ConnectionType::Close);
+        for (connection, expected) in [
+            ("keep-alive, Upgrade", ConnectionType::Upgrade),
+            ("keep-alive\r\nconnection: Upgrade", ConnectionType::Upgrade),
+            ("close, upgrade", ConnectionType::Close),
+            ("upgrade, close", ConnectionType::Close),
+            ("close\r\nconnection: upgrade", ConnectionType::Close),
+            ("upgrade\r\nconnection: close", ConnectionType::Close),
+            ("not-upgrade", ConnectionType::KeepAlive),
+        ] {
+            let raw = format!("GET /test HTTP/1.1\r\nconnection: {connection}\r\n\r\n");
+            let req = parse_ready!(&mut BytesMut::from(raw.as_str()));
+            assert_eq!(req.head().connection_type(), expected, "{connection:?}");
+            assert_eq!(
+                req.upgrade(),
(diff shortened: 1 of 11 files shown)


actix-http/src/h1/decoder.rs
```diff
@@ -175,9 +177,12 @@ pub(crate) trait MessageType: Sized {
                     }
 
                     header::EXPECT => {
-                        let bytes = value.as_bytes();
-                        if bytes.len() >= 4 && &bytes[0..4] == b"100-" {
-                            expect = true;
+                        if let Ok(value) = value.to_str() {
```
JohnTitor: Pre-existing but we need to ignore it for HTTP/1 as per https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1-12:

> A server that receives a 100-continue expectation in an HTTP/1.0 request MUST ignore that expectation.

actix-http/src/h1/decoder.rs
```diff
@@ -175,9 +177,12 @@ pub(crate) trait MessageType: Sized {
...
+                            expect = expect
+                                || value
+                                    .split(',')
```
JohnTitor: A string can be quoted so splitting with a comma can misunderstand an input, I think.

actix-http/src/h1/decoder.rs
```diff
@@ -151,19 +151,21 @@ pub(crate) trait MessageType: Sized {
...
-                        ka = if let Ok(conn) = value.to_str().map(str::trim) {
-                            if conn.eq_ignore_ascii_case("keep-alive") {
-                                Some(ConnectionType::KeepAlive)
-                            } else if conn.eq_ignore_ascii_case("close") {
-                                Some(ConnectionType::Close)
-                            } else if conn.eq_ignore_ascii_case("upgrade") {
-                                Some(ConnectionType::Upgrade)
-                            } else {
-                                None
+                        if let Ok(conn) = value.to_str() {
```
JohnTitor, line 154: I guess we need to align the new bahavior on some other places, consider these headers:
```
connection: keep-alive
connection: Upgrade
upgrade: websocket
```

While `connection_type` returning `Upgrade` thanks to this change, `req.upgrade()` would still return `false`. I haven't confirm it exhaustively (I mean there might also be more places to fix) but we have to eliminate this type of inconsistency.

actix-http/src/h1/decoder.rs
JohnTitor: I think we need an if branch for HTTP/1 as per https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1-12

> A server that receives a 100-continue expectation in an HTTP/1.0 request MUST ignore that expectation.

actix-http/src/h1/decoder.rs
JohnTitor: A value can be quoted so splitting with a comma could misparse some headers like `"foo, bar"`, right?

JohnTitor (commented):
Disclaimer: I'm not a HTTP spec expert so I might be missing something but leaving some concerns. cc @robjtede for additional reviews.

actix-http/src/h1/decoder.rs
WaterWhisperer: Thanks, I updated `Request::upgrade()`, `RequestHead::upgrade()`, and awc's handshake check as well. If anything is missing, thanks for pointing it out.

actix-http/src/h1/decoder.rs
WaterWhisperer: Thanks a lot, I missed it

PR #4186, opened 2026-08-16
fix(multipart): include inner error in Payload and Parse display strings

### LINKED ISSUE #4168: [actix-multipart] Generic "Payload error" when request exceeds size limit
## Expected Behavior

The error message should describe the nature of the "payload error", such as "payload reached size limit" or "can not decode content-encoding".

## Current Behavior

The plaintext error that appears to users is not actionable.

## Possible Solution

Looking at the [`MultipartError`](https://docs.rs/actix-multipart/latest/actix_multipart/enum.MultipartError.html) enum, it seems the relevant details do exist on the `Payload` case, but the `display` attribute doesn't include them.

~~The simplest fix would probably be to add `: {_0}` to the end of the `display` attribute for this and other affected cases (such as `Parse`), such as in [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4ea06803e4731fa7f8ad2fd016010445). The `DuplicateField`, `MissingField`, and `UnknownField` cases already do something similar. I can draft a PR for this if need be.~~
EDIT: This patch was rejected at https://github.com/actix/actix-web/pull/4186#pullrequestreview-4998119280 and reverted in 6a1feec593b9ecdfb6066b471582c3023b595141.

Alternatively, a custom [`error_response`](https://docs.rs/actix-web/latest/actix_web/error/trait.ResponseError.html#method.error_response) method at https://docs.rs/actix-multipart/latest/src/actix_multipart/error.rs.html#99 could work: one that walks the error tree to respond with something reasonable, even in cases where the `Display` impl must remain incomplete per [official Rust idioms and guidelines](https://doc.rust-lang.org/stable/std/error/trait.Error.html#error-source).

In the meantime, users can work around this by `match`-ing against the error in a custom handler, and patch the output text that way.

1. Set up an actix-web service with an endpoint that uses the `actix_multipart::form::MultipartForm` extractor.
2. Try uploading a file that is slightly too large.
3. Observe strange error.

See also this minimal repro: https://git.average.name/AverageHelper/actix_web_multipart_error_display_repro

## Context

I'm trying to handle small image uploads for e.g. profile pictures, which typically don't need to be bigger than ~1 MiB. But even at the default total limit of 50 MiB, the default error should still be somewhat actionable IMO.

## Your Environment

- Rust Version: rustc 1.98.1 (48a229cea 2026-09-01)
- Actix Web Version: 4.15.0

Sruhvx-jpg:
## PR Type

Bug Fix

## PR Checklist

## Overview

Improves error messages and `Display` implementations across `actix-multipart`:

1. **`MultipartError::Payload` and `MultipartError::Parse`**: Include inner error details (`{_0}`).
2. **`MultipartError::Field`**: Include underlying source error (`'{name}': {source}`).
3. **`MultipartError::ContentDispositionNameMissing`**: Fix display message to indicate that the `name` parameter is missing (rather than the entire header).
4. **`JsonFieldError::ContentType` and `TextError::ContentType`**: Provide clear messages describing expected media types (`Field Content-Type is not application/json` / `text/plain`).

Closes #4168

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-- Include inner error in `MultipartError::Payload` and `MultipartError::Parse` display strings. [#4168]
+- Include inner error in `MultipartError::Payload`, `MultipartError::Parse`, and `MultipartError::Field` display strings. [#4168]
+- Fix `MultipartError::ContentDispositionNameMissing` display string to indicate missing name parameter.
+- Improve `JsonFieldError::ContentType` and `TextError::ContentType` error messages.
 
 [#4168]: https://github.com/actix/actix-web/issues/4168
 
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Include inner error in `MultipartError::Payload` and `MultipartError::Parse` display strings. [#4168]
+
+[#4168]: https://github.com/actix/actix-web/issues/4168
+
 ## 0.8.0
 
 - Add multi-field multipart payload builders to `actix_multipart::test`. [#3575]
--- a/actix-multipart/src/error.rs
+++ b/actix-multipart/src/error.rs
@@ -60,11 +60,11 @@ pub enum Error {
     Incomplete,
 
     /// Field parsing failed.
-    #[display("Error during field parsing")]
+    #[display("Error during field parsing: {_0}")]
     Parse(ParseError),
 
     /// HTTP payload error.
-    #[display("Payload error")]
+    #[display("Payload error: {_0}")]
     Payload(PayloadError),
 
     /// Stream is not consumed.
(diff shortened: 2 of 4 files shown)


JohnTitor (approved):
LGTM, thanks!

Sruhvx-jpg:
Hey man thanks.... this made my day

robjtede (commented):
This PR should be reverted. It goes against [official Rust idioms and guidelines](https://doc.rust-lang.org/stable/std/error/trait.Error.html#error-source).

> In error types that wrap an underlying error, the underlying error should be either returned by the outer error’s Error::source(), or rendered by the outer error’s Display implementation, but not both.

In other words for this case, don't append the source's display; it impls Error properly.

If you want error chain printing you should consider using [`err-report`](https://crates.io/crates/err-report) which is a clone of a future std type [`Report`](https://doc.rust-lang.org/stable/std/error/struct.Report.html).

Sruhvx-jpg:
yes right!!! but  instead of reverting it i could just remove that specific change.. what do u say?

PR #4187, opened 2026-08-16
fix(multipart): correctly handle boundary delimiter split across chunks

### LINKED ISSUE #3635: Report a rare case where multipart may fail to parse
We encountered an issue where an iOS client app uploads a large file using multipart requests. The file is split into 32MB chunks, and the client sends a series of multipart HTTP requests. Each request contains several form-data fields to transfer additional information to the server.

On the server side, we are using actix-web with the multipart extractor to handle these requests. However, during transmission, the multipart extractor occasionally fails to parse one of the form-data fields. This issue is rare and difficult to reproduce.

After capturing the traffic using a pcap file, we observed that in one instance, the multipart boundary sequence (--) was split across two TCP frames. It seems this caused the multipart extractor to fail during parsing.

If you want to review the content using Wireshark, rename boundary-problem.pcap.txt to boundary-problem.pcap.

Image

[boudnary-problem.pcap.txt](https://github.com/user-attachments/files/19956041/boudnary-problem.pcap.txt)

Sruhvx-jpg:
## PR Type

PR_TYPE

## PR Checklist

## Overview

Fix https://github.com/actix/actix-web/issues/3635

### DIFF
--- a/actix-multipart/CHANGES.md
+++ b/actix-multipart/CHANGES.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+- Fix multipart field parsing when boundary delimiter is split across payload chunks. [#3635]
 - Improve detail in some error messages.
+
+[#3635]: https://github.com/actix/actix-web/issues/3635
 
 ## 0.8.0
 
--- a/actix-multipart/src/field.rs
+++ b/actix-multipart/src/field.rs
@@ -302,7 +302,7 @@ impl InnerField {
         }
 
         // check boundary
-        if len > 4 && payload.buf[0] == b'\r' {
+        if len >= 4 && payload.buf[0] == b'\r' {
             let b_len = if payload.buf.starts_with(b"\r\n") && &payload.buf[2..4] == b"--" {
                 Some(4)
             } else if &payload.buf[1..3] == b"--" {
@@ -411,7 +411,8 @@ impl InnerField {
 
 #[cfg(test)]
 mod tests {
-    use futures_util::{stream, StreamExt as _};
+    use actix_http::h1;
+    use futures_util::{stream, FutureExt as _, StreamExt as _};
 
     use super::*;
     use crate::Multipart;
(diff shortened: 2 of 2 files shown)


robjtede (changes requested):
Thanks you for the contribution.

This PR does makes too many unnecessary changes to the parsing logic. It turns out a one line fix is sufficient to pass on all the new tests.

```diff
+++ b/actix-multipart/src/field.rs
@@ -302,7 +302,7 @@ impl InnerField {
         // check boundary
-        if len > 4 && payload.buf[0] == b'\r' {
+        if len >= 4 && payload.buf[0] == b'\r' {
```

robjtede:
I'd also say the new tests are quite repetitive. A single new test would suffice:

<details>
<summary>Proposed Test</summary>

```rust
#[actix_rt::test]
async fn boundary_marker_split_across_chunks() {
    let (body, headers) = create_double_request_with_header();
    let boundary_start = memchr::memmem::find_iter(&body, b"\r\n--")
        .nth(1)
        .expect("body should contain a boundary between its fields");
    let boundary_marker_end = boundary_start + 4;

    let (mut tx, rx) = h1::Payload::create(false);
    tx.feed_data(body.slice(..boundary_start));
    tx.feed_data(body.slice(boundary_start..boundary_marker_end));

    let mut multipart = Multipart::new(&headers, rx);

    let mut field = multipart
        .next()
        .await
        .expect("multipart should have two fields")
        .expect("multipart body should be well formatted");
    assert_eq!(field.next().await.unwrap().unwrap(), "one+one+one");
    let next = field.next().now_or_never();
    assert!(
        next.is_none(),
        "partial boundary marker should not be emitted as field data: {next:?}",
    );

    tx.feed_data(body.slice(boundary_marker_end..));
    assert!(field.next().await.is_none());
    drop(field);

    let field = multipart
        .next()
        .await
        .expect("multipart should have two fields")
        .expect("multipart body should be well formatted")
        .bytes(usize::MAX)
        .await
        .expect("field data should not be size limited")
        .expect("reading field data should not error");
    assert_eq!(field, "two+two+two");
}
```

</details>

Sruhvx-jpg:
Thanks for the review and the pointer on `len >= 4`!

Regarding the parsing logic, are you okay with the pattern matching code style, or would you prefer reverting to the standard `if-else` block with just the `len >= 4` change?

Also, for the tests, I took your proposed `boundary_marker_split_across_chunks` test and parameterized the split offset in a loop (from 1 byte up to the full boundary length). That way, a single test covers splits across `\r`, `\r\n`, `\r\n-`, `\r\n--`, and through the boundary string itself, while keeping the `.now_or_never()` check on each iteration.

robjtede:
I usually prefer minimal diffs. This is primarily a fix which can be released quickly. We can discuss the benefits of refactors separately if that's okay 🙂

Sruhvx-jpg:
I will take this as flag to remove the pattern matching code block and replace it with the standard if else block

robjtede (approved):

PR #4191, opened 2026-08-17
refactor: centralize actix-http test service

robjtede:
## Summary

- Move the HTTP/1 dispatcher test `ok_service` helper into `actix_http::test::services`.
- Keep the service module crate-internal and test-only.
- Leave `actix-web` unchanged.

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo clippy -p actix-http --all-targets --all-features`
- `cargo test -p actix-http --lib --all-features` (216 passed, 1 ignored)


robjtede:
Superseded by #4192, which is rebased onto current main.

PR #4198, opened 2026-08-18
refactor(files)!: remove experimental io-uring support

robjtede:
## Summary

- Remove the semver-exempt `actix-files/experimental-io-uring` feature and its implementation.
- Remove `NamedFile::open_async`; callers must use `NamedFile::open` without `.await`.
- Remove the tokio-uring reader and unsafe raw-FD metadata bridge.
- Remove the direct tokio-uring and actix-server dependencies from `actix-files`.
- Remove the dedicated Actix Files io-uring CI job.
- Keep the separate Actix Web server-level io-uring feature unchanged.

## Breaking change

Removing `NamedFile::open_async` is a public API break. This PR is labeled `B-semver-major` and targets the next major Actix Files release (`0.7`).

## Rationale

The focused diagnosis in #4195 localized the Linux allocator failures to cancellation of the io-uring file-read path. This is an alternative to #4197: remove the experimental Actix Files integration and its async-open API instead of retaining the implementation and avoiding reads for HEAD requests.

The standard file implementation already opened files synchronously when the experimental feature was disabled. Internal callers now use that API directly. No futures executor is added.

## Validation

- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `cargo clippy --locked -p actix-files --all-targets --all-features` (passes with existing warnings in other crates)
- `cargo test --locked -p actix-files --all-features`
- `cargo +nightly test --locked --doc -p actix-web`
- `cargo package --locked --allow-dirty -p actix-files`

The previous head passed Linux stable and MSRV CI. The amended head will rerun those checks. `cargo-semver-checks` is expected to report the removed feature and method as a major change; that result matches the PR classification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,24 +93,6 @@ jobs:
         if: matrix.version.name == 'stable' && matrix.target.os == 'ubuntu-latest'
         uses: EmbarkStudios/cargo-deny-action@3c6349835b2b7b196a839186cb8b78e02f7b5f25 # v2.1.1
 
-  io-uring:
-    name: io-uring tests
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          persist-credentials: false
-
-      - name: Install Rust
-        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
-        with:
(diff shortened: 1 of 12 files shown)


robjtede:
@JohnTitor Curious on your thoughts about this. I'm tempted to just strip out all experimental io-uring stuff anyway. Tokio proper seems to be making progress on integration these days. IMO this is causing us too many CI failures and not carrying its weight vs. the maintenance burden.

I prefer this over https://github.com/actix/actix-web/pull/4197 anyway.

JohnTitor (approved):
+1 for removing 👍

PR #4211, opened 2026-08-31
refactor(multipart): use pattern matching on boundary length and EOF state in field polling

Sruhvx-jpg:
## PR Type
Refactor

## PR Checklist

## Overview
Refactors boundary delimiter resolution into a 2D `match (len >= b_size, payload.eof)` pattern match, making the state transitions (full boundary vs. pending stream vs. EOF fallback) explicit and idiomatic compared to nested `if-else` branches (existing test suite already covers all delimiter chunk-split scenarios).

### DIFF
--- a/actix-multipart/src/field.rs
+++ b/actix-multipart/src/field.rs
@@ -313,11 +313,18 @@ impl InnerField {
 
             if let Some(b_len) = b_len {
                 let b_size = boundary.len() + b_len;
-                if len < b_size {
-                    return Poll::Pending;
-                } else if &payload.buf[b_len..b_size] == boundary.as_bytes() {
-                    // found boundary
-                    return Poll::Ready(None);
+                let available = len - b_len;
+                let check_len = cmp::min(available, boundary.len());
+
+                if payload.buf[b_len..b_len + check_len] == boundary.as_bytes()[..check_len] {
+                    match (len >= b_size, payload.eof) {
+                        // full boundary delimiter found
+                        (true, _) => return Poll::Ready(None),
+                        // partial boundary prefix with stream still open; wait for more chunks
+                        (false, false) => return Poll::Pending,
+                        // partial match at EOF is not a boundary; fall through to yield as payload data
+                        (false, true) => {}
+                    }
                 }
             }
         }


Sruhvx-jpg:
hey @robjtede, can u please merge this so can move one of my pr's from draft :)

robjtede:
@Sruhvx-jpg they all target main so no reason this has to go first imo

there's a subtle behavior change introduced by this, it isn't a strict refactor so i'm deciding what to do with it

Sruhvx-jpg:
@robjtede  now that i take a look again at it, i just realized that i wasn returning any Error::incomplete like earlier code did. Well i tried finding something wrong in it, but nor am i as good as u nor do i have a good agent. Maybe the AGI ( astra ) will find something 1 in 67 trillion condtition

robjtede:
The new `Incomplete` return is a fix for another issue I made today. This PR already had a subtle behavior change around its EOF handling.

robjtede:
With #4255 merged, matching partial boundaries at EOF now return `Incomplete` on both main and this branch. The earlier EOF concern is resolved.

There is still a behavior change for a short candidate that already mismatches the boundary. With boundary `abc` and buffered `\r\n--ax`, main waits while the stream is open (or returns `Incomplete` at EOF), whereas this branch yields those bytes as data (then reports `Incomplete` if the message ends). Releasing a known mismatch makes sense, but this is more than a refactor. Could you add a regression test for this behavior, including the EOF case?

<details>
<summary>Test sketch: a known mismatch should be yielded without waiting for more input</summary>

This could go in `field.rs`'s existing test module. It should fail on main because `next` is `None` (`Pending`), and pass on this branch. Keep the payload open until after the data assertion so later bytes cannot hide the difference. This is a sketch, not an executed test.

```rust
#[test]
fn mismatching_boundary_prefix_is_yielded_without_more_input() {
    let mut headers = HeaderMap::new();
    headers.insert(
        header::CONTENT_TYPE,
        header::HeaderValue::from_static("multipart/form-data; boundary=abc"),
    );

    let (mut tx, rx) = h1::Payload::create(false);
    tx.feed_data(Bytes::from_static(
        b"--abc\r\nContent-Disposition: form-data; name=\"field\"\r\n\r\n\r\n--ax",
    ));

    let mut multipart = Multipart::new(&headers, rx);
    let mut field = multipart
        .next()
        .now_or_never()
        .expect("field headers should be ready")
        .expect("multipart should contain a field")
        .expect("field headers should be valid");

    let next = field.next().now_or_never();
    assert!(
        matches!(&next, Some(Some(Ok(bytes))) if bytes.as_ref() == b"\r\n--ax"),
        "a mismatching boundary prefix should be yielded without more input: {next:?}",
    );

    tx.feed_eof();

    let next = field.next().now_or_never();
    assert!(
        matches!(next, Some(Some(Err(crate::error::Error::Incomplete)))),
        "EOF without a closing boundary should report an incomplete message: {next:?}",
    );
}
```

A companion case should make EOF known before the mismatch is parsed and assert that the bytes are yielded before `Incomplete`.

</details>

Sruhvx-jpg:
@robjtede Aight done - made it with   '_Make No Mistake_'  Ultramoderne  prompt ingénierie

Thanks for the guidance! Now back to Yaoi Time

Screenshot From 2026-09-17 19-38-19

robjtede (approved):

PR #4238, opened 2026-09-12
fix(files): apply path_filter to directory listings

### LINKED ISSUE #3207: Setting the path_filter and show_files_listing, but list page still show all files.
```rs
Files::new("/", EXAMPLES_DIR)
    .path_filter(|n, _| {
        println!("{:?}", n);
        let s = n.to_str().unwrap();
        if s == "lib.rs" || s == "" {
            true
        } else {
            false
        }
    })
    .show_files_listing(),
```
## Expected Behavior
Only display the files that pass the filter, as in the example above, only the lib.rs file should be displayed.

## Current Behavior
image
Display all files, but only those passed by the filter can be accessed.

## Possible Solution
Add a filter parameter to the DirectoryRenderer function, or add a filter field to the Directory.

```rs
use actix_files::Files;
use actix_web::{get, guard, middleware, App, HttpServer, Responder};
const EXAMPLES_DIR: &str = concat![env!("CARGO_MANIFEST_DIR"), "/src"];

#[actix_web::main]
    env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));

    log::info!("starting HTTP server at http://localhost:8080");

    HttpServer::new(|| {
        App::new()
            .service(
                Files::new("/", EXAMPLES_DIR)
                    .path_filter(|n, _| {
                        println!("{:?}", n);
                        let s = n.to_str().unwrap();
                        if s == "lib.rs" || s == "" {
                            true
                        } else {
                            false
                        }
                    })
                    .show_files_listing(),
            )
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}

```
## Context
When I set a filter, it indicates that I don't want users to be aware of the existence of certain files. However, these files are still displayed on the list page and are inaccessible, which can confuse users.

## Your Environment
- Rust Version (I.e, output of `rustc -V`): rustc 1.76.0-nightly (1e9dda77b 2023-11-22)
- Actix Web Version: e95c8fe5a6d466ba674ea2ea35c0c4d18e1bb9d0

KayanoLiam:
## PR Type

Bug Fix

## PR Checklist

## Overview

Apply `Files::path_filter` to entries in the default directory listing, while retaining existing request filtering and default-handler behavior.

The filter is passed to the default renderer through a crate-private callback parameter. Entry paths are relative to the serving root and the request head belongs to the listing request. `Directory` retains its original public fields, constructor, derived Debug implementation, and auto traits. No version changes or semver-check suppressions are needed.

Custom renderers retain their existing two-argument signature and remain responsible for their own listing contents, as before.

Tests live in `actix-files/tests/listing_filter.rs` and cover the downstream `Directory` API boundary (struct literals/update syntax/destructuring and Send/Sync/UnwindSafe/RefUnwindSafe/Unpin), root and mounted nested listings with a rejected directory entry, request-header handling, missing-index fallback, both builder orders and the existing hidden-file rule, plus a custom renderer keeping its two-argument callback. Direct request filtering and default-handler dispatch remain covered by the existing `test_path_filter` and `test_default_handler_filter` unit tests.

## Validation

- Rebased onto `main`; this also picks up #4245, which fixes the unrelated `field_try_next_panic` failure seen on the macOS job.
- `cargo test -p actix-files`: passed (78 unit/integration tests; 9 doc tests passed, 1 ignored).
- `cargo clippy -p actix-files --all-targets --all-features -- -D warnings`: passed.
- `cargo +nightly fmt -p actix-files -- --check`: passed.
- `cargo semver-checks --workspace --release-type=patch --baseline-rev f8c80cd8b8887ee111b386025a0fdf6d6522d4ef`: passed on the previous revision using cargo-semver-checks 0.50.0; the public API is unchanged since then and the CI semver-checks job passed as well.

Closes #3207


actix-files/src/lib.rs
```diff
@@ -1326,6 +1326,38 @@ mod tests {
         assert_eq!(resp.status(), StatusCode::NOT_FOUND);
     }
 
+    #[actix_rt::test]
+    async fn test_path_filter_files_listing() {
```
robjtede: Suggestion: consolidate this root-listing case with `nested_listing_filter_is_root_relative_and_request_aware` in the integration test. Keep both root and nested paths, exclusion of a directory entry, request-header handling, missing-index fallback, and both builder orders. The direct-file OK/404 assertions here and in the nested test overlap the existing `test_path_filter`; leave those checks there so the new regression focuses on listing contents.

robjtede (commented):
Test-scope suggestions from the review: retain the distinct listing and public-API coverage, and consolidate overlapping checks. These are cleanup recommendations, not blocking correctness findings.

actix-files/src/lib.rs
```diff
@@ -1326,6 +1326,38 @@ mod tests {
...
+        let tmp = tempfile::tempdir().unwrap();
+        fs::write(tmp.path().join("allowed.txt"), "ok").unwrap();
+        fs::write(tmp.path().join("blocked.txt"), "no").unwrap();
+        fs::create_dir(tmp.path().join("subdir")).unwrap();
+        fs::write(tmp.path().join("subdir").join("nested.txt"), "no").unwrap();
```
robjtede: Suggestion: remove the creation of `subdir/nested.txt`. The request only lists the root and does not recurse, so creating the empty `subdir` is sufficient to verify that rejected directories are omitted.

actix-files/tests/listing_filter.rs
```diff
@@ -0,0 +1,107 @@
...
+    fs,
+    panic::{RefUnwindSafe, UnwindSafe},
+    path::Path,
+};
+
+use actix_files::{Directory, Files};
+use actix_web::{dev::ServiceResponse, http::StatusCode, test, App, HttpResponse};
+
+#[actix_web::test]
+async fn directory_public_api() {
```
robjtede: Suggestion: make this an ordinary synchronous Rust test. It only checks trait bounds and struct syntax and has no async operations, so it does not need an Actix runtime. Keep this check in an integration test to retain the downstream public-API boundary.

actix-files/tests/listing_filter.rs
```diff
@@ -0,0 +1,107 @@
...
+            let mut req = test::TestRequest::with_uri("/static/sub/");
+            if allow {
+                req = req.insert_header(("x-allow", "yes"));
+            }
+            let res = test::call_service(&app, req.to_request()).await;
+            assert_eq!(res.status(), StatusCode::OK);
+            let body = String::from_utf8(test::read_body(res).await.to_vec()).unwrap();
+            assert_eq!(body.contains("allowed.txt"), allow);
+            assert!(!body.contains("blocked.txt"));
+            assert!(!body.contains(".hidden"));
```
robjtede: Suggestion: allow `sub/.hidden` through the path filter if this assertion is intended to protect the existing hidden-file rule. Currently the filter also rejects `.hidden`, so the assertion still passes if hidden-file visibility handling regresses. Otherwise, remove this fixture and assertion from the filter regression.

actix-files/tests/listing_filter.rs
```diff
@@ -0,0 +1,107 @@
...
+                    };
+                    Ok(ServiceResponse::new(
+                        req.clone(),
+                        HttpResponse::Ok().body("custom"),
+                    ))
+                })
+                .path_filter(|_, head| head.headers.contains_key("x-allow"))
+                .default_handler(|req: actix_web::dev::ServiceRequest| async {
+                    Ok(req.into_response(HttpResponse::Forbidden().body("filtered")))
+                }),
```
robjtede, line 120: Suggestion: simplify this to a focused custom-renderer test. Keep a configured filter and verify that the two-argument callback returns the custom response, but remove the default handler and the denied-request assertion below. `test_default_handler_filter` already covers dispatch to the default handler when a filter rejects a request.

actix-files/tests/listing_filter.rs
```diff
@@ -0,0 +1,107 @@
...
+    let app = test::init_service(
+        App::new().service(
+            Files::new("/static", root.path())
+                .show_files_listing()
+                .files_listing_renderer(|dir, req| {
+                    // Downstream struct literals and the existing two-argument callback still work.
+                    let _copy = Directory {
+                        base: dir.base.clone(),
+                        path: dir.path.clone(),
+                    };
```
robjtede: Suggestion: remove this `_copy` construction and use `_dir` for the unused callback parameter. `directory_public_api` already checks downstream struct construction, so this test can focus on the custom-renderer adapter.

KayanoLiam:
Thanks for the review — addressed all six suggestions in 95afce57ed9d32a9d656039c99e568955ddcbd3a:

- Moved the root-listing case out of `lib.rs` and consolidated it with the nested case in `tests/listing_filter.rs` (`default_listing_applies_path_filter`): root + nested listings, a rejected (empty) directory entry, request-header handling, missing-index fallback and both builder orders. The direct-file OK/404 assertions were dropped since `test_path_filter` already covers them, and `subdir/nested.txt` is gone.
- `directory_public_api` is now a plain `#[test]`.
- `.hidden` (root and `sub/`) is now allowed through the filter, so the assertion actually guards the listing's hidden-file rule.
- Custom-renderer test trimmed to the two-argument callback returning the custom response with a filter configured; default handler, denied-request assertion and the `_copy` construction removed (`_dir`).
- Rebased onto `main`, which also picks up #4245 for the unrelated `field_try_next_panic` failure on the macOS job.

Locally: `cargo test -p actix-files`, `cargo clippy -p actix-files --all-targets --all-features -- -D warnings` and nightly `cargo fmt --check` all pass.

KayanoLiam:
Pushed 3b26e98a to fix the workflow startup failures on the previous PR head. GitHub reported `Invalid lockfile: .github/workflows/actions.lock` before the affected workflows could start any jobs. The workflow updates inherited from `main` were not reflected in the lockfile.

This commit refreshes the four changed direct action pins and their two transitive dependencies. It changes only `actions.lock`, preserving the existing SHA-pinned workflow YAML and the `actix-files` implementation/tests.

Validation:
- `gh actions-lock --verify-local` and `gh actions-lock --verify`: valid.
- zizmor 1.24.1, offline workflow scan: no findings.
- `cargo test -p actix-files`: 78 unit/integration tests and 9 doc tests passed; 1 doc test ignored.
- `git diff --check`: passed.

The updated commit is ready for CI and another review.

PR #4242, opened 2026-09-14
chore: remove dependency status badges

xtqqczze:
cc: @robjtede


robjtede:
i'd like to keep them and solve the root issue

